PR #6, opened 2015-11-18
docs/roadmap

diasdavid:
This helps getting a sense of the 'modularizeness' of libp2p and what we need to break go-libp2p into (as in the other implementations, it is more about building the blocks and then making the glue)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ $ GO15VENDOREXPERIMENT=1 go test ./p2p/<path of _test.go you want to run>
 
 # Modules
 
-- [libp2p](https://github.com/diasdavid/rust-libp2p) (entry point)
+- [libp2p](https://github.com/ipfs/go-libp2p) (entry point)
 - **Swarm**
   - [libp2p-swarm]()
   - [libp2p-identify]()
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,20 +1,94 @@
...
+
+```bash
+$ go get github.com/ipfs/go-libp2p
+```
+
+# Run tests
+
+```bash
+$ cd $GOPATH/src/github.com/ipfs/go-libp2p
+$ GO15VENDOREXPERIMENT=1 go test ./p2p/<path of _test.go you want to run>
```
whyrusleeping: just advise people to use the path to the package they want to test. the `-run` flag can select individual tests from there

README.md
```diff
@@ -1,20 +1,94 @@
...
+ - **Add tests**. There can never be enough tests.
+
+# Usage
+
+`go-libp2p` repo will be a place holder for the list of Go modules that compose Go libp2p, as well as its entry point.
+
+## Install
+
+```bash
+$ go get github.com/ipfs/go-libp2p
```
whyrusleeping: I always add `-u` to these installation things, in case they have old source already.

whyrusleeping:
I'd say this is a good starting point. merging.

diasdavid:
thank you :)

PR #8, opened 2015-11-20
make new stream calls accept a context

whyrusleeping:
this will allow dials caused by `NewStream` calls to be cancelled. They have a bad habit of hanging otherwise.

### DIFF
--- a/p2p/net/mock/mock_test.go
+++ b/p2p/net/mock/mock_test.go
@@ -14,8 +14,8 @@ import (
 	protocol "github.com/ipfs/go-libp2p/p2p/protocol"
 	testutil "github.com/ipfs/go-libp2p/testutil"
 
-	context "gx/QmacZi9WygGK7Me8mH53pypyscHzU386aUZXpr28GZgUct/context"
 	detectrace "github.com/jbenet/go-detect-race"
+	context "gx/QmacZi9WygGK7Me8mH53pypyscHzU386aUZXpr28GZgUct/context"
 )
 
 func randPeer(t *testing.T) peer.ID {
@@ -208,21 +208,21 @@ func TestNetworkSetup(t *testing.T) {
 	// p.NetworkConns(n3)
 
 	// can create a stream 2->3, 3->2,
-	if _, err := n2.NewStream(p3); err != nil {
+	if _, err := n2.NewStream(ctx, p3); err != nil {
 		t.Error(err)
 	}
-	if _, err := n3.NewStream(p2); err != nil {
+	if _, err := n3.NewStream(ctx, p2); err != nil {
 		t.Error(err)
 	}
 
 	// but not 1->2 nor 2->2 (not linked), nor 1->1 (not connected)
-	if _, err := n1.NewStream(p2); err == nil {
+	if _, err := n1.NewStream(ctx, p2); err == nil {
 		t.Error("should not be able to connect")
 	}
-	if _, err := n2.NewStream(p2); err == nil {
+	if _, err := n2.NewStream(ctx, p2); err == nil {
 		t.Error("should not be able to connect")
 	}
-	if _, err := n1.NewStream(p1); err == nil {
+	if _, err := n1.NewStream(ctx, p1); err == nil {
 		t.Error("should not be able to connect")
 	}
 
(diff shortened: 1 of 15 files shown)


p2p/net/mock/mock_test.go
```diff
@@ -14,8 +14,8 @@ import (
 	protocol "github.com/ipfs/go-libp2p/p2p/protocol"
 	testutil "github.com/ipfs/go-libp2p/testutil"
 
-	context "gx/QmacZi9WygGK7Me8mH53pypyscHzU386aUZXpr28GZgUct/context"
 	detectrace "github.com/jbenet/go-detect-race"
+	context "gx/QmacZi9WygGK7Me8mH53pypyscHzU386aUZXpr28GZgUct/context"
```
jbenet: why cant this be `ipfs/`? or `gx/ipfs/` ? (preserving paths is important. adding ipns support and other things will be possible. 

please please avoid breaking protocol naming. this is the kinda crap that makes it hard to nest + reuse protocols.

whyrusleeping:
please just give a LGTM and let me merge.

p2p/net/mock/mock_test.go
whyrusleeping: packages will never use ipns paths. that breaks the whole point of using hashes for package references. ipns names can be used for repos just fine, as those are expected to be mutable.

if a package ref were tied to an ipns hash, then installing the same package two different times may result in a broken package, as the entry may update in between installs. I can switch it to *just* ipfs/<hash> instead of gx/<hash>, david and i thought that gx would be a nice prefix for branding and for making it easy to figure out why i have random hash directories sitting around different places.

jbenet: > packages will never use ipns paths. that breaks the whole point of using hashes for package references.

not necessarily. there could be other protocols. 

> if a package ref were tied to an ipns hash, then installing the same package two different times may result in a broken package, as the entry may update in between installs.

agreed, that's how all of Go works today

> I can switch it to just ipfs/ instead of gx/, david and i thought that gx would be a nice prefix for branding and for making it easy to figure out why i have random hash directories sitting around different places.

i dont think it should be `gx/hash`. other protocols may be needed. i think it should be either `gx/ipfs/hash` (proper nesting) or just `ipfs/hash`.

jbenet:
@whyrusleeping it's been merged for half an hour

p2p/net/mock/mock_test.go
whyrusleeping: @noffle @diasdavid, thoughts here? i'm amenable to changing things to `ipfs/hash` but too much testing is less fun.

diasdavid: `gx/ipfs/hash`could be good when we start "mounting everything" and just let the OS figure out someone is requesting for a ipfs link and fetches it from ipfs. Having it gx is kind of nice because keeps things organised on the local fs (+it is really good for devs to look and tell first hand "oh, this uses gx to fetch the dependency)

noffle: `gx/` doesn't encode protocol information re retrieval; @jbenet makes a good point here. However, the import path here doesn't **need** to encode protocol information. @whyrusleeping, I think we talked about something like a `.git` directory (so, `.gx`) that stored each package's local installation FS path (`gx/<hash>` here). And then the `package.json` would define that package's protocol path (e.g. `/ipns/<hash>`). My only qualm with adding another layer of nesting is that diminishes the user experience (try using command line tools in Java projects -- ha).

`gx/[ipfs|ipns]/`: encodes the protocol underneath the `gx/` prefix. Nice that it matches up with what's in `package.json`, but not necessary (as per above). I like the 1:1 matching of FS path and location here.

I don't really like `ipfs/hash/`, since drops the `gx` substring. Having it might be nice for identifying what this folder means when looking at a project using gx.

jbenet: > doesn't need to encode protocol information

this breaks encapsulation and proper layering. it doesnt need another
protocol today, but it makes it way harder for us to evolve gx tomorrow --
supposing we had some other prefix (either someone else's protocol, or even
an ipfs protocol that made sense). embedding the assumption that `gx/` *always
means* `ipfs/` is a source of issues. either allow full paths, or gx-nested
full paths, but not gx-nested partial paths.

PR #12, opened 2015-11-26
WIP: getting interop between js and go swarm

diasdavid:


examples/swarm-setup.go
```diff
@@ -0,0 +1,35 @@
+package main
+
+import (
+	metrics "github.com/ipfs/go-libp2p/p2p/metrics"
+	tu "github.com/ipfs/go-libp2p/testutil"
+	ma "github.com/jbenet/go-multiaddr"
+	swarm "go-libp2p/p2p/net/swarm"
```
diasdavid: Trying to understand why for swarm and peer it didn't complain that I didn't add `github.com/ipfs` to the path

whyrusleeping: does this build? That seems wrong to me...

whyrusleeping: yeah, this doesnt build for me

diasdavid: Shed some light :) I'm trying to get it to work too. I was hoping that I could use gx, but installing globally is hanging https://github.com/whyrusleeping/gx/issues/23

whyrusleeping: just add `github.com/ipfs/` to the imports, and make sure that `GO15VENDOREXPERIMENT` is set to `1`. it *should* work. (i'll try it too)

whyrusleeping: you'll also need to change the multiaddr import to use the vendored one:
```
	ma "gx/QmVUi2ncqnU48zsPgR1rQosDGwY3SSZ1Ndp33j33YjXdsj/go-multiaddr"
```

This is because of go's stupid insistence that the import path of a type is part of the type.

whyrusleeping: If you dont want to type this manually, just run `gx-go rewrite` from the `go-libp2p` directory (the one with the package.json in it)

PR #15, opened 2016-02-16
add remote multiaddr from active connection to peerstore

whyrusleeping:
the main change is question is on line 193, the rest is just a quick readability cleanup.

It doesnt appear that we ever add the actual address we get from the actual connection to the peerstore. That seems like a fairly severe oversight to me...


lgierth:
LGTM :+1:

PR #22, opened 2016-03-04
Add fallback dialer

whyrusleeping:
Previously we were unable to dial on a given transport if there were no
listener addresses defined for that. This meant that most people couldnt
use utp (as they had no utp listener transport to dial from).

### DIFF
--- a/p2p/net/transport/fallback.go
+++ b/p2p/net/transport/fallback.go
@@ -1,0 +1,63 @@
+package transport
+
+import (
+	"fmt"
+
+	manet "gx/ipfs/QmQB7mNP3QE7b4zP2MQmsyJDqG5hzYE2CL8k1VyLWky2Ed/go-multiaddr-net"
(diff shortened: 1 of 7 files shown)


p2p/net/transport/fallback.go
```diff
@@ -0,0 +1,63 @@
...
+func (fbd *FallbackDialer) Matches(a ma.Multiaddr) bool {
+	return mafmt.TCP.Matches(a) || mafmt.UTP.Matches(a)
+}
+
+func (fbd *FallbackDialer) Dial(a ma.Multiaddr) (Conn, error) {
+	if mafmt.TCP.Matches(a) {
+		return fbd.tcpDial(a)
+	}
+	if mafmt.UTP.Matches(a) {
+	}
```
Kubuxu: Isn't something missing here?

whyrusleeping: >.> don't judge me

Kubuxu:
Apart from that comment, LGTM.

PR #36, opened 2016-04-11
Fix address filtering for /ip6, add tests

lgierth:
License: MIT
Signed-off-by: Lars Gierth <larsg@systemli.org>

### DIFF
--- a/p2p/net/filter/filter.go
+++ b/p2p/net/filter/filter.go
@@ -2,7 +2,6 @@ package filter
 
 import (
 	"net"
-	"strings"
 	"sync"
 
 	manet "gx/ipfs/QmYVqhVfbK4BKvbW88Lhm26b3ud14sTBvcm1H7uWUx1Fkp/go-multiaddr-net"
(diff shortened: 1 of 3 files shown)


lgierth:
Fixed the tests

p2p/net/filter/filter.go
```diff
@@ -27,18 +26,24 @@ func (fs *Filters) AddDialFilter(f *net.IPNet) {
...
+		return false
+	}
+	netaddr, err := manet.ToNetAddr(maddr[0])
 	if err != nil {
 		// if we cant parse it, its probably not blocked
 		return false
 	}
+	netip := net.ParseIP(netaddr.String())
+	if netip == nil {
+		return false
```
jbenet: right now these filters are ip-specific. in the future we may want to express filters of other kinds (like tcp ports, etc)

jbenet:
LGTM. @whyrusleeping ?

whyrusleeping:
yeah, LGTM

Kubuxu:
Thanks for that.

PR #38, opened 2016-04-14
refactor swarm dialing logic

whyrusleeping:


whyrusleeping:
A basic idea of the change:
Instead of using channels allocated in the Dial function for both (per peer, and fd consuming) types of rate limiting, we create a rateLimiter object that manages scheduling dials for *all* outgoing connections. It uses a single lock, and the only internal logic it does is for new dials, check if you can take a token, if so, start the dial in a goroutine, if you cant take a token right now, add the dial to a queue to be started as soon as a token is available. This is functionally equivalent to the existing implementation, except for the fact that it uses vastly fewer goroutines. This simplification in concurrency lets us more confidently use a channel of addresses as input to the dial instead of an array of them. Using a channel of addresses with the current implementation increases the locking complexity (And chance for accidental deadlocks) immensely.

whyrusleeping:
when reviewing this, i recommend looking over the changes in swarm_dial.go, particularly in `dialAddrs`, *then* take a look at the dialLimiter code afterwards. It should make more sense that way.

p2p/net/swarm/swarm_dial.go
```diff
@@ -319,32 +319,34 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+	// get live channel of addresses for peer, filtered by the given filters
+	/*
+		remoteAddrChan := s.peers.AddrsChan(ctx, p,
+			addrutil.AddrUsableFilter,
+			subtract_filter,
+			s.Filters.AddrBlocked)
+	*/
+
+	////// TEMP UNTIL PEERSTORE GETS UPGRADED
+	// Ref: https://github.com/ipfs/go-libp2p-peer/pull/1
```
noffle: +1 for tracking issue :heart:

package.json
```diff
@@ -1,5 +1,4 @@
 {
-  "name": "go-libp2p",
```
noffle: no name?

p2p/net/swarm/swarm_dial.go
```diff
@@ -364,98 +366,64 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
-			return connC, nil
-		case <-ctx.Done():
-			// break out and return error
-			break
 		}
 	}
-	return nil, exitErr
+}
+
+func (s *Swarm) limitedDial(ctx context.Context, p peer.ID, a ma.Multiaddr, resp chan dialResult) {
```
noffle: Nit: godoc

p2p/net/swarm/swarm_dial.go
```diff
@@ -364,98 +366,64 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
-
-	// sort addresses so preferred addresses are dialed sooner
-	sort.Sort(AddrList(remoteAddrs))
-
-	// try to connect to one of the peer's known addresses.
-	// we dial concurrently to each of the addresses, which:
-	// * makes the process faster overall
-	// * attempts to get the fastest connection available.
-	// * mitigates the waste of trying bad addresses
+func (s *Swarm) dialAddrs(ctx context.Context, p peer.ID, remoteAddrs <-chan ma.Multiaddr) (conn.Conn, error) {
```
noffle: Nit: godoc

p2p/net/swarm/swarm_dial.go
```diff
@@ -364,98 +366,64 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
 	log.Debugf("%s swarm dialing %s %s", s.local, p, remoteAddrs)
 
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel() // cancel work when we exit func
 
-	conns := make(chan conn.Conn)
-	errs := make(chan error, len(remoteAddrs))
+	// use a single response type instead of errs and conns, reduces complexity *a ton*
```
noffle: Haha, yes. I've learned this lesson a few times. :)

p2p/net/swarm/swarm_dial.go
```diff
@@ -364,98 +366,64 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
-		for _, addr := range remoteAddrs {
-			// returns whatever ratelimiting is acceptable for workerAddr.
-			// may not rate limit at all.
-			rl := s.addrDialRateLimit(addr)
-			select {
-			case <-ctx.Done(): // our context was cancelled
-				return
-			case rl <- struct{}{}:
-				// take the token, move on
+			// limitedDial will start a dial to the given peer when
```
noffle: awesome doc string to have on `limitedDial`!

p2p/net/swarm/swarm_dial.go
```diff
@@ -364,98 +366,64 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+			// it is able, respecting the various different types of rate
+			// limiting that occur without using extra goroutines per addr
+			s.limitedDial(ctx, p, addr, respch)
+			active++
+		case <-ctx.Done():
+			if exitErr == defaultDialFail {
```
noffle: Why wouldn't you want to always use `ctx.Err()` here?

p2p/net/swarm/swarm_dial.go
```diff
@@ -364,98 +366,64 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+			active--
+			if resp.Err != nil {
+				log.Error("got error on dial: ", resp.Err)
+				// Errors are normal, lots of dials will fail
+				exitErr = resp.Err
+
+				if remoteAddrs == nil && active == 0 {
+					return nil, exitErr
+				}
+			} else if resp.Conn != nil {
```
noffle: Is it abnormal behaviour for `resp.Conn == nil`? Should this log/error?

p2p/net/swarm/swarm_dial.go
```diff
@@ -364,98 +366,64 @@ func (s *Swarm) dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
+				return resp.Conn, nil
```
noffle: Will anything be leaked here if we return while `active > 0`?

p2p/net/swarm/addr/addr.go
```diff
@@ -42,10 +42,14 @@ func init() {
 
 // FilterAddrs is a filter that removes certain addresses, according to filter.
```
noffle: "according to *a list* of filters"

"if *all* of the filters return true, the address is kept"

p2p/net/swarm/addr/filter.go
```diff
@@ -0,0 +1,27 @@
...
+	for _, a := range addrs {
+		addrmap[string(a.Bytes())] = true
+	}
+
+	return func(a ma.Multiaddr) bool {
+		return !addrmap[string(a.Bytes())]
+	}
+}
+
+func IsFDCostlyTransport(a ma.Multiaddr) bool {
```
noffle: What does "costly" mean in this context? Could a short godoc here clarify?

p2p/net/swarm/addr/filter.go
```diff
@@ -0,0 +1,27 @@
+package addrutil
+
+import (
+	ma "github.com/jbenet/go-multiaddr"
+	mafmt "github.com/whyrusleeping/mafmt"
+)
+
+func SubtractFilter(addrs ...ma.Multiaddr) func(ma.Multiaddr) bool {
```
noffle: A short godoc here would be helpful, to explain what this does and why it's useful.

p2p/net/swarm/addr/filter.go
```diff
@@ -0,0 +1,27 @@
...
+	return func(a ma.Multiaddr) bool {
+		return !addrmap[string(a.Bytes())]
+	}
+}
+
+func IsFDCostlyTransport(a ma.Multiaddr) bool {
+	return mafmt.TCP.Matches(a)
+}
+
+func FilterNeg(f func(ma.Multiaddr) bool) func(ma.Multiaddr) bool {
```
noffle: nit: quick godoc

p2p/net/swarm/limiter.go
```diff
@@ -0,0 +1,149 @@
...
+		activePerPeer:      make(map[peer.ID]int),
+		dialFunc:           df,
+	}
+}
+
+func (dl *dialLimiter) finishedDial(dj *dialJob) {
+	dl.rllock.Lock()
+	defer dl.rllock.Unlock()
+
+	// release tokens in reverse order than we take them
```
noffle: I see lots of references to the term "token", but only in comments. Could there be a comment somewhere relevant that explains what a token is in this context?

p2p/net/swarm/limiter_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+func mustAddr(t *testing.T, s string) ma.Multiaddr {
+	a, err := ma.NewMultiaddr(s)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return a
+}
+
+func addrWithPort(t *testing.T, p int) ma.Multiaddr {
```
noffle: It's not super important, but you could have the funcs `generateGoodAddr`, `generateBadAddr`, and `isGoodAddr` that use some magic port number underneath but don't expose it.

noffle:
Just a few comments and getting CI to pass -- otherwise LGTM :horse:! Lots of good godocs and thorough tests!

p2p/net/swarm/swarm_dial.go
whyrusleeping: because its next to useless to the user. If we don't have any other better response, then we can use ctx.Err().

p2p/net/swarm/swarm_dial.go
whyrusleeping: nope, the active check is just to know when we're "done" with work. so we don't sit here doing nothing.

PR #51, opened 2016-05-16
Update id.go

efealatan:
Hello, 

i see that protocols are set in the populateMessage()  but for the consumeMessage() there is a line comment for it but no code for it not sure if this intentional during dev.

thanks


whyrusleeping:
@efealatan hrm... I think that was there to remind us that at some point we want to track that information. Maybe instead of removing it we can turn it into a clearer comment saying that?

efealatan:
Cool no worries, added a small todo to the comment to make it clear

thanks

PR #58, opened 2016-05-30
don't leak goroutines when pings complete

whyrusleeping:


whyrusleeping:
@Kubuxu wanna CR for me?

Kubuxu:
LGTM, [this](https://github.com/ipfs/go-libp2p/pull/58/files#diff-8f5503573ee415bb07d6c4dd11299c37R79) change will solve the currently reported issue but rest will prevent hangs in general. Nice.

lgierth:
This LGTM as an adhoc fix, but longer term we should aim for making this independent of time, e.g. by allowing one ping per peer, and aborting earlier pings when a new ping comes in.

PR #61, opened 2016-06-01
revert yamux back to my patch pending hashicorp/yamux#30

whyrusleeping:
reverting this until https://github.com/hashicorp/yamux/issues/30 is fixed

### DIFF
--- a/package.json
+++ b/package.json
@@ -166,9 +166,9 @@
     },
     {
       "author": "whyrusleeping",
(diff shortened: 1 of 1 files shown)


Kubuxu:
SGTM but name field disappeared

whyrusleeping:
dammit. I even filed an issue for it before pushing and i still forgot to fix it: https://github.com/whyrusleeping/gx/issues/75

PR #62, opened 2016-06-01
gx publish version 3.3.0

whyrusleeping:
this version contains the dialer rate limiting changes as well as the yamux revert


Kubuxu:
LGTM

PR #63, opened 2016-06-01
fix consumption of observed remote addrs

whyrusleeping:
This should resolve https://github.com/ipfs/go-ipfs/issues/2771


Kubuxu:
LGTM

PR #67, opened 2016-06-12
add hosts example

whyrusleeping:


whyrusleeping:
closes https://github.com/whyrusleeping/libp2p-examples/issues/2

PR #75, opened 2016-06-27
changes for compatability testing with js-ipfs

whyrusleeping:
DO NOT MERGE

These are changes that are needed for js and go compatibility before we get secio


diasdavid:
We have secio now :D let's make this nice!

PR #76, opened 2016-07-03
expose backoff manager in swarm

whyrusleeping:
Expose this here so we can use it in the swarm commands to clear the dial backoff


Kubuxu:
SGTM

PR #77, opened 2016-07-27
Ran gosimple (github.com/dominikh/go-simple) on go-libp2p

John-Steidley:


p2p/net/conn/dial.go
```diff
@@ -86,7 +86,7 @@ func (d *Dialer) Dial(ctx context.Context, raddr ma.Multiaddr, remote peer.ID) (
 			return
 		}
 
-		if d.PrivateKey == nil || EncryptConnections == false {
+		if d.PrivateKey == nil || !EncryptConnections {
```
John-Steidley: This is the only open question in my mind, why style do you prefer?

whyrusleeping:
LGTM! Thanks!

PR #86, opened 2016-08-16
p2p/host: expose multistream function matching and add semver func

whyrusleeping:
This lets protocols easily match based on semver, so for example, `/bitswap/1.0.0` and `/bitswap/1.0.2` would be compatible.

### DIFF
--- a/p2p/host/match.go
+++ b/p2p/host/match.go
@@ -1,0 +1,35 @@
+package host
+
+import (
+	"strings"
+
+	semver "github.com/coreos/go-semver/semver"
+)
+
+func MultistreamSemverMatcher(base string) (func(string) bool, error) {
+	parts := strings.Split(base, "/")
+	vers, err := semver.NewVersion(parts[len(parts)-1])
+	if err != nil {
+		return nil, err
+	}
+
+	return func(check string) bool {
+		chparts := strings.Split(check, "/")
+		if len(chparts) != len(parts) {
+			return false
+		}
+
+		for i, v := range chparts[:len(chparts)-1] {
+			if parts[i] != v {
+				return false
+			}
+		}
+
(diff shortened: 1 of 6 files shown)


Kubuxu:
I was thinking about prerelease version parts but those are not important in case of compat.

Do we want to do anything special for major version 0 of protocols? Like treating minor as compat breaker?
Either way I would suggest adding test cases with major version 0 to the suite so it is definitive.

There are also no tests for minor version change, which makes it a bit less clear

whyrusleeping:
There is a test for a minor version change, I can add more. I'll also add a test for major version 0, but i don't think we should treat it special in any way

p2p/host/match.go
```diff
@@ -0,0 +1,35 @@
...
+				return false
+			}
+		}
+
+		chvers, err := semver.NewVersion(chparts[len(chparts)-1])
+		if err != nil {
+			return false
+		}
+
+		return vers.Major == chvers.Major && vers.Minor >= chvers.Minor
```
Kubuxu: If both sides are using the same check, and one increments the Minor, then one side will accept connections and other will refuse, is that correct behaviour?

----

Oh, one would leave older handler and install new one with incremented Minor?

whyrusleeping: good point... i'm not sure how best to handle this

whyrusleeping: Ah, i remember now. If you dial to `1.1.0` as `1.0.0`, the other side should reply with `1.0.0`, allowing you to continue on as normal

whyrusleeping: If you dial `1.1.0` to a `1.0.0`, the other side will return `na` and you can retry using `1.0.0`.

whyrusleeping:
@Kubuxu LGTY?

Kubuxu:
Yup, LGTM.

PR #89, opened 2016-08-17
Feat/protocol prefs

whyrusleeping:
This adds support for remembering which of a set of protocols works for a given peer. It also uses protocols learned from the identify handshake to seed this knowledge  if available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 language: go
 
 go:
-    - 1.5.2
+    - 1.7
 
 env:
     - GO15VENDOREXPERIMENT=1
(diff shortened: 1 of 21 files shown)


whyrusleeping:
it appears theres actually a problem on osx here

whyrusleeping:
alright, fixed the osx thing, apparently it was slower for OSX to close out the socket than on linux.

@Kubuxu wanna review?

Kubuxu:
Can you add me to libp2p org so I can assign myself?

whyrusleeping:
invite sent

whyrusleeping:
The one thought I had here is that the preference logic should probably be moved into the peerstore.

whyrusleeping:
@Kubuxu still review this, but i'm gonna merge it so i can move forward here.

Kubuxu:
LGTM, it would be worth to use `protocol.ID` higher in stack instead of string but it isn't that important.

whyrusleeping:
@Kubuxu yeah, i was thinking that too. we can do that later, it was kindof awkward to do the casting between the peerstore and multistream code

PR #91, opened 2016-08-18
swarm: add deadline for connection setup

whyrusleeping:


p2p/net/swarm/swarm_dial.go
```diff
@@ -465,10 +465,21 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (con
...
 // dialConnSetup is the setup logic for a connection from the dial side. it
 // needs to add the Conn to the StreamSwarm, then run newConnSetup
 func dialConnSetup(ctx context.Context, s *Swarm, connC conn.Conn) (*Conn, error) {
 
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(ConnSetupTimeout)
+	}
+
+	if err := connC.SetDeadline(deadline); err != nil {
```
diasdavid: Got it, you use the SetDeadline like a timeout :)

re: adding SetDeadline for muxed streams - A muxed stream should not timeout, since a conn was already established, so do we really need to make it part of the API?

whyrusleeping: @diasdavid can you guarantee that the other side is going to read when you write? I can OOM your node by opening up lots of streams to it and just holding on to them

diasdavid: Good point. Is `SetDeadline` the best way to solve it? Seems a thing that helps towards the solution, but it would be nice to have a way to tell when to close connections and with which criteria.

Issue #93, opened 2016-08-24
update NAT library


whyrusleeping:
We're currently using an old version of github.com/fd/go-nat, There have since been improvements on both it and the underlying library it uses (jackpal/go-nat-pmp). Lets update!

PR #94, opened 2016-08-24
Update go-nat

Kubuxu:
resolves https://github.com/libp2p/go-libp2p/issues/93

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
       "version": "0.3.1"
     },
     {
-      "hash": "QmNLvkCDV6ZjUJsEwGNporYBuZdhWT6q7TBVYQwwRv12HT",
+      "hash": "QmYabcy8kaP658zZRZHLqRquJ37ycNWr4qhBL25tUodZWc",
       "name": "go-nat",
       "version": "0.0.0"
     },


whyrusleeping:
@Kubuxu this updates go-nat, go-nat-pmp, jackpal/gateway and huin/goupnp ?

whyrusleeping:
(i forgot about `gx diff` silly me)

Kubuxu:
Yes, whole stack.

whyrusleeping:
Now i wish i had some `gx changelog-diff` so i can just view whats changed in each of these packages... But that would require gx to track git versions (which it doesnt do yet :( )

whyrusleeping:
Alright, this LGTM. I wish we had tests...

Kubuxu:
Yeah, I can create PR on go-ipfs so we test if it builds with the rest of the system.

Kubuxu:
I've run tests of go-ipfs, they were green: https://github.com/ipfs/go-ipfs/pull/3123

whyrusleeping:
odd... the OSX CI is failing, but the build log only shows successful tests...

PR #95, opened 2016-08-26
Retry NAT punching without duration on mapping failure

Kubuxu:
Some hardware doesn't support UPnP with durations.


Kubuxu:
//c @whyrusleeping for review

whyrusleeping:
what happens if we try to set a timeout on a router that doesnt support timeouts?

Kubuxu:
The request fails, no mapping is created, this means that we didn't punch the NAT.

whyrusleeping:
If a request fails, can we tell that its because timeouts are not supported and not for some other reason?

lgierth:
Oh lol I can see that happening :) Apart from this it's good to regularly retry the mapping even if it succeeded in the first place -- routers reset, forget stuff, etc.

Rule of thumb: assume that your router is crap in one way or another until proven otherwise :)

lgierth:
> Apart from this it's good to regularly retry the mapping even if it succeeded in the first place

But that's for a different PR -- LGTM!

Kubuxu:
@whyrusleeping we can't, go nat gives us only generic error that SOAP call failed.

@lgierth didn't think about that, I will add it, shouldn't be hard.

whyrusleeping:
Yeah, this LGTM. only thing i would prefer is matching the errors more closely, but i doubt that we would get a consistent set of errors back in all cases

Kubuxu:
Hmm, it looks like calling for new permanent mapping on the same port with the same agent returns same mapping.

So I can just not short circuit calls in case of permanent mappings.

@whyrusleeping retrying it with no timeout doesn't cause any damage and has just a benefit, worst case it will fail again, best case it will work.

whyrusleeping:
Yeah, that seems like a good idea to me

Kubuxu:
Updated.

jbenet:
> it looks like calling for new permanent mapping on the same port with the same agent returns same mapping.

For some NATs. Some NATs likely don't do that. Few things are always true in networking but "there is a middlebox that will do that" is a close one.

Basically, try and try again in a variety of ways, and you'll succeed.

Kubuxu:
@lgierth @whyrusleeping does it look good now?

p2p/nat/nat.go
```diff
@@ -333,13 +336,23 @@ func (nat *NAT) NewMapping(maddr ma.Multiaddr) (Mapping, error) {
 
 func (nat *NAT) establishMapping(m *mapping) {
 	oldport := m.ExternalPort()
+
 	log.Debugf("Attempting port map: %s/%d", m.Protocol(), m.InternalPort())
-	newport, err := nat.nat.AddPortMapping(m.Protocol(), m.InternalPort(), "http", MappingDuration)
+	comment := "libp2p"
+	if m.comment != "" {
+		comment = "libp2p-" + comment
```
geoah: @Kubuxu erm is there any change you meant `comment = "libp2p-" + m.comment`?

Kubuxu: Yes, thanks.

p2p/nat/nat.go
```diff
@@ -333,13 +336,23 @@ func (nat *NAT) NewMapping(maddr ma.Multiaddr) (Mapping, error) {
 
 func (nat *NAT) establishMapping(m *mapping) {
 	oldport := m.ExternalPort()
+
 	log.Debugf("Attempting port map: %s/%d", m.Protocol(), m.InternalPort())
-	newport, err := nat.nat.AddPortMapping(m.Protocol(), m.InternalPort(), "http", MappingDuration)
+	comment := "libp2p"
+	if m.comment != "" {
```
whyrusleeping: what if we get back our own comment here? would this end up with "libp2p-libp2p-libp2p-libp2p...." ?

Kubuxu: there is `m.comment `and `comment`.
`comment` is temporary variable. I wanted to use `context.WithValue` to pass it but unfortunately goprocess doesn't have something like it.

whyrusleeping: Ah, i see. I had assumed the comment field came from a prexisting NAT mapping on a router

whyrusleeping:
@Kubuxu Not sure if we should try to fix it here or not, but we have a race condition in the old code: https://github.com/ipfs/go-ipfs/issues/3205

Can you check and see if a similar issue is present in the new code?

whyrusleeping:
I think the fix is to just lock around any time we use the 'nat' object

Kubuxu:
The code didn't change so the issue still will be there.
Will lock around. Or maybe we should chane it into worker loop?

whyrusleeping:
@Kubuxu hrm... i think for now the easiest thing will be to just lock around it.

whyrusleeping:
Those changes LGTM, but things don't appear to build

Kubuxu:
Hmm, I will check if once more.

whyrusleeping:
LGTM

PR #96, opened 2016-08-26
add listing method to backoff for eventual api access

whyrusleeping:


Kubuxu:
LGTM

whyrusleeping:
This code is no longer in this repo, moving over to go-libp2p-swarm

PR #99, opened 2016-09-04
swarm: pass in stream muxer on construct

whyrusleeping:
This lets the 'empty' swarm constructor take in the stream muxer to use, also giving us the option to pass in 'nil' to avoid set up stream multiplexing (for use in the justtcp) example


whyrusleeping:
This also makes the justtcp example a lot cleaner

Issue #41, opened 2016-04-25
libp2p spec link in CONTRIBUTE is a broken link.


mjgarton:

diasdavid:
Fixed with https://github.com/libp2p/go-libp2p/pull/216 and also a lot more documentation to come :)

VMois:
I think you can close this issue. #216 is already merged and link is working.

PR #113, opened 2016-09-15
Fix libp2p spec link

noffle:
Addresses https://github.com/libp2p/go-libp2p/issues/41

### DIFF
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -5,7 +5,7 @@ before reporting things, and help us with the open ones.
 
 Guidelines:
 
-- read the [libp2p spec](https://github.com/ipfs/specs/tree/master/libp2p)
+- read the [libp2p spec](https://github.com/libp2p/specs)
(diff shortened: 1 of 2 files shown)


CONTRIBUTE.md
```diff
@@ -5,7 +5,7 @@ before reporting things, and help us with the open ones.
 
 Guidelines:
 
-- read the [libp2p spec](https://github.com/ipfs/specs/tree/master/protocol/network)
+- read the [libp2p spec](https://github.com/ipfs/specs/tree/master/libp2p)
```
lgierth: This actually lives in github.com/libp2p/specs now (cc @RichardLitt the old ones need to be removed)

noffle: *much table flipping*

All right, I'll fix this one, but there are *many* other places these'll need to be updated.

RichardLitt: I need to automate this.

PR #114, opened 2016-09-16
conn/dial: implement contextDial up the stack

whyrusleeping:


Kubuxu:
There is a mistake in deps of go-reuseport.

I will bubble it once more.

whyrusleeping:
LGTM

PR #115, opened 2016-09-20
don't block on DiscoverHandler

whyrusleeping:
The current code runs the `Connect` calls for each peer discovered over mdns in the same goroutine. If a connect call starts to hang, then it gums up the entire mdns discovery process.

### DIFF
--- a/p2p/discovery/mdns.go
+++ b/p2p/discovery/mdns.go
@@ -1,6 +1,7 @@
 package discovery
 
 import (
+	"context"
 	"errors"
 	"io"
 	"io/ioutil"
@@ -60,7 +61,7 @@ func getDialableListenAddrs(ph host.Host) ([]*net.TCPAddr, error) {
 	return out, nil
 }
 
-func NewMdnsService(peerhost host.Host, interval time.Duration) (Service, error) {
+func NewMdnsService(ctx context.Context, peerhost host.Host, interval time.Duration) (Service, error) {
 
 	// TODO: dont let mdns use logging...
 	golog.SetOutput(ioutil.Discard)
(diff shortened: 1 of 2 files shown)


p2p/discovery/mdns.go
```diff
@@ -109,37 +109,42 @@ func (m *mdnsService) Close() error {
...
+	}()
+
 	ticker := time.NewTicker(m.interval)
 	for range ticker.C {
-		entriesCh := make(chan *mdns.ServiceEntry, 16)
-		go func() {
-			for entry := range entriesCh {
-				m.handleEntry(entry)
-			}
-		}()
```
jbenet: this is incorrect-- how does it pass tests? 

this chan creation and goroutine is here because there may be `mdns.Query` errors for a single peer. the way you moved it out to the top of `pollForEntries` will make it so that it will close the chan exit after the first error-- and it may even call `close(entriesCh)` multiple times (!!!). 

move the chan creation + goroutine back to within the for loop.

p2p/discovery/mdns.go
```diff
@@ -159,7 +164,7 @@ func (m *mdnsService) handleEntry(e *mdns.ServiceEntry) {
 
 	m.lk.Lock()
 	for _, n := range m.notifees {
-		n.HandlePeerFound(pi)
+		go n.HandlePeerFound(pi)
```
jbenet: it looks like _this_ is the only thing this changeset should change. (plus the logs)

p2p/discovery/mdns.go
whyrusleeping: Oh yeah... I was wondering where we closed that

PR #116, opened 2016-09-23
Update go-reuseport, publish 3.5.4

Kubuxu:


lgierth:
Is this for the threads thing? It's easier to write a changelog if dependency updates mention if there's any relevant change

Kubuxu:
It is for a bug we found in it, still working on it as it showed up new bug.

Yes we should have it tracked somewhere, (and also have way to track changes inside our deps that influence the go-ipfs itself).

Kubuxu:
We are having problems with go-reuseport in go-libp2p-kad-dht, will delay it.

PR #121, opened 2016-09-29
update mdns to version with fixed race condition

whyrusleeping:


whyrusleeping:
For some reason the test fail on travis OSX CI, but i cant get them to hang the same way on my laptop (OSX). Not sure whats going on here...

whyrusleeping:
strange. just fine now, no changes

PR #133, opened 2016-10-05
Readme.md: point to the right location for examples.

hsanjuan:
The linked repo seems to have been left behind vs the examples folder which contains working examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ $ make
 
 # Examples
 
-Examples can be found on the [libp2p-examples repo](https://github.com/whyrusleeping/libp2p-examples/)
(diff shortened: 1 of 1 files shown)


RichardLitt:
Should we not move https://github.com/whyrusleeping/libp2p-examples/ to libp2p, instead?

hsanjuan:
The only example in libp2p-examples (`hosts`) is already part of the examples folder in go-libp2p. It would seem to me that libp2p-examples repo is a left over and this was just an old reference. That repo is not maintained and does not really work anymore.

RichardLitt:
Sounds good to me. That seems to be only referenced in this repo, in all of GitHub, so just changing the link should work.

PR #143, opened 2016-11-06
make sure to always close streams when youre done with them

whyrusleeping:
We had random streams building up with no owners.  This should address the issue.


Kubuxu:
Was this the bug we talked about?

whyrusleeping:
@Kubuxu it might have been, i discovered and fixed a few of them recently.

PR #150, opened 2016-11-09
add negotiate timeout option to basichost

whyrusleeping:
We get a fair number of goroutines hanging in negotiate here. This should at least help prevent the buildup.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -23,6 +23,8 @@ import (
 
 var log = logging.Logger("basichost")
 
+var NegotiateTimeout = time.Second * 60
+
 // Option is a type used to pass in options to the host.
 type Option int
 
@@ -48,6 +50,8 @@ type BasicHost struct {
 	relay   *relay.RelayService
 	natmgr  *natManager
 
+	NegotiateTimeout time.Duration
+
 	proc goprocess.Process
 
 	bwc metrics.Reporter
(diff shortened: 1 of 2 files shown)


p2p/net/mock/mock_net.go
```diff
@@ -85,6 +85,7 @@ func (mn *mocknet) AddPeerWithPeerstore(p peer.ID, ps pstore.Peerstore) (host.Ho
 	}
 
 	h := bhost.New(n)
+	h.NegotiateTimeout = 0
```
lgierth: Why disable it, would any of the tests possibly take longer than a minute to negotiate?

whyrusleeping: Disabled for tests because the underlying `net.Conn` pipe that mocknet uses doesnt support `SetDeadline` :/

p2p/host/basic/basic_host.go
```diff
@@ -121,6 +135,15 @@ func (h *BasicHost) newStreamHandler(s inet.Stream) {
 		s.Close()
 		return
 	}
+
+	if h.NegotiateTimeout != 0 {
+		if err := s.SetDeadline(time.Time{}); err != nil {
```
lgierth: If time.IsZero(), it'll be reset?

lgierth:
This LGTM, very minor point about the tests as they might hang if something goes wrong? Maybe it can be disregarded

p2p/host/basic/basic_host.go
whyrusleeping: Yeap, documentation for SetDealine states:
`A zero value for t means I/O operations will not time out.`

PR #152, opened 2016-11-10
Improve echo example

hsanjuan:
* Extended explanations and links to relevant godocs
  * Some renames (the host is not dummy, it's basic)
  * Use addAddr() and let NewStream open the connection to show how to make use of the peerstore.

I had also fixed the example but @lgierth submitted it first. I have merged my fixes with his.


whyrusleeping:
@hsanjuan this is wonderful, thank you!

lgierth:
Thanks :+1:

PR #154, opened 2016-11-16
Update deps for Priv Networks

Kubuxu:
After that I have floodsub kad-dht and go-ipfs left to bubble it through.


Kubuxu:
@whyrusleeping for CR.

PR #155, opened 2016-11-17
Extract netutil to its own package

whyrusleeping:


hsanjuan:
I see this is in-progress. Since the PR has no explanation I have to assume that you needed to quickly re-use some stuff somewhere else.

Overall, having public methods for these kind of things increase the libp2p API footprint. I'm not a fan because when things are part of an API they need to be well documented, well tested, well maintained. However when I go to: https://github.com/libp2p/go-libp2p-netutil I see no tests, no readme, no godoc. despite it being published. That's hardly an incentive to re-use that module. On the downside, we get more complexity by adding extra dependencies to the ecosystem. Since these methods were public and part of the API of this module, we might be also breaking some not-so-obvious things which depended on them.

In short, this is just a kind reminder that when this is no longer in progress, I'd like to see that:
  * The logic behind the change is explained
  * The new module is documented and tested so that other people can profit from the change
  * The PR checks are green (since I have seen precedents of breaking-merges where they were not)

;-)

PR #159, opened 2016-11-18
address protocol downgrade bug

whyrusleeping:
This solves the issue of mis-remembering that a peer supports a given protocol if they downgrade their versions (or no longer support a known preferred version).


whyrusleeping:
cc @Kubuxu @lgierth for review

lgierth:
Intentional dependencies update?

Otherwise LGTM :+1:

whyrusleeping:
Yeah, intentional, needed the new API to fix the problem.

Kubuxu:
LGTM

PR #168, opened 2016-12-17
if protocol selection fails, reset remembered protocols

whyrusleeping:
I beleive this happens if a node downgrades to an older version, and then restarts and reconnects quickly, and then bitswap tries to open a new stream on a misremembered protocol before the identify handshake can complete.

This wouldnt be a problem if identify was inline...

I'll see about adding a test soon, but if anyone has good ideas on how to easily test this first let me know.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -2,6 +2,7 @@ package basichost
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"time"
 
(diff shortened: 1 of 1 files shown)


Kubuxu:
no `fmt` import

lgierth:
LGTM :+1:

whyrusleeping:
This doesnt actually solve the problem... meh

whyrusleeping:
Heres the log file from the reporter: http://sprunge.us/TPJW

The line in question is the 'protocol not supported', which is something that will never be returned by `newStream` since it only creates a `lazyConn` which first off won't actually do anything until its written to, and secondly, the lazyConn doesnt ever return that error through any of its write paths... So the problem isnt coming from where i thought it was. 

Seeing the error the way we're seeing it implies that the other side doesnt actually have *any* compatible bitswap protocols available, which can't be right (or can it?). This requires further investigation...

cc @Kubuxu and @lgierth for help

whyrusleeping:
Interesting... peers exhibiting this issue don't support *any* protocols... i cant even ping between them

whyrusleeping:
Its looking like peers can get into this state where they have no available protocols. I cant ping them, dht to them, or bitswap to them.

whyrusleeping:
I think the peers i'm seeing exhibit this behaviour are mediachain peers.

PR #169, opened 2016-12-26
identify: make agent version mutable

whyrusleeping:
We're starting to have non-standard libp2p nodes in the network, this should help differentiate things for easier testing.


whyrusleeping:
In the future we should make this a field of the identify service, and then have the basichost set it on construction. but for now, this helps solve the problem.

whyrusleeping:
cc @JustinDrake @vyzo @parkan

Kubuxu:
SGTM, IDK why travis failed, it doesn't want to load logs for me.

JustinDrake:
Looks good. OpenBazaar is running a custom version of libp2p. Changing that user agent from the IPFS top level config file would be useful.

parkan:
:+1: for this, however it seems like agentversion and protocolversion are used inconsistently and in some cases (js) their meaning is switched?

PR #170, opened 2017-01-04
update go-multistream and use negotiateLazy 

whyrusleeping:


p2p/host/basic/basic_host.go
```diff
@@ -104,6 +104,7 @@ func New(net inet.Network, opts ...interface{}) *BasicHost {
 
 // newConnHandler is the remote-opened conn handler for inet.Network
 func (h *BasicHost) newConnHandler(c inet.Conn) {
+	h.Peerstore().SetProtocols(c.RemotePeer())
```
whyrusleeping: we have issues around remembering protocols across connection resets. For now we're just clearing them

p2p/host/basic/basic_host_test.go
```diff
@@ -98,7 +98,7 @@ func TestHostProtoPreference(t *testing.T) {
 	protoNew := protocol.ID("/testing/1.1.0")
 	protoMinor := protocol.ID("/testing/1.2.0")
 
-	connectedOn := make(chan protocol.ID, 16)
+	connectedOn := make(chan protocol.ID)
```
whyrusleeping: having the buffer here was allowing some 'slack' in the test which hid failures

p2p/host/basic/basic_host.go
Kubuxu: Do you want to put it in comment in there? It is not really clear from the code.

Kubuxu:
Also travis is failing, is it random or connected?

PR #184, opened 2017-03-12
Remove relay protocol (and roadmap.md)

lgierth:
This has never been used, and is incomplete and insecure.

In the spirit of ipfs/docs#12, this also removes roadmap.md.


lgierth:
I don't care, it was dumped into p2p/protocol/ over two years ago, never touched again, never used or tested. I really don't care about this code :)

Where would you wanna put it? We could keep it alive in a branch or so. (Not that it was ever actually alive.)

Kubuxu:
I would say, if it is in a state the @lgierth says, git history would be good enough.

whyrusleeping:
eh, fine

PR #185, opened 2017-03-15
Libp2p examples

hsanjuan:
Improved echo example

Added example for multicodecs

Related to https://github.com/ipfs/docs/issues/17

### DIFF
--- a/examples/http-proxy/proxy.go
+++ b/examples/http-proxy/proxy.go
@@ -1,0 +1,288 @@
+package main
+
+import (
+	"bufio"
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"strings"
+
+	// We need to import libp2p's libraries that we use in this project.
+	// In order to work, these libraries need to be rewritten by gx-go.
+	crypto "github.com/libp2p/go-libp2p-crypto"
+	host "github.com/libp2p/go-libp2p-host"
+	inet "github.com/libp2p/go-libp2p-net"
+	peer "github.com/libp2p/go-libp2p-peer"
+	ps "github.com/libp2p/go-libp2p-peerstore"
+	swarm "github.com/libp2p/go-libp2p-swarm"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr-net"
+
+	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+)
+
+// Protocol defines the libp2p protocol that we will use for the libp2p proxy
+// service that we are going to provide. This will tag the streams used for
+// this service. Streams are multiplexed and their protocol tag helps
+// libp2p handle them to the right handler functions.
+const Protocol = "/proxy-example/0.0.1"
+
+// makeRandomHost creates a libp2p host with a randomly generated identity.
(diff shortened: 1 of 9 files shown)


hsanjuan:
Added http-proxy example

examples/http-proxy/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+		die(err)
+	}
+	return bhost.New(n)
+}
+
+func log(err error) {
+	fmt.Fprintf(os.Stderr, "%s\n", err)
+}
+
+func die(err error) {
```
whyrusleeping: Instead of `die`, we could use `log.Fatalln` from go's log package

examples/http-proxy/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+	hp := strings.Split(req.Host, ":")
+	if len(hp) > 1 && hp[1] == "443" {
+		req.URL.Scheme = "https"
+	} else {
+		req.URL.Scheme = "http"
+	}
+	req.URL.Host = req.Host
+
+	outreq := new(http.Request)
+	*outreq = *req
```
whyrusleeping: you could just do:
```
outreq := *req
```
for the same effect

hsanjuan: Mm yeah fair enough. I copied from https://golang.org/src/net/http/httputil/reverseproxy.go?s=3845:3920#L121 but since we don't pass it around it's ok as you write it.

examples/http-proxy/proxy.go
hsanjuan: But I feel bad when I log errors to stdout :(

whyrusleeping: the log subpackage logs to stderr though, so you don't have to feel bad!

examples/multicodecs/main.go
```diff
@@ -0,0 +1,181 @@
...
+	ps := ps.NewPeerstore()
+	ps.AddPrivKey(pid, priv)
+	ps.AddPubKey(pid, pub)
+	n, _ := swarm.NewNetwork(context.Background(),
+		[]ma.Multiaddr{listen}, pid, ps, nil)
+	return bhost.New(n)
+}
+
+func main() {
+	// Choose random ports between 10000-10100
```
whyrusleeping: You can also just use port 0

examples/http-proxy/proxy.go
hsanjuan: nah, need to pass *http.Request to RoundTrip. I'll leave it as it is, in case someone wants to improve it borrowing more stuff from the `reverseproxy.go`.

examples/multicodecs/main.go
hsanjuan: that's useful but then I have to use more lines to figure out which ephemeral port was picked up...

hsanjuan:
@whyrusleeping changed the logging. Thanks!

whyrusleeping:
The tests are unhappy in my other PR, lets see if theyre happy here or not

examples/http-proxy/proxy.go
whyrusleeping: SGTM

hsanjuan:
@whyrusleeping heads up, tests are veery happy here :)

PR #187, opened 2017-03-17
add first tutorial, showing creation of a host

whyrusleeping:
Started out making a 'libp2p' tutorial from some of the things posted in this issue: https://github.com/ipfs/docs/issues/17#issuecomment-286908929

Feel free to contribute to this PR and make your own.

### DIFF
--- a/tutorial/001-the-host/README.md
+++ b/tutorial/001-the-host/README.md
@@ -4,16 +4,25 @@ started. This guide will show how to construct and use a simple host.
 The host is an abstraction that manages services on top of a swarm. It provides
 a clean interface to connect to a service on a given remote peer.
 
-First, you'll need an ID, and a place to store that ID. To generate a 'test'
+First, you'll need an ID, and a place to store that ID. To generate an
 ID, you can do the following:
 ```go
 import (
-	testutil "github.com/libp2p/go-testutil"
+	"crypto/rand"
+
+	crypto "github.com/libp2p/go-libp2p-crypto"
+	peer "github.com/libp2p/go-libp2p-peer"
 	pstore "github.com/libp2p/go-libp2p-peerstore"
 )
 
-// For toy applications, this is an easy way to get an identity
-ident, err := testutil.RandIdentity()
+// Generate an identity keypair using go's cryptographic randomness source
+priv, pub, err := crypto.GenerateEd25519Key(rand.Reader)
+if err != nil {
+	panic(err)
+}
+
+// A peers ID is the hash of its public key
+pid, err := peer.IDFromPublicKey(pub)
 if err != nil {
 	panic(err)
 }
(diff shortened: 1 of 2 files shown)


tutorial/001-the-host/README.md
```diff
@@ -0,0 +1,79 @@
...
+First, you'll need an ID, and a place to store that ID. To generate a 'test'
+ID, you can do the following:
+```go
+import (
+	testutil "github.com/libp2p/go-testutil"
+	pstore "github.com/libp2p/go-libp2p-peerstore"
+)
+
+// For toy applications, this is an easy way to get an identity
+ident, err := testutil.RandIdentity()
```
hsanjuan: I don't like using `testutil` here because not what you use in reality and it's not helpful that the tutorial does not show how you do it properly. We should point people to https://godoc.org/github.com/libp2p/go-libp2p-crypto#GenerateKeyPair which can generate proper keys and identitity, just as easily, rather than letting the user figure it out when they're not doing "toys" anymore (imho).

tutorial/001-the-host/README.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+Next, you'll need at least one address that you want to listen on. You can go
+from a string to a multiaddr like this:
+```go
+import ma "github.com/multiformats/go-multiaddr"
+
+...
+
+maddr, err := ma.NewMultiaddr("/ip4/0.0.0.0/tcp/9000")
```
hsanjuan: `127.0.0.0` (maybe with a comment saying use 0.0.0.0 if you want to bind on all interfaces)

tutorial/001-the-host/README.md
```diff
@@ -0,0 +1,79 @@
...
+At this point, we have everything needed to finally construct a host. That call
+is the simplest one so far:
+```go
+import bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+
+myhost := bhost.New(netw)
+```
+
+And thats it, you have a libp2p host and you're ready to start doing some
+awesome p2p networking!
```
hsanjuan: Does not hurt to elaborate here saying that in order to speak to someone else you need to:

* Add other peer's  addresss to the peerstore
* Connect() or *NewStream()
* Set a handler on the other peer

Or simply point to the echo example that show how to do it?

tutorial/001-the-host/README.md
whyrusleeping: In this case I *do* want to bind to all interfaces. This is a swarm listener address

tutorial/001-the-host/README.md
whyrusleeping: Yeah, I was thinking of having the next tutorial (writing ping) show how to do all that. Then mention here what will be covered in the next tutorial

tutorial/001-the-host/README.md
whyrusleeping: Thats fair enough

whyrusleeping:
@hsanjuan I updated the tutorial to use a better way of generating the ID

PR #189, opened 2017-04-11
Resolve /dns4, /dns6, /dnsaddr multiaddrs

lgierth:
I worked on [go-multiaddr-dns](https://github.com/multiformats/go-multiaddr-dns) a bit more and I think it's good enough now. Typing docs and dnsaddr specs too.

This will be used for 1) DNS-enabled bootstrap, and 2) configurable announced swarm addresses.

For a first try I picked BasicHost.Connect as the place for integration. Another way I thought about is async resolution from within Peerstore whenever we learn about a new address.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -18,6 +18,7 @@ import (
 	pstore "github.com/libp2p/go-libp2p-peerstore"
 	protocol "github.com/libp2p/go-libp2p-protocol"
 	ma "github.com/multiformats/go-multiaddr"
+	madns "github.com/multiformats/go-multiaddr-dns"
 	msmux "github.com/multiformats/go-multistream"
 )
 
@@ -55,12 +56,13 @@ const NATPortMap Option = iota
 //  * uses an identity service to send + receive node information
 //  * uses a nat service to establish NAT port mappings
 type BasicHost struct {
-	network inet.Network
-	mux     *msmux.MultistreamMuxer
-	ids     *identify.IDService
-	natmgr  NATManager
-	addrs   AddrsFactory
-	cmgr    connmgr.ConnManager
+	network    inet.Network
+	mux        *msmux.MultistreamMuxer
+	ids        *identify.IDService
+	natmgr     NATManager
+	addrs      AddrsFactory
+	maResolver *madns.Resolver
+	cmgr       connmgr.ConnManager
 
 	negtimeout time.Duration
 
@@ -89,11 +91,16 @@ type HostOpts struct {
 	// If omitted, there's no override or filtering, and the results of Addrs and AllAddrs are the same.
 	AddrsFactory AddrsFactory
 
+	// MultiaddrResolves holds the go-multiaddr-dns.Resolver used for resolving
+	// /dns4, /dns6, and /dnsaddr addresses before trying to connect to a peer.
+	MultiaddrResolver *madns.Resolver
+
 	// NATManager takes care of setting NAT port mappings, and discovering external addresses.
 	// If omitted, this will simply be disabled.
 	NATManager NATManager
 
-	//
+	// BandwidthReporter is used for collecting aggregate metrics of the
+	// bandwidth used by various protocols.
 	BandwidthReporter metrics.Reporter
 
 	// ConnManager is a libp2p connection manager
@@ -113,6 +120,7 @@ func NewHost(ctx context.Context, net inet.Network, opts *HostOpts) (*BasicHost,
 		mux:        msmux.NewMultistreamMuxer(),
 		negtimeout: DefaultNegotiationTimeout,
 		addrs:      DefaultAddrsFactory,
+		maResolver: madns.DefaultResolver,
 	}
 
 	if opts.MultistreamMuxer != nil {
@@ -136,6 +144,10 @@ func NewHost(ctx context.Context, net inet.Network, opts *HostOpts) (*BasicHost,
 
 	if opts.NATManager != nil {
 		h.natmgr = opts.NATManager
+	}
+
+	if opts.MultiaddrResolver != nil {
+		h.maResolver = opts.MultiaddrResolver
 	}
 
 	if opts.BandwidthReporter != nil {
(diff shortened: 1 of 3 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -288,14 +293,32 @@ func (h *BasicHost) newStream(ctx context.Context, p peer.ID, pid protocol.ID) (
...
+// given peer.ID. If there is not an active connection, Connect will issue a
+// h.Network.Dial, and block until a connection is open, or an error is returned.
+// Connect will absorb the addresses in pi into its internal peerstore.
+// It will also resolve any /dns4, /dns6, and /dnsaddr addresses.
 func (h *BasicHost) Connect(ctx context.Context, pi pstore.PeerInfo) error {
+	// TODO(lgierth) figure out whether /ipfs/$peerid needs to be stripped
+	addrs := []ma.Multiaddr{}
+	for _, addr := range pi.Addrs {
+		addrs = append(addrs, addr)
+		p2paddr := ma.Join(addr, ma.StringCast("/ipfs/"+pi.ID.Pretty()))
```
lgierth: Is there a nicer utility func for this? I don't wanna spread "/ipfs" all over the place

lgierth:
Mh might need to update go-multiaddr, or I'll just downgrade it in go-multiaddr-dns's deps.

Also bumps the Go requirement to 1.8 for the net.Resolver API, but I can work around that if neccessary.

p2p/host/basic/basic_host.go
hsanjuan: There's no nice way (i think). I use ma.Encapsulate normally. Also, prefer to use `IDB58Encode` to `Pretty` because there's no compromise that `Pretty` will always produce the right output.

hsanjuan:
I'll have a closer look to this next week.

lgierth:
@whyrusleeping could I have just general feedback on the direction I'm taking? Resolve-at-use (this) vs. resolve-at-learn (more complex).

p2p/host/basic/basic_host.go
whyrusleeping: @lgierth theres a peerstore method now I think

lgierth: There's one for parsing -- agreed, it makes sense to add one for construction too.

p2p/host/basic/basic_host.go
```diff
@@ -288,14 +293,32 @@ func (h *BasicHost) newStream(ctx context.Context, p peer.ID, pid protocol.ID) (
...
-// peerstore. If there is not an active connection, Connect will issue a
-// h.Network.Dial, and block until a connection is open, or an error is
-// returned.
+// given peer.ID. If there is not an active connection, Connect will issue a
+// h.Network.Dial, and block until a connection is open, or an error is returned.
+// Connect will absorb the addresses in pi into its internal peerstore.
+// It will also resolve any /dns4, /dns6, and /dnsaddr addresses.
 func (h *BasicHost) Connect(ctx context.Context, pi pstore.PeerInfo) error {
+	// TODO(lgierth) figure out whether /ipfs/$peerid needs to be stripped
+	addrs := []ma.Multiaddr{}
```
whyrusleeping: I would just do `var addrs []ma.Multiaddr`

p2p/host/basic/basic_host.go
lgierth: Taken care of in libp2p/go-libp2p-peerstore#14, all boy scouts like.

lgierth:
This is ready to go now

lgierth:
Just fixed the comments in HostOpts, that's the only change since yesterday

p2p/host/basic/basic_host.go
```diff
@@ -130,6 +138,10 @@ func NewHost(net inet.Network, opts *HostOpts) *BasicHost {
 		h.natmgr = opts.NATManager
 	}
 
+	if opts.MultiaddrResolver != nil {
```
whyrusleeping: probably don't even need the if statement here. its the same effect either way

lgierth: Don't wanna the overwrite the DefaultResolver

p2p/host/basic/basic_host.go
```diff
@@ -377,9 +390,43 @@ func (h *BasicHost) Connect(ctx context.Context, pi pstore.PeerInfo) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var addrs []ma.Multiaddr
+	for _, addr := range pi.Addrs {
+		addrs = append(addrs, addr)
+
+		reqaddr := addr.Encapsulate(p2paddr)
+		resaddrs, err := h.maResolver.Resolve(ctx, reqaddr)
```
vyzo: I think it would better if there was an address match check before blanket attempting to resolve.

Also, it's not clear what is the behaviour of `Resolve` on non dns address. 
For example, will it resolve to an empty address list if it is not a dns address? will it error out? or will it repeat its input?

lgierth: > I think it would better if there was an address match check before blanket attempting to resolve.

Curious, what would try to match-check here?

> or will it repeat its input?

Yes it'll return its input so it basically ends up as a no-op.

```go
func (r *Resolver) Resolve(ctx context.Context, maddr ma.Multiaddr) ([]ma.Multiaddr, error) {
	resolvable, proto := isResolvable(maddr)
	if !resolvable {
		return []ma.Multiaddr{maddr}, nil
        }
```

lgierth: > Curious, what would try to match-check here?

Oh I see, whether the returned address's peerid matches the input peerid -- the resolver takes care of that. It's pretty neat, see https://github.com/multiformats/go-multiaddr-dns readme for examples, in particular the /dnsaddr examples. For /dns4 and /dns6 there's nothing to match-check since these only do A/AAAA lookups.

vyzo: The issue is that if it just returns its input, then the multiaddr will end up being appended twice, and the second time with needless processing. 

So it's not a noop; it both duplicates the input (the address has already been appended) and has overhead on top, as it needs to perform a few operations that eat cpu cycles.

lgierth: Update after convo on IRC: I'll expose the isResolvable func on the resolver as `Matches()`, and then we can simplify the loop here a lot.

lgierth:
@vyzo fixed the dependency issue, I had forgotten that I had gx-go-link'd go-multiaddr-dns locally. Need to think of ways to show warnings in the case that one of the package's dependencies is gx-go-link'd.

Merging once CI is green as you LGTM'd on IRC.

PR #191, opened 2017-04-22
Improve ObservedAddrSet behaviour

Kubuxu:


p2p/protocol/identify/obsaddr.go
```diff
@@ -8,15 +8,32 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+const ActivationTrsh = 4
```
whyrusleeping: s/Trsh/Thresh/  Looks less like 'Trash'

p2p/protocol/identify/obsaddr.go
```diff
@@ -8,15 +8,32 @@ import (
...
+
 // ObservedAddr is an entry for an address reported by our peers.
 // We only use addresses that:
-// - have been observed more than once. (counter symmetric nats)
-// - have been observed recently (10min), because our position in the
+// - have been observed more than 3 times in last 10 min. (counter symmetric nats)
```
whyrusleeping: This one should be more than ten minutes I think. Can probably get away with making it an hour

Kubuxu: I would extend the time of the aspect explained lower. After it was observed and confirmed (4 observations in 10 min duration) keep it as valid address for an hour. How does that sound?

Kubuxu: And it is an hour that is extended every time address is observed by anyone again.

whyrusleeping: I think the four observations should be able to occur over a longer timespan than ten minutes.

p2p/protocol/identify/obsaddr.go
```diff
@@ -4,19 +4,39 @@ import (
 	"sync"
 	"time"
 
-	pstore "github.com/libp2p/go-libp2p-peerstore"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+const (
+	ActivationTresh = 4
```
whyrusleeping: Close, `s/Tresh/Thresh`

p2p/protocol/identify/obsaddr.go
whyrusleeping: update comment

p2p/protocol/identify/obsaddr.go
```diff
@@ -69,7 +80,7 @@ func (oas *ObservedAddrSet) Add(addr ma.Multiaddr, observer ma.Multiaddr) {
 	// for zero-value.
 	if oas.addrs == nil {
 		oas.addrs = make(map[string]*ObservedAddr)
-		oas.ttl = pstore.OwnObservedAddrTTL
+		oas.ttl = OwnObservedAddrTTL
```
whyrusleeping: Sure we want to disconnect this?

whyrusleeping:
also tests are failing

Kubuxu:
Test are not connected. I will fix them in separate PR/push.

p2p/protocol/identify/obsaddr.go
Kubuxu: I switched back and opened. https://github.com/libp2p/go-libp2p-peerstore/pull/13

Kubuxu:
Oh, there shouldn't be hash update here. It slipped in.

PR #196, opened 2017-05-29
basichost: allow overriding Addrs()

lgierth:
This is for ipfs/go-ipfs#3613. I decided to cheat and just sidestep the whole passing-options-through-n-layers-of-node-initialization. The calling code will simply pass a closure as the `AddrsFactory` option which applies the announced/noAnnounced addrs if neccessary.

(I'm open to better names than AddrsFactory.)

TODO:
- [x] tests

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -23,6 +23,10 @@ var log = logging.Logger("basichost")
 
 var NegotiateTimeout = time.Second * 60
 
+// AddrsFactory functions can be passed to New in order to override
+// addresses returned by Addrs.
+type AddrsFactory func([]ma.Multiaddr) []ma.Multiaddr
+
 // Option is a type used to pass in options to the host.
 type Option int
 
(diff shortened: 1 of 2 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -23,6 +23,8 @@ var log = logging.Logger("basichost")
 
 var NegotiateTimeout = time.Second * 60
 
+type AddrsFactory func([]ma.Multiaddr) []ma.Multiaddr
```
hsanjuan: Document this type please. Why it is here? What is it used for and maybe a practical example of how it can be used.

p2p/host/basic/basic_host.go
```diff
@@ -23,6 +23,8 @@ var log = logging.Logger("basichost")
...
+
 // Option is a type used to pass in options to the host.
```
hsanjuan: great opportunity to also improve Option documentations :)

lgierth: I have a bit of work-in-progress code to refactor construction of BasicHost and Swarm, the Option type will be gone afterwards. Separate PR though :)

lgierth:
Okay here we go

p2p/host/basic/basic_host.go
lgierth: I counted and there's at least 4 different ways of passing in various options to host and swarm, including global variables :) -- it's in bad need of an options struct.

PR #197, opened 2017-05-30
basichost: refactor BasicHost construction

lgierth:
- I'd like to do a similar thing with Swarm
- The first commit (allow overriding Addrs) is from an unrelated PR (#196)

---

There were previously 4 different ways of passing various options
into BasicHost construction.

1. Function parameters to New
2. Option parameters to New
3. Global variables, e.g. NegotiateTimeout
4. Options struct, in the calling code in go-ipfs

This changeset deprecates all of these, and introduces the HostOpts
struct to replace them. It also introduces a new constructor called
NewHost, which accepts a *HostOpts.

The old New constructor continues to work the same way,
but is from now on deprecated too.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -21,23 +21,31 @@ import (
 
 var log = logging.Logger("basichost")
 
-var NegotiateTimeout = time.Second * 60
+var (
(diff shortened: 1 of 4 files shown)


lgierth:
Oh, I shall add some tests for NewHost while I'm at it...

PR #199, opened 2017-06-03
Publicize NAT manager struct

enzoh:


lgierth:
Awesome, thanks for taking this on! Could you make `NATManager` an interface and rename the existing struct to something like `NATPortMap`? That'll make it easier to have multiple different NAT implementations in the future.

lgierth:
Thanks, LGTM :+1:

lgierth:
I'll merge and make a release when #189 is merged, should be today

Issue #205, opened 2017-07-18
Routed Host Hangs


ericxtang:
This used to work, but something happened recently that it no longer works.

For example, if you take example/echo.go, add 
```
func makeKadHost(h host.Host) (host.Host, error) {
	dhtRouting := kad.NewDHT(context.Background(), h, ds.NewMapDatastore())
	if err := dhtRouting.Bootstrap(context.Background()); err != nil {
		glog.Errorf("Error bootstraping dht: %v", err)
		return nil, err
	}

	rHost := rhost.Wrap(h, dhtRouting)
	return rHost, nil
}
```
(rhost is github.com/libp2p/go-libp2p/p2p/host/routed)

If we use the KadHost instead of the basicHost for making connections, the connection hangs.  Am I missing something here?

Stebalien:
Could you post a complete repro (or open a PR that reproduces the issue)?

ericxtang:
@Stebalien yeah - made a PR https://github.com/libp2p/go-libp2p/pull/206

Stebalien:
It works for me after importing/rewriting the deps using gx (we use gx to deal with dependency versioning):

```go
gx-go rewrite

# New deps from your patch.
gx import github.com/libp2p/go-libp2p-kad-dht
gx import github.com/ipfs/go-datastore
```

ericxtang:
Woot!  That worked.  Seems like an issue where gx and github has different versions.

Thanks @Stebalien.  

gx looks like a good way to do versioning. Is there a good tutorial for it besides the github repo?  I'm interested in trying it out, but the hashes scare me.  It would be great to have something like a go-path-to-hash lookup command.

Stebalien:
Sorry for the slow response.

I'm not aware of any good tutorials for it and, unfortunately, the hashes are kind of unavoidable due to how go manages dependencies. @whyrusleeping considered using symlinks but go treats packages with different names as different packages even if they have the same code.

Note: You don't need to check the hashes into your code (just into packages.json). Instead, you can just call `gx-go rw` before building.

PR #206, opened 2017-07-21
Code Example for Issue 205

ericxtang:
Reference code for https://github.com/libp2p/go-libp2p/issues/205

Here is the binary for echo, built on my mac.

[echo.zip](https://github.com/libp2p/go-libp2p/files/1164170/echo.zip)

### DIFF
--- a/examples/echo/main.go
+++ b/examples/echo/main.go
@@ -11,14 +11,18 @@ import (
 	"log"
 	mrand "math/rand"
 
+	"github.com/golang/glog"
+	ds "github.com/ipfs/go-datastore"
 	golog "github.com/ipfs/go-log"
 	crypto "github.com/libp2p/go-libp2p-crypto"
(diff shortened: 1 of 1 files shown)


examples/echo/main.go
```diff
@@ -11,14 +11,18 @@ import (
 	"log"
 	mrand "math/rand"
 
+	"github.com/golang/glog"
```
Kubuxu: we use native go logger in this example, no need for custom one

Stebalien: (this was PR is just a demonstration of a bug)

Kubuxu: Oh, sorry. Today I felt like whole day I was on autopilot so I saw a PR and reviewed it.

Stebalien: That's a good autopilot habit to have :smile:.

Stebalien:
This probably doesn't need to stay open.

PR #208, opened 2017-07-21
RoutedHost: NewStream ensures we have the means to connect to the peer

### LINKED ISSUE #207: RoutedHost.NewStream should ensure we have the means to connect to the peer
Currently, RoutedHost.NewStream directly delegates the call to the underlying (basic) host. 
The idea is that if we don't currently have any open connections to the peer, the basic host will automatically open a connection using Dial and the addresses in the Peerstore.

And that's exactly the problem: if we have not previously had a connection to the peer, then we are likely not to have any address for the peer either. So when the dial is attempted, it will not find any addresses and fail.

This issue directly hinders the usage of relay addresses in the form `/ipfs/QmRelay/p2p-circuit/ipfs/QmPeer` and instead forces us to explicitly specify the address of the the relay (or have previously connected to it somehow).

vyzo:
Fixes #207

### DIFF
--- a/p2p/host/routed/routed.go
+++ b/p2p/host/routed/routed.go
@@ -128,13 +128,12 @@ func (rh *RoutedHost) RemoveStreamHandler(pid protocol.ID) {
 }
 
 func (rh *RoutedHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.ID) (inet.Stream, error) {
-	// check if we need to find some addresses for the peer through the routing system
-	if len(rh.Network().ConnsToPeer(p)) == 0 && len(rh.Peerstore().Addrs(p)) == 0 {
-		addrs, err := rh.findPeerAddrs(ctx, p)
-		if err != nil {
-			return nil, err
-		}
-		rh.Peerstore().AddAddrs(p, addrs, pstore.TempAddrTTL)
+	// Ensure we have a connection, with peer addresses resolved by the routing system (#207)
+	// It is not sufficient to let the underlying host connect, it will most likely not have
+	// any addresses for the peer without any prior connections.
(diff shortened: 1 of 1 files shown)


p2p/host/routed/routed.go
```diff
@@ -119,6 +128,15 @@ func (rh *RoutedHost) RemoveStreamHandler(pid protocol.ID) {
...
 func (rh *RoutedHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.ID) (inet.Stream, error) {
+	// check if we need to find some addresses for the peer through the routing system
+	if len(rh.Network().ConnsToPeer(p)) == 0 && len(rh.Peerstore().Addrs(p)) == 0 {
+		addrs, err := rh.findPeerAddrs(ctx, p)
+		if err != nil {
+			return nil, err
+		}
+		rh.Peerstore().AddAddrs(p, addrs, pstore.TempAddrTTL)
+	}
+
```
Stebalien: Is there any reason to not just call `Connect`? We'll end up calling `Connect` on the host again from within `NewStream` but that should be a no-op.

vyzo: I think we can safely use connect too, although it's redundant once we've put the addresses in the Peerstore. 
Note that BasicHost doesn't call connect explicitly; instead the connection is established by `Dial` when it sees that it has no active connections.

vyzo: it's 3 lines -vs- 5 with using Connect, so I am opting for it; needs a good comment though.

vyzo: Updated to use Connect; so yeah, both shorter and easier to understand with the right comment.

Issue #209, opened 2017-07-21
Circuit Relay Integration


vyzo:
The go-libp2p-circut relay code can integrate with any Host that has a swarm network by using the `circuit.AddRelayTransport` function.

This gives us some alternatives to integrating with go-libp2p:
- We can integrate with `BasicHost`, as an option to `NewBasicHost`.
- We can integrate with `RoutedHost`, by modifying the `Wrap` function or adding a new constructor.
- We can keep relay separate and integrate at the application layer (eg go-ipfs can add the relay transport explicitly)

Ideally, we would want to integrate with libp2p so that it's always available, but there is a difference in the two integrations in terms of capabilities of the relay, modulo #207:
- if `RoutedHost.NewStream` can do peer lookup (#208), then it is preferable to integrate with `RoutedHost` so that we can support dialing to unspecific relay addresses and active relay
- If `RoutedHost.NewStream` will not ever do peer address discovery,, then the distinction is moot and we can just integrate through options in the `NewBasicHost` constructor.

Let's discuss integration approach, in terms of code complexity it's all very straightforward.

vyzo:
Summoning @whyrusleeping @lgierth @Stebalien @diasdavid

vyzo:
cc @dryajov

dryajov:
I'm not familiar with the go terminology or codebase, so can't provide any insights into that, but I can provide a quick overview of where circuit integrates in the js code, and hopefully give some more perspective for the go implementation.

So, circuit is bootstrapped by libp2p, and receives the `libp2p-swarm` for dialing and listening. Right now relay is pretty dumb in that it either has an active connection to the dialing host, or if its an active relay it requires a full address to the host being dialed, otherwise it won't be able to dial it since we're bypassing the DHT. One question i've  been meaning to bring up to the group is weather we should allow the `relay` part of circuit use the DHT, which would require passing `libp2p` instead of `swarm` to realy (Dialer/Listener would still use `swarm`). From the js implementation perspective, the interfaces of both `swarm` and `libp2p` are very similar so swapping them is not a big deal at all, not sure about Go land.

[...]

PR #210, opened 2017-07-24
Circuit Relay integration with basic_host

vyzo:
Integrates Circuit Relay as a transport option in basic_host.NewHost.
See #209

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -107,7 +107,7 @@ type HostOpts struct {
 }
 
 // NewHost constructs a new *BasicHost and activates it by attaching its stream and connection handlers to the given inet.Network.
-func NewHost(net inet.Network, opts *HostOpts) (*BasicHost, error) {
+func NewHost(ctx context.Context, net inet.Network, opts *HostOpts) (*BasicHost, error) {
 	h := &BasicHost{
 		network:    net,
 		mux:        msmux.NewMultistreamMuxer(),
@@ -167,7 +167,7 @@ func NewHost(net inet.Network, opts *HostOpts) (*BasicHost, error) {
 	net.SetStreamHandler(h.newStreamHandler)
 
 	if opts.EnableRelay {
-		relayCtx, relayCancel = context.WithCancel(context.Background())
+		relayCtx, relayCancel = context.WithCancel(ctx)
 		err := circuit.AddRelayTransport(relayCtx, h, opts.RelayOpts...)
 		if err != nil {
 			h.Close()
@@ -200,7 +200,7 @@ func New(net inet.Network, opts ...interface{}) *BasicHost {
 		}
 	}
 
-	h, err := NewHost(net, hostopts)
+	h, err := NewHost(context.Background(), net, hostopts)
 	if err != nil {
 		// this cannot happen with legacy options
 		// plus we want to keep the (deprecated) legacy interface unchanged
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -107,7 +107,7 @@ type HostOpts struct {
 }
 
 // NewHost constructs a new *BasicHost and activates it by attaching its stream and connection handlers to the given inet.Network.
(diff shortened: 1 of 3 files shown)


vyzo:
summoning @Stebalien @whyrusleeping

vyzo:
cc @lgierth

p2p/host/basic/basic_host.go
```diff
@@ -143,10 +150,25 @@ func NewHost(net inet.Network, opts *HostOpts) *BasicHost {
...
 
+	var relayCtx context.Context
+	var relayCancel func()
+	if opts.Relay {
+		relayCtx, relayCancel = context.WithCancel(context.Background())
+		err := circuit.AddRelayTransport(relayCtx, h, opts.RelayOpts...)
+		if err != nil {
+			// perhaps inappropriate, but otherwise we have to change the interface
+			// to return an error, which will nost likely lead to a fatality anyway
+			panic(err)
```
whyrusleeping: Hrm... this construction feels very entangled. we need a host to create a relay that we're using to create a host? Maybe we should set the relay post construction?

whyrusleeping: (also, panics are bad, if we need to lets just switch to returning an error here)

vyzo: Yeah, relay needs a pre-existing host to attach.
Maybe we can leave it outside the basic host construction if you prefer, although it would be nice to have it simply available through the host options.

Re: panic: I can change the interface to return an error.

vyzo: I think we want to keep the legacy interface unchanged however.

vyzo:
@whyrusleeping updated the `NewHost` interface to return an error, so that we don't panic there.

I did not change the legacy `New` constructor, as it would be an interface breaking change -- plus it's a deprecated interface! 
I added a panic there however for handing the `NewHost` error, but it should be impossible with the legacy options the interface supports.

vyzo:
rebased on master and updated go-libp2p-circuit dep to 1.1.3 to keep up with upstream deps.

p2p/host/basic/basic_host.go
```diff
@@ -143,17 +150,30 @@ func NewHost(net inet.Network, opts *HostOpts) *BasicHost {
 		h.cmgr = opts.ConnManager
 	}
 
+	var relayCtx context.Context
+	var relayCancel func()
+	if opts.Relay {
+		relayCtx, relayCancel = context.WithCancel(context.Background())
```
lgierth: I think this context is off -- it's not wired to anything, but is used to dial peers, open streams, and open relayed streams. It should be wired to *something* so it correctly gets cancelled when shutting down go-ipfs.

vyzo: It's loosely wired via the cancel funciton. When the basic host gets closed, it will call cancel.

Regardless, I agree that ideally this context should come from the outside and be tied to the lifetime of the host. It would take an interface change to pass it as a `NewHost` argument though, so that's why I didn't do that initially.

But we _are_ changing the interface already for the error return. It's not widely used so we might as well change it further to take a context argument. 
The legacy interface can pass in a context.Background().

p2p/host/basic/basic_host.go
vyzo: I moved the relay attachment post construction in https://github.com/libp2p/go-libp2p/pull/210/commits/fd23cf6fa544ff10d8070532555eeaaff4eb9ae9

p2p/host/basic/basic_host.go
vyzo: Added base context as parameter to `NewHost` in https://github.com/libp2p/go-libp2p/pull/210/commits/6be81d34c1016ec8b2ac9143893a1908ae997c08

p2p/host/basic/basic_host.go
vyzo: The inappropriate panic was squashed in https://github.com/libp2p/go-libp2p/pull/210/commits/7f1ffcbd4e69394b9babb5dead5b940c1aaca547

PR #211, opened 2017-07-27
tests: update test cases to act like go-peerstream 1.7.0

Stebalien:
this makes the test cases a bit more accurate.


lgierth:
Thanks :)

PR #216, opened 2017-08-19
Summer cleaning

diasdavid:


diasdavid:
image

TODO
```diff
@@ -1 +1 @@
-- pull mdns out as its own separate package
+- [ ] pull mdns out as its own separate package
```
Stebalien: I'd just remove this file and file an issue.

examples/http-proxy/README.md
```diff
@@ -46,7 +43,7 @@ libp2p-peer addresses:
 The run the local peer, indicating that it will need to forward http requests to the remote peer as follows:
 
 ```
-$ ./http-proxy -d /ip4/127.0.0.1/tcp/12000/ipfs/QmddTrQXhA9AkCpXPTkcY7e22NK73TwkUms3a44DhTKJTD
+> ./http-proxy -d /ip4/127.0.0.1/tcp/12000/ipfs/QmddTrQXhA9AkCpXPTkcY7e22NK73TwkUms3a44DhTKJTD
```
Stebalien: Any particular reason for this change? Historically, `$` has meant "prompt where user is not root" and `#` has meant "prompt where user is root".

README.md
```diff
@@ -1,47 +1,80 @@
...
-# Contribute
 
-libp2p implementation in Go is a work in progress. As such, there's a few things you can do right now to help out:
- - Go through the modules below and **check out existing issues**. This would be especially useful for modules in active development. Some knowledge of IPFS/libp2p may be required, as well as the infrasture behind it - for instance, you may need to read up on p2p and more complex operations like muxing to be able to help technically.
- - **Perform code reviews**.
- - **Add tests**. There can never be enough tests.
+## Bundles
 
-# Usage
+There is currently only one bundle of `go-libp2p`, this package. This bundle is used by [`go-ipfs`](https://github.com/ipfs/go-ipfs).
```
diasdavid: We mention bundles on the website and the JavaScript version. I know that right now everything gets a bit hardcoded in go-libp2p, but ideally we would be able to create multiple bundles with different features easily (think different runtimes)

README.md
```diff
@@ -53,15 +86,67 @@ $ make deps
...
+### Packages
+
+List of packages currently in existence for libp2p
 
-## Links
-- [**Specs**](https://github.com/libp2p/specs)
-- [**Website**](https://github.com/libp2p/website)
+| Package            | Version | CI                  |
+|--------------------|---------|---------------------|
+| **Transports**                                     |
```
diasdavid: It is indeed

examples/http-proxy/README.md
diasdavid: There was a conversation a long time ago between Juan and Richard where they questioned what was the most valid character to represent the prompt and the result of the conversation was `>`.

As for `#`, we use it everywhere as comments since it is the character for line comments in bash.

diasdavid:
Thanks for the review @Stebalien :)

examples/http-proxy/README.md
Stebalien: (:roll_eyes:) Unfortunately, that argument kind of makes sense. `$` for variables, `#` for comments, `>` for prompts.


---

But really, 

```bash
𝝺: echo "there can be only one"
```

diasdavid: ahaha :)

can this get merged then?

diasdavid:
Can this be merged? :)

README.md
```diff
@@ -53,15 +86,67 @@ $ make deps
...
+
+Before commiting and pushing to Github, make sure to rewind the gx'ify of dependencies. You can do that with:
+
+```sh
+> make publish
+```
+
+### Tests
+
+Running of individual tests is done through `go test <path to test>`
```
Stebalien: I'd tell users to call `gx test` here as well.

README.md
Stebalien: Can we add a "(WIP)" marker here. Otherwise, this is a bit confusing.

README.md
```diff
@@ -50,18 +83,70 @@ Examples can be found on the [examples folder](examples).
...
+| Package            | Version | CI                  |
+|--------------------|---------|---------------------|
+| **Transports**                                     |
+| **Connection Upgrades**                            |
+| **Stream Muxers**                                  |
+| **Discovery**                                      |
+| **Crypto Channels**                                |
+| **Peer Routing**                                   |
+| **Content Routing**                                |
+| **Generics**                                       |
```
Stebalien: What do you mean by "generics". To me, "generics" usually means something like Java generics but go doesn't have those...

diasdavid: fair point, specially with all the discussion about generics in go land. Renamed to Miscellaneous, it is the packages that don't fit a specific category.

Stebalien:
:sparkling_heart:

PR #228, opened 2017-09-08
Protocol Multiplexing Example: Fixing build instruction so copy and pasting from the Readme works

ssikdar1:
As is go build ./examples/protocol-multiplexing-with-multicodecs
will make the binary name protocol-multiplexing-with-multicodecs
so the next command ./multicodecs won't run
using -o will name the outputfile


Stebalien:
LGTM. Thanks!

PR #233, opened 2017-09-24
allow tag to be passed into NewMdnsService

zippy:
The current implementation uses the constant ServiceTag which is set to an ipfs specific value.  As a generalized p2p library this value should be settable by the caller to NewMdnsService.

### DIFF
--- a/p2p/discovery/mdns.go
+++ b/p2p/discovery/mdns.go
@@ -37,6 +37,7 @@ type mdnsService struct {
 	server  *mdns.Server
 	service *mdns.MDNSService
 	host    host.Host
+	tag     string
 
 	lk       sync.Mutex
 	notifees []Notifee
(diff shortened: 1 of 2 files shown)


zippy:
Hey, thanks for merging.  Currently my use of libp2p seems to be governed by the gx version.  So I don't actually have access to this change until there's a gx bump.  What's your process around that, or perhaps am I doing something wrong on my end in the use of the repo?

whyrusleeping:
Hey @zippy, I can run a gx publish of this for you so you can update it within your package.

whyrusleeping:
Alright, published go-libp2p 5.0.3: https://github.com/libp2p/go-libp2p/blob/master/.gx/lastpubver

In your application that is importing go-libp2p via gx, you will need to run `gx update github.com/libp2p/go-libp2p`. Let me know if you run into any issues, I'm always very interested to see others using gx (and am acutely aware that it needs a good deal of UX work)

PR #241, opened 2017-11-04
first hack at a nice libp2p constructor

whyrusleeping:
Trying to make using go-libp2p as simple as possible

### DIFF
--- a/libp2p.go
+++ b/libp2p.go
@@ -22,13 +22,14 @@ import (
 
 // Config describes a set of settings for a libp2p node
 type Config struct {
-	Transports  []transport.Transport
-	Muxer       mux.Transport
-	ListenAddrs []ma.Multiaddr
-	PeerKey     crypto.PrivKey
-	Peerstore   pstore.Peerstore
-	Protector   pnet.Protector
-	Reporter    metrics.Reporter
+	Transports   []transport.Transport
+	Muxer        mux.Transport
+	ListenAddrs  []ma.Multiaddr
+	PeerKey      crypto.PrivKey
+	Peerstore    pstore.Peerstore
+	Protector    pnet.Protector
+	Reporter     metrics.Reporter
+	DisableSecio bool
 }
 
 func New(ctx context.Context) (host.Host, error) {
@@ -63,8 +64,11 @@ func NewWithCfg(ctx context.Context, cfg *Config) (host.Host, error) {
 		ps = pstore.NewPeerstore()
 	}
 
-	ps.AddPrivKey(pid, cfg.PeerKey)
-	ps.AddPubKey(pid, cfg.PeerKey.GetPublic())
+	// If secio is disabled, don't add our private key to the peerstore
+	if !cfg.DisableSecio {
+		ps.AddPrivKey(pid, cfg.PeerKey)
+		ps.AddPubKey(pid, cfg.PeerKey.GetPublic())
+	}
 
 	swrm, err := swarm.NewSwarmWithProtector(ctx, cfg.ListenAddrs, pid, ps, cfg.Protector, cfg.Muxer, cfg.Reporter)
 	if err != nil {
--- a/libp2p.go
+++ b/libp2p.go
@@ -3,6 +3,7 @@ package libp2p
 import (
 	"context"
 	"crypto/rand"
+	"fmt"
 
 	crypto "github.com/libp2p/go-libp2p-crypto"
 	host "github.com/libp2p/go-libp2p-host"
@@ -34,9 +35,9 @@ type Config struct {
 
 type Option func(cfg *Config) error
 
-func Transports(tpts []transport.Transport) Option {
+func Transports(tpts ...transport.Transport) Option {
 	return func(cfg *Config) error {
-		cfg.Transports = tpts
+		cfg.Transports = append(cfg.Transports, tpts...)
 		return nil
 	}
 }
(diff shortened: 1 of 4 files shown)


libp2p.go
```diff
@@ -0,0 +1,90 @@
...
+	// Set up stream multiplexer
+	tpt := msmux.NewBlankTransport()
+	tpt.AddTransport("/yamux/1.0.0", yamux.DefaultTransport)
+	tpt.AddTransport("/mplex/6.3.0", mplex.DefaultTransport)
+	return tpt
+}
+
+func DefaultConfig() *Config {
+	// Create a multiaddress
+	addr, err := ma.NewMultiaddr("/ip4/127.0.0.1/tcp/0")
```
Stebalien: Should this not be `0.0.0.0`? Maybe, to just accept the address as a parameter?

libp2p.go
```diff
@@ -0,0 +1,90 @@
...
+	transport "github.com/libp2p/go-libp2p-transport"
+	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+	mux "github.com/libp2p/go-stream-muxer"
+	ma "github.com/multiformats/go-multiaddr"
+	mplex "github.com/whyrusleeping/go-smux-multiplex"
+	msmux "github.com/whyrusleeping/go-smux-multistream"
+	yamux "github.com/whyrusleeping/go-smux-yamux"
+)
+
+type Config struct {
```
Stebalien: It would be kind of nice if this (or some subset with the key etc.) could be serialized/deserialized easily.

libp2p.go
```diff
@@ -0,0 +1,90 @@
...
+	Transports  []transport.Transport
+	Muxer       mux.Transport
+	ListenAddrs []ma.Multiaddr
+	PeerKey     crypto.PrivKey
+	Peerstore   pstore.Peerstore
+	Protector   pnet.Protector
+	Reporter    metrics.Reporter
+}
+
+func Construct(ctx context.Context, cfg *Config) (host.Host, error) {
```
dignifiedquire: I wonder if having New(ctx) and NewWithConfig(ctx, cfg) would be nicer. At least for me as a new user it would be more straightforward to understand what those methods do anf how they work

libp2p.go
whyrusleeping: Yeah... that would be nice. Definitely something that can be done later though, As i think it deserves a bit more thought.

libp2p.go
```diff
@@ -0,0 +1,102 @@
...
+	netw := (*swarm.Network)(swrm)
+
+	return bhost.New(netw), nil
+}
+
+func DefaultMuxer() mux.Transport {
+	// Set up stream multiplexer
+	tpt := msmux.NewBlankTransport()
+
+	// By default, support yamux and multiplex
```
hsanjuan: It strikes me that here the supported transports are slightly different than here: https://github.com/libp2p/go-libp2p-swarm/blob/master/swarm.go

I thought that `swarm.NewNetworkWithProtector` would be the *default* way of doing things so any updates to that should happen in `swarm`?

At first sight, it seems that this function is just re-doing what PSTransport provides:

```
// PSTransport is the default peerstream transport that will be used by
// any libp2p swarms.
var PSTransport pst.Transport
```

libp2p.go
hsanjuan: I feel very uneasy about having defaults which open your ports to the world. AWS warns about it, systemd gets shitstorms about it, and anyone building something on top should probably not use the default anyway.

Stebalien: > Anyone building something on top should probably not use the default anyway.

Then, IMO, it should be called "testing" or something like that. In this case, you're setting up a service that's *supposed* to be accessible from the internet. Users will be very confused if it doesn't listen on a public address by default. However, IMO, using a default address probably isn't the best idea anyways as there's no reasonable default port.

libp2p.go
whyrusleeping: Good catch, I'm actually thinking the best path forward here is to remove that logic from the swarm code, and make this place be the default. The reason being that swarm is much deeper in the dependency tree, and making it depend on extra packages makes it harder to update those packages.

hsanjuan: That sounds good to me 👍

whyrusleeping:
I want to try and pull some wisdom from this blog post into this PR: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

whyrusleeping:
I'm just not sure what to make the functions operate on

dignifiedquire:
> I'm just not sure what to make the functions operate on

From my understanding this would be something like this

```go
// current version in the example
basicHost, err := libp2p.NewWithCfg(ctx, &libp2p.Config{
	ListenAddrs: []ma.Multiaddr{addr},
})

// functional version 
basicHost, err := libp2p.New(ctx, libp2p.ListenAddrs(addr))

// functional config - default config
basicHost, err := libp2p.New(ctx)
```

with impl being something like this

```go
func ListenAddrs(addrs ...ma.Multiaddr) func (*Config) error {
  return func (c *Config) error {
    c.ListenAddrs = addrs
  }
}

func New(ctx context.Context, options ...func(*Config) error) (host, error) {
  cfg := Config{}
  for _, option := range options {
    err := option(cfg)
    // handle err
  }
  return NewWithCfg(ctx, cfg)
}
```

whyrusleeping:
@dignifiedquire yeah, having things operate on the config seems like the most reasonable way forward. I'll start on that.

What do you think about still having the option to use the config directly? I'm not really sure because it adds extra things for the user to have to think about.

dignifiedquire:
> What do you think about still having the option to use the config directly

Yeah I would drop that option for users to avoid confusion. We can use `NewWithCfg` and make it private, but having it exposed would be just confusing for users

whyrusleeping:
There, how does that look?

libp2p.go
```diff
@@ -0,0 +1,181 @@
...
+	PeerKey      crypto.PrivKey
+	Peerstore    pstore.Peerstore
+	Protector    pnet.Protector
+	Reporter     metrics.Reporter
+	DisableSecio bool
+}
+
+type Option func(cfg *Config) error
+
+func Transports(tpts []transport.Transport) Option {
```
dignifiedquire: The function signatures are inconsistent, would be nice if they all either take an array or varargs, as currently two of them (Transports and ListenAddrs)  take an array but ListenAddrStrings takes varargs.

libp2p.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func ListenAddrs(addrs []ma.Multiaddr) Option {
+	return func(cfg *Config) error {
+		cfg.ListenAddrs = append(cfg.ListenAddrs, addrs...)
+		return nil
+	}
+}
+
+func NoSecio(cfg *Config) error {
```
dignifiedquire: I think it would be good to have a `Secio` or `WithSecio` one as well, such that you can make it explicit to always use it, independent of the defaults

libp2p.go
```diff
@@ -0,0 +1,181 @@
...
+	cfg.DisableSecio = true
+	return nil
+}
+
+func WithMuxer(m mux.Transport) Option {
```
dignifiedquire: why are these called `With<>` and `Transports` not `WithTransports`?

lgierth:
You might like my ideas for this constructor API, but they're pretty different.
Generally I'd like to be closer js-libp2p, and "leak" fewer of swarm's/host's
details across the surface (in particular we should *not* hardcode secio in
the interface).

Most of the option functions can be simplified as loaders for certain kinds of
modules (as in libp2p modules): transport, muxer, encryption, private networking,
discovery. At a later point, when we have transport capabilities, many of them
can simply be collapsed into the transport module category.

- `Identity(privkey)`
- `ListenAddrs(string...)`
- `AnnounceAddrs(string...)`
- `Transports(tpt.Transport...)`
- `Muxer(msmux.Transport...)`
- `Encryption(...)` / `NoEncryption()`
- `PrivateNetworking(protector)`
- `Discovery(disc.Discovery...)` (the discovery interface is still todo in go-libp2p)
- As for the naming pattern, e.g. WithFoo, FooOpt, etc., I don't have a strong opinion at the moment.

About the default config: why not make it a separate package? We can have
packages for various configurations, while go-libp2p itself depends "mostly"
on interfaces. (The muxers should keep being configurable right here.)

```go
import (
    libp2p "github.com/libp2p/go-libp2p"
    cfg "github.com/libp2p/go-libp2p-default"
)

func main() {
  h, err := libp2p.New(context.TODO(), cfg.Config()...)
}
```

---

Also at a later point, I'd like to be able to construct a node simply by
passing a bunch of protocol identifiers, e.g.:
```go
StrTransports("/ws/1.0.0", "/secio/1.0.0", "/mplex/6.3.0")
StrDiscovery("/mdns/1.0.0", "/udp/1.0.0")
// libp2p modules register themselves similar to how multiaddr protocols do it.
```
But that's really not neccessary at the moment.

libp2p.go
whyrusleeping: Because we overload the term `Transports`. Transports also applies to things like tcp, quic, websockets, etc

libp2p.go
whyrusleeping: I may have been a little tired writing this... >.>

whyrusleeping:
@lgierth there, updated with some of your input

hsanjuan:
Is this go-ish? I don't think I've seen these constructs in Go before. I find it innovative, kind of javascrip-ty.

It's fancy, I might like it, but I find the NewWithCfg initialization way more natural and straightforward in the end (and require less code). Perhaps I'm old fashioned... but let's keep that method exported in any case.

hsanjuan:
oh, just saw the dave cheney post, I guess it's go-ish then :)

whyrusleeping:
alright, gonna merge this for now. The next step is to try out using it in go-ipfs to construct our libp2p node.

PR #243, opened 2017-11-15
Announce and discover peers over simple UDP broadcasts

ehmry:
This patch adds a `broadcastService` that broadcasts the local peer ID and listens for IDs from remote peers. IP broadcasting is limited to IPv4, but is simplier than multicasting and only intended for local area networks.

### DIFF
--- a/p2p/discovery/broadcast.go
+++ b/p2p/discovery/broadcast.go
@@ -1,0 +1,187 @@
+package discovery
+
+import (
+	"context"
+	"math/rand"
+	"net"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/libp2p/go-libp2p-host"
+	"github.com/libp2p/go-libp2p-peer"
+	pstore "github.com/libp2p/go-libp2p-peerstore"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr-net"
+)
+
+type broadcastService struct {
+	host          host.Host
+	listenSock    *net.UDPConn   // Socket listening for broadcasts
+	sendSocks     []*net.UDPConn // Sockets that send broadcasts
+	lk            sync.Mutex
+	notifees      []Notifee
+	broadcastPort int
+	active        bool
+}
+
+//
+// Bind a socket to send and recv broadcasts
+//
+func bindBroadcast(ip net.IP, port int) (*net.UDPConn, error) {
+	conn, err := net.ListenUDP("udp4", &net.UDPAddr{IP: ip, Port: port})
+	if err == nil {
+		file, err := conn.File()
+		if err == nil {
+			err = syscall.SetsockoptInt(int(file.Fd()), syscall.SOL_SOCKET, syscall.SO_BROADCAST, 1)
+		}
+	}
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+func NewBroadcastService(ctx context.Context, peerhost host.Host, defaultPort int, interval time.Duration) (Service, error) {
+
+	// Broadcasts can only be received from 0.0.0.0 or 255.255.255.255
+	// but they can not be sent from 255.255.255.255.
+	// Therefor one listening socket is bound to 255.255.255.255 and a
+	// sending socket will be bound to each address that broadcasts
+	// should be sent from.
+	listenSock, err := bindBroadcast(net.IPv4bcast, defaultPort)
+	if err != nil {
+		return nil, err
+	}
+
+	bs := &broadcastService{
+		host:          peerhost,
+		listenSock:    listenSock,
+		broadcastPort: defaultPort,
+		active:        true,
+	}
+
+	addrs, err := getDialableListenAddrs(peerhost)
+	if err != nil {
+		bs.Close()
+		return nil, err
+	} else {
+		for _, a := range addrs {
+			if a.IP.IsGlobalUnicast() {
+				// binding to the same address and port as the TCP service
+				// embeds that address and port into broadcast messages
+				otherSock, err := bindBroadcast(a.IP, a.Port)
+				if err == nil {
+					bs.sendSocks = append(bs.sendSocks, otherSock)
+				}
+			}
+		}
+	}
+
+	// fudge the interval to dampen oscillations with other nodes
+	margin := int(interval) >> 2
+	interval += time.Duration(rand.Intn(margin))
+	interval -= time.Duration(rand.Intn(margin))
+
+	go bs.sendBroadcasts(ctx, interval)
+	go bs.recvBroadcasts(ctx)
+
+	return bs, nil
+}
+
+func (bs *broadcastService) Close() error {
+	bs.active = false
+	bs.listenSock.Close()
+	for _, conn := range bs.sendSocks {
+		conn.Close()
+	}
+	return nil
+}
+
+func (bs *broadcastService) handleBroadcast(addr *net.UDPAddr, bpeer peer.ID) {
+	// Broadcast messages are UDP, but represent IPv4 TCP sockets
+	maddr, err := manet.FromNetAddr(&net.TCPAddr{
+		IP:   addr.IP,
+		Port: addr.Port,
+	})
+	if err == nil {
+		pi := pstore.PeerInfo{
+			ID:    bpeer,
+			Addrs: []ma.Multiaddr{maddr},
+		}
+
+		bs.lk.Lock()
+		defer bs.lk.Unlock()
+		for _, n := range bs.notifees {
+			if n != nil {
+				go n.HandlePeerFound(pi)
+			}
+		}
+	}
+}
+
+func (bs *broadcastService) recvBroadcasts(ctx context.Context) {
+	var buf = make([]byte, 48)
+	for bs.active {
+		n, rAddr, err := bs.listenSock.ReadFromUDP(buf)
+		if err == nil {
+			if n != 0 {
+				rPeer, err := peer.IDB58Decode(string(buf[0:n]))
+				if err == nil && rPeer != bs.host.ID() {
+					bs.handleBroadcast(rAddr, rPeer)
+				}
+			}
+		}
+	}
+}
+
+func (bs *broadcastService) sendBroadcast() {
+	idBuf := []byte(bs.host.ID().Pretty())
+	broadcastAddr := net.UDPAddr{
+		IP:   net.IPv4bcast,
+		Port: bs.broadcastPort,
+	}
+	for _, conn := range bs.sendSocks {
+		conn.WriteToUDP(idBuf, &broadcastAddr)
+	}
+}
+
+func (bs *broadcastService) sendBroadcasts(ctx context.Context, interval time.Duration) {
+	// send an initial broadcast after a quarter interval
+	time.Sleep(interval >> 2)
+	bs.sendBroadcast()
+
+	// send broadcasts on a periodic timeout
+	ticker := time.NewTicker(interval)
+	for bs.active {
+		select {
+		case <-ticker.C:
+			bs.sendBroadcast()
+
+		case <-ctx.Done():
+			bs.Close()
+		}
+	}
+}
+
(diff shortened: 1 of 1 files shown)


p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,181 @@
...
+					bs.sendSocks = append(bs.sendSocks, otherSock)
+				}
+			}
+		}
+	}
+
+	// fudge the interval to dampen oscillations with other nodes
+	margin := int(interval) >> 2
+	interval += time.Duration(rand.Intn(margin))
+	interval -= time.Duration(rand.Intn(margin))
```
Stebalien: This looks a bit funky. Any reason not to jus to just add a smaller random delay to the interval (rather than add and then subtract one)?

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,181 @@
...
+		pi := pstore.PeerInfo{
+			ID:    bpeer,
+			Addrs: []ma.Multiaddr{maddr},
+		}
+
+		bs.lk.Lock()
+		for _, n := range bs.notifees {
+			go n.HandlePeerFound(pi)
+		}
+		bs.lk.Unlock()
```
Stebalien: Best to defer this call instead.

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,181 @@
...
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (m *broadcastService) RegisterNotifee(n Notifee) {
+	m.lk.Lock()
+	m.notifees = append(m.notifees, n)
+	m.lk.Unlock()
```
Stebalien: Again, we generally defer unlocks when possible.

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,181 @@
...
+	for i, notif := range m.notifees {
+		if notif == n {
+			found = i
+			break
+		}
+	}
+	if found != -1 {
+		m.notifees = append(m.notifees[:found], m.notifees[found+1:]...)
+	}
+	m.lk.Unlock()
```
Stebalien: Defer

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,181 @@
...
+	m.lk.Lock()
+	found := -1
+	for i, notif := range m.notifees {
+		if notif == n {
+			found = i
+			break
+		}
+	}
+	if found != -1 {
+		m.notifees = append(m.notifees[:found], m.notifees[found+1:]...)
```
Stebalien: Since we don't really care about order, you can just swap the entry you want to remove for the last entry. Also, remember to null out the last entry, otherwise, it won't be garbage collected.

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,181 @@
...
+	m.notifees = append(m.notifees, n)
+	m.lk.Unlock()
+}
+
+func (m *broadcastService) UnregisterNotifee(n Notifee) {
+	m.lk.Lock()
+	found := -1
+	for i, notif := range m.notifees {
+		if notif == n {
+			found = i
```
Stebalien: Cleaner to just put the removal code inside here and then return (assuming a deferred unlock).

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,181 @@
...
+	bs.sendBroadcast()
+
+	// send broadcasts on a periodic timeout
+	ticker := time.NewTicker(interval)
+	for {
+		select {
+		case <-ticker.C:
+			bs.sendBroadcast()
+
+		case <-ctx.Done():
```
Stebalien: If we're going to make canceling the context equivalent to calling close, we should call close here. Actually, we should defer a call to close so it gets called regardless.

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,181 @@
...
+		bs.lk.Lock()
+		for _, n := range bs.notifees {
+			go n.HandlePeerFound(pi)
+		}
+		bs.lk.Unlock()
+	}
+}
+
+func (bs *broadcastService) recvBroadcasts(ctx context.Context) {
+	var buf = make([]byte, 48)
```
Stebalien: There are no exit conditions here.

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,181 @@
...
+	interval -= time.Duration(rand.Intn(margin))
+
+	go bs.sendBroadcasts(ctx, interval)
+	go bs.recvBroadcasts(ctx)
+
+	return bs, nil
+}
+
+func (bs *broadcastService) Close() error {
+	bs.listenSock.Close()
```
Stebalien: This should also cancel the context (which means you'll need to make the context cancelable and store the cancel function). I'm not really a fan of using contexts in this way but we're already doing it for the other broadcast services so oh well...

Kubuxu:
I was working on similar feature for cjdns but explicitly IPv6, it allows for better granularity but more over it works on mobile. I think if we are to add something like that we are better off using IPv6.

ehmry:
FWIW I haven't written go for few years so I'm open to any review.

@Kubuxu I had considered trying to do this with multicast and the 'all-systems' address, which is supported for IPv6 as well, but multicast seems overly complicated. My intention here is simply discovering nodes within a small physical network so for the sake of simplicity IPv4 only seems justified.

My immediate use case is actually to discover IPFS nodes that are broadcasting on port 4001 and then connect to them on 5001 as an API client.

p2p/discovery/broadcast.go
ehmry: I add and subtract so the the interval can jitter by a positive or negative amount. And I like getting funky.

ehmry:
@Stebalien I've updated the PR following your recommendations, thanks.

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,187 @@
...
+	return nil
+}
+
+func (bs *broadcastService) handleBroadcast(addr *net.UDPAddr, bpeer peer.ID) {
+	// Broadcast messages are UDP, but represent IPv4 TCP sockets
+	maddr, err := manet.FromNetAddr(&net.TCPAddr{
+		IP:   addr.IP,
+		Port: addr.Port,
+	})
+	if err == nil {
```
Stebalien: Stylistic but... you can reduce the indentation level by using:

```go
if error != nil {
    return
}
// ...
```

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func (bs *broadcastService) sendBroadcasts(ctx context.Context, interval time.Duration) {
+	// send an initial broadcast after a quarter interval
+	time.Sleep(interval >> 2)
+	bs.sendBroadcast()
+
+	// send broadcasts on a periodic timeout
+	ticker := time.NewTicker(interval)
+	for bs.active {
```
Stebalien: Again, can't read/write to a variable from different threads without using some synchronization mechanism (lock, channel, etc.).

Really, I'd just get rid of this variable and use the context.

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	m.notifees = append(m.notifees, n)
+}
+
+func (m *broadcastService) UnregisterNotifee(n Notifee) {
+	m.lk.Lock()
+	defer m.lk.Unlock()
+	for i, notif := range m.notifees {
+		if notif == n {
+			m.notifees[i] = nil
```
Stebalien: Having an array filled with nils is probably not the best way to go. Instead, use https://github.com/golang/go/wiki/SliceTricks#delete-without-preserving-order. That will simplify `RegisterNotifee` and `handleBroadcast` as well.

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,187 @@
...
+				go n.HandlePeerFound(pi)
+			}
+		}
+	}
+}
+
+func (bs *broadcastService) recvBroadcasts(ctx context.Context) {
+	var buf = make([]byte, 48)
+	for bs.active {
+		n, rAddr, err := bs.listenSock.ReadFromUDP(buf)
```
Stebalien: I'm pretty sure most errors from this function are fatal. That is, we can't continue listening on the socket (we'd end up in a rapid, infinite loop). It's probably better to log the error, stop the service, and return.

You could also try checking to see if the error is temporary (cast to `net.Error` and check if `err.Temporary()` is true).

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,187 @@
...
+	broadcastPort int
+	active        bool
+}
+
+//
+// Bind a socket to send and recv broadcasts
+//
+func bindBroadcast(ip net.IP, port int) (*net.UDPConn, error) {
+	conn, err := net.ListenUDP("udp4", &net.UDPAddr{IP: ip, Port: port})
+	if err == nil {
```
Stebalien: The go idiom is to (almost) always write:

```go
if err != nil {
    return nil, err
}
```

That helps prevent nesting/confusing control flow.

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,187 @@
...
+			if a.IP.IsGlobalUnicast() {
+				// binding to the same address and port as the TCP service
+				// embeds that address and port into broadcast messages
+				otherSock, err := bindBroadcast(a.IP, a.Port)
+				if err == nil {
+					bs.sendSocks = append(bs.sendSocks, otherSock)
+				}
+			}
+		}
+	}
```
Stebalien: We may want to return an error if we fail to bind to *anything*. I wouldn't stop IPFS from starting in that case but, if we return an error and *log* it, it'll make debugging easier.

ehmry: When I looked at the mdns file I saw three different methods of logging errors, so I thought it cleaner not to log at all.

p2p/discovery/broadcast.go
ehmry: It's a bool.

p2p/discovery/broadcast.go
ehmry: I would argue that typing the same early return idiom over and over and over is worse for understanding control flow.

p2p/discovery/broadcast.go
Stebalien: We usually use `github.com/ipfs/go-log`. Any other methods of logging should be fixed. However, I'd just suggest returning an error and logging it from go-ipfs.

p2p/discovery/broadcast.go
Stebalien: > It's a bool.

I know, I'm referring to https://golang.org/ref/mem. For example, given:

```go
package main

func main() {
	var active = true

	go func() {
		active = false
	}()

	for active {
	}
}
```

Go may choose to allow the second go routine to loop forever (optimizing out the check of `active`) because no synchronization primitives were used. Try running the above code with the race detector.

p2p/discovery/broadcast.go
Stebalien: It's just how go is written. However, I'd disagree:

1. It makes it very easy to recognize where errors are being returned.
2. It reduces indentation.
3. It's consistent with all other go code.

Now, one can argue that there should be a *better* way to return errors in go but that's a separate issue (and not something we can do anything about).

p2p/discovery/broadcast.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func (bs *broadcastService) recvBroadcasts(ctx context.Context) {
+	var buf = make([]byte, 48)
+	for bs.active {
+		n, rAddr, err := bs.listenSock.ReadFromUDP(buf)
+		if err == nil {
```
vyzo:
```
if err != nil {
 continue
}
```

Again, this reduces indentation level and cognitive overhead.

ehmry:
I'm just not going to think of go at all.

PR #249, opened 2017-11-27
Added an example for a multi-protocol app using strongly typed data and protobufs

avive:
Here's a little examples that shows how to build a node with multiple app protocols using protobufs as the wire format. The example creates a node that implements two protocols - ping and echo and the example demonstrate exchanging methods using the two protocols between two peers.


examples/multipro/protocol.go
```diff
@@ -0,0 +1,42 @@
...
+
+// helper method - writes a protobuf go data object to a network stream
+// data - reference of protobuf go data object to send (not the object itself)
+// s - network stream to write the data to
+func sendProtoMessage(data proto.Message, s inet.Stream) bool {
+	writer := bufio.NewWriter(s)
+	enc := protobufCodec.Multicodec(nil).Encoder(writer)
+	err := enc.Encode(data)
+	if err != nil {
+		log.Fatal(err)
```
whyrusleeping: log.Fatal will halt the program immediately, we would never get to return false

avive: good catch - got rid of all log.Fatal in latest check-in

avive:
I'd like to do some cleanups before this can be merged. I will commit here when I think this is 100% ready for merging. ok?

avive:
I can't really get the example to go build when replacing the GX deps with the canonical github repo names. Can one of you guys help me out with that? I assume we don't want code checkins with GX deps correct?

hsanjuan:
If you use the canonical github repo, and run `gx-go rewrite` on the base of this repo, and then try to build it (with the gx-rewritten paths), does it work? In that case, just add that step to the instructions... otherwise you may need to add your own package.json to the example folder...

avive:
I'm still having trouble with GX and deps :-(

Stebalien:
@avive I believe your problem is the `github.com/avive/go-libp2p/...` import (two different paths to import the same package). Would you mind if I pushed a version with gx fixes to your branch (gx can be a bit tricky, probably easier to learn by example)?

Stebalien:
So... Apparently, force-pushing master to a PR will close it. Worse, because this PR was made master to master, it appears that I can't reopen it. New PR: #280.

PR #264, opened 2018-01-18
Added default muxer in case when none was provided.

dmbreaker:
Should fix 'echo' example.

### DIFF
--- a/libp2p.go
+++ b/libp2p.go
@@ -180,13 +180,19 @@ func newWithCfg(ctx context.Context, cfg *Config) (host.Host, error) {
 		ps = pstore.NewPeerstore()
 	}
 
+	// Set default muxer if none was passed in
+	muxer := cfg.Muxer
+	if muxer == nil {
+		muxer = DefaultMuxer()
+	}
+
 	// If secio is disabled, don't add our private key to the peerstore
 	if !cfg.DisableSecio {
 		ps.AddPrivKey(pid, cfg.PeerKey)
 		ps.AddPubKey(pid, cfg.PeerKey.GetPublic())
 	}
 
-	swrm, err := swarm.NewSwarmWithProtector(ctx, cfg.ListenAddrs, pid, ps, cfg.Protector, cfg.Muxer, cfg.Reporter)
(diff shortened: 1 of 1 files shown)


dmbreaker:
Ok, it will allow example pass, and won't display stacktrace.
But example will use another (default) protocol instead.

libp2p.go
```diff
@@ -180,13 +180,19 @@ func newWithCfg(ctx context.Context, cfg *Config) (host.Host, error) {
...
+	}
+
 	// If secio is disabled, don't add our private key to the peerstore
 	if !cfg.DisableSecio {
 		ps.AddPrivKey(pid, cfg.PeerKey)
 		ps.AddPubKey(pid, cfg.PeerKey.GetPublic())
 	}
 
-	swrm, err := swarm.NewSwarmWithProtector(ctx, cfg.ListenAddrs, pid, ps, cfg.Protector, cfg.Muxer, cfg.Reporter)
+	swrm, err := swarm.NewSwarmWithProtector(ctx, cfg.ListenAddrs, pid, ps, cfg.Protector, muxer, cfg.Reporter)
```
vyzo: this looks inconsistent.

libp2p.go
```diff
@@ -180,13 +180,19 @@ func newWithCfg(ctx context.Context, cfg *Config) (host.Host, error) {
 		ps = pstore.NewPeerstore()
 	}
 
+	// Set default muxer if none was passed in
+	muxer := cfg.Muxer
```
vyzo: for consistency with the code below, it would be better to set it in the `cfg` object directly:
```go
if cfg.Muxer == nil {
   cfg.Muxer = DefaultMuxer()
}
```

Stebalien: Hm. But that would modify the user's config. I prefer it this way.

libp2p.go
Stebalien: Not really. With the protector/reporter, we default to nil. This is the one case where we don't.

Stebalien:
@vyzo given that we provide the DefaultMuxer method, I see no reason not to use it here.

vyzo:
fair enough.

Stebalien:
Thanks!

PR #267, opened 2018-01-27
mDNS discovery starts announcing right away

### LINKED ISSUE #256: mDNS discovery should start announcing right away
At the moment it waits for the first interval to pass, so it only starts announcing after 10 seconds or so.

alholm:
Does not wait for first interval to elapse as in case we don't need frequent announcements and interval is set e.g. to several minutes - first announcement will happen only after it.

Closes #256

### DIFF
--- a/p2p/discovery/mdns.go
+++ b/p2p/discovery/mdns.go
@@ -118,29 +118,32 @@ func (m *mdnsService) pollForEntries(ctx context.Context) {
 
 	ticker := time.NewTicker(m.interval)
 	for {
+		//execute mdns query right away at method call and then with every tick
+		entriesCh := make(chan *mdns.ServiceEntry, 16)
+		go func() {
+			for entry := range entriesCh {
+				m.handleEntry(entry)
+			}
+		}()
+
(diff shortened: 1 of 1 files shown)


p2p/discovery/mdns.go
```diff
@@ -116,31 +116,33 @@ func (m *mdnsService) Close() error {
...
-				}
-			}()
-
-			log.Debug("starting mdns query")
-			qp := &mdns.QueryParam{
-				Domain:  "local",
-				Entries: entriesCh,
-				Service: m.tag,
-				Timeout: time.Second * 5,
+	for ticker := time.NewTicker(m.interval); ; {
```
vyzo: It would reach much nicer if left the ticker initialization out of the loop and just use `for { ...}`.

So:
```
ticker := ...
for {
  ...
}
```

vyzo: It would also simplify the diff.

Stebalien:
macos test failure is unrelated (faulty test written by yours-truly).

Issue #257, opened 2017-12-14
Switch mdns library


Stebalien:
Either (a) switch to a fork or, even better, switch to https://github.com/grandcat/zeroconf. The current library is unmaintained and has at least one data race (and parts of it are incomprehensible).

Stebalien:
Closing in favor of #623.

kpp:
Why not switch to https://github.com/hashicorp/mdns ? https://github.com/whyrusleeping/mdns is a fork of https://github.com/hashicorp/mdns and https://github.com/hashicorp/mdns fixed some data races.

Stebalien:
We've fixed all the ones fixed upstream.

kpp:
Where?

Stebalien:
https://github.com/whyrusleeping/mdns/commit/9d49aa4bf4b2dada709ec1c3390cef586036eafe

Looking at a diff between the two versions, it looks like the only remaining change we need is the ability to turn off (or redirect) logging. We should probably upstream that and then get off the fork (eventually).

kpp:
> whyrusleeping/mdns@9d49aa4

Mutex is a bad solution because of:

```
$ go vet
# github.com/whyrusleeping/mdns
./client.go:165:41: call of logf copies lock value: github.com/whyrusleeping/mdns.client contains sync.Mutex
```

> Looking at a diff between the two versions, it looks like the only remaining change we need is the ability to turn off (or redirect) logging.

What do you mean? I am new to go

Issue #271, opened 2018-02-08
TestMdnsDiscovery fails


magik6k:
I'm not sure where to start debugging this.

Arch Linux / No {nf,ip}tables / go1.9.3

```
--- FAIL: TestMdnsDiscovery (2.07s)
        mdns_test.go:51: dial attempt failed: failed to dial <peer.ID VJcuj6> (default failure)
FAIL
FAIL    github.com/libp2p/go-libp2p/p2p/discovery       2.104s
```

Stebalien:
I was able to reproduce this. Our mDNS library is known to be racy (data races) so this isn't that surprising. See #257.

Stebalien:
Closing in favor of #623.

PR #272, opened 2018-02-14
Skip TestMdnsDiscovery

magik6k:
This tests fails about 9 out of 10 times for me, making the use of gx-workspace on libp2p related trees rather annoying.

Should get re-enabled after https://github.com/libp2p/go-libp2p/issues/257 is done.


whyrusleeping:
Can we file an issue (or two) to make sure we get this passing soon? and have tests just for this

Stebalien:
The issues are #271 and #257.

PR #278, opened 2018-02-19
Example of echo server using routed hosts with local and ipfs bootstrap

jvsteiner:
Hi, many thanks to @Stebalien for helping me understand the `libp2p` family of libraries.  I incorporated some of what I've learned in this example code.  I think it should work pretty well, but currently it's not set up to use a real world-reachable ip address, due to what I believe are issues with NAT.  If you would like to have anything additional added, I'd be happy to help.


examples/routed-echo/README.md
```diff
@@ -0,0 +1,49 @@
+# Routed Host: echo client/server
+
+This example is intended to follow up the basic host and echo examples by adding use of the ipfs distributed hash table to lookup peers.
+
+Functionally, this example works similarly to the echo example, however setup of the host includes wrapping it with a Kademila hash table, so it can find peers using only their ID.
```
Stebalien: Nit: "their IDs"

examples/routed-echo/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+> ./routed-echo -secio -l 10000
+2018/02/19 12:22:32 I can be reached at:
+2018/02/19 12:22:32 /ip4/127.0.0.1/tcp/10000/ipfs/QmfRY4vuKpU2tApACrbmYFn9xoeNzMQhLXg7nKnyvnzHeL
+2018/02/19 12:22:32 /ip4/192.168.1.203/tcp/10000/ipfs/QmfRY4vuKpU2tApACrbmYFn9xoeNzMQhLXg7nKnyvnzHeL
+2018/02/19 12:22:32 Now run "./routed-echo -l 10001 -d QmfRY4vuKpU2tApACrbmYFn9xoeNzMQhLXg7nKnyvnzHeL -secio" on a different terminal
+2018/02/19 12:22:32 listening for connections
+```
+
+The listener libp2p host will print its randomly generated Base58 encoded ID string, which combined with the ipfs DHT, can be used to reach the host, despite lacking other connection details.  You can also pass the `-global` flag to bootstrap using the same peers as ipfs.
```
Stebalien: Maybe: "By default, this example will bootstrap off your local IPFS peer (assuming one is running). If you'd rather bootstrap off the same peers `go-ipfs` uses, pass the `-global` flag."

That is, something to indicate the default behaviour.

examples/routed-echo/main.go
```diff
@@ -0,0 +1,203 @@
...
+
+	out, err := ioutil.ReadAll(s)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	log.Printf("read reply: %q\n", out)
+}
+
+// doEcho reads a line of data a stream and writes it back
```
Stebalien: nit: "from a stream".

examples/routed-echo/README.md
jvsteiner: note: I couldn't get it to work reliably if the first host was bootstrapped locally, and the second was bootstrapped globally.

jvsteiner:
updated

Stebalien:
So, @whyrusleeping, opinions... this example adds a dependency on the DHT. I'm wondering if we should just move all the examples out of this repo into a separate `go-libp2p-examples` package (possibly leaving either a README or a git submodule behind). Thoughts?

whyrusleeping:
I'm happy to move them to a separate repo, as long as we make sure to point to it from here

jvsteiner:
I see some CI checks failed, not sure if this is related to my code - either way, if there's anything I can do to improve this, let me know!

Stebalien:
Those problems are my bad. I don't know exactly what happened but I'll fix and rebase.

Stebalien:
I've filed an issue for breaking out the examples. We can do that separately. #292

Stebalien:
So... actually, it turns out that `go-libp2p-kad-dht` depends on *this* repo (for some test cases) so we have a circular dependency (hence the gx issue). It looks like we'll *have* to move this into a separate repo (although, really, we should move out the basic host and the mock network instead...).

This may take a while.

jvsteiner:
added NAT traversal, and using `bhost.NewHost` constructor

agahEbrahimi:
Is it possible to broadcast a message onto the whole network to see if they have a certain resource so you can then have a direct channel between you and the node closest to you with that resource?

Stebalien:
@agahEbrahimi please try to keep comments on-topic. As that appears to be a question about ipfs, you should probably ask it on https://discuss.ipfs.io.

Stebalien:
Ok, I've finally created [go-libp2p-examples](https://github.com/libp2p/go-libp2p-examples). Mind re-opening this PR there?

jvsteiner:
sure, I'll try to get to it this week!

PR #285, opened 2018-02-27
discovery: swap mdns lib for grandcat/zeroconf

rargulati:
Specifically addressing the request here: https://github.com/libp2p/go-libp2p/issues/257

This commit makes the minimal amount of changes to switch us from
whyrusleeping/mdns to grandcat/zeroconf. Of note, rather than asking the
`host.Host` which `addrs` we can listen on, we push this to the
zeroconf library (which runs across available ifaces).

I haven't had a chance to test this on something complicated to see if it retains the desired behavior (to be honest, still new to this codebase, unsure of what all of those would be!), but tests pass locally. 

TODO: 
 - [x] Revert PORT behavior
 - [x] ensure that grandcat/zerconf dep is in gx; tests pass

### DIFF
--- a/p2p/discovery/mdns.go
+++ b/p2p/discovery/mdns.go
@@ -2,21 +2,21 @@ package discovery
 
 import (
 	"context"
-	"errors"
 	"io"
 	"io/ioutil"
 	golog "log"
 	"net"
+	"strings"
 	"sync"
 	"time"
 
+	mdns "github.com/grandcat/zeroconf"
 	logging "github.com/ipfs/go-log"
 	"github.com/libp2p/go-libp2p-host"
 	"github.com/libp2p/go-libp2p-peer"
 	pstore "github.com/libp2p/go-libp2p-peerstore"
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr-net"
-	"github.com/whyrusleeping/mdns"
 )
 
 var log = logging.Logger("mdns")
@@ -34,32 +34,13 @@ type Notifee interface {
 }
 
 type mdnsService struct {
-	server  *mdns.Server
-	service *mdns.MDNSService
-	host    host.Host
-	tag     string
+	server *mdns.Server
+	host   host.Host
+	tag    string
 
 	lk       sync.Mutex
 	notifees []Notifee
 	interval time.Duration
-}
-
-func getDialableListenAddrs(ph host.Host) ([]*net.TCPAddr, error) {
-	var out []*net.TCPAddr
-	for _, addr := range ph.Addrs() {
-		na, err := manet.ToNetAddr(addr)
-		if err != nil {
-			continue
-		}
-		tcp, ok := na.(*net.TCPAddr)
-		if ok {
-			out = append(out, tcp)
-		}
-	}
-	if len(out) == 0 {
-		return nil, errors.New("failed to find good external addr from peerhost")
-	}
-	return out, nil
 }
 
 func NewMdnsService(ctx context.Context, peerhost host.Host, interval time.Duration, serviceTag string) (Service, error) {
@@ -67,39 +48,22 @@ func NewMdnsService(ctx context.Context, peerhost host.Host, interval time.Durat
 	// TODO: dont let mdns use logging...
 	golog.SetOutput(ioutil.Discard)
 
-	var ipaddrs []net.IP
 	port := 4001
-
-	addrs, err := getDialableListenAddrs(peerhost)
-	if err != nil {
-		log.Warning(err)
-	} else {
-		port = addrs[0].Port
-		for _, a := range addrs {
-			ipaddrs = append(ipaddrs, a.IP)
-		}
-	}
-
 	myid := peerhost.ID().Pretty()
 
 	info := []string{myid}
 	if serviceTag == "" {
 		serviceTag = ServiceTag
 	}
-	service, err := mdns.NewMDNSService(myid, serviceTag, "", "", port, ipaddrs, info)
-	if err != nil {
-		return nil, err
-	}
 
 	// Create the mDNS server, defer shutdown
-	server, err := mdns.NewServer(&mdns.Config{Zone: service})
+	server, err := mdns.Register(myid, serviceTag, "", port, info, nil)
 	if err != nil {
 		return nil, err
 	}
 
 	s := &mdnsService{
 		server:   server,
-		service:  service,
 		host:     peerhost,
 		interval: interval,
 		tag:      serviceTag,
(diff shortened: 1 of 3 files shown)


p2p/discovery/mdns.go
```diff
@@ -152,8 +120,13 @@ func (m *mdnsService) pollForEntries(ctx context.Context) {
 }
 
 func (m *mdnsService) handleEntry(e *mdns.ServiceEntry) {
-	log.Debugf("Handling MDNS entry: %s:%d %s", e.AddrV4, e.Port, e.Info)
-	mpeer, err := peer.IDB58Decode(e.Info)
+	// pull out the txt
+	info := strings.Join(e.Text, "|")
+
+	// addripv4 potential issue
+	log.Debugf("Handling MDNS entry: %s:%d %s", e.AddrIPv4[0], e.Port, info)
```
rargulati: This is the only decision I wasn't certain of.

p2p/discovery/mdns.go
```diff
@@ -111,34 +75,38 @@ func NewMdnsService(ctx context.Context, peerhost host.Host, interval time.Durat
...
+	m.server.Shutdown()
+	// grandcat/zerconf swallows error, satisfy interface
+	return nil
 }
 
 func (m *mdnsService) pollForEntries(ctx context.Context) {
 
 	ticker := time.NewTicker(m.interval)
 	for {
+		resolver, err := mdns.NewResolver(nil)
```
Kubuxu: I think it might be beast to put resolver in `mdnsService` struct and reuse it, instead of creating a new one every iteration.

p2p/discovery/mdns.go
```diff
@@ -34,72 +34,36 @@ type Notifee interface {
...
-	return out, nil
-}
-
 func NewMdnsService(ctx context.Context, peerhost host.Host, interval time.Duration, serviceTag string) (Service, error) {
 
 	// TODO: dont let mdns use logging...
 	golog.SetOutput(ioutil.Discard)
 
-	var ipaddrs []net.IP
 	port := 4001
```
Kubuxu: This is old but it is very wrong.

p2p/discovery/mdns.go
```diff
@@ -165,7 +138,7 @@ func (m *mdnsService) handleEntry(e *mdns.ServiceEntry) {
 	}
 
 	maddr, err := manet.FromNetAddr(&net.TCPAddr{
-		IP:   e.AddrV4,
+		IP:   e.AddrIPv4[0],
```
Kubuxu: As there is possibility of multiple addresses, I would iterate over them, crate maddr for each and add it to the peerstore.

p2p/discovery/mdns.go
rargulati: Yikes, good catch. Fixed.

rargulati:
@Stebalien @Kubuxu I've addressed the feedback in c92f1f6

In regards to the port, what would you recommend here? Skimmed through https://tools.ietf.org/html/rfc6762 and see that 5353 is used for the listener, but not seeing anything for the port that's announcing.

I notice that I'm failing tests due to a gx issue. Any recommendations? I'll need to learn that right quick :)

Kubuxu:
@rargulati It was my mistake, I thought it was the port we were announcing to the world but it is mDNS discovery port.

@Stebalien correct if I am wrong, but currently mDNS only publishes peerID in local network, not the full multiaddr, is that right? If it is so, is there a reason not to publish maddr so we can connect to it? Publishing just peerID sees little fruitless as it won't work without DHT capabilities (in case that bootstrap didn't work).

Stebalien:
Port *is* the port of the service. Also, we do announce the IP addresses and we convert port + IP into `/ip4/IP/tcp/PORT`. However, that's... suboptimal. It doesn't account for any other potential transports.

Note: For *now*, I'd be fine preserving the existing behavior. The primary concern here is upgrading to a *working* mDNS library (one that doesn't have memory safety issues).

However, we *also* have a bunch of bugs/inconsistencies that we need to fix. I've written them up here:https://github.com/libp2p/libp2p/issues/28

rargulati:
Just to clarify on final things:
-[ ] Revert PORT to 4001 - the fix to this and other issues will be addressed as detailed in libp2p/libp2p#28
-[ ] Get grandcat/zerconf into gx (I'm struggling with this - how do I get a new hash for a new dependency - either of you mind doing a quick how-to in irc/email/here?) 

Does that sound about right @Stebalien @Kubuxu - I'll get this across the line asap.

Stebalien:
>  Revert PORT to 4001 - the fix to this and other issues will be addressed as detailed in libp2p/libp2p#28

Actually, restore the previous "figure out what port we're using" code. It's not perfect, but it's better than just assuming port 4001.

>  Get grandcat/zerconf into gx (I'm struggling with this - how do I get a new hash for a new dependency - either of you mind doing a quick how-to in irc/email/here?)

I'll import this into gx tomorrow. We need to write some proper tutorials but it's on our never ending backlog.

Stebalien:
> I'll import this into gx tomorrow. We need to write some proper tutorials but it's on our never ending backlog.

This will actually require updating some of our dependencies so it may take a week (blocked on a bunch of other stuff).

deckbsd:
Hi, do you know when this commit will be done ?

Stebalien:
It's been blocked on:

* [x] Cutting a new release (0.4.14)
* [ ] Fixing a namesys issue: https://github.com/ipfs/go-ipfs/pull/4506
* [ ] Updating the datastores and the DHT (in go-ipfs).

Once those are done, I'll finally be able to bubble a few low-level updates through the entire tree. The problem here is that this zeroconf library requires a newer version of `golang.org/x/net` than we use in go-ipfs.

Stebalien:
Still need to fix https://github.com/libp2p/go-libp2p/issues/376 but I'd like to punt on that.

Stebalien:
Waiting on https://github.com/cenkalti/backoff/pull/59

p2p/discovery/mdns.go
```diff
@@ -44,62 +45,76 @@ type mdnsService struct {
 	interval time.Duration
 }
 
-func getDialableListenAddrs(ph host.Host) ([]*net.TCPAddr, error) {
-	var out []*net.TCPAddr
-	for _, addr := range ph.Addrs() {
+// Tries to pick the best port.
+func getBestPort(addrs []ma.Multiaddr) (int, error) {
```
Stebalien: This should be fixed by https://github.com/libp2p/go-libp2p/issues/376

p2p/discovery/mdns.go
```diff
@@ -152,8 +165,10 @@ func (m *mdnsService) pollForEntries(ctx context.Context) {
 }
 
 func (m *mdnsService) handleEntry(e *mdns.ServiceEntry) {
-	log.Debugf("Handling MDNS entry: %s:%d %s", e.AddrV4, e.Port, e.Info)
-	mpeer, err := peer.IDB58Decode(e.Info)
+	// pull out the txt
+	info := strings.Join(e.Text, "|")
```
bigs: is this designed to fail if there are less than or more than one entry?

bigs: to be clear, because of the decoding afterwards

Stebalien: Hm. Yeah, that looks wrong. We should just take the first entry and ignore the rest.

Stebalien:
Blocked on https://github.com/grandcat/zeroconf/issues/38. Fixing that was the reason we went down this road in the first place so there's no real reason to merge this till we get that fixed.

Stebalien:
Given https://github.com/libp2p/specs/pull/80, we should probably just abort this whole thing and implement our own library that:

1. Implements that spec.
2. Doesn't have random race conditions.
3. Can handle things like network interface changes.

PR #289, opened 2018-03-02
Fix contributing link

mark-rushakoff:
Looks like it was changed from lowercase to uppercase recently in
https://github.com/ipfs/community/commit/ff38ebf65448353587e938e61ca36829fbecc8b4


mark-rushakoff:
That rename was reverted, so this PR is no longer necessary.

PR #298, opened 2018-03-12
Update README.md

qidu:
to get the source right


Stebalien:
Any reason to remove the `/...`? That fetches the dependencies of the sub-packages (e.g., examples).

Issue #297, opened 2018-03-08
Transport + Swarm refactor


Stebalien:
Moved from https://github.com/libp2p/go-libp2p-transport/issues/21 for visibility.

Related to #279 (but this one is *just* about the refactor itself, not the overall task of QUIC + TLS + refactor).

State: In review. go-ipfs currently builds and runs (well) with these patches.

Branch `feat/refactor` in:

* [x] [ipfs/go-ipfs](https://github.com/ipfs/go-ipfs)
  * [x] PR open
  * [x] Reviewed
  * [x] Merged
* [x] [libp2p/go-libp2p-metrics](https://github.com/libp2p/go-libp2p-metrics)
  * [x] PR open - https://github.com/libp2p/go-libp2p-metrics/pull/12
  * [x] Reviewed - @whyrusleeping 
  * [x] Merged
* [x] [libp2p/go-libp2p-swarm](https://github.com/libp2p/go-libp2p-swarm)
  * [x] PR open -  https://github.com/libp2p/go-libp2p-swarm/pull/63
  * [x] Reviewed - @Kubuxu, @whyrusleeping, @magik6k
  * [x] Merged
* [x] [libp2p/go-libp2p-circuit](https://github.com/libp2p/go-libp2p-circuit)
  * [x] PR open - https://github.com/libp2p/go-libp2p-circuit/pull/33
  * [x] Reviewed - @vyzo 
  * [x] Merged
* [x] [libp2p/go-ws-transport](https://github.com/libp2p/go-ws-transport)
  * [x] PR open - https://github.com/libp2p/go-ws-transport/pull/31
  * [x] Reviewed
  * [x] Merged
* [x] [libp2p/go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht) -- netutil changes + changed notification guarantees.
  * [x] PR open - https://github.com/libp2p/go-libp2p-kad-dht/pull/126
  * [x] Reviewed
  * [x] Merged
* [x] [libp2p/go-libp2p-netutil](https://github.com/libp2p/go-libp2p-netutil)
  * [x] PR open - https://github.com/libp2p/go-libp2p-netutil/pull/19
  * [x] Reviewed
  * [x] Merged
* [x] [multiformats/go-multiaddr-net](https://github.com/multiformats/go-multiaddr-net)
  * [x] PR open - https://github.com/multiformats/go-multiaddr-net/pull/37
  * [x] Reviewed
  * [x] Merged
[...]

PR #299, opened 2018-03-13
refactor for transport changes

Stebalien:
Also, make the libp2p constructor fully useful. There should now be no need to manually construct a swarm/host. That's actually where most of the new code comes from (hence the +450 lines of (actual) code). This allows me to actually use the libp2p constructor in go-ipfs.

DO NOT MERGE (yet)
Part of: libp2p/go-libp2p#297

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,0 +1,176 @@
+package config
+
+import (
+	"context"
+	"fmt"
+
+	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+
+	logging "github.com/ipfs/go-log"
+	circuit "github.com/libp2p/go-libp2p-circuit"
+	crypto "github.com/libp2p/go-libp2p-crypto"
+	host "github.com/libp2p/go-libp2p-host"
+	ifconnmgr "github.com/libp2p/go-libp2p-interface-connmgr"
+	pnet "github.com/libp2p/go-libp2p-interface-pnet"
+	metrics "github.com/libp2p/go-libp2p-metrics"
+	inet "github.com/libp2p/go-libp2p-net"
+	peer "github.com/libp2p/go-libp2p-peer"
+	pstore "github.com/libp2p/go-libp2p-peerstore"
+	swarm "github.com/libp2p/go-libp2p-swarm"
+	tptu "github.com/libp2p/go-libp2p-transport-upgrader"
+	filter "github.com/libp2p/go-maddr-filter"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+var log = logging.Logger("p2p-config")
+
+// AddrsFactory is a function that takes a set of multiaddrs we're listening on and
+// returns the set of multiaddrs we should advertise to the network.
+type AddrsFactory = bhost.AddrsFactory
+
+// NATManagerC is a NATManager constructor.
+type NATManagerC func(inet.Network) bhost.NATManager
+
+// Config describes a set of settings for a libp2p node
+//
+// This is *not* a stable interface. Use the options defined in the root
+// package.
+type Config struct {
+	PeerKey crypto.PrivKey
+
+	Transports         []TptC
+	Muxers             []MsMuxC
+	SecurityTransports []MsSecC
+	Insecure           bool
+	Protector          pnet.Protector
+
+	Relay     bool
+	RelayOpts []circuit.RelayOpt
+
+	ListenAddrs  []ma.Multiaddr
+	AddrsFactory bhost.AddrsFactory
+	Filters      *filter.Filters
+
+	ConnManager ifconnmgr.ConnManager
+	NATManager  NATManagerC
+	Peerstore   pstore.Peerstore
+	Reporter    metrics.Reporter
+}
+
+// NewNode constructs a new libp2p Host from the Config.
+//
+// This function consumes the config. Do not reuse it (really!).
+func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
+	// Check this early. Prevents us from even *starting* without verifying this.
+	if pnet.ForcePrivateNetwork && cfg.Protector == nil {
+		log.Error("tried to create a libp2p node with no Private" +
+			" Network Protector but usage of Private Networks" +
+			" is forced by the enviroment")
+		// Note: This is *also* checked the upgrader itself so it'll be
+		// enforced even *if* you don't use the libp2p constructor.
+		return nil, pnet.ErrNotInPrivateNetwork
+	}
+
+	if cfg.PeerKey == nil {
+		return nil, fmt.Errorf("no peer key specified")
+	}
+
+	// Obtain Peer ID from public key
+	pid, err := peer.IDFromPublicKey(cfg.PeerKey.GetPublic())
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.Peerstore == nil {
+		return nil, fmt.Errorf("no peerstore specified")
+	}
+
+	if !cfg.Insecure {
+		cfg.Peerstore.AddPrivKey(pid, cfg.PeerKey)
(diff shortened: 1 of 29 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -97,25 +93,14 @@ type HostOpts struct {
...
-	// BandwidthReporter is used for collecting aggregate metrics of the
-	// bandwidth used by various protocols.
-	BandwidthReporter metrics.Reporter
+	NATManager func(inet.Network) NATManager
 
 	// ConnManager is a libp2p connection manager
 	ConnManager ifconnmgr.ConnManager
-
-	// Relay indicates whether the host should use circuit relay transport
-	EnableRelay bool
```
Stebalien: Handled by the libp2p constructor.

p2p/host/basic/basic_host.go
```diff
@@ -97,25 +93,14 @@ type HostOpts struct {
 
 	// NATManager takes care of setting NAT port mappings, and discovering external addresses.
 	// If omitted, this will simply be disabled.
-	NATManager NATManager
-
-	// BandwidthReporter is used for collecting aggregate metrics of the
-	// bandwidth used by various protocols.
-	BandwidthReporter metrics.Reporter
```
Stebalien: Was only needed by the identity service because the old swarm didn't enable the bandwidth reporter early enough. That's no longer the case so the host doesn't need this now.

p2p/host/basic/basic_host.go
```diff
@@ -536,11 +497,6 @@ func (h *BasicHost) Close() error {
 	return h.proc.Close()
 }
 
-// GetBandwidthReporter exposes the Host's bandiwth metrics reporter
-func (h *BasicHost) GetBandwidthReporter() metrics.Reporter {
```
Stebalien: This isn't a part of the Host interface and I can't find anything that actually uses it. Really, users should just hang onto the bandwidth reporter that they passed in.

libp2p.go
```diff
@@ -155,87 +231,65 @@ func Identity(sk crypto.PrivKey) Option {
...
-	// Set up stream multiplexer
-	tpt := msmux.NewBlankTransport()
-
-	// By default, support yamux and multiplex
-	tpt.AddTransport("/yamux/1.0.0", yamux.DefaultTransport)
-	tpt.AddTransport("/mplex/6.3.0", mplex.DefaultTransport)
-
-	return tpt
+// NATPortMap configures libp2p to use the default NATManager. The default
+// NATManager will attempt to punch holes in your NAT.
```
vyzo: this comment is technically incorrect -- doesn't the default nat manager attempt to open ports using UPnP? 
It doesn't really use hole punching.

config/config.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a new blank peerstore if none was passed in
+	ps := cfg.Peerstore
+	if ps == nil {
+		ps = pstore.NewPeerstore()
+	}
+	if !cfg.Insecure {
```
vyzo: Why do we keep lugging around the insecure option?

Stebalien: We use it for debugging. It's actually really useful. However, it's designed to be an all or nothing "disable all security" option.

libp2p.go
Stebalien: Yes, good point.

config/config.go
vyzo: fair enough! I can see it being useful in analysing wire protocols without encryption.

Stebalien:
@whyrusleeping massively reduced the code duplication for the reflection stuff. It turned out to be much easier than I had thought. I left the optional error in the transport constructor return type but broke the code int a bunch of smaller, readable functions.

hsanjuan:
I was looking to bubble go-reuseport with a fixed `go-log` that does not panic randomly. Or do you prefer to keep libp2p frozen until all these changes come in?

Stebalien:
Feel free (that's why I'm merging these all at the same time).

config/config.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+var log = logging.Logger("p2p-config")
+
+// AddrsFactory is a function that takes a set of multiaddrs we're listening on and
+// returns the set of multiaddrs we should advertise to the network.
+type AddrsFactory = bhost.AddrsFactory
+
+// NATManagerC is a NATManager constructor.
+type NATManagerC func(inet.Network) bhost.NATManager
```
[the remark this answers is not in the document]
Kubuxu: Still better than in Java world.

config/config.go
```diff
@@ -0,0 +1,171 @@
...
+	Peerstore          pstore.Peerstore
+	Protector          pnet.Protector
+	Reporter           metrics.Reporter
+	Relay              bool
+	RelayOpts          []circuit.RelayOpt
+	Insecure           bool
+	ConnManager        ifconnmgr.ConnManager
+	AddrsFactory       bhost.AddrsFactory
+	Filters            *filter.Filters
+	NATManager         NATManagerC
```
Kubuxu: Some spacing ques would be good here. Like `Relay ` and `RelayOpts` can have their own section.

config/config.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// NewNode constructs a new libp2p Host from the Config.
+//
+// This function consumes the config. Do not reuse it (really!).
+func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
+	// Check this early. Prevents us from even *starting* without verifying this.
+	if pnet.ForcePrivateNetwork && cfg.Protector == nil {
+		log.Error("tried to dial with no Private Network Protector but usage" +
```
Kubuxu: `dial` is not accurate here

config/config.go
```diff
@@ -0,0 +1,171 @@
...
+		return nil, fmt.Errorf("no peer key specified")
+	}
+
+	// Obtain Peer ID from public key
+	pid, err := peer.IDFromPublicKey(cfg.PeerKey.GetPublic())
+	if err != nil {
+		return nil, err
+	}
+
+	// Create a new blank peerstore if none was passed in
```
Kubuxu: Comment not accurate, either change behaviour or the comment.

config/muxer.go
```diff
@@ -0,0 +1,62 @@
...
+// using reflection.
+func MuxerConstructor(m interface{}) (MuxC, error) {
+	// Already constructed?
+	if t, ok := m.(mux.Transport); ok {
+		return func(_ host.Host) (mux.Transport, error) {
+			return t, nil
+		}, nil
+	}
+
+	fn, err := makeConstructor(m, muxType, muxArgTypes)
```
Kubuxu: IMO `s/fn/ctor/`

package.json
```diff
@@ -293,6 +276,24 @@
       "hash": "QmcBWojPoNh4qm7zvv4qiepvCnnc7ALS9qcp7TNwwxT1gT",
       "name": "go.uuid",
       "version": "1.1.0"
+    },
+    {
+      "author": "whyrusleeping",
+      "hash": "QmYkxWtBWqkj6SWCiFN2a3xnnk5womAQYd9Kh68LJJR7jz",
+      "name": "go-ws-transport",
```
Kubuxu: do we have to depend on a transport?

Stebalien: It's a default transport so yes unless we want to make it not a default transport. Note: we used to depend on it in go-libp2p-swarm but I moved all default transport construction here.

config/config.go
Stebalien: Good catch. I moved the defaults to a special `FallbackDefaults` option.

Stebalien:
> I am not a fan of voodoo magic but I don't think there is a better way.

I completely agree. My first attempt involved defining some nice interfaces but it turned into a mess.

PR #302, opened 2018-03-20
Examples/bootstrapping railing

florianlenz:
## Changes
- Added example that describes how to use bootstrapping nodes

## Verification
- When you follow the instructions from the readme in I committed you should get an executable that connect's to the bootstrapping nodes from the `railing.go` file.

### DIFF
--- a/examples/bootstrapping-railing/railing.go
+++ b/examples/bootstrapping-railing/railing.go
@@ -3,9 +3,10 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/libp2p/go-libp2p"
-	"github.com/libp2p/go-libp2p-peerstore"
-	"github.com/ipfs/go-ipfs-addr"
+	
+	libp2p "github.com/libp2p/go-libp2p"
+	peerstore "github.com/libp2p/go-libp2p-peerstore"
+	ipfsaddr "github.com/ipfs/go-ipfs-addr"
 )
 
 var bootstrapPeers = []string{
--- a/examples/bootstrapping-railing/railing.go
+++ b/examples/bootstrapping-railing/railing.go
@@ -3,15 +3,15 @@ package main
 import (
 	"context"
 	"fmt"
-	
+
 	libp2p "github.com/libp2p/go-libp2p"
-	peerstore "github.com/libp2p/go-libp2p-peerstore"
 	ipfsaddr "github.com/ipfs/go-ipfs-addr"
+	peerstore "github.com/libp2p/go-libp2p-peerstore"
 )
 
-//The bootstrapping nodes are just long running nodes with a static IP address.
-//That means you can easily have your own bootstrapping nodes. Everything you need is
-//a server with a static IP address.
+// The bootstrapping nodes are just long running nodes with a static IP address.
+// That means you can easily have your own bootstrapping nodes. Everything you need is
+// a server with a static IP address.
 var bootstrapPeers = []string{
 	"/ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
 	"/ip4/104.236.76.40/tcp/4001/ipfs/QmSoLV4Bbm51jM9C4gDYZQ9Cy3U6aXMJDAbzgu2fzaDs64",
@@ -24,12 +24,12 @@ var bootstrapPeers = []string{
 	"/ip4/104.236.151.122/tcp/4001/ipfs/QmSoLju6m7xTh3DuokvT3886QRYqxAzb1kShaanJgW36yx",
 }
 
-//This example show's you how you can connect to a list of bootstrapping nodes.
+// This example show's you how you can connect to a list of bootstrapping nodes.
 func main() {
 
 	ctx := context.Background()
 
-	//Create the host
+	// Create the host
 	host, err := libp2p.New(ctx, libp2p.Defaults)
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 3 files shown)


examples/bootstrapping-railing/railing.go
```diff
@@ -0,0 +1,61 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"github.com/libp2p/go-libp2p"
+	"github.com/libp2p/go-libp2p-peerstore"
+	"github.com/ipfs/go-ipfs-addr"
```
Kubuxu: It is prefered to use named import (even the same name as pkg name) if the pkg name is different than last segment of the path.

Kubuxu:
Another things is, in future we will be taking down some of the bootstrap nodes so it would be better to try connecting to few in parallel so you don't have to wait for full timeout on all of them.

Also it is generally safe to resume the rest of initialisation after being connected to 2 or 3 bootstrap nodes, it would be nice if the example code reflected that.

examples/bootstrapping-railing/railing.go
```diff
@@ -0,0 +1,61 @@
+package main
+
+import (
+	"context"
+	"fmt"
```
Kubuxu: Please separate stdlib and ipfs imports by a empty line for clarity.

examples/bootstrapping-railing/railing.go
```diff
@@ -0,0 +1,61 @@
...
+	"/ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
+	"/ip4/104.236.76.40/tcp/4001/ipfs/QmSoLV4Bbm51jM9C4gDYZQ9Cy3U6aXMJDAbzgu2fzaDs64",
+	"/ip4/104.236.176.52/tcp/4001/ipfs/QmSoLnSGccFuZQJzRadHn95W2CrSFmZuTdDWP8HXaHca9z",
+	"/ip4/104.236.179.241/tcp/4001/ipfs/QmSoLPppuBtQSGwKDZT2M73ULpjvfd3aZ6ha4oFGL1KrGM",
+	"/ip4/162.243.248.213/tcp/4001/ipfs/QmSoLueR4xBeUbY9WZ9xGUUxunbKWcrNFTDAadQJmocnWm",
+	"/ip4/128.199.219.111/tcp/4001/ipfs/QmSoLSafTMBsPKadTEgaXctDQVcqN88CNLHXMkTNwMKPnu",
+	"/ip4/178.62.158.247/tcp/4001/ipfs/QmSoLer265NRgSp2LA3dPaeykiS1J6DifTC88f5uVQKNAd",
+	"/ip4/178.62.61.185/tcp/4001/ipfs/QmSoLMeWqB7YGVLJN3pNLQpmmEk35v6wYtsMGLzSr5QBU3",
+	"/ip4/104.236.151.122/tcp/4001/ipfs/QmSoLju6m7xTh3DuokvT3886QRYqxAzb1kShaanJgW36yx",
+}
```
Kubuxu: It would also be great to add a comment here that any long running node with stable external address can be used as bootstrap node.

examples/bootstrapping-railing/railing.go
```diff
@@ -3,11 +3,15 @@ package main
...
-	"github.com/ipfs/go-ipfs-addr"
+	
+	libp2p "github.com/libp2p/go-libp2p"
+	peerstore "github.com/libp2p/go-libp2p-peerstore"
+	ipfsaddr "github.com/ipfs/go-ipfs-addr"
 )
 
+//The bootstrapping nodes are just long running nodes with a static IP address.
+//That means you can easily have your own bootstrapping nodes. Everything you need is
+//a server with a static IP address.
```
Kubuxu: General go style is to have a space between // and comment.

florianlenz: @Kubuxu I added it now. Do you know what `go fmt` didn't fix it?

florianlenz:
Is there a specific reason why there is no go lib for railing like here in the JS [implementation](https://github.com/libp2p/js-libp2p-railing)? I am willing to develop such an module for go libp2p in the case it's needed.

Stebalien:
We call this bootstrap in go and actually have an example in-progress (https://github.com/libp2p/go-libp2p/pull/278). Unfortunately, it's blocked on some other changes.

We *definitely* want a separate bootstrap package.

General design/requirements:

1. Take in a `go-libp2p-net` `Network`, a list of bootstrap peers, and a minimum peer count.
2. While running, monitor the number of peers we're connected to. If we drop below some number, automatically start bootstrapping. We'd probably want to listen to connect/disconnect events on the Network to do this.
3. If bootstrapping fails, backoff (probably some kind of bounded exponential backoff).
4. When we notice that we're having bootstrapping issues, periodically (e.g., once a minute?) poll `network.InterfaceAddresses()`. If we notice a change, try bootstrapping.

Ideally, we'd also try bootstrapping whenever we notice the user attempting to dial but we don't have events in place for that (yet).

florianlenz:
@Stebalien Perfect. I will work on the package.

Stebalien:
Awesome! I've wanted that for a while.

Kubuxu:
@Stebalien in light of your refactor, do you want to merge it or hold it off?

Stebalien:
@Kubuxu my plan was to wait for #304.

agahEbrahimi:
@Stebalien does https://github.com/libp2p/go-libp2p/pull/278 work?

agahEbrahimi:
Do all bootstrapping nodes hold a DHT for the multiaddress of all other nodes?

Stebalien:
@agahEbrahimi yes. However, that's blocked on fixing some dependency issues.

marten-seemann:
This is most likely severely outdated. Closing.

PR #306, opened 2018-03-27
misc: downgrade a warning to info

Kubuxu:
It is source of most warning noise in Warning channel and we would like
to enable warning channel some time in future by default


p2p/host/basic/basic_host.go
```diff
@@ -250,7 +250,7 @@ func (h *BasicHost) newStreamHandler(s inet.Stream) {
 			}
 			logf("protocol EOF: %s (took %s)", s.Conn().RemotePeer(), took)
 		} else {
-			log.Warningf("protocol mux failed: %s (took %s)", err, took)
```
Stebalien: How long does it usually take to get this warning. This could be a sign of something weird. However, I agree we should be less noisy for something that the user can't really do anything about.

PR #326, opened 2018-05-07
Updated host example README to match code

mgoelzer:
The README.md text of the `examples/libp2p-host` tutorial does not match the code provided in the corresponding `host.go`.  It looks like the README.md is outdated.  At some point we simplified the host constructor to hide implementation details like:

* `peerstore`
* the storing of defaults in `github.com/libp2p/go-libp2p/p2p/host/basic`
* any explicit reference to multiaddrs

These are great simplifications to the libp2p API, and the README.md should reflect them.

However, I didn't want to completely remove the concepts of `peerstore` and `multiaddr`, since these concepts may be exposed to users of the API elsewhere.  So, what I did was move that stuff to a section at the end of the README that basically explains that `libp2p.New()` is doing this stuff behind the scenes.

I would appreciate a review for accuracy of the updated README content.

### DIFF
--- a/examples/libp2p-host/README.md
+++ b/examples/libp2p-host/README.md
@@ -4,7 +4,7 @@ For most applications, the host is the basic building block you'll need to get s
 
 The host is an abstraction that manages services on top of a swarm. It provides a clean interface to connect to a service on a given remote peer.
 
-First, you'll need an ID, and a place to store that ID. To generate an ID, you can do the following:
+First, you'll need an ID. A host's ID is the hash of its public key. To generate an ID, you can do the following:
 
 ```go
 import (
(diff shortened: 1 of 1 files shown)


mgoelzer:
In particular, I'm pretty sure what I'm saying about a peerstore being created internally within the host constructor is at least somewhat wrong.  The host constructor only receives the public key of the host, so it cannot possible be setting up a peerstore with both the public and private keys, as it does not know the latter.  Should I just take this out?

If the concept of a peerstore is something that higher level users of libp2p need to know about, I'd like to keep it in somehow.  But if this is purely an implementation detail, then it could be taken out.

mgoelzer:
Closing this because @Stebalien's refactoring is going to change a lot of this.

PR #331, opened 2018-05-23
Add configured transports to swarm

bigs:
Not sure if we want to merge this, but this patch enables user-configured `Transport`s by actually adding them to the `Swarm`. Note that in order to avoid invalid errors when binding listen addresses, I defer calling `Listen` (passing `nil` to the swarm initializer instead) until after the transports have been added.

I know this will be deprecated by the upcoming transport refactor, but figured it could be useful in the meantime!

### DIFF
--- a/libp2p.go
+++ b/libp2p.go
@@ -200,12 +200,17 @@ func newWithCfg(ctx context.Context, cfg *Config) (host.Host, error) {
 		ps.AddPubKey(pid, cfg.PeerKey.GetPublic())
 	}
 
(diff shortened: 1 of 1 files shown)


whyrusleeping:
Ah yeah, good catch. I'll leave this one to @Stebalien to merge as it may conflict with his changes

whyrusleeping:
@bigs @Stebalien is this still a valid change post refactor?

bigs:
It's not! Closing. Tomorrow, I'll check to see what I need to change in my example so we can merge that.

Stebalien:
We'll need to fix the tor transport.

PR #332, opened 2018-05-23
Add updated chat example with tor transport

bigs:
This PR introduces an updated chat example with support for DHT bootstrapping (no longer relies on the IPFS gateway nodes) and usage of the tor transport, as provided by OpenBazaar's `go-onion-transport`.

In order for tor support to function in its current implementation, this patch depends on the changes introduced in my other PR, #331. This dependency exists because the current `Host` creation interface offered by this library doesn't actually add user-supplied transports to the swarm after creation. If #331 is deemed pointless or otherwise unsavory, I can refactor this example to create the swarm and host more manually, though this would be a significant departure from @whyrusleeping 's original gist and would up the complexity of the demo. Happy to do whatever is preferred!

Furthermore, I'm not sure if we want to keep the old `chat/` example around, but I didn't want to blow it away without asking first! Happy to rename this folder to replace the old example if that's desired.


bigs:
@whyrusleeping something odd happened with the merge here, but this is generally good to go. i've tested that the tor example is working properly under the fork. for the time being, depends on OpenBazaar/go-onion-transport#3. if the merge history is of concern, i can just kill this and graft the code in again. no important change history to preserve with an example like this.

whyrusleeping:
@bigs Yeah, whats going on here github? Its showing what looks like us merging master into some other older branch

whyrusleeping:
@bigs maybe you rebased on an old master or something?

whyrusleeping:
As a related aside, @karalabe is working on https://github.com/ipsn/go-libtor which is a statically compiling, fully embedded tor library in go. Might be worth checking out.

karalabe:
I managed to make it build on linux (amd64, 386, arm64, arm) and android (amd64, 386, arm64, arm) with `gomobile` over the weekend (didn't have time to run tests on android yet, so it builds, not sure if correctly). I don't have an osx machine at all to test with, and windows I'm not a big fan of ;P

That said, I'd gladly accept PRs to the repo ;P The entire thing is assembled by this nightly Travis cron wrapper https://github.com/ipsn/go-libtor/blob/master/build/wrap.go.

bigs:
@whyrusleeping i've un-borked this branch (through some evil git magics). should be good to go 😄

Stebalien:
Looks like this needs a `go vet` run.

bigs:
@Stebalien addressed three context leak concerns w/ go vet

Stebalien:
The tests are failing as the onion transport hasn't been imported. See https://github.com/libp2p/go-libp2p/pull/381.

bigs:
@Stebalien ahhhhh totally. want me to PR this to the examples repo?

Stebalien:
Yes please!

bigs:
moved to examples

PR #349, opened 2018-06-06
add a link to options.go in the host example

zaibon:
so reader can see all the supported configuration for the
host constructor


bigs:
@zaibon Thanks! Definitely helpful.

PR #358, opened 2018-06-12
improve error message in libp2p reflection-based transport option

Stebalien:


Stebalien:
Mostly. I wish there were a way to ask the compiler *why* a type doesn't implement an interface. Mismatched dependency versions (e.g., two different peer ID versions) can cause this to error with "PerfectlyValidTransport doesn't implement Transport". :disappointed:

PR #365, opened 2018-06-23
Add Go standard dep support for libp2p

iouts:
Currently the go-libp2p are using `gx` for dependency management which have some disadvantages when new dev env setup:
* Windows user must have cygwin-like env for `make`.
* Must install gx + gx-go.
* The gx will rewrite go code's import to hash value  which make hard to do change management.
* The gx deps rely on IPFS.  The IPFS gateway hard to visit(ex China).  Otherwise, you will need install IPFS as daemon for install deps.
* Even `gx --verbose install --global` success developer still need deal with dependency issues to make [example](https://github.com/libp2p/go-libp2p/tree/master/examples) work.

It is easy to add [dep+vendor](https://github.com/golang/dep) to make it feel better.  Now it is very quick to setup the dev env by
1. clone the codes.
2. `dep ensure`
3. `go build ./examples/echo`


lanzafame:
@iouts probably isn't the best time to move to `dep` right now, considering the flux in the space with the acceptance of the `vgo` proposal. 

I agree the developer workflow could be improved, but I would say that this isn't an ideal time for this kind of change due to the flux and how much `gx` is used across `libp2p`.

Stebalien:
So, we don't (and won't) vendor our dependencies internally because, we have a *lot* of small packages and checking 0.5-1.0M lines of code into each of our repositories is just not going to happen.

However, you absolutely *can* use go-libp2p without `gx`. You'll just want to vendor the deps in your own package.

---

We're also working on improving `gx`. Eventually, I'd like to make autogenerate configurations for other package managers for compatibility. However, we have no intention of ever vendoring deps.

---

Feel free to continue asking questions. However, I'm going to close this PR as vendoring is a solution we've considered and discarded.

PR #366, opened 2018-06-24
Routed hosts should search for new addresses if the available adresses do not work issue#351

jmintb:
Currently if a routed host attempts a connection, it would only search for new addresses if no addresses were present in the peerinfo for that peer. 
This means that if the connection using the known addresses failed, there was no attempt to search for new addresses so that the connection might succeed.

This pr alters the Connect function for the routed host, so it searches for new addresses if it cannot connect using the known addresses. If new addresses are found, it attempts to connect using these.

I have also implemented a test that covers this case. It uses a mock test router which keeps peerinfo in a map and implements the Routing interface. Using the mock router makes it easy to manipulate the available peerinfo and avoids depending on an actual routing package.

### DIFF
--- a/p2p/host/routed/routed.go
+++ b/p2p/host/routed/routed.go
@@ -58,8 +58,11 @@ func (rh *RoutedHost) Connect(ctx context.Context, pi pstore.PeerInfo) error {
 
 	// Check if we have some addresses in our recent memory.
 	addrs := rh.Peerstore().Addrs(pi.ID)
+	hasSearched := false // Used to determine if we have used to router to search for addresses.
+
 	if len(addrs) < 1 {
 		// no addrs? find some with the routing system.
+		hasSearched = true
 		var err error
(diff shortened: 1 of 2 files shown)


Stebalien:
I'm not sure if this will work:

1. Due to our backoff logic in go-libp2p-swarm, I'm kind of surprised the tests work. If we fail a dial, we should refuse to dial for a period of time.
2. It has to *retry* all the old addresses.

Before we can fix this, we may need to fix https://github.com/libp2p/go-libp2p-swarm/issues/37. That way, we can backoff on a per-multiaddr basis.

jmintb:
In that case should I leave this until https://github.com/libp2p/go-libp2p-swarm/issues/37 is fixed?

Stebalien:
`libp2p.New` uses `go-libp2p-swarm` under the hood. I'll have to look at those tests in depth to figure out what's going on.

Issue #377, opened 2018-07-25
Default listeners


Stebalien:
It would be nice to enable listening by default (i.e., fill in some reasonable listen addresses) as not doing this tends to confuse users. However, we'll want to make it easy to *turn this off*.

upperwal:
@Stebalien I am on it.

Adding a new element in this array with a new default function
https://github.com/libp2p/go-libp2p/blob/9e5ef7a4e6ced5eb5eb9982b2a0541b331ba5387/defaults.go#L59-L83

```go
var DefaultListenAddrs = func(cfg *Config) error {
	defaultListenAddr, err := multiaddr.NewMultiaddr("/ip4/0.0.0.0/tcp/0")
	if err != nil {
		return err
	}
	return cfg.Apply(ListenAddrs(defaultListenAddr))
}
```

> However, we'll want to make it easy to turn this off.

What do you mean by this? If a user adds a new ListenAddr will it be appended to ```ListenAddrs  []ma.Multiaddr``` or the default will be overwritten?

Stebalien:
I mean we need to make it easy to override the default. Putting it in the `deafults` variable with a correct "fallback" function will do the trick. However, we'll also have to think about what we want to do if the user manually specifies transports as we can't currently just "do the right thing" for random transports.

upperwal:
Got it. For now I am adding ```/ip4/0.0.0.0/tcp/0``` as the default ListenAddrs.

Stebalien:
Thanks @upperwal!

PR #378, opened 2018-07-25
Added a Default listener

upperwal:
Solving #377 

Added IPV4 - TCP on a random port as a default listener (```/ip4/0.0.0.0/tcp/0```)

@Stebalien can you please review.


defaults.go
```diff
@@ -52,6 +53,15 @@ var RandomIdentity = func(cfg *Config) error {
 	return cfg.Apply(Identity(priv))
 }
 
+// DefaultListenAddrs configures libp2p to use default listen address
+var DefaultListenAddrs = func(cfg *Config) error {
+	defaultListenAddr, err := multiaddr.NewMultiaddr("/ip4/0.0.0.0/tcp/0")
```
Stebalien: This should also support `ip6`.

defaults.go
```diff
@@ -80,6 +90,10 @@ var defaults = []struct {
 		fallback: func(cfg *Config) bool { return cfg.Peerstore == nil },
 		opt:      DefaultPeerstore,
 	},
+	{
+		fallback: func(cfg *Config) bool { return cfg.ListenAddrs == nil },
```
Stebalien: We should probably only apply these if we are using the default transports. I'd put this *before* the default transport section and check to see if `Transports` is nil.

upperwal:
@Stebalien done.

libp2p_test.go
```diff
@@ -39,6 +40,39 @@ func TestInsecure(t *testing.T) {
 	h.Close()
 }
 
+func TestDefaultListenAddrs(t *testing.T) {
```
Stebalien: Should have a "correctly set the default listeners" test (i.e., make sure this actually works).

upperwal: Something like this?
```go
h, err := New(ctx)
if err != nil {
	t.Fatal(err)
}
if len(h.Addrs()) != 2 {
	t.Error("expected 2 default listen addrs")
}
h.Close()
```

Is length of ```h.Addrs()``` sufficient for tests or should they be compared with the actual strings?

upperwal:
@Stebalien Is jenkins only responsible for tests w/o compiling?

libp2p_test.go
Stebalien: That's probably good enough. Thanks!

Stebalien: I think that test error is due to how we expand ALL addresses using the interface addresses in `Host.Addrs()`. Try calling `h.Network().ListenAddresses()`. In that case, we should even be able to verify that they are the correct addresses.

upperwal:
@Stebalien My assumption was incorrect. I was assuming that if ```New``` is executed with two listener addresses, something like

```go
host, err := libp2p.New(
	ctx,
	libp2p.ListenAddrStrings("ip4/0.0.0.0/tcp/0", "ip6/0.0.0.0/tcp/0"),
)
```

then ```len(host.Addrs())``` would be 2 where one would be for ```ip4``` interface and the other one for ```ip6``` interface but it turns out that because I don't have an ipv6 connection ```host.Addrs()``` corresponds to my ```ipv4``` interface with one being ```127.0.0.1``` and the other one ```192.168.x.x```. 

For ```ip6/0.0.0.0/tcp/0```, ```AddListenAddr``` returns ```no suitable address found``` because I don't have any ```ipv6``` connection.

So all my tests are incorrect. What do you suggest. How should I test?

Test could be designed in a way that if either one of ```/ip4/0.0.0.0/tcp/x``` or ```/ip6/0.0.0.0/tcp/x``` is found in ```host.Network().ListenAddresses()``` it would be a success. This could be done by searching for ```/ip4/0.0.0.0/tcp/``` and ```/ip6/0.0.0.0/tcp/``` string in ```host.Network().ListenAddresses()```. Do you agree?

Stebalien:
I'd check for `/ip4/0.0.0.0/tcp/x` in `Network().ListenAddresses()` and leave it at that. Nobody disables IPv4.

defaults.go
```diff
@@ -52,6 +53,23 @@ var RandomIdentity = func(cfg *Config) error {
...
 }
 
+// DefaultListenAddrs configures libp2p to use default listen address
+var DefaultListenAddrs = func(cfg *Config) error {
+	defaultIP4ListenAddr, err := multiaddr.NewMultiaddr("/ip4/0.0.0.0/tcp/0")
+	if err != nil {
+		return err
+	}
+
+	defaultIP6ListenAddr, err := multiaddr.NewMultiaddr("/ip6/0.0.0.0/tcp/0")
```
Stebalien: This should be `/ip6/::/tcp/0` (sorry for not noticing earlier). You'll also have to fix the test. `0.0.0.0` will use the IP4 over IP6 ALL address.

upperwal: aah, my bad.

PR #380, opened 2018-07-27
always reset ping streams when done

Stebalien:
Currently, we leak a stream. We could use FullClose but, IMO, it's not worth it. We might as well just reset and walk away.

### DIFF
--- a/p2p/protocol/ping/ping.go
+++ b/p2p/protocol/ping/ping.go
@@ -44,19 +44,14 @@ func (p *PingService) PingHandler(s inet.Stream) {
 		select {
 		case <-timer.C:
 			log.Debug("ping timeout")
-			s.Reset()
 		case err, ok := <-errCh:
 			if ok {
 				log.Debug(err)
-				if err == io.EOF {
-					s.Close()
-				} else {
-					s.Reset()
-				}
 			} else {
 				log.Error("ping loop failed without error")
 			}
 		}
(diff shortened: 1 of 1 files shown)


Stebalien:
@marten-seemann thoughts?

marten-seemann:
I guess it works this way, but we're kind of misusing `Reset` here, don't we? `Reset` is intended for abnormal stream termination (e.g. a peer sends me stuff that I don't need).

Stebalien:
Yeah, that's why I pulled you into the conversation. In this case, I think reset may be the correct solution because we only bail when:

1. The other side closes the stream (they shouldn't do that).
2. The other side sends something we don't want (they shouldn't do that).
3. The user interrupts us (we should terminate immediately).

marten-seemann:
Ok, that means that in `Ping` we have to close the stream on `ctx.Done()`, and reset under all other conditions.

Thinking about this, pings should be independent of each other, so we really should open a new stream for every ping. I opened an issue (#391) for that.

Stebalien:
> Ok, that means that in Ping we have to close the stream on ctx.Done(), and reset under all other conditions.

Well, the context can still be canceled at any time. I guess we could close nicely if the context is canceled between pings but I don't really see any reason not to reset then. The reason *to* reset is that:

1. That way, we don't have to wait around for an EOF.
2. The other side doesn't care either way.

PR #381, opened 2018-07-27
extract libp2p examples to go-libp2p-examples

Stebalien:
Repo: https://github.com/libp2p/go-lib2p-examples


Stebalien:
To avoid dependency issues, I'd like to move examples to a separate repo. This will allow us to:

1. Depend on go-libp2p-kad-dht (which depends on go-libp2p).
2. Depend on the onion transport without adding it to the gx deps.
3. Not use gx so as to avoid confusing new users.

PR #388, opened 2018-08-08
reset when we fail in identify

Stebalien:


Stebalien:
1. I think this *is* the correct way to do this.
2. This works around a half-closed-stream bug in the mock stream system made visible by the recent change to multistream: https://github.com/multiformats/go-multistream/pull/27

PR #395, opened 2018-08-15
Issue #370 Check if empty peer ID

### LINKED ISSUE #370: print better error when trying to open stream to empty peer ID
Maybe i'm just being dumb. But it took me about 10 minutes to debug the fact that I was trying to open a new stream to an empty peerID.

It kept complaining about not being able to dial a peer. Which I should have picked up on it failing to dial `<peer.ID >` as meaning it was an empty peer ID... but I didnt.

kishansagathiya:
On trying to open a new stream against an empty peer ID, we get an
error saying that failed to dial a peer.

This commit changes that to a more informative error message:
`Empty peer ID`

Should be merged after https://github.com/libp2p/go-libp2p-peer/pull/34
Fixes #370

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -307,6 +307,11 @@ func (h *BasicHost) RemoveStreamHandler(pid protocol.ID) {
 // to create one. If ProtocolID is "", writes no header.
 // (Threadsafe)
 func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.ID) (inet.Stream, error) {
+	err := p.Validate()
+	if err != nil {
+		return nil, err
+	}
+
 	pref, err := h.preferredProtocol(p, pids)
 	if err != nil {
 		return nil, err


kishansagathiya:
Currently all a lot of tests are failing with
```
p2p/host/basic/basic_host.go:310:10: p.Validate undefined (type peer.ID has no field or method Validate)
```
despite using modified go-libp2p-peer. My guess is that this will be fixed after https://github.com/libp2p/go-libp2p-peer/pull/34 is merged.

Stebalien:
The issue is that we use a tool called `gx` for dependency management so we need to bubble the update up the dependency tree.

---

I'm not sure if this is the right place for this fix. Personally, I'd check in `Swarm.dialPeer` (github.com/libp2p/go-libp2p-swarm). That way, we catch all attempted dials to invalid peers.

kishansagathiya:
On go-libp2p-swarm https://github.com/libp2p/go-libp2p-swarm/pull/84

Stebalien:
>  Though this one too fails with the same error

It will. Now that I've merged your go-libp2p-peer PR, I'll bubble the dependency updates (although I may wait a day as it's easier to do that kind of thing in batches).

Stebalien:
Superseded by the swarm PR.

PR #400, opened 2018-08-21
Ensure duplicate transports are filtered

rargulati:
Without adding the transport to the set, we'll never identify if there
are duplicates in the slice of secure transports provided by the config.

### DIFF
--- a/config/security.go
+++ b/config/security.go
@@ -61,6 +61,7 @@ func makeSecurityTransport(h host.Host, tpts []MsSecC) (security.Transport, erro
 		if _, ok := transportSet[tptC.ID]; ok {
 			return nil, fmt.Errorf("duplicate security transport: %s", tptC.ID)
 		}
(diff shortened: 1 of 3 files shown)


config/security.go
```diff
@@ -61,6 +61,7 @@ func makeSecurityTransport(h host.Host, tpts []MsSecC) (security.Transport, erro
 		if _, ok := transportSet[tptC.ID]; ok {
 			return nil, fmt.Errorf("duplicate security transport: %s", tptC.ID)
 		}
+		transportSet[tptC.ID] = tptC
```
Stebalien: This should be `= struct{}{}`.

Stebalien:
Thanks for catching this! Mind also fixing the same issue in `muxer.go` and adding some simple tests?

Stebalien:
Thanks!

PR #403, opened 2018-08-22
feat: use package-table

hacdias:
@diasdavid as I'm not much familiar with each package, I might have put some under the wrong category. Please take a look!

PR #404, opened 2018-08-24
README: fix go-libp2p-blankhost row

raulk:
Tiny follow-up from #403; the fix hadn't made it to the README.md

License: MIT
Signed-off-by: Raúl Kripalani <raul@protocol.ai>


raulk:
@hacdias – mind to take a look here? Tiny follow-up. Not sure if a PR is needed, but just in case ;-)

PR #407, opened 2018-08-30
implement ping/2.0.0, which uses a new stream for every ping

### LINKED ISSUE #391: use a separate stream for every ping
Pings should be delivered independent of each other. It doesn't make a difference on transports that introduce HoL between streams, but it does on transports that offer streams at the transport level (i.e. QUIC).

marten-seemann:
Fixes #391.

This PR adds a new ping protocol `ping/2.0.0`. This ping protocol uses a separate stream for every ping, thereby getting rid of the head-of-line blocking present in `ping/1.0.0`, where all pings were sent on the same stream. Obviously, this will only make any real-world difference when using QUIC, since multiplexing streams on a single TCP connection reintroduces head-of-line blocking between streams.

In order to not break compatibility, nodes still support both the `ping/1.0.0` and the `ping/2.0.0` stream handler for incoming pings. For incoming pings, we first try `ping/2.0.0`, and if that fails fall back to `ping/1.0.0`. We should remove the `ping/1.0.0` code in a later release, when we're confident that the majority of nodes was upgraded to support `ping/2.0.0`.

### DIFF
--- a/p2p/protocol/ping/ping.go
+++ b/p2p/protocol/ping/ping.go
@@ -12,67 +12,59 @@ import (
 	host "github.com/libp2p/go-libp2p-host"
 	inet "github.com/libp2p/go-libp2p-net"
 	peer "github.com/libp2p/go-libp2p-peer"
+	multistream "github.com/multiformats/go-multistream"
 )
 
 var log = logging.Logger("ping")
 
+// PingSize is the size of ping in bytes
 const PingSize = 32
 
-const ID = "/ipfs/ping/1.0.0"
+// ID is the protocol ID of the ping/2.0.0 protocol
+// This version uses a new stream for every ping.
+const ID = "/ipfs/ping/2.0.0"
 
-const pingTimeout = time.Second * 60
+// IDLegacy is the protocol ID of the ping/1.0.0 protocol
+// This version uses a single stream for all pings.
+const IDLegacy = "/ipfs/ping/1.0.0"
 
 type PingService struct {
 	Host host.Host
 }
 
+// NewPingService creates a new PingService.
+// It uses supports both ping/1.0.0 and ping/2.0.0 for incoming pings
 func NewPingService(h host.Host) *PingService {
 	ps := &PingService{h}
+	h.SetStreamHandler(IDLegacy, ps.PingHandlerLegacy)
 	h.SetStreamHandler(ID, ps.PingHandler)
 	return ps
 }
 
-func (p *PingService) PingHandler(s inet.Stream) {
+// PingHandler handles a ping/2.0.0 ping.
+func (ps *PingService) PingHandler(s inet.Stream) {
+	defer s.Close()
+
 	buf := make([]byte, PingSize)
+	if _, err := io.ReadFull(s, buf); err != nil {
+		log.Debug(err)
+		return
+	}
 
-	errCh := make(chan error, 1)
-	defer close(errCh)
-	timer := time.NewTimer(pingTimeout)
-	defer timer.Stop()
-
-	go func() {
-		select {
-		case <-timer.C:
-			log.Debug("ping timeout")
-		case err, ok := <-errCh:
-			if ok {
-				log.Debug(err)
-			} else {
-				log.Error("ping loop failed without error")
-			}
-		}
-		s.Reset()
-	}()
-
-	for {
-		_, err := io.ReadFull(s, buf)
-		if err != nil {
-			errCh <- err
-			return
-		}
-
-		_, err = s.Write(buf)
-		if err != nil {
-			errCh <- err
-			return
-		}
-
-		timer.Reset(pingTimeout)
+	if _, err := s.Write(buf); err != nil {
+		log.Debug(err)
+		return
 	}
 }
 
+// Ping pings a peer.
+// It first attempts to use ping/2.0.0, and falls back to ping/1.0.0
+// if the peer doesn't support it yet.
 func (ps *PingService) Ping(ctx context.Context, p peer.ID) (<-chan time.Duration, error) {
 	s, err := ps.Host.NewStream(ctx, p, ID)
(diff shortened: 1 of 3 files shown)


Stebalien:
I'm not really sure what the *advantage* of this is, other than being (in theory) slightly simpler. If we were implementing this from scratch, had quic from the beginning, and had a better protocol negotiation protocol, I would have done it this way. However, we already have a ping protocol so this seems like we're introducing a bunch of complexity for no gain.

marten-seemann:
The advantage is that we're getting more accurate RTT measurements when packet loss occurs.
The additional complexity is only temporary (as long as we keep `ping/1.0.0` for backwards compatibility reasons).

Stebalien:
> The advantage is that we're getting more accurate RTT measurements when packet loss occurs.

I see, no retransmission.

---

Let's put this on hold until we have a better multistream select and we've merged the QUIC implementation.

p2p/protocol/ping/ping.go
```diff
@@ -12,92 +12,90 @@ import (
...
+		return
 	}
 }
 
+// Ping pings a peer.
+// It first attempts to use ping/2.0.0, and falls back to ping/1.0.0
+// if the peer doesn't support it yet.
 func (ps *PingService) Ping(ctx context.Context, p peer.ID) (<-chan time.Duration, error) {
 	s, err := ps.Host.NewStream(ctx, p, ID)
+	if err == multistream.ErrNotSupported {
```
Stebalien: This should be `ps.Host.NewStream(ctx, p, ID, LegacyID)`. You can then check the protocol we *actually chose*, possibly passing the stream to the legacy driver.

marten-seemann: That's really convenient, I didn't know that `NewStream` can do that.

Stebalien: Yeah, it allows us to negotiate iff we need to.

Stebalien:
LGTM but on-hold till we can make use of it.

p2p/protocol/ping/ping.go
```diff
@@ -16,88 +16,85 @@ import (
...
+// It first attempts to use ping/2.0.0, and falls back to ping/1.0.0
+// if the peer doesn't support it yet.
 func (ps *PingService) Ping(ctx context.Context, p peer.ID) (<-chan time.Duration, error) {
-	s, err := ps.Host.NewStream(ctx, p, ID)
+	s, err := ps.Host.NewStream(ctx, p, ID, IDLegacy)
 	if err != nil {
 		return nil, err
 	}
+	if s.Protocol() == IDLegacy {
+		return ps.PingLegacy(ctx, p)
```
raulk: It seems that `PingLegacy` opens a second stream. Is there a reason why we can't use `s`, which has negotiated the legacy protocol already?

p2p/protocol/ping/ping.go
```diff
@@ -16,88 +16,85 @@ import (
...
+			t, err := ping(s)
+			if err != nil {
+				log.Debugf("ping error: %s", err)
+				return
+			}
+
+			ps.Host.Peerstore().RecordLatency(p, t)
+			select {
+			case out <- t:
+				s, err = ps.Host.NewStream(ctx, p, ID)
```
raulk: It's a pity that we have to incur in protocol negotiation with every single PING :-(

I wonder if we have thought of introducing a concept of "templated streams" or something like that, where new streams declare themselves as being a child prototype of an existing one, hence bypassing proto negotiation? (CC @Stebalien @whyrusleeping @bigs)

I know the PING protocol specifies that a random payload will be sent and echoed back by the responder, but if we are _merely_ trying to measure RTT, the ACK from the [proto handshake](https://github.com/multiformats/go-multistream/blob/master/client.go#L74) would suffice for that measurement?

p2p/protocol/ping/ping_legacy.go
```diff
@@ -0,0 +1,113 @@
...
+			errCh <- err
+			return
+		}
+
+		timer.Reset(pingTimeout)
+	}
+}
+
+// PingLegacy pings a peer using ping/1.0.0
+func (ps *PingService) PingLegacy(ctx context.Context, p peer.ID) (<-chan time.Duration, error) {
```
raulk: Would it make sense to explicitly refer to the version number? If we end up with a v3 at some point, the term "legacy" would become ambiguous. I much rather call the files `ping_v1.go`, `ping_v2.go`, as well as the functions. That also reduces cognitive load on the reader, as it's one less association they need to make (legacy == v1).

p2p/protocol/ping/ping.go
Stebalien: This'll only have to negotiate once. Once we've learned the supported protocol, we remember it and never have to negotiate again.

Stebalien:
> > The advantage is that we're getting more accurate RTT measurements when packet loss occurs.

> I see, no retransmission.

Actually, I'm not sure about this anymore. We'll *still* have retransmissions and head of line blocking isn't going to be an issue for us (in QUIC at least) with the *current* ping protocol because we fully complete each ping before we start the next. That is:

* With the current protocol, we start the timer, send a ping, wait for a response, stop the timer; repeat. Dropped packets should only affect one round trip (assuming QUIC).
* With the new ping protocol, we open a new stream to do this but *still* open each stream in series. A dropped packet will affect the timing in the exact same way.

As @raulk pointed out, the benefit of the current protocol is that we negotiate the protocol up-front, once. While protocol negotiation should be "free" (assuming we already know our peer's protocols), it would be *nice* to not rely on it (for exact timing). That is, I'm *fine* relying on this assumption for performance, but relying on it for *correctness* seems a bit fishy.

marten-seemann:
You're right. I missed that we're just sending a single ping at a time, so HoL shouldn't be an issue.

PR #408, opened 2018-08-30
use the new DisableLogging mdns option, don't turn off logging globally

### LINKED ISSUE #392: don't turn off logging
https://github.com/libp2p/go-libp2p/blob/cc70c4be36f765ba9fb92d6e0528722773af607d/p2p/discovery/mdns.go#L68 turns off logging completely.
quic-go uses the go standard logger (but introduces a log level which can be set via an env variable, so quic-go is quiet by default). However, this line prevents quic-go from logging at all.

marten-seemann:
Fixes #392. Depends on https://github.com/whyrusleeping/mdns/pull/3.

We should merge this *before* rolling out an IPFS release with QUIC support. This fix is necessary to be able to activate QUIC logging.

### DIFF
--- a/p2p/discovery/mdns.go
+++ b/p2p/discovery/mdns.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"errors"
 	"io"
-	"io/ioutil"
-	golog "log"
 	"net"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


marten-seemann:
@Stebalien: whyrusleeping/mdns#3 was merged. We can merge this PR now.

PR #412, opened 2018-09-07
Deduplicate addresses of host output

cannium:
Sometimes `ipfs id` show duplicate addresses since peer observed addresses are likely same as interface listen addresses or NAT external addresses.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -474,10 +474,11 @@ func mergeAddrs(addrLists ...[]ma.Multiaddr) (uniqueAddrs []ma.Multiaddr) {
 	exists := make(map[string]bool)
 	for _, addrList := range addrLists {
 		for _, addr := range addrList {
-			if exists[addr.String()] {
+			k := string(addr.Bytes())
+			if exists[k] {
 				continue
 			}
-			exists[addr.String()] = true
+			exists[k] = true
 			uniqueAddrs = append(uniqueAddrs, addr)
 		}
 	}
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -469,26 +469,41 @@ func (h *BasicHost) Addrs() []ma.Multiaddr {
 	return h.addrs(h.AllAddrs())
 }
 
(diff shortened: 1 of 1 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -469,26 +469,41 @@ func (h *BasicHost) Addrs() []ma.Multiaddr {
 	return h.addrs(h.AllAddrs())
 }
 
+// mergeAddrs merges input address lists, leave only unique addresses
+func mergeAddrs(addrLists ...[]ma.Multiaddr) (uniqueAddrs []ma.Multiaddr) {
+	exists := make(map[string]bool)
```
Stebalien: It would be nice to both preallocate this and `uniqueAddrs` as we know *approximately* the number of addresses we'll get. It's probably not *that* much of a performance issue but still best practices (IIRC, we'll run this function every time we connect to a new peer).

p2p/host/basic/basic_host.go
```diff
@@ -469,26 +469,41 @@ func (h *BasicHost) Addrs() []ma.Multiaddr {
...
+	for _, addrList := range addrLists {
+		for _, addr := range addrList {
+			if exists[addr.String()] {
```
Stebalien: We can actually just call `string(addr)`. calling `addr.String()` will format the address as a string (e.g., `/ip4/.../tcp/.../ipfs/...`) which will be significantly more expensive than just converting the bytes representation to a string.

Stebalien:
Nice find. Thanks!

p2p/host/basic/basic_host.go
Stebalien: Basically, `addr.String()` requires a bunch of allocations and complicated logic while `string(thing)` is almost free (at most one allocation).

p2p/host/basic/basic_host.go
cannium: I did a small performance test https://gist.github.com/cannium/dab43ef7b2fc7fb272b7ca583403920a, and found the version without pre-allocation is faster. So I'll leave it as is.

Issue #416, opened 2018-09-08
Keep more information when tracking observed addresses (in identify)


Stebalien:
We should record:

* The complete mapping. That is, instead of mapping "observed address -> observer", we should map "internal address -> observed address -> observer".
* The connection direction. We can now get this information by calling `conn.Stat()`.

Given this, we should be able to better manage this list.

Issue #421, opened 2018-09-10
Remove unreachable addresses from the observed address list.


Stebalien:
Depends on: https://github.com/libp2p/go-libp2p/issues/416

Given more information about our observed addresses, we can try removing addresses observed from *outbound* connections once we start getting confirmations from *inbound* connections.

Also related: https://github.com/libp2p/go-libp2p-autonat

PR #427, opened 2018-09-18
Track more info for observed addresses

cannium:
Fix https://github.com/libp2p/go-libp2p/issues/416

### DIFF
--- a/p2p/protocol/identify/obsaddr.go
+++ b/p2p/protocol/identify/obsaddr.go
@@ -62,7 +62,6 @@ func (oas *ObservedAddrSet) Addrs() (addrs []ma.Multiaddr) {
 	}
 
 	now := time.Now()
-	filteredAddrMap := make(map[string][]*ObservedAddr)
 	for local, observedAddrs := range oas.addrs {
 		filteredAddrs := make([]*ObservedAddr, 0, len(observedAddrs))
 		for _, a := range observedAddrs {
@@ -74,9 +73,8 @@ func (oas *ObservedAddrSet) Addrs() (addrs []ma.Multiaddr) {
 				}
 			}
 		}
-		filteredAddrMap[local] = filteredAddrs
+		oas.addrs[local] = filteredAddrs
 	}
-	oas.addrs = filteredAddrMap
 	return addrs
 }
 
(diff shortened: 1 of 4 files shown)


cannium:
ping @vyzo @Stebalien for review.
Add internal address and connection direction to observed address info

vyzo:
is this actually useful?

cannium:
This patch pave the way for https://github.com/libp2p/go-libp2p/issues/421, so peers would see less addresses that are actually not reachable.

Stebalien:
> is this actually useful?

Very. Without this:

1. We don't know the direction of the mapping. This matters on many NATs.
2. The actual mapping. That is, which internal address maps to which external address. Currently, we only record our external addresses but we don't actually *remember* which internal addresses they appear to map to.

Stebalien:
I believe the we need to map internal to external to observation. This currently records the *first* observed internal address (and the first observed direction) but that may not be consistent. That is, I believe we'd want something like:

```
map[<internal address>]map[<observed/external address>]Observation { SeenBy: map[<peer addr>] {time, direction}}
```

That way, for each internal address, we can list out all the observed external addresses and figure out where from and how they've been observed.

cannium:
I think an internal address could be observed as multiple external addresses, not the opposite. Say my node is listening on `10.0.0.1:4001`, I can configure multiple port mappings for it, so observed address could be `200.0.0.1:5000`, `200.0.0.1:6000`, etc.

Stebalien:
You can get both (technically). These observed addresses include observed addresses from *outbound* connections. It's possible for two sequential (non-overlapping) connections from different internal ports to end up using the same external port.

cannium:
That's possible, but from my experience, is very rare. A reasonable NAT implementation would avoid port reuse in a very short time, so if two internal ports are mapped to the same external port, it's highly possible that the first connection is considered as closed.

Stebalien:
It's also possible for a peer to lie. IMO, that's the real issue here. The only thing we *know* is the internal address. The external address is just an observation being reported by a peer.

cannium:
That's still a `1:n` relationship between internal address and external addresses, the map would be like `external address -> internal address`, or you mean we should do `internal address -> list of external addresses`?

Stebalien:
The latter. Specifically: `internal -> {external -> { observer -> [time, direction] } }`. That way, for any port I'm explicitly listening on, I can see which external addresses have been reported. For each external address, I can see who reported them, when, and the connection direction.

cannium:
I think I misunderstood your intention, updated now. Another idea occurs to me that `ObservedAddrSet` might be better tracked together with peerstore.

cannium:
Any comments? Still invalid implementation?

p2p/protocol/identify/obsaddr.go
```diff
@@ -41,60 +46,85 @@ func (oa *ObservedAddr) TryActivate(ttl time.Duration) bool {
...
+		for _, a := range observedAddrs {
+			// leave only alive observed addresses
+			if now.Sub(a.LastSeen) <= oas.ttl {
+				filteredAddrs = append(filteredAddrs, a)
+				if a.activated(oas.ttl) {
+					addrs = append(addrs, a.Addr)
+				}
+			}
 		}
+		filteredAddrMap[local] = filteredAddrs
```
Stebalien: Any reason not to update `oas.addrs` directly?

p2p/protocol/identify/obsaddr.go
```diff
@@ -41,60 +46,85 @@ func (oa *ObservedAddr) TryActivate(ttl time.Duration) bool {
...
+	observedAddr := &ObservedAddr{
+		Addr: observed,
+		SeenBy: map[string]observation{
+			observerString: {
+				seenTime:      now,
+				connDirection: direction,
+			},
+		},
+		LastSeen: now,
+	}
```
Stebalien: This logic feels a bit out-of-order; we're constructing something we may not use. How about:

```
// construct an `observation` (not the full `observedAddr`).

observedAddrs := oas.addrs[localString] // no need for "found"

// Will be a no-op if observedAddrs is nil
for i, previousObserved := range observedAddrs {
    // ...
}

// Now construct the full `ObservedAddr` and insert it.
```

I haven't tried reordering this so it may not be any cleaner but I *think* it'll be a bit easier to follow.

cannium:
Re-run CI, tests passed now.

PR #441, opened 2018-09-28
Enable relay by default in New

### LINKED ISSUE #440: Add relay dialer and listener (not hop) to the default libp2p constructor
Seems like a pretty reasonable thing to do. Thoughts @Stebalien @bigs @raulk @vyzo ?

vyzo:
Closes #440

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -44,8 +44,9 @@ type Config struct {
 	Insecure           bool
 	Protector          pnet.Protector
 
-	Relay     bool
-	RelayOpts []circuit.RelayOpt
+	RelayCustom bool
+	Relay       bool
+	RelayOpts   []circuit.RelayOpt
 
 	ListenAddrs  []ma.Multiaddr
 	AddrsFactory bhost.AddrsFactory
--- a/options.go
+++ b/options.go
@@ -201,11 +201,21 @@ func AddrsFactory(factory config.AddrsFactory) Option {
 	}
 }
 
-// EnableRelay configures libp2p to enable the relay transport.
+// EnableRelay configures libp2p to enable the relay transport with configuration options.
 func EnableRelay(options ...circuit.RelayOpt) Option {
 	return func(cfg *Config) error {
+		cfg.RelayCustom = true
 		cfg.Relay = true
 		cfg.RelayOpts = options
+		return nil
+	}
+}
+
+// DisableRelay configures libp2p to disable the relay transport
+func DisableRelay() Option {
+	return func(cfg *Config) error {
+		cfg.RelayCustom = true
+		cfg.Relay = false
 		return nil
 	}
 }
@@ -245,9 +255,15 @@ func NATManager(nm config.NATManagerC) Option {
 // NoListenAddrs will configure libp2p to not listen by default.
 //
 // This will both clear any configured listen addrs and prevent libp2p from
-// applying the default listen address option.
+// applying the default listen address option. It also disables relay, unless the
+// user explicitly specifies with an option, as the transport creates an implicit
+// listen address that would make the node diable through any relay it was connected to.
 var NoListenAddrs = func(cfg *Config) error {
 	cfg.ListenAddrs = []ma.Multiaddr{}
+	if !cfg.RelayCustom {
+		cfg.RelayCustom = true
+		cfg.Relay = false
+	}
 	return nil
 }
 
--- a/options.go
+++ b/options.go
@@ -257,7 +257,7 @@ func NATManager(nm config.NATManagerC) Option {
 // This will both clear any configured listen addrs and prevent libp2p from
 // applying the default listen address option. It also disables relay, unless the
 // user explicitly specifies with an option, as the transport creates an implicit
-// listen address that would make the node diable through any relay it was connected to.
+// listen address that would make the node dialable through any relay it was connected to.
 var NoListenAddrs = func(cfg *Config) error {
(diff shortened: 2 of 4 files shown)


config/config.go
```diff
@@ -44,8 +44,9 @@ type Config struct {
 	Insecure           bool
 	Protector          pnet.Protector
 
-	Relay     bool
-	RelayOpts []circuit.RelayOpt
+	RelayCustom bool
```
raulk: `RelayCustom` seems like an bookkeeping flag we use only to track if the user has set relay options explicitly. Can we make it private?

options.go
```diff
@@ -245,9 +255,15 @@ func NATManager(nm config.NATManagerC) Option {
 // NoListenAddrs will configure libp2p to not listen by default.
 //
 // This will both clear any configured listen addrs and prevent libp2p from
-// applying the default listen address option.
+// applying the default listen address option. It also disables relay, unless the
+// user explicitly specifies with an option, as the transport creates an implicit
+// listen address that would make the node diable through any relay it was connected to.
```
raulk: nit: s/diable/dialable

options.go
```diff
@@ -245,9 +255,15 @@ func NATManager(nm config.NATManagerC) Option {
...
 var NoListenAddrs = func(cfg *Config) error {
 	cfg.ListenAddrs = []ma.Multiaddr{}
+	if !cfg.RelayCustom {
+		cfg.RelayCustom = true
```
raulk: Is it necessary to set `RelayCustom` to true here?

config/config.go
vyzo: I would have liked  this to, but the readers are in a different package.

options.go
vyzo: sure, will fix; typos are bugs of sorts :)

options.go
vyzo: yes, otherwise the fallback will set it to default and enable it.

config/config.go
raulk: 👍

raulk: Maybe a docstring would clarify the role of this flag, also stating it's not intended for explicit use.

Stebalien: Maybe we should change `Relay` to an enum (ish) with `Enabled`, `Disabled`, and `Default` variants.

raulk: I thought about that too, but didn’t propose it because it’d imply breakage.

Stebalien:
So, I'd *like* libp2p to be as silent as possible on start (by default). Unfortunately, go-libp2p-circuit tries to figure out if each peer supports relying which adds some additional chatter (that's probably unnecessary in most cases). Could we add an option to disable this first? That is, add an option to disable:

1. Discovering if our peers are willing to relay.
2. Dialing `/p2p-circuit/ipfs/QmTarget` (no relay specified) addresses.

Really, I've still yet to see a use-case for using a random peer as a relay when dialing. In theory, it could help bypass firewalls and censorship but it's effectively a random walk. It may make sense to have some "discover N relays" function that tries to setup N relay "entry points" so we can announce them, but that's a different issue (and something that will happen periodically, not on every connection event).

config/config.go
Stebalien: This struct actually has an explicit warning saying that it's not stable. Users should use the *options*, they shouldn't modify the config directly.

vyzo: ok, I got started in making it an enum and it's extremely cumbersome. 
Better to keep the RelayCustom contraption.

vyzo:
Then again it could be ok as an _option_ -- but it can't be the default, as it has far reaching implications to break unspecific relay dialing by default.

vyzo:
We could also have an option to specify fixed relays, which can be passed to p2p-circuit as known relay nodes so that we can at least have working unspecific relay address dialing.

Stebalien:
> Or do you want to disable the unspecific relay address dialing entirely?
...
> Then again it could be ok as an option -- but it can't be the default, as it has far reaching implications to break unspecific relay dialing by default.

How is the current behavior of unspecified relay addresses useful outside of testing and demos?

---

Basically, I see exactly two use-cases for unspecified relay addresses:

1. Autorelay. In theory, we could insert an address transformer between the routing layer and the swarm, prepending `/p2p-circuit/` to every address once we have enough relays. However, IMO, relay discovery would likely happen in the autorelay module (e.g., on connect, call `circuit.CanHop(p)`).
2. Testing/demoing relay.

vyzo:
Opened #444 for autorelay module.

Stebalien:
So, we should definitely leave the current behavior as the default in go-ipfs until we have a better relay discovery mechanism (as users may be relying on it). It's just that that doing this by default will affect all users of libp2p itself.

vyzo:
Yeah, if we turn it off right now we will break relay dialing completely in ipfs.
So until we have autorelay ready and deployed, we'll have to keep the discovery (and then probably for a while longer).

But we can certainly have a `circuit.DisableRelayDiscovery` option that libp2p users could pass for quiet operation.

vyzo:
Or maybe it should be the default and modify ipfs to pass `EnableRelayDiscovery` so that it keeps working as it currently does.

Stebalien:
> Or maybe it should be the default and modify ipfs to pass EnableRelayDiscovery so that it keeps working as it currently does.

I'd prefer this one. That way, both IPFS and libp2p continue doing what they've always done. The only difference is that libp2p nodes would, by default, be able to dial/accept relayed connections.

PR #443, opened 2018-10-01
downgrade protocol mux failure log to debug

vyzo:
There is no real information conveyed (so it's really not INFO) and there is tons of it.

The more pertinent question is why I see so many of them -- probably because of prevalence of undialable addrs


Stebalien:
It could either be that the connection was killed or that the stream was reset. I'm not sure which.

Stebalien:
Actually, do you have time to look into why we're getting these and document that in an issue? It would be nice if we actually *knew*.

vyzo:
In the meantime, should we merge this? It's really spamming at INFO level.

Stebalien:
Yeah... fine. As long as we don't forget...

vyzo:
I'll open issue for the investigation, I don't have any good ideas yet.

PR #449, opened 2018-10-05
fix relay-specific address dialing in routed host

### LINKED ISSUE #448: Dialing specific relay addresses through routed hosts
We have a problem with dialing specific relay addresses of the form `/ipfs/QmRelay/p2p-circuit/ipfs/QmTarget`: the relay transport is attached to the basic host, so it will not be able to find `QmRelay` unless we have connected to the peer in the past or it somehow made its way to our Peerstore.
This will become of major importance as we enable relay by default, phase out unspecific relay addresses and move to specific relay address advertising (see #441 and #444).

Suggested Solution:
- Modify the routed host so that it comprehends relay addrs in `Connect`, and issues a `FindPeer` for the relay when it sees a relay-specific address.

vyzo:
Fixes #448.

### DIFF
--- a/p2p/host/routed/routed.go
+++ b/p2p/host/routed/routed.go
@@ -90,6 +90,11 @@ func (rh *RoutedHost) Connect(ctx context.Context, pi pstore.PeerInfo) error {
 			continue
 		}
 
+		if relayID == pi.ID {
+			// it's an old style p2p-circuit address that includes the peer
+			continue
+		}
+
 		if len(rh.Peerstore().Addrs(relayID)) > 0 {
 			// we already have addrs for this relay
 			continue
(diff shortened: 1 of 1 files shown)


p2p/host/routed/routed.go
```diff
@@ -67,6 +68,43 @@ func (rh *RoutedHost) Connect(ctx context.Context, pi pstore.PeerInfo) error {
...
+			continue
+		}
+
+		relayAddrs, err := rh.findPeerAddrs(ctx, relayID)
+		if err != nil {
+			log.Debugf("failed to find relay %s: %s", relay, err)
+			continue
+		}
+
+		rh.Peerstore().AddAddrs(relayID, relayAddrs, pstore.AddressTTL)
```
Stebalien: This should use the temp TTL.

vyzo: The rationale wass to keep it around long enough to do all the resolutions in the address set, which may take more than 10s because of potentially multiple `findPeers`.
Also, it would be nice to avoid having to re-lookup for multiple dials to the same relay.

vyzo: will make it use the temp ttl for now, and see if this causes problems.

Issue #375, opened 2018-07-23
NAT Traversal in chat example


Harrywang55666:
Hi, I'm new to libp2p
https://github.com/libp2p/go-libp2p/tree/master/examples/chat
This example shows peer A (A behind NAT) connect peer B (B has public IP), but B couldn't connect A. I don't think it is a typical P2P example. How could a peer to connect a peer which behinds NAT or firewall, give us an example please.

Stebalien:
The almost-always works answer is:

1. Enable the circuit relay protocol (the `EnableRelay` option).
2. Pick some set of nodes that you want to use as relays.
3. Advertise these addresses using the `AddrsFactory` option.

Eventually, we'd like to automate the second two pieces but that means we'd need an automated way to find and pick good relays.

---

Canned NAT traversal explanation: We have three nat-traversal solutions at the moment.

**UPnP/NATPortMap**

When NAT traversal is enabled (in go-libp2p, pass the `NATPortMap()` option to the libp2p constructor), libp2p will use UPnP and NATPortMap to ask the NAT's router to open and forward a port for libp2p. If your router supports either UPnP or NATPortMap, this is by far the best option.

**STUN/hole-punching**

LibP2P has it's own version of the "STUN" protocol using peer-routing, external address discovery, and reuseport. LibP2P:

1. Uses the same source port for both dialing and listening (reuseport).
2. Receives observed address information from all connected peers (external address discovery).
3. Publishes these addresses to the network (peer routing).

* On *some* NATs, the first feature allows us to accept new inbound connections on the source port used by the *router* (the "mapped" source port).
* The second feature allows us to discover this mapped source port and our external address.
* The third feature allows us to *tell* the network about this external address.

**TURN-like protocol**

Finally, we have a TURN like protocol called p2p-circuit. This protocol allows libp2p nodes to "proxy" through other p2p nodes. Unfortunately, neither go-libp2p nor js-libp2p *announce* p2p-circuit addresses at the moment (by default). This means that it doesn't just automagically bypass NATs (yet).

The next steps here are to:

1. Have all nodes behind NATs connect to "hub" nodes that volunteer to act as p2p-circuit relays.
2. Announce to the network that we're accepting relay connections from these nodes.

**Corporate Firewalls**

While technically a different issue, this still falls under the category of working around meddlesome middle boxes.

[...]

PR #454, opened 2018-10-17
autorelay

### LINKED ISSUE #444: autorelay module
It is becoming apparent that advertising unspecific relay addresses and dialing them through trying known relays is not a scalable solution (in the number of relays).

We need an `autorelay` module that attaches to a host and does the following:
- Discovers relays through a high level interface -- eg DHT providers, rendezvous
- Selects relays (eg for minimum latency) and keeps connections open to them, even in the face of connection manager pruning
- Advertises relay-specific addresses.

Similarly, HopRelays will have to advertise themselves through the appropriate discovery mechanism.

vyzo:
Implements autorelay; Closes #444.

Depends on:
- [X] https://github.com/libp2p/go-libp2p-discovery/pull/1
- [X] https://github.com/libp2p/go-libp2p-autonat/pull/2
- [X] https://github.com/multiformats/go-multiaddr-net/pull/46

TBD:
- [X] gx import discovery
- [X] gx import autonat
- [X] gx update multiaddr-net
- [X] Update `New` constructor to automatically create relay/autorelay/routed hosts
- [X] tests!

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,10 +5,13 @@ import (
 	"fmt"
 
 	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+	relay "github.com/libp2p/go-libp2p/p2p/host/relay"
+	routed "github.com/libp2p/go-libp2p/p2p/host/routed"
 
 	logging "github.com/ipfs/go-log"
 	circuit "github.com/libp2p/go-libp2p-circuit"
 	crypto "github.com/libp2p/go-libp2p-crypto"
+	discovery "github.com/libp2p/go-libp2p-discovery"
 	host "github.com/libp2p/go-libp2p-host"
 	ifconnmgr "github.com/libp2p/go-libp2p-interface-connmgr"
 	pnet "github.com/libp2p/go-libp2p-interface-pnet"
@@ -16,6 +19,7 @@ import (
 	inet "github.com/libp2p/go-libp2p-net"
 	peer "github.com/libp2p/go-libp2p-peer"
 	pstore "github.com/libp2p/go-libp2p-peerstore"
+	routing "github.com/libp2p/go-libp2p-routing"
 	swarm "github.com/libp2p/go-libp2p-swarm"
 	tptu "github.com/libp2p/go-libp2p-transport-upgrader"
 	filter "github.com/libp2p/go-maddr-filter"
@@ -30,6 +34,13 @@ type AddrsFactory = bhost.AddrsFactory
 
 // NATManagerC is a NATManager constructor.
 type NATManagerC func(inet.Network) bhost.NATManager
+
+type Routing interface {
+	routing.ContentRouting
+	routing.PeerRouting
+}
+
+type RoutingC func(host.Host) (Routing, error)
 
 // Config describes a set of settings for a libp2p node
 //
@@ -58,6 +69,8 @@ type Config struct {
 	Reporter    metrics.Reporter
 
 	DisablePing bool
+
+	Routing RoutingC
 }
 
 // NewNode constructs a new libp2p Host from the Config.
@@ -99,8 +112,8 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 		swrm.Filters = cfg.Filters
 	}
 
-	// TODO: make host implementation configurable.
-	h, err := bhost.NewHost(ctx, swrm, &bhost.HostOpts{
+	var h host.Host
+	h, err = bhost.NewHost(ctx, swrm, &bhost.HostOpts{
 		ConnManager:  cfg.ConnManager,
 		AddrsFactory: cfg.AddrsFactory,
 		NATManager:   cfg.NATManager,
@@ -158,7 +171,34 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 		return nil, err
 	}
 
-	// TODO: Configure routing (it's a pain to setup).
+	if cfg.Routing != nil {
+		router, err := cfg.Routing(h)
+		if err != nil {
+			h.Close()
+			return nil, err
+		}
+
+		if cfg.Relay {
+			discovery := discovery.NewRoutingDiscovery(router)
+
+			hop := false
+			for _, opt := range cfg.RelayOpts {
+				if opt == circuit.OptHop {
+					hop = true
+					break
+				}
+			}
+
+			if hop {
+				h = relay.NewRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery)
+			} else {
+				h = relay.NewAutoRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery)
+			}
+		}
+
+		h = routed.Wrap(h, router)
+	}
+
 	// TODO: Bootstrapping.
 
 	return h, nil
--- a/config/config.go
+++ b/config/config.go
@@ -35,12 +35,7 @@ type AddrsFactory = bhost.AddrsFactory
 // NATManagerC is a NATManager constructor.
 type NATManagerC func(inet.Network) bhost.NATManager
 
-type BasicRouting interface {
-	routing.ContentRouting
-	routing.PeerRouting
-}
-
-type RoutingC func(host.Host) (BasicRouting, error)
+type RoutingC func(host.Host) (routing.PeerRouting, error)
 
 // Config describes a set of settings for a libp2p node
 //
@@ -178,21 +173,24 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 			return nil, err
 		}
 
-		if cfg.Relay {
-			discovery := discovery.NewRoutingDiscovery(router)
+		crouter, ok := router.(routing.ContentRouting)
+		if ok {
+			if cfg.Relay {
+				discovery := discovery.NewRoutingDiscovery(crouter)
 
-			hop := false
-			for _, opt := range cfg.RelayOpts {
-				if opt == circuit.OptHop {
-					hop = true
-					break
+				hop := false
+				for _, opt := range cfg.RelayOpts {
+					if opt == circuit.OptHop {
+						hop = true
+						break
+					}
 				}
-			}
 
-			if hop {
-				h = relay.NewRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery)
-			} else {
-				h = relay.NewAutoRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery)
+				if hop {
+					h = relay.NewRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery)
+				} else {
+					h = relay.NewAutoRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery)
+				}
 			}
 		}
 
--- a/config/config.go
+++ b/config/config.go
@@ -35,12 +35,12 @@ type AddrsFactory = bhost.AddrsFactory
 // NATManagerC is a NATManager constructor.
 type NATManagerC func(inet.Network) bhost.NATManager
 
-type Routing interface {
+type BasicRouting interface {
 	routing.ContentRouting
 	routing.PeerRouting
 }
 
-type RoutingC func(host.Host) (Routing, error)
+type RoutingC func(host.Host) (BasicRouting, error)
 
 // Config describes a set of settings for a libp2p node
 //
--- a/options.go
+++ b/options.go
@@ -260,6 +260,17 @@ func Ping(enable bool) Option {
 	}
 }
 
+// Routing will configure libp2p to use routing.
+func Routing(rt config.RoutingC) Option {
+	return func(cfg *Config) error {
+		if cfg.Routing != nil {
+			return fmt.Errorf("cannot specified multiple routing options")
+		}
+		cfg.Routing = rt
+		return nil
+	}
+}
+
 // NoListenAddrs will configure libp2p to not listen by default.
 //
 // This will both clear any configured listen addrs and prevent libp2p from
--- a/options.go
+++ b/options.go
@@ -264,7 +264,7 @@ func Ping(enable bool) Option {
 func Routing(rt config.RoutingC) Option {
 	return func(cfg *Config) error {
 		if cfg.Routing != nil {
-			return fmt.Errorf("cannot specified multiple routing options")
+			return fmt.Errorf("cannot specify multiple routing options")
 		}
 		cfg.Routing = rt
 		return nil
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -59,9 +59,10 @@ type BasicHost struct {
 	ids        *identify.IDService
 	pings      *ping.PingService
 	natmgr     NATManager
-	addrs      AddrsFactory
 	maResolver *madns.Resolver
 	cmgr       ifconnmgr.ConnManager
+
+	AddrsFactory AddrsFactory
 
 	negtimeout time.Duration
 
@@ -106,11 +107,11 @@ type HostOpts struct {
 // NewHost constructs a new *BasicHost and activates it by attaching its stream and connection handlers to the given inet.Network.
 func NewHost(ctx context.Context, net inet.Network, opts *HostOpts) (*BasicHost, error) {
 	h := &BasicHost{
-		network:    net,
-		mux:        msmux.NewMultistreamMuxer(),
-		negtimeout: DefaultNegotiationTimeout,
-		addrs:      DefaultAddrsFactory,
-		maResolver: madns.DefaultResolver,
+		network:      net,
+		mux:          msmux.NewMultistreamMuxer(),
+		negtimeout:   DefaultNegotiationTimeout,
+		AddrsFactory: DefaultAddrsFactory,
+		maResolver:   madns.DefaultResolver,
 	}
 
 	h.proc = goprocessctx.WithContextAndTeardown(ctx, func() error {
@@ -136,7 +137,7 @@ func NewHost(ctx context.Context, net inet.Network, opts *HostOpts) (*BasicHost,
 	}
 
 	if opts.AddrsFactory != nil {
-		h.addrs = opts.AddrsFactory
+		h.AddrsFactory = opts.AddrsFactory
 	}
 
 	if opts.NATManager != nil {
@@ -254,6 +255,11 @@ func (h *BasicHost) newStreamHandler(s inet.Stream) {
 	log.Debugf("protocol negotiation took %s", took)
 
 	go handle(protoID, s)
+}
+
+// PushIdentify pushes an identify update through the identify push protocol
+func (h *BasicHost) PushIdentify() {
+	h.ids.Push()
 }
 
 // ID returns the (local) peer.ID associated with this Host
@@ -474,7 +480,7 @@ func (h *BasicHost) ConnManager() ifconnmgr.ConnManager {
 // Addrs returns listening addresses that are safe to announce to the network.
 // The output is the same as AllAddrs, but processed by AddrsFactory.
 func (h *BasicHost) Addrs() []ma.Multiaddr {
-	return h.addrs(h.AllAddrs())
+	return h.AddrsFactory(h.AllAddrs())
 }
 
 // mergeAddrs merges input address lists, leave only unique addresses
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -258,6 +258,7 @@ func (h *BasicHost) newStreamHandler(s inet.Stream) {
 }
 
 // PushIdentify pushes an identify update through the identify push protocol
+// Warning: this interface is unstable and may disappear in the future.
 func (h *BasicHost) PushIdentify() {
 	h.ids.Push()
 }
--- a/p2p/host/relay/autorelay.go
+++ b/p2p/host/relay/autorelay.go
@@ -116,7 +116,8 @@ func (h *AutoRelayHost) findRelays(ctx context.Context) {
 	h.mx.Unlock()
 
 	limit := 20
-	for ; need > limit; limit *= 2 {
+	if need > limit/2 {
+		limit = 2 * need
 	}
 
 	dctx, cancel := context.WithTimeout(ctx, 60*time.Second)
--- a/p2p/host/relay/autorelay.go
+++ b/p2p/host/relay/autorelay.go
@@ -10,6 +10,7 @@ import (
 	basic "github.com/libp2p/go-libp2p/p2p/host/basic"
 
 	autonat "github.com/libp2p/go-libp2p-autonat"
+	_ "github.com/libp2p/go-libp2p-circuit"
 	discovery "github.com/libp2p/go-libp2p-discovery"
 	host "github.com/libp2p/go-libp2p-host"
 	inet "github.com/libp2p/go-libp2p-net"
@@ -23,7 +24,17 @@ var (
 	DesiredRelays = 3
 
 	BootDelay = 60 * time.Second
+
+	unspecificRelay ma.Multiaddr
 )
+
+func init() {
+	var err error
+	unspecificRelay, err = ma.NewMultiaddr("/p2p-circuit")
+	if err != nil {
+		panic(err)
+	}
+}
 
 // AutoRelayHost is a Host that uses relays for connectivity when a NAT is detected.
 type AutoRelayHost struct {
@@ -61,7 +72,7 @@ func (h *AutoRelayHost) hostAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
 	if h.addrs != nil && h.autonat.Status() == autonat.NATStatusPrivate {
 		return h.addrs
 	} else {
-		return h.addrsF(addrs)
+		return filterUnspecificRelay(h.addrsF(addrs))
 	}
 }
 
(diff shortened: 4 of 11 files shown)


vyzo:
cc @whyrusleeping @mgoelzer

vyzo:
rebased on master

mgoelzer:
cc @dignifiedquire 
Fyi this is in progress now, following the recent NAT-ing problems.

options.go
```diff
@@ -252,6 +252,17 @@ func NATManager(nm config.NATManagerC) Option {
 	}
 }
 
+// Routing will configure libp2p to use routing.
+func Routing(rt config.RoutingC) Option {
+	return func(cfg *Config) error {
+		if cfg.Routing != nil {
+			return fmt.Errorf("cannot specified multiple routing options")
```
magik6k: s/specified/specify

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,278 @@
...
+	h.mx.Lock()
+	if len(h.relays) >= DesiredRelays {
+		h.mx.Unlock()
+		return
+	}
+	need := DesiredRelays - len(h.relays)
+	h.mx.Unlock()
+
+	limit := 20
+	for ; need > limit; limit *= 2 {
```
magik6k: This feels like there should be a better way to do this

vyzo: I guess we could just take some factor of need (say 2x) and go from there.

vyzo: simplified in https://github.com/libp2p/go-libp2p/pull/454/commits/f495b8a4b70ea59067262ea35b38c1f797c911ac

mgoelzer:
@vyzo, this is coming along really nicely.  Before final merge, can you add some docs about how it works with usage examples for downstreams?  Also, I’m thinking we should get a reviewer from Filecoin to verify a build off this branch fixes the problem they were having that raised the priority of this? (or if it doesn’t, we should open new issues to cover what else they need for their use case) Maybe @phritz?

vyzo:
We still need to bubble up some recent changes in multiaddr-net before we can merge, but that's just procedural at this point.

Re: documentation
It all works through the libp2p constructor.
If you construct a host with relay enabled (now the default) and supply the new `Routing` option, then it will instantiate an autorelay host that will discover relays through routing and advertise relay addrs when the presence of impenetrable NAT is detected.
If you enable relay with OptHop, which is the option to construct an actual relay, then it will construct a relay host that advertises relay capabilities through the routing object.
The `Routing` option takes a constructor function that builds a routing instance given a host; this can simply be an instance of `IpfsDHT`, but anything that implements the routing interface is acceptable.

Note that an autonat service instance is needed to detect the presence of NAT -- see https://github.com/libp2p/go-libp2p-autonat-svc/pull/1 (which is also waiting for the bubble up before merge).

So, in a nutshell:
- you need to instantiate an AutoNAT service somewhere in the network, and connect to it so that it is discovered by the system (this will probably run on the bootstrappers)
- the actual relays need to be constructed with `libp2p.New(libp2p.EnableRelay(circuit.OptHop), libp2p.Routing(...))`
- the autorelayed hosts need to be constructed with `libp2p.New(libp2p.Routing(...))`

vyzo:
I think we want to rename the `config.Routing` interface to `BasicRouting` and type alias it in the libp2p top module (for better usability).
Maybe we should define it in `go-libp2p-routing` though, it's the combination of `PeerRouting` and `ContentRouting`.

phritz:
@vyzo I think @mgoelzer was proxying the request from go-filecoin around docs. Basically when we went to try to implement relaying ourselves we spent a lot of time trying to figure out how to do it. We had confusion around:
- semantics of constructor flags (hop vs relay what to configure where -- do we need both hop and relay on the natted peer or just relay? what does "active" mean and do we need to set it?)
- whether we needed to advertise relay addresses manually and how to do that (yes use sprintf to do it manually and use addrsfactor -- which precipitated "how do you use addrsfactory?")
- what format the relay address should have
- whether we should be using /p2p or /ipfs, the docs are inconsistent
- whether we needed to manually keep connected to relay nodes open (i think yes)
- how relaying interacted with dht
- if we have multiple relay nodes how/if peers could find the right one to connect through (i think you encode the relay node you are connected to in the address that you advertise)
- whether we should suppress the non-relayed addresses
- whether the ten zillion lines of errors that libp2p logs were actual errors (thing X that should not happen did happen) or whether they were really more like warnings that are to be expected in the course of events and we could ignore (eg, couldn't connect to a host, connection reset, etc)

We ended up spending many hours trying to get this working, even with the benefit of early input from why and stebalien, and relied mostly on trying to find places in the code that used this stuff and trial and error. 

It would be great to have a little more explanation of how this works (https://github.com/libp2p/go-libp2p/pull/454#issuecomment-431551627 is a great start -- promote it to documentation!) along with some sample code. *The sample code would be most helpful!* This is the case we were after:
- I want natted peers to be able to talk to each other via a fixed set of relay nodes known ahead of time. I also want not-natted peers to be able to talk to all peers, ideally we dont have to relay connections for peers who can speak directly to each other.

The case of talking to each other via a set of discoverable relay nodes is a nice to have, as is auto-detection of whether my peer is reachable or needs to use a relay (I think you have this WIP).

Docs we used:
- https://github.com/libp2p/go-libp2p/issues/375 <= this was actually the most helpful, consider promoting a lot of its content to first-class documentation
- https://github.com/libp2p/specs/tree/master/relay <= we were confused by the discussion of address format, a little more explanation might be helpful

vyzo:
rebased on master

vyzo:
I will add a `doc.go` in `p2p/host/relay` documenting the usage of autorelay.

mgoelzer:
@vyzo Thanks for adding the docs PR, and @phritz  thanks for engaging in this discussion to clarify what the questions are.

Long term, I'd like us to find a way that each major PR has documentation independent of its source code files or Github PR comments page.  We could either house this in a single location like https://docs.libp2p.io (which we could drive off GH pages or something) or we need to put it in each module's repo (e.g., https://github.com/libp2p/go-libp2p-circuit/blob/master/README.md).  We can hire a professional docs writer to clean up the writing and index it somewhere, as long as we capture the facts somewhere.

Since `docs.libp2p.io` is blocked on infra, my suggestion for now would be to take the discussion here and use it to fill in https://github.com/libp2p/go-libp2p-circuit/blob/master/README.md.  Would that be workable?

p2p/protocol/identify/id.go
```diff
@@ -140,6 +145,26 @@ func (ids *IDService) responseHandler(s inet.Stream) {
 	go inet.FullClose(s)
 }
 
+func (ids *IDService) pushHandler(s inet.Stream) {
+	ids.responseHandler(s)
+}
+
+func (ids *IDService) Push() {
```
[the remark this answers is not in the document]
vyzo: This is only called when the address set changes, and it reaches all connected peers provided they support the push protocol. 
There is no periodic push or need to bootstrap.

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,279 @@
...
+	host "github.com/libp2p/go-libp2p-host"
+	inet "github.com/libp2p/go-libp2p-net"
+	peer "github.com/libp2p/go-libp2p-peer"
+	pstore "github.com/libp2p/go-libp2p-peerstore"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr-net"
+)
+
+var (
+	DesiredRelays = 3
```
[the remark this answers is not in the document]
vyzo: It already is configurable, it's a variable :)

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,279 @@
...
+	need := DesiredRelays - len(h.relays)
+	h.mx.Unlock()
+
+	limit := 20
+	if need > limit/2 {
+		limit = 2 * need
+	}
+
+	dctx, cancel := context.WithTimeout(ctx, 60*time.Second)
+	pis, err := discovery.FindPeers(dctx, h.discover, "/libp2p/relay", limit)
```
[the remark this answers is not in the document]
vyzo: ok, will do. I don't think we need a version for this, relay is pretty fixed functionality.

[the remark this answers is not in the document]
vyzo: called it `RelayRendezvous` and it's a constant now.

vyzo:
> Since docs.libp2p.io is blocked on infra, my suggestion for now would be to take the discussion here and use it to fill in https://github.com/libp2p/go-libp2p-circuit/blob/master/README.md. Would that be workable?

@mgoelzer Sure, but let's merge first. Then we can add pretty comprehensive documentation there.

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,279 @@
...
+	cancel()
+	if err != nil {
+		log.Debugf("error discovering relays: %s", err.Error())
+		return
+	}
+
+	// TODO better relay selection strategy; this just selects random relays
+	//      but we should probably use ping latency as the selection metric
+	shuffleRelays(pis)
```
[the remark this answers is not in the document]
vyzo: moved the relay selection out of line, into a `selectRelays` function.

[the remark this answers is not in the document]
vyzo: Ultimately, I think we also need to take into account relay load in the selection strategy.
The selection metric could be `latency * load`, and select to minimize.
But this will need some changes in the relay protocol (ie ability to measure load and report that), so I am going to leave it as a big fat TODO and stay with randomized selection for initial deployment.

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,289 @@
...
+	for _, pi := range pis {
+		h.mx.Lock()
+		if _, ok := h.relays[pi.ID]; ok {
+			h.mx.Unlock()
+			continue
+		}
+		h.mx.Unlock()
+
+		cctx, cancel := context.WithTimeout(ctx, 60*time.Second)
+		err = h.Connect(cctx, pi)
```
raulk: We should be tagging these connections in the connection manager. 

Also, are we covering the case where a connection to a relay in `h.relays` has been terminated by the conn manager?

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,289 @@
...
+	shuffleRelays(pis)
+	return pis
+}
+
+func (h *AutoRelayHost) updateAddrs() {
+	h.doUpdateAddrs()
+	h.PushIdentify()
+}
+
+func (h *AutoRelayHost) doUpdateAddrs() {
```
raulk: docs here please ;-) `doUpdateAddrs` is not self descriptive, i.e. which addrs? where are the updated?

p2p/host/relay/autorelay.go
raulk: Nevermind about point 2, I see this is covered in the Notifee. Point 1 stays active.

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	need := DesiredRelays - len(h.relays)
+	h.mx.Unlock()
+
+	limit := 20
+	if need > limit/2 {
+		limit = 2 * need
+	}
+
+	dctx, cancel := context.WithTimeout(ctx, 60*time.Second)
```
raulk: Edge case: we could get overlapping executions of findRelays if user has changed the value of `autonat.AutoNATRefreshInterval` (default: 15 min) to be lower than the DHT lookup (fixed 60 secs). It's unlikely but feasible; probably worth covering for robustness.

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,289 @@
...
+func (h *AutoRelayHost) updateAddrs() {
+	h.doUpdateAddrs()
+	h.PushIdentify()
+}
+
+func (h *AutoRelayHost) doUpdateAddrs() {
+	h.mx.Lock()
+	defer h.mx.Unlock()
+
+	addrs := filterUnspecificRelay(h.addrsF(h.AllAddrs()))
```
raulk: Can't we call `Addrs()` instead, which already does the `AddrFactory` drill for us?

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,289 @@
...
+				continue
+			}
+		} else {
+			raddrs = append(raddrs, addr)
+		}
+	}
+
+	// add relay specific addrs to the list
+	for _, pi := range h.relays {
+		circuit, err := ma.NewMultiaddr(fmt.Sprintf("/ipfs/%s/p2p-circuit", pi.ID.Pretty()))
```
raulk: Can we use `/p2p` which is equivalent for `/ipfs` in multiaddr?

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,289 @@
...
+		if err != nil {
+			panic(err)
+		}
+
+		for _, addr := range pi.Addrs {
+			if !manet.IsPrivateAddr(addr) {
```
raulk: I'm not sure I follow this logic. The resulting self addresses via relays would end up taking the form: `/ipfs/QmRelay/p2p-circuit/{maddr of *relay* if not private}`

Are we sure we want to encapsulate the relay's addrs instead of our own?

p2p/host/relay/autorelay_test.go
```diff
@@ -0,0 +1,185 @@
...
+func connect(t *testing.T, a, b host.Host) {
+	pinfo := pstore.PeerInfo{ID: a.ID(), Addrs: a.Addrs()}
+	err := b.Connect(context.Background(), pinfo)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+// and the actual test!
+func TestAutoRelay(t *testing.T) {
```
raulk: Sorry to be a stickler for testing, but I think a component like this deserves more exhaustive testing at the time of its submission of the PR, or at least a commitment from the author to add more tests shortly thereafter.

Particularly I'd like to see coverage regarding:

* our actual own advertised addresses as a result of autorelay (we are only testing that there's a P_CIRCUIT component, but we should be testing it's entirety, especially because we're manipulating maddrs)
* behaviour with unspecific relays.
* multiple rounds of autorelay.
* timeouts.
* different AutoNAT statuses.
* connection terminations.
* etc.

p2p/host/relay/autorelay_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	h4, err := libp2p.New(ctx, libp2p.EnableRelay())
+
+	connect(t, h1, h3)
+	time.Sleep(3 * time.Second)
+
+	// verify that we advertise relay addrs
+	haveRelay := false
+	for _, addr := range h3.Addrs() {
+		_, err := addr.ValueForProtocol(circuit.P_CIRCUIT)
```
raulk: Let's assert the entire multiaddr, not only that a component exists.

p2p/host/relay/autorelay.go
vyzo: Sure, what's the magic incantation for the tagging?

p2p/host/relay/autorelay.go
vyzo: ugh, ok -- it's the only addrs object we manage.

p2p/host/relay/autorelay.go
vyzo: No, we can't get overlapping executions; they are only launched by the `background` goroutine, serially.
At worst, what could happen is that we get back-to-back executions.

p2p/host/relay/autorelay.go
vyzo: No, we have overriden the AddrsFactory.

p2p/host/relay/autorelay.go
vyzo: sure.

p2p/host/relay/autorelay.go
vyzo: There is nothing after the `p2p-circuit` part, it's just a transport signifier.
The relayed multiaddr will be of the form `/ipfs/QmRelay/p2p-circuit`.

vyzo: So just to be clear, it's the non-private address of the relay that is encapsulating the `/p2p/QmRelay/p2p-circuit` part, not the other way around.

p2p/host/relay/autorelay.go
bigs: [here](https://github.com/libp2p/go-libp2p-interface-connmgr/blob/master/interface.go#L18) though i'm not sure what magic string we want to use

p2p/host/relay/autorelay_test.go
bigs: this is a good use case for a testbed/daemon environmen

p2p/host/relay/autorelay.go
vyzo: done.

p2p/host/relay/autorelay.go
vyzo: what's a good value for the weight though? Is it increasing weight = less likely to be severed?

p2p/host/relay/autorelay_test.go
vyzo: yeah, I would very much see these integration tested in a testbed than having to write reams of borderline useless testing code.

p2p/host/relay/autorelay.go
vyzo: Our implementation states as such, so I'll give it a round 42.

vyzo: tagged in https://github.com/libp2p/go-libp2p/pull/454/commits/863316d1cea124fcabac77be431b3b8314bed481

p2p/host/relay/autorelay.go
vyzo: documented in https://github.com/libp2p/go-libp2p/pull/454/commits/d7bc5ddf47b92479885298e2387903787163ff07

p2p/host/relay/autorelay.go
raulk: 👍 missed that

p2p/host/relay/autorelay.go
raulk: Got it, thanks.

p2p/host/relay/autorelay_test.go
raulk: Ok, I didn't realise it was tested for dialability 👍

p2p/host/relay/autorelay_test.go
raulk: can't tell you how many times what appeared to be overtesting at the time caught regressions later on ;-)

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,289 @@
...
+	if h.addrs != nil && h.autonat.Status() == autonat.NATStatusPrivate {
+		return h.addrs
+	} else {
+		return filterUnspecificRelay(h.addrsF(addrs))
+	}
+}
+
+func (h *AutoRelayHost) background(ctx context.Context) {
+	select {
+	case <-time.After(autonat.AutoNATBootDelay + BootDelay):
```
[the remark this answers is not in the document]
whyrusleeping: Yeah, we do eventually need a better solution. Having to know that if you try doing X before Y finishes means it will fail is kinda rough.

vyzo:
rebased on master and updated the gx deps to discovery and autonat; we are no longer blocked now that the multiaddr-net stuff has bubbled up!

vyzo:
@Stebalien care to review? this is ready for merge.

Stebalien:
I know we need this feature, but we keep making that excuse and we rarely, if ever, go back and fix interfaces.

Stebalien:
I'll take a look at this again tomorrow when I'm less grumpy (spent too much time working on the commands library today).

phritz:
As a future consumer of this functionality we are happy to have it later rather than sooner if that means it makes your lives easier or puts the functionality on more principled footing. We're finding on our end that hacking stuff in has painful long-term costs.

config/config.go
```diff
@@ -31,6 +35,13 @@ type AddrsFactory = bhost.AddrsFactory
 // NATManagerC is a NATManager constructor.
 type NATManagerC func(inet.Network) bhost.NATManager
 
+type BasicRouting interface {
+	routing.ContentRouting
+	routing.PeerRouting
+}
+
+type RoutingC func(host.Host) (BasicRouting, error)
```
Stebalien: What if we had this return `routing.PeerRouting` router and then used a type switch for everything else? Basically, peer routing is the minimum functionality but, if we have content routing, we can construct a discovery service based on it. In the future, we could check to see if the router supported *other* interfaces (i.e., it may directly implement the discovery interface).

p2p/host/relay/autorelay.go
```diff
@@ -0,0 +1,301 @@
...
+	addrsF   basic.AddrsFactory
+
+	disconnect chan struct{}
+
+	mx     sync.Mutex
+	relays map[peer.ID]pstore.PeerInfo
+	addrs  []ma.Multiaddr
+}
+
+func NewAutoRelayHost(ctx context.Context, bhost *basic.BasicHost, discover discovery.Discoverer) *AutoRelayHost {
```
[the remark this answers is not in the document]
Stebalien: Do we *just* need this for pushing addrs? How difficult would some kind of evented "addrs change" system be? If we had that, we could register a callback (at the top layer) that pushes address updates (doesn't even need to be identify, whatever's simplest).

Really, I don't know what the best solution is but this whole "PushAddrs" thing feels backwards.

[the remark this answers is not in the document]
vyzo: We also need to override the AddrsFactory, besides pushing addrs.

config/config.go
```diff
@@ -31,6 +35,13 @@ type AddrsFactory = bhost.AddrsFactory
 // NATManagerC is a NATManager constructor.
 type NATManagerC func(inet.Network) bhost.NATManager
 
+type BasicRouting interface {
+	routing.ContentRouting
```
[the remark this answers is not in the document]
vyzo: Do you want to reduce this to just `PeerRouting` and cast our way to the content router?
That's doable.

config/config.go
vyzo: Actually, it's not so bad; we can pass `PeerRouting` as a minimum signifying that we want a routed host.  The additional functionality is semi optional, and as you say maybe we'll implement the discovery interface directly from the DHT one day.

config/config.go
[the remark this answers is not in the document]
vyzo: Removed the `BasicRouting` interface in https://github.com/libp2p/go-libp2p/pull/454/commits/c91a0d34406fbb70c689e1b3e8ddc06270cb0cca

config/config.go
vyzo: Reduced to `PeerRouting` + cast per your suggestion in https://github.com/libp2p/go-libp2p/pull/454/commits/c91a0d34406fbb70c689e1b3e8ddc06270cb0cca

vyzo:
So I have removed the `BasicRouting` interface altogether in favor of using the minimal `PeerRouting` and type casting for capabilities.

mgoelzer:
Preface:  always be weary of engineering suggestions from the PM 😀

>that would significantly delay deployment of this, while not adding anything to the end user.

This sounds a lot like a technical debt tradeoff question.  What I can say from the business perspective is that we're not in the kind of crisis or semi-crisis situation where it's rational to be taking on significant tech debt.  AFAIK, this is not on the critical path for FC to launch a public testnet since their early users will presumably be advanced enough to open ports on a router or cloud host (@phritz feel free to correct me).  I'm not qualified to make the engineering decision, but those are the data I have.

>I know we need this feature, but we keep making that excuse and we rarely, if ever, go back and fix interfaces.

Nobody fixes public interfaces (unless the fix is purely additive).  This one is worth thinking about imho.

vyzo:
Just to quantify the "level" of tech debt, i am pretty certain i can rewrite it in an afternoon for any future host refactoring. The code and concepts are simple, so I would say in terms of "debt" the effect is minimal.

p2p/host/basic/basic_host.go
```diff
@@ -256,6 +257,11 @@ func (h *BasicHost) newStreamHandler(s inet.Stream) {
 	go handle(protoID, s)
 }
 
+// PushIdentify pushes an identify update through the identify push protocol
+func (h *BasicHost) PushIdentify() {
```
Stebalien: Let's add a comment saying that this is an unstable interface. Ideally, we'd have an address manager, a protocol muxer, etc. that all emit events. Then, the identify service would subscribe to these events to determine when and if to push information to peers.

vyzo: Sure, will add a comment.

vyzo:
As the author of the gnarly hacks I hereby commit to making the service thingie and disentanglement happen!

vyzo:
Design discussion in #467.

@bigs @raulk any objections to merging? Let's deliver a feature at least :)
I'll add documentation in `go-libp2p-circuit` as well.

bigs:
yep. perhaps we can have a sync design session to talk about how we can make a more long-term friendly design? and perhaps we should leave some more breadcrumb comments to highlight things to change?

vyzo:
Let's have a sync session to discuss the next gen host design!
Re: more comments -- well, we already have lots of discussion in the PR and also the follow up issue, dunno if it will add anything.

bigs:
soundss great @vyzo you've got my 👍

vyzo:
rebased on master and updated deps.

raulk:
As long as we explicitly state in godoc that these interfaces are UNSTABLE, I can live with merging now and revisiting once the service abstraction is ready.

PR #458, opened 2018-10-19
Add ping service in basic host and libp2p constructor option (enabled by default)

vyzo:
Adds the ping service to basic_host, with a libp2p constructor option to control its instantiation; it's enabled by default.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -56,6 +56,9 @@ type Config struct {
 	NATManager  NATManagerC
 	Peerstore   pstore.Peerstore
 	Reporter    metrics.Reporter
+
+	PingCustom bool
+	Ping       bool
 }
 
 // NewNode constructs a new libp2p Host from the Config.
@@ -102,6 +105,7 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 		ConnManager:  cfg.ConnManager,
 		AddrsFactory: cfg.AddrsFactory,
 		NATManager:   cfg.NATManager,
+		EnablePing:   cfg.Ping,
 	})
 	if err != nil {
 		swrm.Close()
--- a/p2p/protocol/ping/ping_test.go
+++ b/p2p/protocol/ping/ping_test.go
@@ -1,4 +1,4 @@
-package ping
+package ping_test
 
 import (
 	"context"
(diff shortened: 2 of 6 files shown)


config/config.go
```diff
@@ -56,6 +56,9 @@ type Config struct {
 	NATManager  NATManagerC
 	Peerstore   pstore.Peerstore
 	Reporter    metrics.Reporter
+
+	PingCustom bool
+	Ping       bool
```
raulk: Not sure I'm liking the proliferation of the `...Custom` fields. If we make the actual flags pointers `*bool`, a `nil` could indicate no user-provided value. WDYT? Same for Relay/RelayCustom?

p2p/protocol/ping/ping_test.go
```diff
@@ -1,4 +1,4 @@
-package ping
+package ping_test
```
raulk: Why is this change needed?

vyzo: we get an import cycle without it, because the test uses basic hosts.

config/config.go
vyzo: I tried with a ternary enum, but it was disgusting to work with :)

vyzo:
IPFS already enables ping, but they construct the PingService explicitly; this change obviates the need for that construction.
It is also very useful for autorealy, as we'd like to eventually choose relays based on latency.
I don't think there is a such a great concern for ping DoS right now, but let's open an issue to discuss the implications.

config/config.go
raulk: How about modelling the ternary state like this? https://github.com/libp2p/go-libp2p/commit/4758a9552e20447bf34e7543d7607e7a5fb9ce76

vyzo: ugh, that's ugly too. I don't mind the Custom thingies that much, because the struct is only used for configuring the host and nothing else.

p2p/protocol/ping/ping_test.go
lanzafame: I will not that this is a good practice for testing a package from the perspective of a 'user' of a package as it only gives you access to the exported functionality of the `ping` package. Though it can make testing certain things more difficult especially in the domain of ipfs and libp2p.

lanzafame: I will note, that this is a good practice for testing a package from the perspective of a 'user' of a package as it only gives you access to the exported functionality of the package. Though it can make testing certain things more difficult especially in the domain of ipfs and libp2p.

config/config.go
marten-seemann: Why not `DisablePing`?

PR #481, opened 2018-11-10
gx publish 6.0.27

hannahhoward:
Updating go-ipfs-delay


Stebalien:
(thanks for catching the missing routing dep)

Issue #433, opened 2018-09-25
Proposal for NAT UDP hole-punching


cannium:
### Reason
To increase connectivity between nodes as much as we can. 

Relay is good, but is not so "peer-to-peer", and at the cost of community maintained relay servers. Also, as we have tested, many of our customers are behind some kind of NATs. Hopefully hole-punching would solve some other percents of the problem. 

### Technical details
Hole-punching is better understood at protocol level. To track an NAT mapping, a gateway(or some other device) would record at least 3 addresses(and ports): local internal address, local external address and remote external address. For an outbound packet, the gateway would replace its local internal address with local external address; for a inbound packet, the gateway would replace its local external address with local internal address. From the perspective of a internal node, the connection is between local internal address and remote external address.

If a gateway only do some translations, the problem would be much easier. But a gateway would also apply some restrictions to inbound packets, mainly for security reasons. This [wiki](https://en.wikipedia.org/wiki/Network_address_translation#Methods_of_translation) lists 4 types of NATs and their restrictions for inbound packets, and theoretically UDP hole-punching could solve 3 types. For TCP packets, I observed additional restrictions, like sequence number check, so UDP is a better choice.
 
The diagram below is a punching example.
![udp hole-punching](https://user-images.githubusercontent.com/1171383/46002031-7a7c3300-c0df-11e8-9e6a-1e0a55839e70.jpeg)

`Node1` and `Node2` are two nodes behind different NAT gateways, with internal address `192.168.0.1:4001` and `10.0.0.2:4001`, respectively. A helper node with public IP `3.0.0.1` would assist connection establishment, but not relay data for them. 

For the first step, the public IP node would see `Node1` has external address `100.100.0.1:10000`, `Node2` has `200.200.0.1:20000`, and send this information to one another side. Knowing the other's external address, `Node1` would send packets to `200.200.0.1:20000`, on receiving `Node1`'s packet, `NAT1` would allow inbound packets from `200.200.0.1:20000` (otherwise normal communication is disturbed). `Node2` and `NAT2` are the same. The first 1 or more packets might be dropped because of timing (inbound packets not already allowed), but eventually a 2-way communication would be established.

And for multiple layers of NATs, the situation is similar, only take longer time.
![multiple-layer-nat](https://user-images.githubusercontent.com/1171383/46003514-44d94900-c0e3-11e8-9336-49f6cda0683a.jpg)

### Limitations
[...]

Issue #188, opened 2017-03-31
Support for WebRTC transport


JustinDrake:
So the Duo team (working on OpenBazaar) was all excited that we can now have truly peer-to-peer communications between js-ipfs and go-ipfs using WebSockets. It turns out that, in practice, WebSockets doesn't work direct p2p. The reason is that Firefox (since always) and Chrome (since recently) only allow for `wss://` (secure WebSockets) on an `https://` page. So the Duo team has to use a `wss://` to `ws://` proxy, which truly sucks.

Our hope is that go-libp2p will support WebRTC. The Duo team would be happy to review and test the implementation. 👍

whyrusleeping:
@JustinDrake I would love to have this too. The issue is that there is no decent implementation of webrtc in go (to my knowledge). We definitely want this too.

As an aside, we're adding support for QUIC soon, which should help some (chrome will have support for QUIC, and other browsers should follow)

JustinDrake:
Crazy idea: could we use a WebRTC-to-WebSockets Node proxy, given that we have WebRTC support in Node?

> The issue is that there is no decent implementation of webrtc in go

Cool. Out of curiosity, what is wrong with https://github.com/keroserene/go-webrtc? It seems to be relatively active. Maybe it's worth talking to the team behind it?

> we're adding support for QUIC soon

Oh awesome! Is there a PR/branch/issue you can link to?

> chrome will have support for QUIC

Do you know when? (I ask because I'm expecting the launch of OpenBazaar 2.0 in a few months.)

lgierth:
> So the Duo team has to use a wss:// to ws:// proxy, which truly sucks.

Put nginx in front of the /ws listener, for SSL termination

lgierth:
> Put nginx in front of the /ws listener, for SSL termination

What I meant to imply apart from SSL termination, is: I'd strongly advise against running js-ipfs or go-ipfs things directly exposed to the internet. Better put nginx/haproxy/apache in front.

JustinDrake:
Thanks @lgierth. My colleague @jackkleeman has adapted a Go `wss` to `ws` proxy here https://github.com/duomarket/websocketproxy

Using nginx may be more scalable/resilient/well-tested. Do you happen to have an nginx config file readily available for SSL termination?

> I'd strongly advise against running js-ipfs or go-ipfs things directly exposed to the internet

Is that for security reasons? https://github.com/OpenBazaar/openbazaar-go is running go-ipfs without anything in front.

[...]

PR #490, opened 2018-11-21
[WIP] UDP-based NAT hole-punching

cannium:
This patch utilizes relayed connection as signaling channel, and run `/libp2p/punching/1.0.0` protocol on it. If both nodes are behind NAT gateways, they send self-observed QUIC addresses to each other. On receiving such messages, nodes would dial the addresses, hoping to establish connections directly.

I think the patch could be deployed and tested now,  so I publish it for review. I understand that some major refactors would happen in go-libp2p, so I'm prepared for corresponding refactors. Also as a "WIP", expect nits here and there:)

Part of https://github.com/libp2p/go-libp2p/issues/433


p2p/host/relay/autorelay.go
```diff
@@ -92,7 +92,9 @@ func (h *AutoRelayHost) background(ctx context.Context) {
 
 	for {
 		wait := autonat.AutoNATRefreshInterval
-		switch h.autonat.Status() {
+		status := h.autonat.Status()
+		h.Ids.SetNatStatus(status)
```
vyzo: that's gross -- we should embed AutoNAT in basic_host if identify needs it or pass it a reference.

vyzo:
Please don't duplicate the AutoNAT constants and functionality in identify by second hand.
Instead, instantiate it in the `basic_host` and pass it as reference.

vyzo:
Also, it's not clear to me why identify needs to have the NATStatus for this protocol.
How is this surfaced?

cannium:
> Please don't duplicate the AutoNAT constants and functionality in identify by second hand.
> Instead, instantiate it in the `basic_host` and pass it as reference.

I understand that the info `NATStatus` is owned by AutoNATService, but currently `AutoNATService` doesn't expose any method for query and I'd like to limit the patch inside `go-libp2p`(at least for now). 

> Also, it's not clear to me why identify needs to have the NATStatus for this protocol.
> How is this surfaced?

The NAT status is exchanged by Identify protocol, so we only do punching if both peers are behind NAT gateways.

vyzo:
> I understand that the info NATStatus is owned by AutoNATService, but currently AutoNATService doesn't expose any method for query and I'd like to limit the patch inside go-libp2p(at least for now).

That's not the AutoNATService you want -- move the AutoNAT client which lives in the AutoRelayedHost.
The way you have written it right now it won't work at all if the user doesn't costruct an autorelay host (ie doesn't pass the `Routing` parameter in the constructor).

p2p/protocol/punching/punching.go
```diff
@@ -0,0 +1,178 @@
...
+	stream, err := p.host.NewStream(ctx, peerID, PunchingProtocol)
+	cancel()
+	if err != nil {
+		log.Debug("Failed to create new stream to", peerID)
+		return
+	}
+	allAddresses := p.host.Addrs()
+	quicAddress := make([]string, 0, len(allAddresses))
+	for _, a := range allAddresses {
+		s := a.String()
```
vyzo: You should work with the multiaddr representation directly instead of converting to strings.

vyzo:
I think we may want to punt a bit on this, I am going to be working on the service refactoring the next few days.

cannium:
> I think we may want to punt a bit on this, I am going to be working on the service refactoring the next few days.

Sure. I'd also expect a refactor after changes landed on libp2p. I publish it earlier so maybe more people interested could help test it.

gaillard:
Is this full duplication with what is being done in https://github.com/libp2p/go-libp2p/issues/188? Is this closer timeframe?

cannium:
> Is this full duplication with what is being done in #188? Is this closer timeframe?

No. WebRTC is a standardized protocol that happen also solves NAT traversal problem. This patch is much simpler. For the time being, I'm waiting for the refactoring of libp2p.

denis4net:
Are you going to merge it into the master?

gaillard:
Does this support the symmetric to restricted address cone case?

cannium:
> Are you going to merge it into the master?

I'll try my best.
> Does this support the symmetric to restricted address cone case?

Theoretically it could handle restricted cone, but not symmetric case.

marten-seemann:
OBE (Project Flare).

PR #495, opened 2018-11-29
Add delay in initial relay advertisement to allow the dht time to bootstrap

vyzo:
This came up in https://github.com/libp2p/go-libp2p-discovery/issues/5

Also increases the test autonat identify delay to 500ms to make it less flaky in macos.

### DIFF
--- a/p2p/host/relay/relay.go
+++ b/p2p/host/relay/relay.go
@@ -31,8 +31,11 @@ func NewRelayHost(ctx context.Context, bhost *basic.BasicHost, advertise discove
 	}
 	bhost.AddrsFactory = h.hostAddrs
 	go func() {
-		time.Sleep(AdvertiseBootDelay)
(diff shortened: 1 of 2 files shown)


p2p/host/relay/relay.go
```diff
@@ -25,7 +30,10 @@ func NewRelayHost(ctx context.Context, bhost *basic.BasicHost, advertise discove
 		advertise: advertise,
 	}
 	bhost.AddrsFactory = h.hostAddrs
-	discovery.Advertise(ctx, advertise, RelayRendezvous)
+	go func() {
+		time.Sleep(AdvertiseBootDelay)
```
bigs:
```go
select {
case <-time.After(AdvertiseBootDelay):
  discovery.Advertise(ctx, advertise, RelayRendezvous)
case <-ctx.Done():
}
```

vyzo:
@bigs done.

PR #498, opened 2018-11-30
suppressing error msg for NoSecurity option

upperwal:
```identity``` displays an error if ```libp2p.NoSecurity``` option is used. This change will suppress it by also checking for private key.

### CODE REVIEW

File: p2p/protocol/identify/id.go [hunk 1]
```diff
@@ -186,9 +186,11 @@ func (ids *IDService) populateMessage(mes *pb.Identify, c inet.Conn) {
 
 	// set our public key
 	ownKey := ids.Host.Peerstore().PubKey(ids.Host.ID())
-	if ownKey == nil {
+	prKey := ids.Host.Peerstore().PrivKey(ids.Host.ID())
```
Stebalien: Let's make it clear *why* we're checking for the private key. Something like:

```go
if ownKey == nil {
  // check if we're even operating in "secure mode"
  if ids.Host.Peerstore().PrivKey(ids.Host.ID()) != nil {
    // warn
  }
} else ...
```

(this also avoid storing the private key to make it clear that we're *just* checking to see if we have it)

PR #520, opened 2019-01-14
Updating usage docs for relay/autorelay

mvid:


p2p/host/relay/autorelay_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	autonatpb "github.com/libp2p/go-libp2p-autonat/pb"
 	circuit "github.com/libp2p/go-libp2p-circuit"
 	host "github.com/libp2p/go-libp2p-host"
+	dht "github.com/libp2p/go-libp2p-kad-dht"
```
vyzo: no, we can't depend on kad-dht; this creates a circular reference.

vyzo:
also, can we put the examples in their own file? I don't want them in the tests.

mvid:
> also, can we put the examples in their own file? I don't want them in the tests.

@vyzo Unfortunately it seems that for usage examples to be properly formatted in GoDoc, they need to be in a file that ends in "_test". I can move them to something like "example_test", if you prefer

p2p/host/relay/autorelay_test.go
mvid: @vyzo I brought the `autorelay_test` file into the `relay` package so that I could use the mock functions in the usage example

vyzo: can't do that either; the reason it is not there is that it causes  a circular dependency between the top level package and the test...

mvid: I see... relay -> libp2p -> config -> relay

I put the examples in `relay_test`. This seems a little brittle, no?

vyzo: it's ok, it's more of a go failing.

p2p/host/relay/autorelay.go
```diff
@@ -54,6 +54,10 @@ type AutoRelayHost struct {
 	addrs  []ma.Multiaddr
 }
 
+// Warning:  This function should not be used directly by clients activating
+// autorelay. NewAutoRelayHost is used internally by the libp2p.config.NewNode
```
vyzo: this should be `libp2p.New` constructor, the users don't call `libp2p.conig.NewNode` directly either.

p2p/host/relay/examples_test.go
```diff
@@ -0,0 +1,32 @@
...
+	libp2p "github.com/libp2p/go-libp2p"
+
+	host "github.com/libp2p/go-libp2p-host"
+	routing "github.com/libp2p/go-libp2p-routing"
+)
+
+func ExampleNewRelayHost() {
+	ctx := context.Background()
+
+	libp2p.New(ctx, libp2p.EnableRelay())
```
vyzo: that's incorrect in autorelay context, you need to pass `EnableRelay(circuit.OptHop)` and the routing option.

p2p/host/relay/relay.go
```diff
@@ -22,7 +22,10 @@ type RelayHost struct {
 	addrsF    basic.AddrsFactory
 }
 
-// New constructs a new RelayHost
+// Warning:  This function should not be used directly by clients activating
+// relay. NewRelayHost is used internally by the libp2p.config.NewNode
```
vyzo: same here about `libp2p.New`.

p2p/host/relay/examples_test.go
vyzo: and `EnableAutoRelay` of course.

p2p/host/relay/relay.go
mvid: Corrected

p2p/host/relay/autorelay.go
mvid: Corrected

p2p/host/relay/examples_test.go
mvid: I have two examples, one for relay and one for autorelay. The autorelay example includes `EnableAutoRelay` and `Routing`. I changed the examples to provide an empty list of RelayOpts, since I can't find anywhere in the code where RelayOpts are required. 

Are they required? What is the common case for relay options?

mvid: Ok, to clarify, are there cases where libp2p consuming developers will create instances of RelayHost outside of `libp2p.New`? If so, what are they? I can add a number of examples, but my assumption was that like with AutoRelay, the supported case was only `libp2p.New`

vyzo: No, it's only through the constructor; but there are use cases where developers enable relay without autorelay.

mvid: Yea, definitely. Check `examples_test.go` again. There are two main examples, one for Relay and one for AutoRelay, via the libp2p.New constructor. I tried to use as few optional arguments as possible, but I can add RelayOpts if necessary.

vyzo: I think we need another example that constructs the relay node with autorelay.

mvid:
Right, but I’m confused. Am I not looking at a relay with autorelay example in examples_test.go:20 ?

vyzo:
It does, but I would like two examples for autorealy:
- one for constructing a "client" node
- and one for constructing an actual relay

mvid:
I see, I’ll add that soon

mvid:
We have a full Relay example in go-libp2p-examples:
https://github.com/libp2p/go-libp2p-examples/pull/49

Also working on getting a reasonable autorelay example in the examples repo:
https://github.com/libp2p/go-libp2p-examples/pull/50

Any comments on the AR branch would be appreciated

marten-seemann:
Closing this PR due too age.

PR #531, opened 2019-02-12
Use PeerRouting in autorelay to find relay peer addresses

vyzo:
Fixes a subtle bug: if the discovery module returns no addresses for the relay peers, then we might not be able to connect to them because we only have a basic host.


vyzo:
cc @ZenGround0

raulk:
Ah, went to approved but it was already done ;-)

PR #536, opened 2019-02-14
gx publish 6.0.33

vyzo:


vyzo:
cc @Stebalien: Important fixes for autorelay.

PR #539, opened 2019-02-15
Resolve addrs with P2P instead of IPFS

anacrolix:
It has no functional difference but it shakes the tree and asserts that p2p is the correct code. I did fix something ages back rewrote p2p codes to ipfs, even when they originally were not. This is a similar assertion.


raulk:
Late to the party, but LGTM. There’s probably other places in the codebase to switch too.

Issue #428, opened 2018-09-18
Handle `/quic` and `/ws` in NAT mappings


Stebalien:
The *mappings* report `/ip{4,6}/.../{udp,tcp}/...` *without* the `/quic` or `/ws` suffix. This means that, while we do map these protocols, we don't actually advertise the correct addresses.

We need to:

1. Fix `BasicHost.AllAddrs` to pick the right mappings based on the internal port, not the full address.
2. Fix `go-libp2p-nat` to only accept tcp/udp addresses (quic, ws, etc. don't make any sense at the NAT level anyways). Really, we may want to switch the interfaces to protocol + port instead of multiaddr but we can do that later.

overbool:
@Stebalien Did you mean that `/ip4/.../tcp/4001` and `/ip4/.../tcp/4001/ws` will generate two mappings, but latter was redundant？

vyzo:
> Fix go-libp2p-nat to only accept tcp/udp addresses (quic, ws, etc. don't make any sense at the NAT level anyways).

But we do want to open port mappings in this case too.

overbool:
`/ip4/.../tcp/4001/ws` is redundant, because `tcp+4001` had been mapped (`/ip4/.../tcp/4001`). However, `/ip4/.../tcp/4002/ws` will be mapped, because `tcp+4002` is a new mapping.

just like:

> Really, we may want to switch the interfaces to protocol + port instead of multiaddr but we can do that later.

Stebalien:
No. We actually *can't* listen on both `/ip4/.../tcp/4001` and `/ip4/.../tcp/4001/ws`. Those two listeners would conflict.

The issue is much simpler:

1. When we see `/ip4/.../tcp/4002/ws`, we ask the NAT manager to map it. This works as intended.
2. When we advertise our addresses, we ask the NAT manager which addresses we have mapped. Unfortunately, these addresses *don't* include the `/ws` part because the NAT manager really only cares about the IP/port/protocol tripple.

This means that even though we might be listening on `/ip4/.../tcp/4002/ws`, we'll end up *advertising* `/ip4/.../tcp/4002` (no `/ws`).

vyzo:
that's definitely a bug!

PR #549, opened 2019-03-06
better nat mapping

Stebalien:
1. Update to work with https://github.com/libp2p/go-libp2p-nat/pull/14.
2. Avoid observed addrs when our NAT tells us about external addrs.
3. Ignore bad addrs reported by our NAT. Substitute with observed addrs.
4. Fix https://github.com/libp2p/go-libp2p/issues/428.

(really, I'm just trying to get NAT port mapping working on my network...)

### DIFF
--- a/p2p/host/basic/natmgr.go
+++ b/p2p/host/basic/natmgr.go
@@ -43,7 +43,7 @@ type natManager struct {
 
 	ready chan struct{} // closed once the nat is ready to process port mappings
 
-	refreshMu sync.Mutex
+	syncMu sync.Mutex
 
 	proc goprocess.Process // natManager has a process + children. can be closed.
 }
@@ -111,11 +111,13 @@ func (nmgr *natManager) discoverNAT() {
 		// we need to sign up here to avoid missing some notifs
 		// before the NAT has been found.
 		nmgr.net.Notify((*nmgrNetNotifiee)(nmgr))
-		nmgr.refresh()
+		nmgr.sync()
 	})
 }
 
-func (nmgr *natManager) refresh() {
+// syncs the current NAT mappings, removing any outdated mappings and adding any
+// new mappings.
+func (nmgr *natManager) sync() {
 	nat := nmgr.NAT()
 	if nat == nil {
 		// Nothing to do.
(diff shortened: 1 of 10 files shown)


Stebalien:
This PR has some pretty nasty functions and could use some multiaddr helpers.

p2p/host/basic/natmgr.go
```diff
@@ -87,151 +90,151 @@ func (nmgr *natManager) discoverNAT() {
...
+			case ma.P_IP6, ma.P_IP4:
+			default:
+				continue
+			}
 
-	defer log.EventBegin(ctx, "natMgrAddPortMappingWait", lm).Done()
+			// Only bother if we're listening on a
+			// unicast/unspecified IP.
+			ip := net.IP(maIP.RawValue())
+			if !(ip.IsGlobalUnicast() || ip.IsUnspecified()) {
```
raulk: Just to confirm. This doesn't `continue` if the listen address is public (i.e. we don't need a port mapping). That's because we're presuming in that case there's no NAT device at all, and hence this func would return at line 124? What would happen if the host is listening on multiple network interfaces, of which one is public and one is private (behind a NAT)?

p2p/host/basic/natmgr.go
```diff
@@ -87,151 +90,151 @@ func (nmgr *natManager) discoverNAT() {
...
+				go func(proto string, port int) {
+					defer wg.Done()
+					_, err := nat.NewMapping(proto, port)
+					if err != nil {
+						log.Errorf("failed to port-map %s port %d: %s", proto, port, err)
+					}
+				}(proto, port)
+			}
+		}
+	})
```
raulk: Not relevant to the changes in this PR, but shouldn't we be triggering an "identify push" here if there were changes (maybe not _exactly_ here, but when the mappings are done)?

Stebalien: Probably (https://github.com/libp2p/go-libp2p/issues/553). However, we also need to push when:

1. Our observed addresses change significantly.
2. Our interface addresses change.

etc...

raulk: Eventbus 🚌🚌🚌

PR #551, opened 2019-03-06
dep: update autonat

Stebalien:


Stebalien:
@raulk how long is libp2p planning on supporting both gx and go modules? If we're planning on supporting both for a while, we may want to add some gx/CI tooling.

PR #558, opened 2019-03-08
identify: open stream via host

raulk:
minor change to simplify code and avoid unnecessary dances.


raulk:
@Stebalien – makes sense. Thanks for the clarification!

PR #563, opened 2019-03-18
relay: simplify declaration of multiaddr var

lgierth:
StringCast will do the panicking for us.


lgierth:
>  codecov/project — 49.95% (-0.03%) compared to 0e4b173 

How is that even possible, I *deleted code* :X

Line-based code coverage stinks.

PR #570, opened 2019-04-03
Reduce autorelay boot delay and correctly handle private->public transition

vyzo:
Reduces the autorelay boot delay to take advantage of recent autonat optimizations and also correctly handles the private->public transition by scheduling an identify push in that case.

Depends on https://github.com/libp2p/go-libp2p-autonat/pull/20 [go mod update needed]


vyzo:
cc @ZenGround0

vyzo:
Updated the gomod dependency for go-libp2p-autonat as well, will tag a release right after merge.

PR #574, opened 2019-04-05
identify: only store _reported_ multiaddrs

Stebalien:
We still tell the remote host about the observed addr but we don't store it. That way, we give them a chance to decide if they want to actually use and advertise it.

Ideally, we'd distinguish between local information and signed routing information but we don't do that yet.

Hopefully this will prevent nodes behind relays from telling everyone else to connect to their peers through these relays.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -238,11 +238,8 @@ func (ids *IDService) consumeMessage(mes *pb.Identify, c inet.Conn) {
 		lmaddrs = append(lmaddrs, maddr)
 	}
 
-	// if the address reported by the connection roughly matches their annoucned
-	// listener addresses, its likely to be an external NAT address
-	if HasConsistentTransport(c.RemoteMultiaddr(), lmaddrs) {
-		lmaddrs = append(lmaddrs, c.RemoteMultiaddr())
-	}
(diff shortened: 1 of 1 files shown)


Stebalien:
Actually, this is a bit interesting... We only add this remote address if it has the same protocol sequence as an address advertised by the peer. That means we _shouldn't_ add the address here unless the peer adds the relay to their observed address set.

So, this may still may prevent the address explosion issue however, it won't fix the actual problem. To do that, we just need to remove observed relay addresses.

p2p/protocol/identify/id.go
```diff
@@ -412,6 +409,11 @@ func (ids *IDService) consumeObservedAddress(observed []byte, c inet.Conn) {
 		return
 	}
 
+	if !HasConsistentTransport(c.RemoteMultiaddr(), ids.Host.Addrs()) {
+		log.Debugf("ignoring observed multiaddr that doesn't match the transports of any addresses we're announcing", c.RemoteMultiaddr())
+		return
+	}
```
vyzo: so what happens if we are behind a NAT and advertise a relay addr, and connect to a peer through a different relay (one which we don't advertise)? It seems that we might pollute our advertised set to include this relay too.

Stebalien:
Let's discuss in #575. I'm not sure if this is right.

Stebalien:
Re-opened in #577 but not as a fix for #575.

PR #577, opened 2019-04-05
identify: be more careful about the addresses we store

### LINKED ISSUE #575: Relay Swarm
This morning, we noticed a bunch of peers swarming and overloading our relays. We initially thought this might be malicious but, on further inspection, it looks like there were just some popular nodes running behind relays (causing a bunch of peers to connect to them).

However, we're still concerned some peers may be routing traffic through relays when they shouldn't be. This issue exists so we can report our findings as we investigate this.

Stebalien:
1. Never record addresses _we_ observe. Only store addresses if a peer tells us to. This could make dialing back a bit harder but should give nodes more control over the addresses they care to advertise.
2. Only record reported observed addresses if they match the transports of some address we're already advertising. For example, we should only advertise relay addresses if we're already advertising other relay addresses.

Note: I don't believe this is related to #575 but, IMO, it's still a good idea.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -240,6 +240,11 @@ func (ids *IDService) consumeMessage(mes *pb.Identify, c inet.Conn) {
 
 	// NOTE: Do not add `c.RemoteMultiaddr()` to the peerstore if the remote
 	// peer doesn't tell us to do so. Otherwise, we'll advertise it.
+	//
+	// This can cause an "addr-splosion" issue where the network will slowly
+	// gossip and collect observed but unadvertised addresses. Given a NAT
+	// that picks random source ports, this can cause DHT nodes to collect
+	// many undialable addresses for other peers.
 
 	// Extend the TTLs on the known (probably) good addresses.
 	// Taking the lock ensures that we don't concurrently process a disconnect.
@@ -412,6 +417,11 @@ func (ids *IDService) consumeObservedAddress(observed []byte, c inet.Conn) {
 	log.Debugf("identify identifying observed multiaddr: %s %s", c.LocalMultiaddr(), ifaceaddrs)
 	if !addrInAddrs(c.LocalMultiaddr(), ifaceaddrs) && !addrInAddrs(c.LocalMultiaddr(), ids.Host.Network().ListenAddresses()) {
 		// not in our list
+		return
+	}
+
+	if !HasConsistentTransport(maddr, ids.Host.Addrs()) {
+		log.Debugf("ignoring observed multiaddr that doesn't match the transports of any addresses we're announcing", c.RemoteMultiaddr())
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


p2p/protocol/identify/id.go
```diff
@@ -418,6 +415,11 @@ func (ids *IDService) consumeObservedAddress(observed []byte, c inet.Conn) {
 		return
 	}
 
+	if !HasConsistentTransport(maddr, ids.Host.Addrs()) {
+		log.Debugf("ignoring observed multiaddr that doesn't match the transports of any addresses we're announcing", c.RemoteMultiaddr())
+		return
+	}
```
vyzo: Do we care about the expansion of relay addrs we are advertising? Some of them maybe transient, because we happened to connect to some relay in order to talk to another NATed peer. But this is not a relay we are keeping connected because we selected it (and soon RESERVE capacity), but rather it's an incidental connection.

To be more specific here: If we are NATed/relayed and we connect to another NATed/relayed host, we will end up adding the other peer's relay into our observed address set and end up advertising that together with our primary relay connections.

Stebalien: IMO, we should just get better at pushing updates. I actually like automatically discovering new relays like this as it helps spread the load a bit.

In the meantime, you're probably right. We don't want to advertise these addresses. Luckily, the default relay address filter _should_ filter out all those addresses at the next layer up.

vyzo: Yeah, I have big hopes with that patch.
The issue with using the random relays we happened to be connected is that these connections are not expected to be stable. 
It's great that we are spreading the load though.

vyzo:
@whyrusleeping thoughts here? This is a subtle little change that may have pervasive effects.

p2p/protocol/identify/id.go
```diff
@@ -238,11 +238,8 @@ func (ids *IDService) consumeMessage(mes *pb.Identify, c inet.Conn) {
 		lmaddrs = append(lmaddrs, maddr)
 	}
 
-	// if the address reported by the connection roughly matches their annoucned
-	// listener addresses, its likely to be an external NAT address
-	if HasConsistentTransport(c.RemoteMultiaddr(), lmaddrs) {
-		lmaddrs = append(lmaddrs, c.RemoteMultiaddr())
-	}
+	// NOTE: Do not add `c.RemoteMultiaddr()` to the peerstore if the remote
+	// peer doesn't tell us to do so. Otherwise, we'll advertise it.
```
Stebalien: @whyrusleeping 

Proposal: Don't even _record_ the remote peer's address in the peerstore if they aren't advertising it.
Rational: If we record it, we'll share it. I'm worried this is causing us to gossip bad ephemeral addresses around the network. I'm still seeing some addr-splosion peers.

Objections?

vyzo: SGTM.

whyrusleeping: This seems like a good idea, though I hope we're not forgetting the reason we added these in the first place.

I can't think of any good reason to be doing this right now, but let's make sure that the rationale for this decision is well documented (including what happens when we advertise these addresses without the authors consent)

Stebalien: I've extended the comment with more rational so we don't run into this situation again.

PR #578, opened 2019-04-05
relay: turn autorelay into a service and always filter out relay addresses

Stebalien:
Also, remove the "RelayHost".

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -128,7 +128,7 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 		// TODO: We shouldn't be doing this here.
 		oldFactory := h.AddrsFactory
 		h.AddrsFactory = func(addrs []ma.Multiaddr) []ma.Multiaddr {
-			return relay.Filter(oldFactory(addrs))
+			return oldFactory(relay.Filter(addrs))
 		}
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -109,16 +109,27 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 		swrm.Filters = cfg.Filters
 	}
 
-	var h host.Host
-	h, err = bhost.NewHost(ctx, swrm, &bhost.HostOpts{
+	h, err := bhost.NewHost(ctx, swrm, &bhost.HostOpts{
 		ConnManager:  cfg.ConnManager,
 		AddrsFactory: cfg.AddrsFactory,
 		NATManager:   cfg.NATManager,
 		EnablePing:   !cfg.DisablePing,
 	})
+
 	if err != nil {
 		swrm.Close()
 		return nil, err
+	}
+
+	if cfg.Relay {
+		// If we've enabled the relay, we should filter out relay
+		// addresses by default.
+		//
+		// TODO: We shouldn't be doing this here.
+		oldFactory := h.AddrsFactory
+		h.AddrsFactory = func(addrs []ma.Multiaddr) []ma.Multiaddr {
+			return relay.Filter(oldFactory(addrs))
+		}
 	}
 
 	upgrader := new(tptu.Upgrader)
@@ -206,18 +217,24 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 		}
 
 		if hop {
-			h = relay.NewRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery)
+			// advertise ourselves
+			// TODO: Why do we only do this when EnableAutoRelay is
+			// set? This has absolutely _nothing_ to do with
+			// autorelay.
+			relay.Advertise(ctx, discovery)
 		} else {
-			h = relay.NewAutoRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery, router)
+			// TODO
+			// 1. Stop abusing contexts like this.
+			// 2. Introduce a service management system (e.g.,
+			// uber's fx) so we can actually manage the lifetime of
+			// this service.
+			_ = relay.NewAutoRelay(swrm.Context(), h, discovery, router)
 		}
 	}
 
 	if router != nil {
-		h = routed.Wrap(h, router)
+		return routed.Wrap(h, router), nil
 	}
-
-	// TODO: Bootstrapping.
-
 	return h, nil
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -218,16 +218,8 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 
 		if hop {
 			// advertise ourselves
-			// TODO: Why do we only do this when EnableAutoRelay is
-			// set? This has absolutely _nothing_ to do with
-			// autorelay.
 			relay.Advertise(ctx, discovery)
 		} else {
-			// TODO
-			// 1. Stop abusing contexts like this.
-			// 2. Introduce a service management system (e.g.,
-			// uber's fx) so we can actually manage the lifetime of
-			// this service.
 			_ = relay.NewAutoRelay(swrm.Context(), h, discovery, router)
 		}
 	}
--- a/options.go
+++ b/options.go
@@ -235,7 +235,7 @@ func DisableRelay() Option {
 //    (circuit.OptHop is passed to EnableRelay), this node will advertise itself
 //    as a public relay using the provided routing system.
 // 2. When this libp2p node is _not_ configured as a relay "hop", it will
-//    automatically if it is unreachable (e.g., behind a NAT). If so, it will
+//    automatically detect if it is unreachable (e.g., behind a NAT). If so, it will
 //    find, configure, and announce a set of public relays.
 func EnableAutoRelay() Option {
 	return func(cfg *Config) error {
--- a/options.go
+++ b/options.go
@@ -201,7 +201,12 @@ func AddrsFactory(factory config.AddrsFactory) Option {
 	}
 }
 
-// EnableRelay configures libp2p to enable the relay transport with configuration options.
+// EnableRelay configures libp2p to enable the relay transport with
+// configuration options. By default, this option only configures libp2p to
+// accept inbound connections from relays and make outbound connections
+// _through_ relays when requested by the remote peer. (default: enabled)
+//
+// To _act_ as a relay, pass the circuit.OptHop option.
 func EnableRelay(options ...circuit.RelayOpt) Option {
 	return func(cfg *Config) error {
 		cfg.RelayCustom = true
@@ -211,7 +216,7 @@ func EnableRelay(options ...circuit.RelayOpt) Option {
 	}
 }
 
-// DisableRelay configures libp2p to disable the relay transport
+// DisableRelay configures libp2p to disable the relay transport.
 func DisableRelay() Option {
 	return func(cfg *Config) error {
 		cfg.RelayCustom = true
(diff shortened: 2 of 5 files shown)


config/config.go
```diff
@@ -206,7 +227,11 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 		}
 
 		if hop {
-			h = relay.NewRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery)
+			// advertise ourselves
+			// TODO: Why do we only do this when EnableAutoRelay is
+			// set? This has absolutely _nothing_ to do with
+			// autorelay.
```
Stebalien: @vyzo?

Stebalien:
Hm. Actually, I'm going to try to simply remove both of these custom hosts. I have a feeling we don't actually need them.

p2p/host/relay/autorelay.go
```diff
@@ -274,34 +260,29 @@ func shuffleRelays(pis []pstore.PeerInfo) {
...
-	if _, ok := h.relays[p]; ok {
-		delete(h.relays, p)
+
+	ar.mx.Lock()
+	defer ar.mx.Unlock()
+
+	if ar.host.Network().Connectedness(p) == inet.Connected {
+		// We have a second connection.
+		return
+	}
```
Stebalien: Fixes a small race.

config/config.go
```diff
@@ -109,18 +109,29 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
...
+	if cfg.Relay {
+		// If we've enabled the relay, we should filter out relay
+		// addresses by default.
+		//
+		// TODO: We shouldn't be doing this here.
+		oldFactory := h.AddrsFactory
+		h.AddrsFactory = func(addrs []ma.Multiaddr) []ma.Multiaddr {
+			return relay.Filter(oldFactory(addrs))
+		}
+	}
```
Stebalien: This was the real issue. Ideally we'd inject this from the relay code but we can't assert that we have a basic host there.

config/config.go
vyzo: Sure it has to do with autorelay -- it's the advertisement of relay services for autorelayed hosts!

config/config.go
```diff
@@ -206,18 +217,24 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
...
+			// autorelay.
+			relay.Advertise(ctx, discovery)
 		} else {
-			h = relay.NewAutoRelayHost(swrm.Context(), h.(*bhost.BasicHost), discovery, router)
+			// TODO
+			// 1. Stop abusing contexts like this.
+			// 2. Introduce a service management system (e.g.,
+			// uber's fx) so we can actually manage the lifetime of
+			// this service.
+			_ = relay.NewAutoRelay(swrm.Context(), h, discovery, router)
```
vyzo: we need to keep the autorelay host, otherwise we will wrap/return the basic host and all the address manipulation will be voided.

vyzo:
So why are we killing `RelayHost`? I think we should keep it instead of having the advertisement hiding in the constructor.

p2p/host/relay/autorelay.go
```diff
@@ -29,13 +28,11 @@ var (
...
-
-	unspecificRelay = ma.StringCast("/p2p-circuit")
 )
 
-// AutoRelayHost is a Host that uses relays for connectivity when a NAT is detected.
-type AutoRelayHost struct {
-	*basic.BasicHost
+// AutoRelay is a Host that uses relays for connectivity when a NAT is detected.
+type AutoRelay struct {
+	host     *basic.BasicHost
```
vyzo: also this, we are no longer a host without the embedding.

p2p/host/relay/autorelay.go
```diff
@@ -48,37 +45,37 @@ type AutoRelayHost struct {
...
-	if h.addrs != nil && h.autonat.Status() == autonat.NATStatusPrivate {
-		return h.addrs
+func (ar *AutoRelay) hostAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
+	ar.mx.Lock()
+	defer ar.mx.Unlock()
+	if ar.addrs != nil && ar.autonat.Status() == autonat.NATStatusPrivate {
+		return ar.addrs
 	} else {
-		return filterUnspecificRelay(h.addrsF(addrs))
+		return ar.addrsF(addrs)
```
vyzo: where does this get filtered now?

config/config.go
vyzo: actually it's probably ok, we are overriding the `AddressFactory` at the end of the day.

p2p/host/relay/doc.go
```diff
@@ -14,11 +14,6 @@ How it works:
 - `AutoNATService` instances are instantiated in the
   bootstrappers (or other well known publicly reachable hosts)
 
-- `RelayHost`s are constructed with
-  `libp2p.New(libp2p.EnableRelay(circuit.OptHop), libp2p.Routing(makeDHT))`.
-  They provide Relay Hop services, and advertise through the DHT
-  in the `/libp2p/relay` namespace
-
 - `AutoRelayHost`s are constructed with `libp2p.New(libp2p.Routing(makeDHT))`
```
vyzo: this needs an update too, it's no longer a host.

vyzo:
The documentation has become obsolete.

p2p/host/relay/autorelay.go
```diff
@@ -90,105 +87,105 @@ func (h *AutoRelayHost) background(ctx context.Context) {
...
+	if len(ar.relays) >= DesiredRelays {
+		ar.mx.Unlock()
 		return
 	}
-	need := DesiredRelays - len(h.relays)
-	h.mx.Unlock()
+	need := DesiredRelays - len(ar.relays)
+	ar.mx.Unlock()
 
 	limit := 20
```
vyzo: can we bump this limit to 50? We have 25 relays deployed plus whatever user-run relays, so we should try to see them all.

vyzo: Pushed a commit that bumps that.

config/config.go
Stebalien: If you look at this from an "I'm enabling the autorelay subsystem" perspective, yes. However, users will just see an autorelay _option_. When looking at these as _options_,

1. If I enable relay, I expect my node to make use of relays and accept relay connections.
2. If I enable relay _hop_, I expect my node to _act_ as a relay. That _includes_ telling other users that I'm willing to be a relay.
3. If I enable autorelay, I expect my node to automatically pick and advertise a relay.

Having to enable both relay hop + autorelay to get other users to use me as a relay is really not what I would expect.

vyzo: We do need a mechanism to allow users to enable relay for their own purposes (ie their own little fleet of nodes) but _without_ having to be inundated by the torrent of traffic that comes when announcing the relay functionality.

config/config.go
vyzo: And how are we going to implement `Stop` or `Close` when there are background goroutines doing work? Why, with _contextx_ of course!
So we can't really get rid of them and I think it's better to have one global context for the entire host and tree of services rather than have each service make its own background one with cancel.

config/config.go
vyzo: Maybe we should reverse the order of these two filters here. Perhaps the user address factory wants to add some (specific) relay addrs to advertise.

config/config.go
Stebalien: Sure, but that should really be an option on the relay service. I'll try to document all this to be less confusing but I know we're still going to be confusing users.

config/config.go
Stebalien: > And how are we going to implement Stop or Close when there are background goroutines doing work? Why, with contextx of course!

That depends on the case but usually, yes.

>  So we can't really get rid of them and I think it's better to have one global context for the entire host and tree of services rather than have each service make its own background one with cancel.

And how are we going to _wait_ for services to stop? How can I cleanly restart a node? How can I make sure one service stops before I stop a dependency?

At the end of the day, we _need_ close methods and feeding through a context just means we have two ways to close things.

---

But this is an issue that has been plaguing all of go-ipfs and go-libp2p since the very beginning (the first attempt to solve it was actually https://github.com/jbenet/goprocess) so I'll drop the comment.

options.go
```diff
@@ -220,8 +225,18 @@ func DisableRelay() Option {
...
+// routing (not enabled by default).
+//
+// This subsystem performs two functions:
+//
+// 1. When this libp2p node is configured to act as a relay "hop"
+//    (circuit.OptHop is passed to EnableRelay), this node will advertise itself
+//    as a public relay using the provided routing system.
+// 2. When this libp2p node is _not_ configured as a relay "hop", it will
+//    automatically if it is unreachable (e.g., behind a NAT). If so, it will
+//    find, configure, and announce a set of public relays.
```
Stebalien: @vyzo, this is why I don't like this option design. I now have to explain to the user that the "AutoRelay" option can do two _very_ different things depending on how they've configured their relay.

vyzo: AutoRelay is a system service. 
If you are a hop relay, then the functionality is to advertise relay services. If you are a client, then the functionality is to use automatically use the services advertised by hop relays.

config/config.go
vyzo: I am not arguing against `Close` methods; all I am saying is that contexts will inevitably be used for this and it's convenient to have a global context with user cancellation.
When we go services, we might have `GlobalContext` service dependency that provides this context.

options.go
```diff
@@ -220,8 +225,18 @@ func DisableRelay() Option {
...
+// error to enable AutoRelay without enabling relay (enabled by default) and
+// routing (not enabled by default).
+//
+// This subsystem performs two functions:
+//
+// 1. When this libp2p node is configured to act as a relay "hop"
+//    (circuit.OptHop is passed to EnableRelay), this node will advertise itself
+//    as a public relay using the provided routing system.
+// 2. When this libp2p node is _not_ configured as a relay "hop", it will
+//    automatically if it is unreachable (e.g., behind a NAT). If so, it will
```
vyzo: missing "detect" word.

vyzo: fixed.

options.go
Stebalien: What service? There are, and always have been, two _completely separate and orthogonal services_: one for "hop" relays and one for clients. That's why this is so damn confusing.

vyzo: I tend to think of `autorelay` as a single service. But yes, at some level you are right that these are two seperate services and that could potentially be confusing.

config/config.go
Stebalien: And I'm explicitly arguing against using contexts for canceling services _at all_ because:

1. It causes all services to shut down all at once. This has been causing _real_ and _really annoying_ problems in go-ipfs where we'll, e.g., close the blockservice before we stop all services trying to write blocks.
2. It makes it really easy to incorrectly implement services. Before the transport refactor, we had quite a few services that wouldn't actually stop unless you both canceled the context and called Close.

vyzo: Having a `Close` method for services is undeniably a useful abstraction, as it could handle orderly dependency shutdown as well (but it won't be easy).
The global context could be used by services that need a context to manager their internal goroutines, there is nothing stopping them from wrapping this context in a `WithCancel`.

Stebalien: The issue is that canceling this context will cause the service to start shutting down _before_ close is called, throwing any orderly shutdown out the window. Now, of course, we could just say "don't cancel this context till you've finished the orderly shutdown" but that's not exactly user-friendly.

options.go
raulk: > There are, and always have been, two completely separate and orthogonal services: one for "hop" relays and one for clients.

Actually, it's even more confusing than that. A _client_ can be a _consumer_ of a relay service, or a _target_ of a relayed connection. Currently, all of these flow over the same thick protocol. I really have a hard time wrapping my head around this, and so do our users.

I'd like to segregate these protocols if possible:

* `/libp2p/circuit/relay/hop/0.1.0`: exposed only by relay nodes. Used inside the initiator -> relay connection.
* `/libp2p/circuit/relay/target/0.1.0`: exposed by clients accepting relayed connections. Used in the relay -> target connection.

This would allow for better implicit observability too. Would've helped a lot in the incident we're analysing.

config/config.go
raulk: Let's put together a requirements/design document outlining _what we actually need_ from the refactor. That will help us drive alignment and consensus.

I sense we all have slightly different views, understandings and wishes, and it'll cause frustration if we jump into technical solutions before we decide what problems we want to solve.

Things we have discussed:

* dependency injection
* service lifecycles
* hot replacement of runtime services
* service tree (lifecycle dependencies)
* eventbus
* etc.

PR #582, opened 2019-04-08
move things outside of the lock in obsaddr

whyrusleeping:
The test added here is just to benchmark the effect of lock contention. In my tests, this patch makes things ~4x faster (54 seconds to 14.5 seconds)

### DIFF
--- a/p2p/protocol/identify/obsaddr_test.go
+++ b/p2p/protocol/identify/obsaddr_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	detectrace "github.com/ipfs/go-detect-race"
 	net "github.com/libp2p/go-libp2p-net"
 	ma "github.com/multiformats/go-multiaddr"
 )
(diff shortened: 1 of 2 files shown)


whyrusleeping:
okay, after rebasing the perf gains arent quite as nice, but still 25 seconds to 11 seconds.

p2p/protocol/identify/obsaddr_test.go
```diff
@@ -117,3 +118,48 @@ func TestObsAddrSet(t *testing.T) {
...
+	b5 := m("/ip4/1.2.3.10/tcp/1237")
+
+	_ = []ma.Multiaddr{a1, a2, a3, a4, a5, b1, b2, b3, b4, b5}
+
+	var wg sync.WaitGroup
+	for i := 0; i < 1000; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for j := 0; j < 10000; j++ {
```
Stebalien:
```suggestion
			for j := 0; j < 1000; j++ {
```

(so this can actually pass on travis)

PR #583, opened 2019-04-09
identify: avoid parsing/printing multiaddrs

Stebalien:
1. Don't bother formatting the address as a _string_.
2. Use SplitFirst to avoid parsing the entire address.


Stebalien:
_slight_ perf improvement (1.4x speedup).

PR #585, opened 2019-04-09
Improve observed address handling

### LINKED ISSUE #584: Address observations should be dropped when overloaded
We should consider rewriting the address observation system to:

1. Use a single observation worker.
2. Drop observations when we get overloaded (i.e., write to a channel with a select and a default case).

(cc @vyzo)

vyzo:
Overhauls the observed address set to be more friendly to relays:
- Optimizes the lock contention in Addrs, AddrsFor by stopping the eager clean up of expired entries
- Adds a background worker for adding and gc'ing observed address, with a very short buffer

Closes #584

### DIFF
--- a/p2p/protocol/identify/obsaddr.go
+++ b/p2p/protocol/identify/obsaddr.go
@@ -160,7 +160,11 @@ func (oas *ObservedAddrSet) gc() {
 				filteredAddrs = append(filteredAddrs, a)
 			}
 		}
-		oas.addrs[local] = filteredAddrs
+		if len(filteredAddrs) > 0 {
+			oas.addrs[local] = filteredAddrs
+		} else {
+			delete(oas.addrs, local)
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


vyzo:
I think we can improve further by using a read/write lock, as there is now a single writer and many readers.

p2p/protocol/identify/obsaddr.go
```diff
@@ -42,23 +45,40 @@ func (oa *ObservedAddr) activated(ttl time.Duration) bool {
...
+
+	// this is the worker channel
+	wch chan newObservation
+}
+
+func NewObservedAddrSet(ctx context.Context) *ObservedAddrSet {
+	oas := &ObservedAddrSet{
+		addrs: make(map[string][]*ObservedAddr),
+		ttl:   pstore.OwnObservedAddrTTL,
+		wch:   make(chan newObservation, 1),
```
Stebalien: Let's make this longer, like 64 or something (or at least 16). We want as many observed addrs as we can on initial start.

vyzo: Ok, will set to 16 (= 4 * activation-threshold)

PR #590, opened 2019-04-10
Respect nodial option in routed host

vyzo:
Also includes the relevant gomod updates and kills gx travis builds.


vyzo:
@stebalien LGTM'ed over slack as he is mobile.

PR #593, opened 2019-04-12
dep: update go-libp2p-nat

Stebalien:
Switches to libp2p/go-nat which has some improved gateway detection logic.


Stebalien:
(just a dep update)

PR #597, opened 2019-04-13
Periodically schedule identify push if the address set has changed

vyzo:
Adds a background task to basic host, that periodically schedules an identify push if the address set has changed.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -72,7 +72,8 @@ type BasicHost struct {
 
 	proc goprocess.Process
 
-	bgcancel  func()
+	ctx       context.Context
+	cancel    func()
 	mx        sync.Mutex
 	lastAddrs []ma.Multiaddr
 }
@@ -171,8 +172,8 @@ func NewHost(ctx context.Context, net inet.Network, opts *HostOpts) (*BasicHost,
 	net.SetStreamHandler(h.newStreamHandler)
 
 	bgctx, cancel := context.WithCancel(ctx)
-	h.bgcancel = cancel
-	go h.background(bgctx)
+	h.ctx = bgctx
+	h.cancel = cancel
 
 	return h, nil
 }
@@ -212,6 +213,11 @@ func New(net inet.Network, opts ...interface{}) *BasicHost {
 	}
 
 	return h
+}
+
+// Start starts background tasks in the host
+func (h *BasicHost) Start() {
+	go h.background()
 }
 
 // newConnHandler is the remote-opened conn handler for inet.Network
@@ -288,10 +294,7 @@ func (h *BasicHost) PushIdentify() {
 	}
 }
 
-func (h *BasicHost) background(ctx context.Context) {
-	// wait a bit for the host to initialize (avoid race with libp2p constructor)
-	time.Sleep(1 * time.Second)
-
+func (h *BasicHost) background() {
 	// periodically schedules an IdentifyPush to update our peers for changes
 	// in our address set (if needed)
 	ticker := time.NewTicker(1 * time.Minute)
@@ -309,7 +312,7 @@ func (h *BasicHost) background(ctx context.Context) {
 		case <-ticker.C:
 			h.PushIdentify()
 
-		case <-ctx.Done():
+		case <-h.ctx.Done():
 			return
 		}
 	}
@@ -719,7 +722,7 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
 
 // Close shuts down the Host's services (network, etc).
 func (h *BasicHost) Close() error {
-	h.bgcancel()
+	h.cancel()
 	return h.proc.Close()
 }
 
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -289,6 +289,9 @@ func (h *BasicHost) PushIdentify() {
 }
 
 func (h *BasicHost) background(ctx context.Context) {
+	// wait a bit for the host to initialize (avoid race with libp2p constructor)
+	time.Sleep(1 * time.Second)
+
 	// periodically schedules an IdentifyPush to update our peers for changes
 	// in our address set (if needed)
 	ticker := time.NewTicker(1 * time.Minute)
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"net"
+	"sync"
 	"time"
 
 	logging "github.com/ipfs/go-log"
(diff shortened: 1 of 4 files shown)


vyzo:
ci failed with a data race...

vyzo:
```

**Comment by vyzo** [comment]
```
WARNING: DATA RACE
Read at 0x00c0000b6190 by goroutine 48:
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).Addrs()
      /home/travis/gopath/src/github.com/libp2p/go-libp2p/p2p/host/basic/basic_host.go:557 +0x6e
  github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).background()
      /home/travis/gopath/src/github.com/libp2p/go-libp2p/p2p/host/basic/basic_host.go:300 +0x201
Previous write at 0x00c0000b6190 by goroutine 46:
  github.com/libp2p/go-libp2p/config.(*Config).NewNode()
      /home/travis/gopath/src/github.com/libp2p/go-libp2p/config/config.go:134 +0x13d1
  github.com/libp2p/go-libp2p.NewWithoutDefaults()
      /home/travis/gopath/src/github.com/libp2p/go-libp2p/libp2p.go:68 +0x17a
  github.com/libp2p/go-libp2p.New()
      /home/travis/gopath/src/github.com/libp2p/go-libp2p/libp2p.go:54 +0x147
  github.com/libp2p/go-libp2p.TestNoTransports()
      /home/travis/gopath/src/github.com/libp2p/go-libp2p/libp2p_test.go:50 +0xc8
  testing.tRunner()
      /home/travis/.gimme/versions/go1.11.9.linux.amd64/src/testing/testing.go:827 +0x162
```

insanity...

**Comment by vyzo** [comment]
I added a delay in the background task, but maybe we should have an explicit Start method that gets called by the constructor instead.

**Comment by vyzo** [comment]
Now we get this failure:
```
=== RUN   TestFuzzManyPeers
race: limit on 8128 simultaneously alive goroutines is exceeded, dying
FAIL	github.com/libp2p/go-libp2p/p2p/net/mock	30.895s
```

I'll gate this test.

**Comment by vyzo** [comment]
It's already gated, so I reduced the peer count further.

**Comment by Stebalien** [comment]
Let's do an explicit start.

**Comment by vyzo** [comment]
Ok, will do.

**Comment by vyzo** [comment]
Added `Start` method in basic_host.

**Comment by vyzo** [comment]
rebased on master.

**Comment by vyzo** [comment]
summoning @Stebalien

File: p2p/protocol/identify/id.go [hunk 1]
```diff
@@ -156,19 +159,35 @@ func (ids *IDService) pushHandler(s inet.Stream) {
...
+			defer wg.Done()
+
 			s, err := ids.Host.NewStream(ctx, p, IDPush)
 			if err != nil {
-				log.Debugf("error opening push stream: %s", err.Error())
+				log.Debugf("error opening push stream to %s: %s", p, err.Error())
 				return
 			}
 
 			ids.requestHandler(s)
```
Stebalien: TODO: Not a bug in this PR but we should probably:

1. Do this in a goroutine.
2. Reset the stream if the context expires.

At the moment, we can hang the _entire_ push if a single peer stalls.

Stebalien: Actually, this is now an issue because we're blocking everything else (which we weren't doing before).

vyzo: I can move the context back into the goroutine and remove the wait group, so that we don't stall.

p2p/host/basic/basic_host.go
```diff
@@ -263,7 +279,67 @@ func (h *BasicHost) newStreamHandler(s inet.Stream) {
...
+		}
+	}
+}
+
+func sameAddrs(a, b []ma.Multiaddr) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	// this is O(n*m), might be worth using a map to turn into O(n+m)
```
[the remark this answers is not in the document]
vyzo: I think I'll fix it now, I don't like having perf bombs waiting to go off.

p2p/protocol/identify/id.go
vyzo: Note that the only reason I did this dance with the wait group was my desire to reuse the context instead of creating one for each peer.

vyzo: Ah, you want the `ids.requestHandler` call in a goroutine, so that we don't stall; much better than exploding the contexts.

vyzo: so resetting the stream will need an extra goroutine to look after it.

vyzo: It's probably more reasonable to wait in the dispatch goroutine for up to the context and reset the stream there.

vyzo: Also note we are not blocking anything, there is a supervisory goroutine waiting for the WaitGroup.

vyzo: We might simply get away with setting a `Deadline` in the stream.

vyzo: Done in https://github.com/libp2p/go-libp2p/pull/597/commits/5503cd7ff3cb79f9c1f2d3b11822a962352f79d4
I moved the `ids.requestHandler` into a goroutine and added a wait for the context cancellation to reset the stream if the handler doesn't complete by then.

p2p/host/basic/basic_host.go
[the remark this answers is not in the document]
vyzo: Implemented the map in https://github.com/libp2p/go-libp2p/pull/597/commits/b17ba3100e530f604345a27d73963c3ae7decdad

p2p/protocol/identify/id.go
Stebalien: > Also note we are not blocking anything, there is a supervisory goroutine waiting for the WaitGroup.

Ah, got it. But better to reset the stream anyways.

Stebalien:
Note: multiple contexts is actually pretty cheap.

p2p/host/basic/basic_host.go
```diff
@@ -263,7 +279,63 @@ func (h *BasicHost) newStreamHandler(s inet.Stream) {
...
+		}
+	}
+}
+
+func sameAddrs(a, b []ma.Multiaddr) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	bmap := make(map[string]struct{})
```
Stebalien: Nit: we could preallocate.

vyzo: ok, will pass a length.

vyzo: done

p2p/protocol/identify/id.go
vyzo: Yeah, let's not build up goroutines if we have a stuck peer.

vyzo:
rebased on master for merging.

PR #598, opened 2019-04-13
autorelay: curtail addrsplosion

vyzo:
Relays, as discovered in the DHT, are _addrsploded_. The current (pre-)computation of advertised relay addrs fails to account for this fact, resulting in the advertisement of many dead addresses. In addition, many relays discovered in the DHT are actually dead.

This patch adds logic to deal with the situation
- relay addrs are dynamically recomputed by the address factory using the relay addrs present in the peerstore instead of pre-computing based on what was advertised.
- a special clean up pass is done on the relay address set to remove addrsploded addrs.
- a retry pass is added to `findRelays` that deals with the case of no live relays in the discovered set.
- we also increase the relay advertisement boot delay to 15 min so that some minimum node stability is ensured.

Note: the patch was originally much more extensive, adding a complex relay selection logic that tried to parallelize to deal with dead relays. That proved to be contentious, and biased against dedicated relays, so it has been reverted.

### DIFF
--- a/p2p/host/relay/addrsplosion.go
+++ b/p2p/host/relay/addrsplosion.go
@@ -118,16 +118,30 @@ func sanitizeAddrsplodedSet(public, private []ma.Multiaddr) []ma.Multiaddr {
 	var result []ma.Multiaddr
 	for _, pas := range pubaddrs {
 		if len(pas) == 1 {
+			// it's not addrsploded
 			result = append(result, pas[0].addr)
 			continue
 		}
 
+		haveAddr := false
 		for _, pa := range pas {
 			if pa.port == 4001 || pa.port == 4002 {
+				// it's a default port, use it
 				result = append(result, pa.addr)
+				haveAddr = true
 				continue
 			}
+
 			if _, ok := privports[pa.port]; ok {
+				// it matches a privately bound port, use it
+				result = append(result, pa.addr)
+				haveAddr = true
+			}
+		}
+
+		if !haveAddr {
+			// we weren't able to select a port; bite the bullet and use them all
+			for _, pa := range pas {
 				result = append(result, pa.addr)
 			}
 		}
--- a/p2p/host/relay/addrsplosion.go
+++ b/p2p/host/relay/addrsplosion.go
@@ -1,0 +1,137 @@
+package relay
+
+import (
+	"encoding/binary"
+
+	pstore "github.com/libp2p/go-libp2p-peerstore"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr-net"
+)
+
+// This function cleans up a relay's address set to remove private addresses and curtail
+// addrsplosion.
+func cleanupAddressSet(pi pstore.PeerInfo) pstore.PeerInfo {
+	var public, private []ma.Multiaddr
+
+	for _, a := range pi.Addrs {
+		if manet.IsPublicAddr(a) || isDNSAddr(a) {
+			public = append(public, a)
+			continue
+		}
+
+		// discard unroutable addrs
+		if manet.IsPrivateAddr(a) {
+			private = append(private, a)
+		}
+	}
+
+	if !hasAddrsplosion(public) {
+		return pstore.PeerInfo{ID: pi.ID, Addrs: public}
+	}
+
+	addrs := sanitizeAddrsplodedSet(public, private)
+	return pstore.PeerInfo{ID: pi.ID, Addrs: addrs}
+}
+
+func isDNSAddr(a ma.Multiaddr) bool {
+	dnsaddr := false
+	ma.ForEach(a, func(c ma.Component) bool {
+		switch c.Protocol().Code {
+		case 54, 55, 56:
+			dnsaddr = true
+		}
+
+		return false
+	})
+
+	return dnsaddr
+}
+
+// we have addrsplosion if for some protocol we advertise multiple ports on
+// the same base address.
+func hasAddrsplosion(addrs []ma.Multiaddr) bool {
+	aset := make(map[string]int)
+
+	for _, a := range addrs {
+		key, port := addrKeyAndPort(a)
+		xport, ok := aset[key]
+		if ok && port != xport {
+			return true
+		}
+		aset[key] = port
+	}
+
+	return false
+}
+
+func addrKeyAndPort(a ma.Multiaddr) (string, int) {
+	var (
+		key  string
+		port int
+	)
+
+	ma.ForEach(a, func(c ma.Component) bool {
+		switch c.Protocol().Code {
+		case ma.P_TCP, ma.P_UDP:
+			port = int(binary.BigEndian.Uint16(c.RawValue()))
+			key += "/" + c.Protocol().Name
+		default:
+			val := c.Value()
+			if val == "" {
+				val = c.Protocol().Name
+			}
+			key += "/" + val
+		}
+		return true
+	})
+
+	return key, port
+}
+
+// clean up addrsplosion
+// the following heuristic is used:
+// - for each base address/protocol combination, if there are multiple ports advertised then
+//   only accept the default port if present.
+//   - If the default port is not present, we check for non-standard ports by tracking
+//     private port bindings if present.
+//   - If there is no default or private port binding, then we can't infer the correct
+//     port and give up and return all addrs (for that base address)
+func sanitizeAddrsplodedSet(public, private []ma.Multiaddr) []ma.Multiaddr {
+	type portAndAddr struct {
+		addr ma.Multiaddr
+		port int
+	}
+
+	privports := make(map[int]struct{})
+	pubaddrs := make(map[string][]portAndAddr)
+
+	for _, a := range private {
+		_, port := addrKeyAndPort(a)
+		privports[port] = struct{}{}
+	}
+
+	for _, a := range public {
+		key, port := addrKeyAndPort(a)
+		pubaddrs[key] = append(pubaddrs[key], portAndAddr{addr: a, port: port})
+	}
+
+	var result []ma.Multiaddr
+	for _, pas := range pubaddrs {
+		if len(pas) == 1 {
+			result = append(result, pas[0].addr)
+			continue
+		}
+
+		for _, pa := range pas {
+			if pa.port == 4001 || pa.port == 4002 {
+				result = append(result, pa.addr)
+				continue
+			}
+			if _, ok := privports[pa.port]; ok {
+				result = append(result, pa.addr)
+			}
+		}
+	}
+
+	return result
+}
--- a/p2p/host/relay/addrsplosion.go
+++ b/p2p/host/relay/addrsplosion.go
@@ -6,6 +6,7 @@ import (
 	circuit "github.com/libp2p/go-libp2p-circuit"
 	pstore "github.com/libp2p/go-libp2p-peerstore"
 	ma "github.com/multiformats/go-multiaddr"
+	dns "github.com/multiformats/go-multiaddr-dns"
 	manet "github.com/multiformats/go-multiaddr-net"
 )
 
@@ -57,7 +58,7 @@ func isRelayAddr(a ma.Multiaddr) bool {
 func isDNSAddr(a ma.Multiaddr) bool {
 	if first, _ := ma.SplitFirst(a); first != nil {
 		switch first.Protocol().Code {
-		case 54, 55, 56:
+		case dns.P_DNS4, dns.P_DNS6, dns.P_DNSADDR:
 			return true
 		}
 	}
--- a/p2p/host/relay/addrsplosion.go
+++ b/p2p/host/relay/addrsplosion.go
@@ -142,15 +142,15 @@ func sanitizeAddrsplodedSet(public, private []ma.Multiaddr) []ma.Multiaddr {
 
 		haveAddr := false
 		for _, pa := range pas {
-			if pa.port == 4001 || pa.port == 4002 {
-				// it's a default port, use it
+			if _, ok := privports[pa.port]; ok {
+				// it matches a privately bound port, use it
 				result = append(result, pa.addr)
 				haveAddr = true
 				continue
 			}
 
-			if _, ok := privports[pa.port]; ok {
-				// it matches a privately bound port, use it
+			if pa.port == 4001 || pa.port == 4002 {
+				// it's a default port, use it
 				result = append(result, pa.addr)
 				haveAddr = true
 			}
--- a/p2p/host/relay/addrsplosion.go
+++ b/p2p/host/relay/addrsplosion.go
@@ -4,7 +4,6 @@ import (
 	"encoding/binary"
 
 	circuit "github.com/libp2p/go-libp2p-circuit"
-	pstore "github.com/libp2p/go-libp2p-peerstore"
 	ma "github.com/multiformats/go-multiaddr"
 	dns "github.com/multiformats/go-multiaddr-dns"
 	manet "github.com/multiformats/go-multiaddr-net"
@@ -12,10 +11,10 @@ import (
 
 // This function cleans up a relay's address set to remove private addresses and curtail
 // addrsplosion.
-func cleanupAddressSet(pi pstore.PeerInfo) pstore.PeerInfo {
+func cleanupAddressSet(addrs []ma.Multiaddr) []ma.Multiaddr {
 	var public, private []ma.Multiaddr
 
-	for _, a := range pi.Addrs {
+	for _, a := range addrs {
 		if isRelayAddr(a) {
 			continue
 		}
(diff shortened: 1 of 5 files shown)


vyzo:
I think we might want to try parallel connects, as some of the relays in the discovered set (I have 183) appear to be dead.

vyzo:
If not parallel connects, at least parallel find peers.

vyzo:
Added a relay address set clean up phase to curtail addrsplosion.

vyzo:
Maybe 15s to `FindPeer` and `Connect` is too aggressive though, might be better served with a 30s timeout.

vyzo:
I will revert it to 30s.

vyzo:
Reworked the relay selection logic to perform `FindPeer` in parallel.

vyzo:
cc @ZenGround0

vyzo:
I limited the number of `FindPeer` queries to 50.

vyzo:
I am going to revert https://github.com/libp2p/go-libp2p/pull/598/commits/113dcf91d2cecc7120adfb6b59d5e5a013c2cb01 -- I don't think it is sane to trust some random peer's peerstore just to avoid a `FindPeer` query.

p2p/host/relay/autorelay.go
```diff
@@ -128,18 +129,23 @@ func (ar *AutoRelay) background(ctx context.Context) {
 }
 
 func (ar *AutoRelay) findRelays(ctx context.Context) {
+again:
```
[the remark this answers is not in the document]
vyzo: omg, dogma against gotos! this way we don't rightwards drift.

p2p/host/relay/autorelay.go
```diff
@@ -128,18 +129,23 @@ func (ar *AutoRelay) background(ctx context.Context) {
...
 	ar.mx.Lock()
-	if len(ar.relays) >= DesiredRelays {
+	haveRelays := len(ar.relays)
+	if haveRelays >= DesiredRelays {
 		ar.mx.Unlock()
+		// this dance is necessary to cover the Private->Public->Private transition
+		// where we were already connected to enough relays while private and dropped
+		// the addrs while public
+		if ar.addrs == nil {
```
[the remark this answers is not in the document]
vyzo: I thought about it, and I concluded that the lock is not needed -- the only writer of addrs is the current goroutine.

p2p/host/relay/autorelay.go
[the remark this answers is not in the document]
vyzo: there is also scope complexity with the for loop, all to avoid a simple and _beautiful_ language construct :p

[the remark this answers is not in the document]
vyzo: I really think that the use of the verboten verb `goto` results in cleaner code here, and it doesn't rightwards drift.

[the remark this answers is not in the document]
Stebalien: I can live with this but this is _mostly_ an issue because this function is way too complicated.

p2p/host/relay/autorelay.go
```diff
@@ -194,16 +192,83 @@ func (ar *AutoRelay) findRelays(ctx context.Context) {
...
+		}(pi.ID)
+	}
+
+	rcount := 0
+	for len(result) < count && rcount < maxq {
+		select {
+		case qr := <-resultCh:
+			rcount++
+			if qr.err == nil {
+				pi := cleanupAddressSet(qr.pi)
```
Stebalien: The bad addresses are already going to be in our peerstore, right?

p2p/host/relay/autorelay.go
```diff
@@ -253,6 +318,54 @@ func (ar *AutoRelay) doUpdateAddrs() {
 	ar.addrs = raddrs
 }
 
+// This function cleans up a relay's address set to remove private addresses and curtail
+// addrsplosion. For the latter, we use the following heuristic:
+// - if the address set includes a (tcp) address with the default port 4001,
+//   we remove all tcp addrs with a different port
+func cleanupAddressSet(pi pstore.PeerInfo) pstore.PeerInfo {
```
Stebalien: What about stuff like `/ip4/.../tcp/4002/ws`? What about different IP addresses? Let's allow one port per public IP + transport (ws, quic, raw tcp, etc.).

We can also look at, e.g., the localhost advertisement (if it exists) to try to pick the best (non-standard) port.

p2p/host/relay/autorelay.go
```diff
@@ -253,6 +318,54 @@ func (ar *AutoRelay) doUpdateAddrs() {
...
+		}
+
+		newAddrs = append(newAddrs, addr)
+	}
+
+	return pstore.PeerInfo{ID: pi.ID, Addrs: newAddrs}
+}
+
+func tcpPort(addr ma.Multiaddr) (int, error) {
+	val, err := addr.ValueForProtocol(ma.P_TCP)
```
Stebalien: This function should really just be avoided. I know it's annoying but it's much better to actually parse the multiaddr.

p2p/host/relay/autorelay.go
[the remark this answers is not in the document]
Stebalien: We should comment on this.

p2p/host/relay/autorelay.go
```diff
@@ -194,16 +192,83 @@ func (ar *AutoRelay) findRelays(ctx context.Context) {
...
+
+	qctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	// shuffle to randomize the order of queries
 	shuffleRelays(pis)
-	return pis
+	for _, pi := range pis[:maxq] {
+		go func(p peer.ID) {
+			pi, err := ar.router.FindPeer(qctx, p)
```
Stebalien: Why?

p2p/host/relay/autorelay.go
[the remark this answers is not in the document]
vyzo: ok, I'll add a comment.

p2p/host/relay/autorelay_test.go
```diff
@@ -133,7 +145,7 @@ func connect(t *testing.T, a, b host.Host) {
 
 // and the actual test!
 func TestAutoRelay(t *testing.T) {
-	t.Skip("fails 99% of the time")
+	//t.Skip("fails 99% of the time")
```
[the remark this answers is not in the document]
vyzo: I tuned the timings a bit, and it passes 90% of the time in this pr. It is still a little flaky, but unfortunately I can't reproduce the failures on my laptop to effectively debug the failure condition on travis.

p2p/host/relay/autorelay.go
vyzo: Maybe, but they are not going to be used in the address announcement (which is what addrsplodes te relay addr advertisement).

p2p/host/relay/autorelay.go
vyzo: ok, should I parse by component and look for tcp?

p2p/host/relay/autorelay.go
vyzo: Ok, I'll concoct a better heuristic.

p2p/host/relay/autorelay.go
vyzo: I want to parallelize the `FindPeer`s so that we don't do them serially (and fail slow on dead peers).
I also have a commit I reverted (https://github.com/libp2p/go-libp2p/commit/113dcf91d2cecc7120adfb6b59d5e5a013c2cb01) which uses the addresses returned by discovery -- I can reinstate it.

vyzo: One problem with using the addrs returned by discovery is that it penalizes relay whose peers don't happen to be in the peerstore of the responding peer.
I want to do a true random selection over _live_ relays and that short-circuits.

p2p/host/relay/autorelay.go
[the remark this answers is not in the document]
vyzo: Added comment.

p2p/host/relay/autorelay_test.go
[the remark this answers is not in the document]
vyzo: If we find it still being an issue, we can disable again. But I do want it active in this pr, so that it runs and we get some coverage.

p2p/host/relay/autorelay.go
vyzo: I reworked the heuristic and moved it to its own file.
It works as follows:
- it declares addrsplosion if there are multiple ports for some base-address/protocol combination
- for each base-address/protocol combination with addrsplosion it selects the addrs where the port is 4001 or 4002 or it matches a privately bound port
- if we can't select any default or private port, bite the bullet and accept them all.

p2p/host/relay/autorelay.go
vyzo: I killed the function altogether.

p2p/host/relay/autorelay.go
vyzo: Also, hotspot creation.

p2p/host/relay/autorelay.go
Stebalien: I introduced `ma.ForEach` and friends to try to make this a bit easier. It's still _extremely_ painful and I'd like to try some kind of pattern matching solution but that's a bit tricky in go (no macros :cry:).

vyzo: Yeah, the new parsing code uses `ma.ForEach`.

p2p/host/relay/autorelay.go
vyzo: that would be a very nice improvement with `RESERVE`!

p2p/host/relay/addrsplosion.go
```diff
@@ -0,0 +1,151 @@
...
+
+	addrs := sanitizeAddrsplodedSet(public, private)
+	return pstore.PeerInfo{ID: pi.ID, Addrs: addrs}
+}
+
+func isDNSAddr(a ma.Multiaddr) bool {
+	dnsaddr := false
+	ma.ForEach(a, func(c ma.Component) bool {
+		switch c.Protocol().Code {
+		case 54, 55, 56:
```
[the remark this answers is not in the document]
vyzo: Unfortunately `go-multiaddr-dns` does not define them as constants :(

p2p/host/relay/addrsplosion.go
```diff
@@ -0,0 +1,151 @@
...
+		return pstore.PeerInfo{ID: pi.ID, Addrs: public}
+	}
+
+	addrs := sanitizeAddrsplodedSet(public, private)
+	return pstore.PeerInfo{ID: pi.ID, Addrs: addrs}
+}
+
+func isDNSAddr(a ma.Multiaddr) bool {
+	dnsaddr := false
+	ma.ForEach(a, func(c ma.Component) bool {
```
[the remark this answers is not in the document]
vyzo: It does check only the first component, but yeah doing it with `SplitFirst` is probaby better.

p2p/host/relay/addrsplosion.go
```diff
@@ -0,0 +1,151 @@
...
+// This function cleans up a relay's address set to remove private addresses and curtail
+// addrsplosion.
+func cleanupAddressSet(pi pstore.PeerInfo) pstore.PeerInfo {
+	var public, private []ma.Multiaddr
+
+	for _, a := range pi.Addrs {
+		if manet.IsPublicAddr(a) || isDNSAddr(a) {
+			public = append(public, a)
+			continue
+		}
```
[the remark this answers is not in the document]
vyzo: good point, although relays really should advertise relay addrs.

p2p/host/relay/addrsplosion.go
```diff
@@ -0,0 +1,151 @@
...
+				result = append(result, pa.addr)
+				haveAddr = true
+				continue
+			}
+
+			if _, ok := privports[pa.port]; ok {
+				// it matches a privately bound port, use it
+				result = append(result, pa.addr)
+				haveAddr = true
+			}
```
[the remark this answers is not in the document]
vyzo: ok, will do.

p2p/host/relay/autorelay.go
vyzo: Hrm, that might be ok -- although it does prefer relays for which we have addrs, and this may include failed connection attempts in a retry because `TempAddrTTL` is now long enough to cause problems.
But then again, these will be results form a `FindPeer` query so we might as well retry some relays to see if they were temporarily down (assuming they were up because they were found).

p2p/host/relay/addrsplosion.go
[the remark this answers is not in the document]
vyzo: done in https://github.com/libp2p/go-libp2p/pull/598/commits/9d7f6b83e5566ebbe379efaedae4f7096b4ef979

p2p/host/relay/addrsplosion.go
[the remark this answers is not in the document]
vyzo: done in https://github.com/libp2p/go-libp2p/pull/598/commits/528c473840c7f24f93302133fd133fadd02cdcfc

p2p/host/relay/addrsplosion.go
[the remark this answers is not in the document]
vyzo: done in https://github.com/libp2p/go-libp2p/pull/598/commits/21c4e1d298be6601a4921b20cec963a5ea8d8adf

p2p/host/relay/autorelay.go
vyzo: done in https://github.com/libp2p/go-libp2p/pull/598/commits/a8d14f9b025d3073c60938257735e60ff8e23b89

vyzo: Perhaps we should _add_ the addrs from the result of the query in the peerstore too, but I don't know what's the right TTL here.

Stebalien: Random idea: we could recompute the relay's addresses from our peerstore every time the user calls Addrs. That way, we update them based on new addresses from the relay. That'll make this code significantly less important (unnecessary?).

p2p/host/relay/addrsplosion.go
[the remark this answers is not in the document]
Stebalien: `madns.P_DNS4` etc?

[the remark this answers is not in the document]
vyzo: found them! My locally cached copy was older than when they were added.

p2p/host/relay/autorelay.go
vyzo: we'll still have to clean up addrsploded relay, but we need this code so that we put a strict upper bound in the discovery time of relay addresses from parallelization.
Otherwise, we'll have to do it serially for every relay we tried (the old code) and be very slow when we get dead relays in the query result.

p2p/host/relay/addrsplosion.go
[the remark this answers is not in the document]
vyzo: done.

Stebalien:
WRT https://github.com/libp2p/go-libp2p/pull/598#discussion_r276539352

The bad peer addresses will already get put into the peerstore anyways so this patch won't actually help fix _dialing_ relays. This patch will allow us to trim down the address list auto-relay users announce but a better solution might be to re-calculate `/p2p-cirucit` addresses dynamically from the peerstore. That way, we get the latest addresses sent by our relay's identify service.

In terms of _this_ code, @vyzo and I discussed generalizing it and using it from within either the dialer and/or the peerstore. Specifically, we can use the addrsplosion detection logic to trim down the set of addresses we're dialing. However, the current heuristics around circuit and private addresses obviously won't generalize (although we _should_ drop addresses of the form `/ip4/private/tcp/.../p2p-circuit/...`).

So, the question is, is merging this patch worth it? IMO, not really given that it doesn't affect dialing and there's an (IMO) better option on the advertising end.

vyzo:
So the patch fixes two problems with the current code:
- It fixes the addrsplosion in the announced addrs, which is a real problem as most of the address sets returned by discovery are addrsploded (this has been experimentally determined)
- It fixes the handling of dead peers in the discovery set, which are processed serially and could result to very long times in determining a relay to use.

vyzo:
Wrt to moving forward, we can stop tracking the relay addrs discovered by `FindPeer` and use the set returned by identify after connecting (that is when constructing the address set in `updateAddrs`).
We will still have to clean it up though, as we might have garbage in the peerstore from the queries.

vyzo:
I have updated the code to dynamically compute the relay address set on demand.

p2p/host/relay/autorelay.go
vyzo: I think I'll bite the bullet and use addrs from discovery; we are shuffling anyway.
This should reduce the number of `FindPeers` queries with large sets.

vyzo: but I still think it's a bad idea; the quality of addrs we get from discovery is very poor.
flip-flop.

vyzo: ah, it's unnecessary, `FindProviders` stashes in the peerstore anyway. So yeah, nothing we can do anyway.

vyzo: I think we can use a blacklist to avoid using the same peers within the context of a single `findRelays` if we failed to connect.

vyzo: maybe not a good idea, as a relay may be temporarily down.

vyzo: but on the same time, it's just a single pass over the relay space and the known addrs short-circuit might result in the same (dead) relays being hit all over again.

vyzo: I added the ignore list, which gets cleaned up between retries if we fail to select any relays.

vyzo: and I reverted the whole thing, as it exhibits very complex behaviour that is biased towards full DHT clients and penalizes dedicated relays like ours.

vyzo:
I have significantly simplified the patch, removing the contentious parallel query selection logic.
The behaviour wrt to relay selection and dialing matches the old behaviour now.

PR #599, opened 2019-04-15
Replace peer addresses in identify

### LINKED ISSUE #579: Identify should _replace_ addresses
Currently, we add all addresses a peer tells us about using identify. Really, we should _replace_ the addresses we know for the peer as we likely have out-of-date information.

May relate to: https://github.com/libp2p/go-libp2p/issues/575

vyzo:
Closes #579.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -252,8 +252,10 @@ func (ids *IDService) consumeMessage(mes *pb.Identify, c inet.Conn) {
 	ids.addrMu.Lock()
 	switch ids.Host.Network().Connectedness(p) {
 	case inet.Connected:
+		ids.Host.Peerstore().UpdateAddrs(p, pstore.ConnectedAddrTTL, 0)
 		ids.Host.Peerstore().AddAddrs(p, lmaddrs, pstore.ConnectedAddrTTL)
 	default:
+		ids.Host.Peerstore().UpdateAddrs(p, pstore.ConnectedAddrTTL, 0)
 		ids.Host.Peerstore().AddAddrs(p, lmaddrs, pstore.RecentlyConnectedAddrTTL)
 	}
 	ids.addrMu.Unlock()
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -33,6 +33,9 @@ const LibP2PVersion = "ipfs/0.1.0"
 
 var ClientVersion = "go-libp2p/3.3.4"
 
+// transientTTL is a short ttl for invalidated previously connected addrs
+const transientTTL = 10 * time.Second
+
 // IDService is a structure that implements ProtocolIdentify.
 // It is a trivial service that gives the other peer some
 // useful information about the local peer. A sort of hello.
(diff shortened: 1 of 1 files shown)


vyzo:
A concern here is that we might be clobbering permanent addresses.

vyzo:
It looks like we'll need a change in the peerstore interface to avoid clobbering them.

vyzo:
An approach that doesn't require a change is to set the `ConnectedAddrTTL` address to be `TempAddrTTL` with `UpdateAddrs` and then do `AddAddrs` for the new address with `ConnectedAddrTTL`.
A bit hacky, but it might just work.

vyzo:
In terms of interface changes, we can perhaps add `GetAddrs(p, ttl)` call to peerstore.
This way we can do `GetAddrs(p, PermenentTTL)` before the `SetAddrs` and then re-add the permenent addresses.

This would look like this:
```go
	permaddrs := id.Host.Peerstore().GetAddrs(p, pstore.PermenentAddrTTL)
	switch ids.Host.Network().Connectedness(p) {
	case inet.Connected:
		ids.Host.Peerstore().SetAddrs(p, lmaddrs, pstore.ConnectedAddrTTL)
	default:
		ids.Host.Peerstore().SetAddrs(p, lmaddrs, pstore.RecentlyConnectedAddrTTL)
	}
	if len(permaddrs) > 0 {
		id.Host.Peerstore.AddAddrs(p, permaddrs, pstore.PermenentAddrTTL)
	}
```

thoughts @Stebalien @raulk ?

raulk:
Relates to https://github.com/libp2p/go-libp2p-identify/issues/2 cc @hsanjuan.

p2p/protocol/identify/id.go
```diff
@@ -252,9 +252,9 @@ func (ids *IDService) consumeMessage(mes *pb.Identify, c inet.Conn) {
 	ids.addrMu.Lock()
 	switch ids.Host.Network().Connectedness(p) {
 	case inet.Connected:
-		ids.Host.Peerstore().AddAddrs(p, lmaddrs, pstore.ConnectedAddrTTL)
+		ids.Host.Peerstore().SetAddrs(p, lmaddrs, pstore.ConnectedAddrTTL)
```
Stebalien: This function is really deceptive. It sets the TTLs for the given addrs but doesn't actually swap out the old addrs for a new set.

As far as I can tell, nobody actually _uses_ this. We now use `UpdateAddrs` to atomically change TTLs.

vyzo: I don't think `UpdateAddrs` is sufficient here: it only updates the given addrs and requires us to know the old TTL.
Specifically, if the most recent `IdentifyPush` _removed_ addrs from the set, ie maybe because autorelay has rewritten the addr set, then the old addrs not in the new set will stick around forever!

vyzo: My proposal is to modify `SetAddrs` to discard the old addrs and act as a real setter (and be its first user), and introduce a new `GetAddrs` method that allows us to retrieve the old addrs associated with some TTL. The latter will allow us to avoid clobbering permanent Addrs.

vyzo: Actually, there might be a way to do the dance with `UpdateAddrs` and `AddAddrs`. 
First update addrs of `ConnectedTTL` to `TempTTL` and then `AddAddrs` with connected ttl.

It doesn't have the same effect as removing addrs, but it's close as it will mark them eligible for garbage collection and reduce the exposure time for gossiping invalid addrs.

vyzo: We can go one step further and set the TTL to 0 to invalidate them, no interface change necessary.

hsanjuan:
My interest here boils down to never deleting or changing the TTL on addresses added with PermanentAddrTTL, even if they don't work.

p2p/protocol/identify/id.go
vyzo: Fixed and rebased/squased//force-pushed.

vyzo:
The implementation in https://github.com/libp2p/go-libp2p/pull/599/commits/06391d4f2f4af216f2b46489eeba56de3ad61e42 never touches permanent addr TTLs.

It only replaces addresses with `ConnectedAddrTTL`.

p2p/protocol/identify/id.go
```diff
@@ -252,8 +252,10 @@ func (ids *IDService) consumeMessage(mes *pb.Identify, c inet.Conn) {
 	ids.addrMu.Lock()
 	switch ids.Host.Network().Connectedness(p) {
 	case inet.Connected:
+		ids.Host.Peerstore().UpdateAddrs(p, pstore.ConnectedAddrTTL, 0)
```
Stebalien: I wasn't actually suggesting that we _use_ `UpdateAddrs` (I was also thinking about fixing `SetAddrs`) but I actually like this more.

Maybe we should downgrade these to `TempAddrTTL` instead of 0? Honestly, I'm fine either way and `TempAddrTTL` is long enough that that might cause problems. My only concern is that we'll have a short period where we'll have _no_ good addresses.

We could also use a constant 10s timeout (for _really_ temporary addresses).

vyzo: It's a valid concern I think, so I'll update to use a 10s ttl to cover the gap.

vyzo: I called it `transientTTL` and it's 10s long; maybe we should have this constant in the peerstore.

Stebalien: Well, that's what `TempAddrTTL` was _supposed_ to be...

PR #603, opened 2019-04-17
constructor: allow nil options

Stebalien:
This can make it significantly easier to configure libp2p with optional options.

### DIFF
--- a/libp2p_test.go
+++ b/libp2p_test.go
@@ -133,3 +133,28 @@ func makeRandomHost(t *testing.T, port int) (host.Host, error) {
 
 	return New(ctx, opts...)
 }
+
+func TestChainOptions(t *testing.T) {
+	var cfg Config
+	var optsRun []int
+	optcount := 0
+	newOpt := func() Option {
+		index := optcount
+		optcount++
+		return func(c *Config) error {
+			optsRun = append(optsRun, index)
+			return nil
+		}
+	}
+	if err := cfg.Apply(newOpt(), nil, ChainOptions(newOpt(), newOpt(), ChainOptions(), ChainOptions(nil, newOpt()))); err != nil {
+		t.Fatal(err)
+	}
+	if optcount != len(optsRun) {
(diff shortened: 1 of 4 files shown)


libp2p_test.go
```diff
@@ -133,3 +133,28 @@ func makeRandomHost(t *testing.T, port int) (host.Host, error) {
 
 	return New(ctx, opts...)
 }
+
+func TestChainOptions(t *testing.T) {
+	var cfg Config
+	var optsRun []int
+	optcount := 0
+	newOpt := func() Option {
+		index := optcount
```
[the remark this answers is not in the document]
Stebalien: Sure.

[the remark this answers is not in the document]
Stebalien: Ah, wait, no. We need to _capture_ the current value.

libp2p_test.go
```diff
@@ -133,3 +133,28 @@ func makeRandomHost(t *testing.T, port int) (host.Host, error) {
...
+		optcount++
+		return func(c *Config) error {
+			optsRun = append(optsRun, index)
+			return nil
+		}
+	}
+	if err := cfg.Apply(newOpt(), nil, ChainOptions(newOpt(), newOpt(), ChainOptions(), ChainOptions(nil, newOpt()))); err != nil {
+		t.Fatal(err)
+	}
+	if optcount != len(optsRun) {
```
bigs: let's compare against a constant here, i.e. 4

libp2p_test.go
```diff
@@ -133,3 +133,28 @@ func makeRandomHost(t *testing.T, port int) (host.Host, error) {
...
+		t.Errorf("expected to have handled %d options, handled %d", optcount, len(optsRun))
+	}
+	for i, x := range optsRun {
```
bigs: if this is meant to test ordering, definitely drop a comment so future programmers understand the behavior

libp2p_test.go
Stebalien: SGTM.

PR #608, opened 2019-04-22
cache relayAddrs for a short period of time

vyzo:
`relayAddrs` is not the cheapest of methods and can be called potentially hundreds of times within a short period of time whenever there is a push.
This patch adds a bit of logic to cache the result for a short period of time, enough to cover a single push.


vyzo:
It's a bit more invasive (and subtle change) that needs some thought, so let's start with the big offender.

PR #613, opened 2019-04-26
add a peer to relays list if it has relay-hop tag

### LINKED ISSUE #612: Proposal: auto-add relay peer on connection
For now, we only looking for relay peers via DHT and it doesn't work all the times.
In case you have connected somehow to the relay peer, it will be not used for your relay address. 

Is it any special reason behind this? 
BTW I made the PR so you can get the idea

requilence:
Closes #612


vyzo:
Note: the "relay-hop" tag is added through passive relay discovery in go-libp2p-circuit, it's not the same tag that autorelay adds.
Regardless, I don't think this is in the right direction per discussion in #612.

PR #627, opened 2019-05-07
test: disable flaky autorelay test in CI

Stebalien:
This keeps breaking travis.

### DIFF
--- a/p2p/host/relay/autorelay_test.go
+++ b/p2p/host/relay/autorelay_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 	"time"
 
+	cienv "github.com/jbenet/go-cienv"
 	libp2p "github.com/libp2p/go-libp2p"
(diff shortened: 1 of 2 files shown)


p2p/host/relay/autorelay_test.go
```diff
@@ -144,7 +145,9 @@ func connect(t *testing.T, a, b host.Host) {
 
 // and the actual test!
 func TestAutoRelay(t *testing.T) {
-	//t.Skip("fails 99% of the time")
+	if cienv.IsRunning() {
+		t.Skip("disabled on CI: fails 99% of the time")
```
vyzo: well it's not 99% of the time, it's more like 50%.... but it's ok, we can exaggerate a bit :)

Stebalien: (fair enough)

Issue #328, opened 2018-05-09
Peer dialing to self


hsanjuan:
Is it do-able to enable peers to dial themselves?

We have Raft using a libp2p-transport and it wants to dial itself all the time. It does not seem to cause undesired side-effects, but it does not seem crazy that any cluster-application using RPCs ends up trying to contact itself for some things. (same issue in ipfs-cluster RPC, but we worked around it).

Perhaps we can provide some shortcut at a higher level so that a peer contacting itself does not even attempt touch the network?

Pointers appreciated...

vyzo:
perhaps we should have a virtual network interface for this? 
I don't think it makes sense to dial oneself and setup the entire network stack, it's a smell.

hsanjuan:
If you point me to the right repos, I can put it on my to-do list

vyzo:
@Stebalien thoughts and pointers?

Stebalien:
So, we'd have to be *very* careful here. All of our services (DHT, bitswap, identify, etc.) will break if we start dialing ourself.

Somewhat off topic...

We may need some form of "self dial" in the future for a libp2p daemon however, it won't look like a true self-dial. Basically, given a daemon and some service registered to handle protocol A, the some other *local* service may ask the daemon to dial itself and then speak protocol A. However, this is really just a special case that needs to be handled by the libp2p daemon dialer proxy.

hsanjuan:
> We may need some form of "self dial" in the future for a libp2p

I need it, now..

Happy if it is a non-default option.

bigs:
@hsanjuan @Stebalien Would this have a special API (i.e. `DialSelf`) or do we want to be able to dial our own peer ID? Perhaps we make an option for the latter and fail to dial ourselves when it's not enabled.

hsanjuan:
We want to be able to dial our own peer ID.

Stebalien:
> but it does not seem crazy that any cluster-application using RPCs ends up trying to contact itself for some things

In my experience, self-RPCs tend to cause more trouble than they're worth (mostly deadlocks).

I assume you're trying to vote for yourself? The correct way to do this is to just vote for yourself internally before sending any RPCs. You should be doing this anyways so that you don't ask a bunch of peers to vote for you but then end up voting for a different peer.

---

Basically, we can support this but I don't want to build you a footgun.

hsanjuan:
[...]

PR #638, opened 2019-05-17
[WIP] Self dialing

bigs:
The beginning of self dialing work, eventually closing #328.

This work starts with the transports. I developed a ring buffer based `MemoryTransport`, but then, in documenting it, realized I really should just use pipes. I've implemented both and left them in for posterity. I'd be happy to delete the `MemoryTransport` if we decide there is no use for it.

The memory transport expects to listen to `/p2p` or `/ipfs` multiaddrs populated with the peer's identity.


vyzo:
I think we should stick to a single transport for this, pipes look nice and simple.

p2p/transport/pipe/pipetransport.go
```diff
@@ -0,0 +1,93 @@
...
+
+	peer "github.com/libp2p/go-libp2p-peer"
+	tpt "github.com/libp2p/go-libp2p-transport"
+	tptu "github.com/libp2p/go-libp2p-transport-upgrader"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+type PipeTransport struct {
+	mlistenchans *sync.RWMutex
+	listenchans  map[string]chan *PipeConn
```
vyzo: do we really care to support multiple connections for this?

p2p/transport/pipe/pipetransport.go
```diff
@@ -0,0 +1,93 @@
...
+	if !ok {
+		return
+	}
+	close(ch)
+	delete(t.listenchans, addr)
+}
+
+func (t *PipeTransport) CanDial(addr ma.Multiaddr) bool {
+	protocols := addr.Protocols()
+	return len(protocols) == 1 && protocols[0].Code == ma.P_P2P
```
vyzo: This filter is too wide I think, it will also catch arbitrary `/p2p/QmTarget` addresses which almost certainly cannot be dialed with the pipe.
Maybe we should check the peer Id matching our own here?

p2p/transport/pipe/pipetransport.go
```diff
@@ -0,0 +1,93 @@
...
+	ch, ok := t.listenchans[raddrStr]
+	if !ok {
+		return nil, fmt.Errorf("no memorylistener for %s", raddrStr)
+	}
+
+	connA, connB := net.Pipe()
+	manetConnA := WrapNetConn(connA, raddr)
+	manetConnB := WrapNetConn(connB, raddr)
+	ch <- manetConnA
+	return t.upgrader.UpgradeOutbound(ctx, t, manetConnB, p)
```
vyzo: We probably want to disable encryption for this, it's a self-pipe after all and it only adds overhead.

p2p/transport/pipe/pipetransport.go
```diff
@@ -0,0 +1,93 @@
...
+	laddrStr := laddr.String()
+	if _, ok := t.listenchans[laddrStr]; ok {
+		return nil, fmt.Errorf("already listening on %s", laddrStr)
+	}
+
+	ch := make(chan *PipeConn)
+	t.listenchans[laddrStr] = ch
+
+	listener := NewPipeListener(laddr, ch, t)
+	upgradedListener := t.upgrader.UpgradeListener(t, listener)
```
vyzo: same here.

p2p/transport/pipe/pipetransport.go
bigs: the upgrader provided can be the "insecure" upgrader which just adds muxing

bigs:
thanks @vyzo! def agree, will remove the memorytransport.

p2p/transport/pipe/pipetransport.go
bigs: i wanted to do this, but we wont always know the peer ID at the time of creating the transport, so i left it blank. do we think we should lock it in?

p2p/transport/pipe/pipetransport.go
vyzo: How will this work in practice? Let's take care to avoid the encryption overhead.

p2p/transport/pipe/pipetransport.go
vyzo: Hrm, tricky. My worry is that we'll start trying to pipe-dial arbitrary addrs (I've seen naked `/p2p` addrs in the wild).

p2p/transport/mem/memoryconn.go
```diff
@@ -0,0 +1,156 @@
...
+// MemoryConn is a view into a pair of GrowableRingBuffers representing a fully
+// duplexed connection. One is used for reading and the other is used for
+// writing.
+type MemoryConn struct {
+	readbuf  *GrowableRingBuffer
+	writebuf *GrowableRingBuffer
+	parent   *MemoryConnPair
+}
+
+func NewMemoryConn(reader, writer *GrowableRingBuffer, parent *MemoryConnPair) *MemoryConn {
```
Stebalien: We may not need any buffering at all. In that case, we can implement this entirely with channels:

```go
func (c *ChanConn) Write(b []byte) (int, error) {
  n := 0
  for n < len(b) {
    select {
    case <-c.writeTimer.C:
      return n, ErrTimeout
    case <-c.closed:
      ...
    case remote, ok := <-c.write:
      if !ok { ... }
      coppied := copy(remote, b[:n])
      c.writeResult <- coppied
      n += coppied
    }
  }
}

func (c *ChanConn) Read(b []byte) (int, error) {
  select {
    case <-c.readTimer.C:
      return 0, ErrTimeout
    case <-c.closed:
      ...
    case c.read <- b:
  }
  return <-c.readResult, nil
}


// avoids any intermediate buffers in `io.Copy`.
func (c *ChanConn) ReadFrom(r io.Reader) (int64, error) {
  n := 0
  for {
    select {
    case <-c.writeTimer.C:
      return n, ErrTimeout
    case <-c.closed:
      ...
    case remote, ok := <-c.write:
      if !ok { ... }
      coppied, err := r.Read(remote)
      c.writeResult <- coppied
      n += coppied
      if err != nil {
        return n, err
      }
    }
  }
}


// c.write is connected to c.read, c.writeResult is connected to c.readResult.
```

That should use very little memory and should avoid copying when at all possible.

Stebalien: Actually, it looks like `net.Pipe` does everything except `ReadFrom` (which isn't absolutely necessary. We should probably just use that.

p2p/transport/pipe/pipetransport.go
Stebalien: We should always know our own peer ID when creating the transport. You can define the constructor to accept it as a parameter (instead of accepting an upgrader).

p2p/transport/mem/memoryconn.go
bigs: yeahhhhhh i was thinking about this too! i'm happy to just keep the pipe.

bigs:
> To actually hook this in, I think we'll need a `/memory` multiaddr. We strip out the `/p2p/Qm...` part when looking up transports as that isn't actually a part of the transport address.
> 
> We can even introduce a memory network concept so all peers within a process can communicate through this.
> 
>     * Listening on `/memory` would cause this transport to register a listener (for its peer ID) with the memory" network.
> 
>     * Dialing a peer with `/memory` would cause this transport to try to connect to that peer using the memory network.
> 
> 
> We could even name these networks `/memory/SOME_NAME` if we want to support multiple local networks but we shouldn't get _too_ fancy (we can also control this in the in-memory network itself).

i wanted to avoid having to extend the multiformats repo, though it seems it's unavoidable. i'd advocate for both `/memory` and `/pipe`, even if we don't choose to implement memory at the moment (or reconfigure to use channels).

should the pipe transport have a muxer and no security or should i simply create a new pipe for every new stream?

i think we should ditch the memory transport, since your proposed channel based solution is almost exactly what `net.Pipe()` is doing

Stebalien:
> i wanted to avoid having to extend the multiformats repo, though it seems it's unavoidable. i'd advocate for both /memory and /pipe, even if we don't choose to implement memory at the moment (or reconfigure to use channels).

By "memory" I just mean in-memory, not necessarily a specific in-memory transport (the user can pick which transport they want to use for process-local connections). Maybe `/local` is better? The idea is that any application would get to define what the "local" protocol is and how it should work.

> should the pipe transport have a muxer and no security or should i simply create a new pipe for every new stream?

I'd just create a new pipe for every stream. That'll be a _lot_ less overhead in practice.

lgierth:
> By "memory" I just mean in-memory, not necessarily a specific in-memory transport

tomaka needs the same: multiformats/multiaddr#71

bigs:
oh wow @lgierth thanks for the link in

bigs:
> > i wanted to avoid having to extend the multiformats repo, though it seems it's unavoidable. i'd advocate for both /memory and /pipe, even if we don't choose to implement memory at the moment (or reconfigure to use channels).
> 
> By "memory" I just mean in-memory, not necessarily a specific in-memory transport (the user can pick which transport they want to use for process-local connections). Maybe `/local` is better? The idea is that any application would get to define what the "local" protocol is and how it should work.

should these slots be named, then? i.e. `/memory/foo` should `/memory` be a path multiaddr? so we sandbox by peer ID but allow arbitrary named ports?

Stebalien:
> should these slots be named, then? i.e. /memory/foo should /memory be a path multiaddr? so we sandbox by peer ID but allow arbitrary named ports?

I wouldn't make it a path multiaddr but having arbitrary slots may make sense. However, I can't think of any reason I'd want to use slots where I _wouldn't_ just want to setup a per-peer policy (i.e., tell the local network service "allow peer X to dial peer Y").

What I'm thinking of is basically @tomaka's approach.

bigs:
@Stebalien i've fixed the implementation of the pipetransport to eschew upgraders and the like. i'm going to focus my attention to the multiformats side to get that set up, then i'll continue work on this change. remaining:

- [ ] update multiaddr handling to use /memory
- [ ] add options and new transport to libp2p hosts
- [ ] implement self dialing via pipetransport when enabled on a host

raulk:
@bigs – how is this going? do you need a review from me or should I wait for the next iteration? In the latter case, when is it due, so I can plan it? Thanks!

bigs:
@raulk the only blocker right now is consensus around /memory multiaddrs at
pierre’s linked issue. i’ve already implemented a version that uses opaque
uint64 slots should we thumbs up that!

bigs:
regarding the dialer work, i think the transport is still necessary (you still have to listen somewhere in your code) but it will be much cleaner determining where we want to use it, just as you’ve described.

raulk:
@bigs 

> i think the transport is still necessary (you still have to listen somewhere in your code)

Hm, I think listening is not necessary. I think of this as a persistent, in-memory bridge rather than a traditional transport.

There's only ever a singleton _virtual_ connection. We create it at system startup, and whenever the swarm processes a dial to ourselves, we return _that_ connection.

We probably don't even want to generate any Notifee events, nor trigger the connection handler. It doesn't need to be registered in the swarm either, and it shouldn't be subject to connection management.

We'll need access to the [swarm's `StreamHandler`](https://github.com/libp2p/go-libp2p-swarm/blob/b0cc226042a89b22d74ac9cd17f5aa052d9c4185/swarm_conn.go#L115), so we can invoke it when a new stream is opened. That should give way to multistream-select negotiation, and eventually a callback to the appropriate stream handler.

One detail here is that we should only consider handlers that have annotated themselves as "supporting self-dialling", as per the original design: https://github.com/libp2p/go-libp2p/issues/328#issuecomment-465275237

bigs:
i suppose so long as a stream handler exists, listening becomes irrelevant, though that does break pierre’s testing use case, which could be quite nice. it allows us to allocate an address that doesn’t consume any resources beyond cpu/memory which could potentially enable large tests. do see what you mean, though.

bigs:
doesn't look like my comment posted but as of last friday this is ready for final review. only thing left after review is go.mod directive replacement and publish.

@raulk @Stebalien @vyzo

p2p/transport/pipe/pipestream.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (s *PipeStream) Read(b []byte) (int, error) {
+	return s.inbound.Read(b)
+}
+
+func (s *PipeStream) Write(b []byte) (int, error) {
+	return s.outbound.Write(b)
+}
+
+func (s *PipeStream) SetDeadline(t time.Time) error {
```
Stebalien: When setting deadlines, we can probably improve perf by only setting deadlines on the side we care about (read from inbound, write from outbound).

p2p/transport/pipe/pipestream.go
```diff
@@ -0,0 +1,77 @@
...
+
+	streammux "github.com/libp2p/go-stream-muxer"
+)
+
+var ErrHalfClosed = fmt.Errorf("tried to write to stream that was half closed")
+var ErrAlreadyHalfClosed = fmt.Errorf("tried to half close stream that was already half closed")
+
+type PipeStream struct {
+	inbound  net.Conn
+	outbound net.Conn
```
Stebalien: I wonder if go would accept a patch to expose a `CloseWrite` function...

p2p/transport/pipe/pipestream.go
```diff
@@ -0,0 +1,77 @@
...
+		inbound:  inbound,
+		outbound: outbound,
+	}
+}
+
+func (s *PipeStream) Close() error {
+	return s.outbound.Close()
+}
+
+func (s *PipeStream) Reset() error {
```
Stebalien: Hm. With a normal pipe conn, we'd be able to call `CloseWithError`. Unfortunately, that doesn't look possible here.

We either need to:

1. Re-implement this type.
2. Set some atomic "reset" flag _before_ closing. and check that flag _after_ reading/writing.

p2p/transport/pipe/pipestream.go
bigs: sounds good!

p2p/transport/pipe/pipestream.go
bigs: what exactly do you mean here? the implementation as stands seems reasonable to me

Stebalien: Resetting a stream needs to close it with an error. We don't want the other side to think we _successfully_ finished writing, we want them to see that something went wrong. For example, `ioutil.ReadAll(stream)` should fail of the remote side calls `Reset()`.

bigs: ahhh makes total sense. missed this comment. implemented a simple version of this with buffered channels.

bigs: updated friday @Stebalien

go.mod
```diff
@@ -35,3 +39,7 @@ require (
 	github.com/multiformats/go-multistream v0.1.0
 	github.com/whyrusleeping/mdns v0.0.0-20180901202407-ef14215e6b30
 )
+
+replace github.com/libp2p/go-libp2p-swarm => ../go-libp2p-swarm
+
+replace github.com/multiformats/go-multiaddr => ../../workspace-multiformats/go-multiaddr
```
Jorropo: Nothing else to say :)

p2p/transport/pipe/pipetransport.go
Jorropo: Even if that not a thing here I think we can expect some people or other maybe dialling multiple process via shared memory or multiple instance of libp2p in a same process for debugging other layer of libp2p in the future.

mffrench:
Hi, 

is there any progress on that effort ? seems to me it has been abandonned since more a year now and I'm facing similar issue on my own project now and it seems I will have to spent time on a feature which would be definitively great to have in libp2p ...

In my case I would like to simulate several identity in a tiny environment regarding the amount of p2p node roles I've to manage. I'm currently starting many p2p nodes (currently 12) but it has its limit (many current IOs leading to many timeouts ... Note: libp2p nodes are not the only IOs consuming component in my test scenario). Let me know if you see some helpers...

Thank you

Stebalien:
If you're doing local testing, I recommend using a "mock network": https://pkg.go.dev/github.com/libp2p/go-libp2p@v0.12.0/p2p/net/mock#Mocknet.

mffrench:
Well this is not only local testing, I would like some of my libp2p node to handles several identity/IPFS addresses a little like nginx can handles several URLs. 

Currently for each identity I'm running one libp2p node because of `dial to self attempted` issue. But in my business case, most of these identity should be managed by a super libp2p node regarding their roles on the network...

Stebalien:
You want multiple identities but one node? Every node will always have it's own identity. Are you sure you need multiple identities not just multiple protocols? Let's move this discussion over to https://discuss.libp2p.io, I think we can solve your issue without any new features.

MarcoPolo:

PR #639, opened 2019-05-21
Close the connection manager

vyzo:


vyzo:
ok, I'll add the README update. Care to do the announcement on discss?

PR #642, opened 2019-05-22
dep: import go-smux-* into the libp2p org

### LINKED ISSUE #640: libp2p cannot build
https://github.com/whyrusleeping/go-smux-multiplex/issues/15
libp2p go mod depends on many whyrusleeping projects with deleted version tags.
This causes build failure and all projects depend on libp2p fail to build as well

Stebalien:
0. I imported whyrusleeping/go-smux-* into go mod.
1. To handle post-1.0 tags, I renamed all of our "gx" tags to "gx/v..." as we did with the rest of libp2p.
2. This broke _everything_ because `go mod` absolutely can't deal with deleted tags if any transitive go.sum file mentions the tag.
3. To fix this, I first got rid of all mentions of these 2.x+incompatible tags.
4. Unfortunately, I realized this was still going to be a nightmare for downstream users as a `go get github.com/libp2p/go-libp2p@latest` on a fresh clone of said downstream project (no go mod cache) would fail. I didn't want to have to ask all of our downstream users to "please delete your go.sum file".
5. So I did what I should have done in the first place:
  a. I restored those tags.
  b. I forked all the whyrusleeping/go-smux-* repositories (and whyrusleeping/yamux) into the libp2p org. We planned on doing this anyways.
  c. I removed all tags but the _new_ go-mod tags on those new libp2p repos.

What does this mean for downstream users? If they manually import any of these stream muxer repos, they'll have to fix their imports. Otherwise, it shouldn't make a difference (as long as they skip version 0.0.26).

fixes #640


Stebalien:
Merging as everyone is asleep and everything is broken...

PR #664, opened 2019-06-28
if we already have a peers key, don't bother checking identify for it

whyrusleeping:
Obviously this change requires more change than this, but I want to discuss it a little bit first.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -317,6 +317,10 @@ func (ids *IDService) consumeReceivedPubKey(c network.Conn, kb []byte) {
 		return
 	}
 
+	if ids.Host.Peerstore().PubKey(rp) != nil {
+		return
+	}
+
 	newKey, err := ic.UnmarshalPublicKey(kb)
 	if err != nil {
 		log.Warningf("%s cannot unmarshal key from remote peer: %s, %s", lp, rp, err)


raulk:
Shouldn't we be checking the pubkey we received is equal to the pubkey we have?

whyrusleeping:
@raulk if we do that, that means we have to parse it. Parsing public keys is showing up as our biggest source of allocations right now.

whyrusleeping:
@Stebalien technically yes, theres code farther down in the method that pulls the key out of the peerstore and checks that they are the same.

Stebalien:
@whyrusleeping let's double check that, I may have been wrong. We shouldn't be decoding RSA keys with golang's crypto if we're building with openssl.

whyrusleeping:
@Stebalien youre right, its not the decoding thats the problem, its the serialization. `currKey.Equals(newKey)` at the end of the method is the culprit.

p2p/protocol/identify/id.go
```diff
@@ -317,6 +317,10 @@ func (ids *IDService) consumeReceivedPubKey(c network.Conn, kb []byte) {
 		return
 	}
 
+	if ids.Host.Peerstore().PubKey(rp) != nil {
+		return
+	}
```
jbenet: I think this will introduce problems. i believe that we need to verify the key matches here (not just use it), and recall having issues that led to this

Stebalien: Things this code can currently do:

1. If `conn.RemotePeer()` returns `""` (insecure transport?) try to associate `""` with the new key (WTF?) which will fail (good) because `""` isn't a valid peer ID (phew).
2. If the peerstore doesn't have a key for `conn.RemotePeer()`, it adds it.
3. Log an error.

This change leaves 2 unchanged, 1 is just buggy (probably should have been `AddPubKey(np, newKey)`, and 3 shouldn't affect anything.

Stebalien:
This is no longer a performance hot-spot.

PR #667, opened 2019-07-02
mdns: always use interface addresses

Stebalien:
We don't want to use the transformed/munged host addresses for local announcements. Ideally, we'd take a more scientific approach to this (i.e., "host, please give me addresses relative to X") but we can't do that yet.

Note: this is causing mdns to break when used with autorelay


Stebalien:
ping @vyzo

PR #674, opened 2019-07-10
Fix racy and failing test cases

### LINKED ISSUE #671: TestStreamStress failing with race detector
Due to too many goroutines. I believe the issue is that we now open even _more_ streams for identify+ protocols.

### LINKED ISSUE #672: TestStreamStress doesn't wait for the test to finish
`makePinger` and `makePonger` shouldn't launch goroutines as we don't end up _waiting_ for these goroutines which means we finish the test before it's actually done.

https://github.com/libp2p/go-libp2p/blob/198633b285eb5daaf6507506bfa1f28e4f1c20f9/p2p/net/mock/mock_test.go#L322-L365

raulk:
Fixes #671, fixes #672, and integrates a peerstore fix to prevent a race.


p2p/net/mock/mock_test.go
```diff
@@ -368,7 +374,7 @@ func TestStreamsStress(t *testing.T) {
 	ctx := context.Background()
 	nnodes := 100
 	if detectrace.WithRace() {
-		nnodes = 50
+		nnodes = 30
```
raulk: This is OK (and even better) because we're stress testing _streams_, so reducing the host count increases stream density per host.

raulk:
Tests are green 💚🎉 (although `TestStreamsStress` takes ages on Travis).

p2p/net/mock/mock_test.go
```diff
@@ -319,28 +318,35 @@ func TestStreams(t *testing.T) {
...
+		done := make(chan struct{}, 1)
 		go func() {
+			defer func() { done <- struct{}{} }()
 			defer helpers.FullClose(s)
 
 			for i := 0; i < n; i++ {
 				b := make([]byte, 4+len(st))
 				if _, err := s.Write([]byte("ping" + st)); err != nil {
-					panic(err)
+					t.Fatal(err)
```
Stebalien: nit: go doesn't like calling fatal from a goroutine. We should probably just Error and return.

p2p/net/mock/mock_test.go
```diff
@@ -351,13 +357,13 @@ func makePonger(st string) func(network.Stream) {
 					if err == io.EOF {
 						return
 					}
-					panic(err)
+					t.Fatal(err)
```
Stebalien: Same.

p2p/net/mock/mock_test.go
```diff
@@ -319,28 +318,35 @@ func TestStreams(t *testing.T) {
 
 }
 
-func makePinger(st string, n int) func(network.Stream) {
-	return func(s network.Stream) {
+func makePinger(t *testing.T, st string, n int) func(network.Stream) chan struct{} {
+	t.Helper()
+
+	return func(s network.Stream) chan struct{} {
+		done := make(chan struct{}, 1)
```
Stebalien: Why not just drop the goroutine?

p2p/net/mock/mock_test.go
raulk: Fixed.

p2p/net/mock/mock_test.go
raulk: Fixed.

p2p/net/mock/mock_test.go
raulk: Fixed.

PR #679, opened 2019-07-14
WIP - Provide private key to insecure.New()

mgoelzer:
insecure.New() requires a private key argument.  The key is used only for identity generation.  (See [this line in github.com/libp2p/go-libp2p-core/sec/insecure/insecure.go](https://github.com/libp2p/go-libp2p-core/blob/1d45af25d96204b2f5fd6ea13a6ebe573ae25a49/sec/insecure/insecure.go#L35))  This PR passes the host's private key through to the chain of functions that lead to a call to insecure.New() to correct a compile-time error in `go-libp2p/config/config.go` on the line `upgrader.Secure = makeInsecureTransport(pid)`.  

When the insecure transport is not being used, the compiler seems to optimize that line away before it can emit an error, which may be why this error hasn't been noticed previously.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -142,7 +142,7 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 	upgrader.Protector = cfg.Protector
 	upgrader.Filters = swrm.Filters
 	if cfg.Insecure {
-		upgrader.Secure = makeInsecureTransport(pid)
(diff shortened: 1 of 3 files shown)


mgoelzer:
Hmm, I need to study what test this is breaking and why... relabeled as WIP.

mgoelzer:
Closing this PR because https://github.com/libp2p/go-libp2p-core/pull/38 reverts the change that made `insecure.New()` take a second argument.

PR #683, opened 2019-07-22
Added missing error check in test

muesli:
Since err gets assigned, I assume we also want to check for it being nil.


muesli:
@raulk Actually I tried finding them with static code analysis. Looks like that's all. I fixed a few more unused but assigned values in a separate PR: https://github.com/libp2p/go-libp2p/pull/682

PR #697, opened 2019-08-02
Update deps, mocknet tests

bigs:
Bumps go-libp2p-core and go-libp2p-testing, updates mocknet tests to use Ed25519.

@Stebalien only weird thing here is the `TestStreamsWithLatency` fails as the test now takes a bit more time to execute.


Stebalien:
> @Stebalien only weird thing here is the TestStreamsWithLatency fails as the test now takes a bit more time to execute.

Let's just bump the times.

bigs:
Ready for review @Stebalien

bigs:
This needs to go after secio

PR #702, opened 2019-08-13
better user-agent handling

Stebalien:
This PR improves user-agent handling by:

1. Deprecating the global variable.
2. Setting the default using go mod.
3. Adding an option for overriding the user-agent.


p2p/protocol/identify/id.go
```diff
@@ -31,9 +33,27 @@ const ID = "/ipfs/id/1.0.0"
...
+func init() {
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		return
+	}
+	version := bi.Main.Version
+	if version == "(devel)" {
+		ClientVersion = bi.Main.Path
+	} else {
+		ClientVersion = fmt.Sprintf("%s@%s", bi.Main.Path, bi.Main.Version)
```
Stebalien: This is the path/version of the _main_ module (application), not libp2p itself.

p2p/protocol/identify/id.go
```diff
@@ -31,9 +33,27 @@ const ID = "/ipfs/id/1.0.0"
...
 // TODO(jbenet): fix the versioning mess.
+// XXX: Don't change this till 2020. You'll break all go-ipfs versions prior to
+// 0.4.17 which asserted an exact version match.
 const LibP2PVersion = "ipfs/0.1.0"
 
-var ClientVersion = "go-libp2p/3.3.4"
+// ClientVersion is the default user agent.
+//
+// Deprecated: Set this with the UserAgent option.
+var ClientVersion = "github.com/libp2p/go-libp2p"
```
vyzo: let's just call this `go-libp2p`.

p2p/protocol/identify/id.go
```diff
@@ -31,9 +33,27 @@ const ID = "/ipfs/id/1.0.0"
...
+
+func init() {
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		return
+	}
+	version := bi.Main.Version
+	if version == "(devel)" {
+		ClientVersion = bi.Main.Path
```
vyzo: let's keep the `devel` designator somewhere in the reported path.

p2p/protocol/identify/id.go
Stebalien: Sure but users using go mod will still use their full paths by default. E.g., `github.com/libp2p/go-libp2p-examples@v0.1.1`.

p2p/protocol/identify/id.go
Stebalien: Anything built manually (i.e., built from a manually checked-out source tree) will have a `(devel)` version. It doesn't mean it's _actually_ a development version.

To get a proper version, one would have to build with `go get github.com/libp2p/go-libp2p-examples@latest` (i.e., build with go mod).

Users that care can override this.

p2p/protocol/identify/id.go
vyzo: oh ok, nevermind then.

p2p/protocol/identify/id.go
vyzo: ok.

Stebalien:
@raulk this:

1. Changes the default user agent for downstream projects from a static string to 'github.com/foo/bar@VERSION`.
2. Bumps the minimum go version to 1.12 (which we may implicitly require anyways?).

p2p/protocol/identify/id.go
raulk: This will increase the size of the identify messages by 15 bytes. It's ok but worth noting.

p2p/protocol/identify/id.go
raulk: Can we suffix with @unknown?

p2p/protocol/identify/id.go
```diff
@@ -47,7 +67,8 @@ const transientTTL = 10 * time.Second
 //  * Our IPFS Agent Version
 //  * Our public Listen Addresses
 type IDService struct {
-	Host host.Host
+	Host      host.Host
+	UserAgent string
```
raulk: Keeping this exported could lead to races if the user modifies the value at runtime. Not a biggie, but just pointing it out.

p2p/protocol/identify/id.go
```diff
@@ -31,9 +33,27 @@ const ID = "/ipfs/id/1.0.0"
 
 // LibP2PVersion holds the current protocol version for a client running this code
 // TODO(jbenet): fix the versioning mess.
+// XXX: Don't change this till 2020. You'll break all go-ipfs versions prior to
```
raulk: We should make this configurable too. Many projects using libp2p don't care about compatibility with IPFS and their networks are deliberately segregated.

raulk: See #714

Issue #694, opened 2019-07-30
95% of DHT find provider requests are trying to find relays


Stebalien:
When combined with the issues outlined in https://github.com/libp2p/go-libp2p-kad-dht/issues/146, this is spamming the network. Worse, we're doing this _constantly_.

Every 15 minutes, if we're connected to fewer than 5 relays, we search for relays ~5~ *3* times until we give up, waiting 30s between each search.

That means:

1. 30 seconds searching, 30 seconds waiting. 5 times (5 minutes).
2. 10 minutes waiting.

That means that a full _third_ of the network (assuming they have auto-relay enabled) is asking for this key at any given time unless they happen to find ~5~ *3* stable nodes.

---

Key issues:

* Requiring 5 is probably too many.
* This module makes _no_ effort to differentiate between stable/unstable relays. This means that crappy relays with low connection limits could get picked repeatedly, causing relay users to constantly have to try to find more relays.

Stebalien:
For now, let's just hard-code a set of relays and turn this off.

vyzo:
It tries to find 3 relays, but we can reduces this to 1 or 2.

vyzo:
And yeah, it should try to estimate relay stability to pick good relays.

Stebalien:
> It tries to find 3 relays, but we can reduce this to 1 or 2.

Ah, got it. It just tries 5 times. I'm pretty sure one of the issues here is that there are groups running relays with low connection limits. That's why we need to make sure we pick a stable relay.

In the mean-time, I'd like to switch to hard-coded relays.

vyzo:
I don't think hard-coded relays are a very good idea, we can't just route all the traffic to our 3 medium relays and call it day. Who else is a stable relay operator? 
I think it's better to try and estimate relay stability and pick accordingly.

Stebalien:
The current issue is both stability _and_ https://github.com/libp2p/go-libp2p-kad-dht/issues/146. I'm not kidding when I say that 95% of the DHT requests are trying to find relays.

vyzo:
This is a problem indeed, but hard-coding relays doesn't seem like the right solution. What relays to hard code? We don't have operating guidelines or known relay operators we can contact, and we (as pl) can't run enough relays to support the entire network.

Stebalien:
[...]

PR #705, opened 2019-08-16
options to configure static relays for autorelay

vyzo:
So that we can avoid hitting the DHT; also changes the desired relays parameter to 1 (from 3).
See #694

### DIFF
--- a/options.go
+++ b/options.go
@@ -17,6 +17,7 @@ import (
 	circuit "github.com/libp2p/go-libp2p-circuit"
 	config "github.com/libp2p/go-libp2p/config"
 	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+	autorelay "github.com/libp2p/go-libp2p/p2p/host/relay"
 
 	filter "github.com/libp2p/go-maddr-filter"
 	ma "github.com/multiformats/go-multiaddr"
@@ -251,19 +252,15 @@ func EnableAutoRelay() Option {
 // discover relays
 func StaticRelays(relays []peer.AddrInfo) Option {
 	return func(cfg *Config) error {
-		cfg.StaticRelays = relays
+		cfg.StaticRelays = append(cfg.StaticRelays, relays...)
 		return nil
 	}
 }
 
-// DefaultRelays configures the static relays to use the known PL-operated relays
-func DefaultRelays() Option {
+// DefaultStaticRelays configures the static relays to use the known PL-operated relays
+func DefaultStaticRelays() Option {
 	return func(cfg *Config) error {
-		for _, addr := range []string{
-			"/ip4/147.75.80.110/tcp/4001/p2p/QmbFgm5zan8P6eWWmeyfncR5feYEMPbht5b1FW1C37aQ7y",
-			"/ip4/147.75.195.153/tcp/4001/p2p/QmW9m57aiBDHAkKj9nmFSEn7ZqrcF1fZS4bipsTCHburei",
-			"/ip4/147.75.70.221/tcp/4001/p2p/Qme8g49gm3q4Acp7xWBKg3nAa9fxZ1YmyDJdyGgoG6LsXh",
-		} {
+		for _, addr := range autorelay.DefaultRelays {
 			a, err := ma.NewMultiaddr(addr)
 			if err != nil {
 				return err
--- a/options.go
+++ b/options.go
@@ -10,6 +10,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/connmgr"
 	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/metrics"
+	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/peerstore"
 	"github.com/libp2p/go-libp2p-core/pnet"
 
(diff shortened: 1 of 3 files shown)


albrow:
Does the `StaticRelays` option mean that those are the *only* relays that can be used? Ideally, for our network we would like to use static relays to reduce the time to first connection but we still want to discover other relays over time. Otherwise our dedicated relay hosts would be overworked.

vyzo:
With this patch yes, when specified only those will be used.
Perhaps we should have a fallback to the DHT, but the issue is that we are having excessive DHT traffic because of autorelay.

vyzo:
ping @Stebalien

vyzo:
Moving forward we also want to use relays discovered by passive discovery (when the option `circuit.OptDiscovery` is specified).

options.go
```diff
@@ -245,6 +246,39 @@ func EnableAutoRelay() Option {
...
+	}
+}
+
+// DefaultRelays configures the static relays to use the known PL-operated relays
+func DefaultRelays() Option {
+	return func(cfg *Config) error {
+		for _, addr := range []string{
+			"/ip4/147.75.80.110/tcp/4001/p2p/QmbFgm5zan8P6eWWmeyfncR5feYEMPbht5b1FW1C37aQ7y",
+			"/ip4/147.75.195.153/tcp/4001/p2p/QmW9m57aiBDHAkKj9nmFSEn7ZqrcF1fZS4bipsTCHburei",
+			"/ip4/147.75.70.221/tcp/4001/p2p/Qme8g49gm3q4Acp7xWBKg3nAa9fxZ1YmyDJdyGgoG6LsXh",
```
raulk: Can we have these multiaddrs in a global var? You can initialize them with an IIFE, or in an init function.

p2p/host/relay/autorelay.go
```diff
@@ -48,12 +50,13 @@ type AutoRelay struct {
 	cachedAddrsExpiry time.Time
 }
 
-func NewAutoRelay(ctx context.Context, bhost *basic.BasicHost, discover discovery.Discoverer, router routing.PeerRouting) *AutoRelay {
+func NewAutoRelay(ctx context.Context, bhost *basic.BasicHost, discover discovery.Discoverer, router routing.PeerRouting, static []peer.AddrInfo) *AutoRelay {
```
raulk: This starts screaming "functional options".

p2p/host/relay/autorelay.go
```diff
@@ -245,6 +248,10 @@ func (ar *AutoRelay) connect(ctx context.Context, pi peer.AddrInfo) bool {
 }
 
 func (ar *AutoRelay) discoverRelays(ctx context.Context) ([]peer.AddrInfo, error) {
+	if len(ar.static) > 0 {
```
raulk: I don't think setting static relays should be exclusive from dynamic relay discovery. You probably want static relays for a speedier bootup, but as you dynamically discover relays, you should preferentially switch to those.

raulk: I see that @albrow expected the same thing: https://github.com/libp2p/go-libp2p/pull/705#issuecomment-522708344

options.go
vyzo: sure.

p2p/host/relay/autorelay.go
vyzo: *shrug*

p2p/host/relay/autorelay.go
vyzo: This defeats the purpose; the main impetus for the static relay configuration is to avoid hitting the DHT _altogether_.
We might want to leverage passive discovery with `CanHop` in the relay module, but I don't think it's a good idea to have both static relays _and_ hit the dht.

p2p/host/relay/autorelay.go
vyzo: it's not really user visible, so it doesn't offer much in terms of ergonomics.

vyzo:
Updated to use a global variable in the autorelay module for the default relays; also renamed the `DefaultRelays` to `DefaultStaticRelays` and made `StaticRelays` append instead of overwriting.

Stebalien:
Agreed. Static relays are really just an emergency measure until we can implement a better system.

RubenKelevra:
> Agreed. Static relays are really just an emergency measure until we can implement a better system.

How about flipping the process, having a node subscribe to pubsub when they need a relay.

All nodes which offer the relay service will send a message on this pubsub on how many connections they can accept.

The number is multiplied with a number x like 0.01 in the beginning.

On connection to the relay we send a 'request' for relay services. When the relay is already full, it will send a deny message and close the connection.

On each failure to connect to a relay, the number x is cut down in half, to simply estimate the 'load' on all the relays.

After a random period between 30 and 600 minutes the number x is multiplied by 2 again.

Stebalien:
We should advertise relays through pubsub (and maybe even just have relays advertise other relays). However, there's still the issue of reputation.

RubenKelevra:
@Stebalien I think that's an issue for another time. I think it's best to just replace the DHT method with something that's responding dynamically to the network size, by lowering the connection attempts.

The optimal solution for fixing this would be to add peers in a Web Of Trust fashion and allow 'friends' to use the relay function.

We could also hold those connections all the time and bootstrap with the friend connections first, instead of the static IPs and domains we currently use.

Stebalien:
I'm not sure if that's an issue we _can_ punt on. It takes a bit of effort to attack the DHT. It would take very little effort to add enough bad relays to affect the network.

We don't really need a full WoT system, even just something that verifies that a relay is doing the right thing would be enough. For example, some way to ask other peers to dial back through the relay and verify that it's working.

RubenKelevra:
@Stebalien if we change the amount of desired relays back to 3, we could check each of those relays by requesting some data:

One relay is asked to provide some random stuff, and we request it thru both other relays.

We cycle this thru all 3 relays to confirm all relays are working fine.

---

Another possibility is to request everything round robin thru the 3 relays, and if we identify a relay which is sending data which doesn't match the checksums for example, we drop that connwc and create a new one.

p2p/host/relay/autorelay.go
```diff
@@ -48,12 +57,13 @@ type AutoRelay struct {
...
 }
 
-func NewAutoRelay(ctx context.Context, bhost *basic.BasicHost, discover discovery.Discoverer, router routing.PeerRouting) *AutoRelay {
+func NewAutoRelay(ctx context.Context, bhost *basic.BasicHost, discover discovery.Discoverer, router routing.PeerRouting, static []peer.AddrInfo) *AutoRelay {
 	ar := &AutoRelay{
 		host:       bhost,
 		discover:   discover,
 		router:     router,
 		addrsF:     bhost.AddrsFactory,
+		static:     static,
```
aschmahmann: @vyzo I know I'm late to the party here, but this felt a bit small for an issue.

Why did we bother changing the constructor and using a `[]Addr.Info` here instead of just creating a `Discoverer` implementation that returns a the static list?

Sure it can be more efficient to use the slice, but now we have this thing we're going to be stuck removing.

PR #710, opened 2019-08-25
use TLS by default, secio as a fallback

marten-seemann:


.travis.yml
```diff
@@ -4,7 +4,7 @@ os:
 language: go
 
 go:
-  - 1.11.x
+  - 1.12.x
```
marten-seemann: We need TLS 1.3, which was added in Go 1.12.

marten-seemann:
Makes sense.
What's the threshold for rolling this out, considering that selecting TLS will save peers one network round trip if both of them support it, whereas falling back to secio will cost one network round trip?

raulk:
Besides the RTT tradeoffs, we need to characterise handshake cost (CPU, memory, allocs, etc.) and ongoing AEAD costs once the secure channel is established. A good test case for our fancy new testbed.

marten-seemann:
> I'd like to wait for ~90%, but I don't feel strongly about this. No latter than October and probably in go-ipfs 0.5.0 (the next feature release).

@Stebalien I rebased and updated the PR. Do you think we can merge this now?

Stebalien:
Let's set it as the default in testground first. I'm worried that https://github.com/libp2p/go-libp2p-tls/pull/37 will be problematic for switching to TLS by default.

marten-seemann:
Isn’t that one fixed?

Stebalien:
Yes, but that fix hasn't been released. If we _dial_ one of these nodes with the TCP transport, we could connect then have the connection closed (by the remote peer) out from under us.

Stebalien:
We will do this. At a _minimum_, we need to support TLS by default, even if it's not _the_ default.

raulk:
We won't be defaulting to TLS, as that would introduce an interoperability tax.

PR #717, opened 2019-09-04
Add Filters option

albrow:
The new `Filters` option allows users to take ownership of the `Filters` struct and is suitable for cases where the addresses you want to accept/deny can change over time. We're using it in [0x Mesh](https://github.com/0xProject/0x-mesh) to implement IP-based rate-limiting/banning (see https://github.com/0xProject/0x-mesh/pull/392).

### DIFF
--- a/options.go
+++ b/options.go
@@ -246,7 +246,8 @@ func EnableAutoRelay() Option {
 }
 
 // FilterAddresses configures libp2p to never dial nor accept connections from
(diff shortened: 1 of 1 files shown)


albrow:
Ping @raulk.

raulk:
Thanks for this! Requesting a sanity check from @stebalien and @bigs.

PR #726, opened 2019-09-20
Add bridged chats

swedneck:


swedneck:
Well that's up to you, i just want people to be aware that you don't need IRC to access the libp2p chat. Only linking to IRC in README's can easily make people think it's the only option.

PR #735, opened 2019-10-31
Added utp transport

Jorropo:
Blocked: libp2p/go-utp-transport#5

Still need to be rebased with the correct go.mod when libp2p/go-utp-transport newer version is released.
I added in default and not in ipfs/go-ipfs because it was here in the past (see c441a14b693d7316700caae2ead204d7de770dd3).


Stebalien:
The downstream repo has been archived. At one point, UTP was going to be our UDP transport, but since QUIC has landed this transport no longer fills a niche.

PR #740, opened 2019-11-05
Identify: Make activation threshold configurable

hsanjuan:
Also: fix the comments and try to explain how the activation/aliveness of
addresses happen (and write the correct defaults)


hsanjuan:
Thankity thanks

PR #746, opened 2019-11-21
Fix #745.

### LINKED ISSUE #745: Need to know if a stream handler was set.
I'm making a new test protocol, `webrtc-aside` the goal is to use any stream and upgrade it to webrtc, (it simply use stream for sdp exchange) so it would be instantly compatible with relay, ws, ...

The only problem is finding stun server but I'm still thinking to it.

The problem is if 2 listen are made the second one overide the first (stalling the first, (and a close of the first would stall the second)) but I want listen to return an error if there is already an stream handler.
So I want `AddHandler` (or a new func `AddHandlerSafe` avoiding a refactor) to return error if there is already an stream set (and not setting the new handler).
It also need to be TOCTOU free, so adding a func `IsHandlerSet` might be usefull else where but not in this case.

Jorropo:
Closes #745.

Still need to be rebased with correct multistream and yamux version.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -398,6 +398,24 @@ func (h *BasicHost) SetStreamHandler(pid protocol.ID, handler network.StreamHand
 	})
 }
 
+// Same as `SetStreamHandler` but return an error if there was already an
+// handler.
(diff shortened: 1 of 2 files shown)


Jorropo:
As @Stebalien says, its impossible to correctly match `SetStreamHandlerMatchSafe` currently.

PR #747, opened 2019-11-21
exchange signed routing records in identify

yusefnapora:
This is still in progress & gomod is pointing at these PR branches:

- https://github.com/libp2p/go-libp2p-core/pull/73
- https://github.com/libp2p/go-libp2p-peerstore/pull/98

TODO:
- [x] test
- [ ] land dependency PRs and update gomod

@vyzo this should be good enough to start hacking on


vyzo:
thank you!

p2p/protocol/identify/id.go
```diff
@@ -316,6 +317,16 @@ func (ids *IDService) populateMessage(mes *pb.Identify, c network.Conn) {
...
 
+	// Generate a signed routing record containing our listen addresses and
+	// send it along with the unsigned addrs
+	signedState, err := host.SignedRoutingStateFromHost(ids.Host)
+	if err != nil {
+		envelopeBytes, err := signedState.Marshal()
+		if err != nil {
+			mes.SignedRoutingState = envelopeBytes
+		}
+	}
```
vyzo: let's not swallow the errors here, warn on them?

p2p/protocol/identify/id.go
```diff
@@ -374,18 +385,35 @@ func (ids *IDService) consumeMessage(mes *pb.Identify, c network.Conn) {
 	// that picks random source ports, this can cause DHT nodes to collect
 	// many undialable addresses for other peers.
 
+	// add certified addresses for the peer, if they sent us a signed routing
+	// state record
+	routingState, err := signedRoutingStateFromMsg(mes)
+	if err != nil {
+		log.Warning("error getting routing state from Identify message: %v", err)
```
vyzo: you probably meant `Warningf` here.

p2p/protocol/identify/id.go
yusefnapora: good call. I totally had the if condition backwards too, lol.

yusefnapora:
@vyzo how do you feel about my last commit (https://github.com/libp2p/go-libp2p/pull/747/commits/028fe9880995a25beb7631c78a834ee016c9db17)?

I added a config option to disable sending or parsing signed address records so that we could test the behavior where one peer supports them but the other doesn't. It kind of bugs me though, since the flag is only really useful for testing and it clutters up the code. I couldn't figure out a better way to test it though, short of spinning up both versions for real in something like testground.

vyzo:
Hrm, it's not ideal but we do need to test somehow.
Perhaps give it a more explicit name that indicates it's only used for testing?

yusefnapora:
@vyzo I updated this branch based on @raulk's [design blueprint](https://github.com/libp2p/go-libp2p/issues/761#issue-536486675). It points to a new commit from the -core PR that defines some new event types, and I rebased this branch and the -core PR onto master, since there were a few dependency updates since I opened the PRs.

The gist is that now `BasicHost` emits a new `EvtLocalAddressesUpdated` event when it detects addresses have changed, instead of directly calling `ids.Push()`.

There's also a new `routingStateManager` component in the identify service that listens for `EvtLocalAddressesUpdated` events and builds a new `SignedRoutingState` record. It emits a `LocalPeerRoutingStateUpdated` event, which the main identify service listens for. When the id service gets that event, it triggers the identify/push.

This adds a bit more indirection, but it should give you a way to grab new routing state record in other places outside the ID service (e.g. peer exchange).

Unfortunately, the mock network tests seem to have gotten flakier - I had to re-run them on Travis, and they seem to fail pretty often on my laptop. I'm not sure if I actually made it worse, or if there's just more spam in the error log - there's a ton of error messages from `boguskey` ("this better be a test!"), since it's using bogus testing keys to sign/verify the routing records.

Since the mock net tests have some assertions about timing (e.g. "expected to take ~2 seconds, took x"),  it's possible that I made performance regress enough to fail the tests. It's odd that they'd fail with the bogus keys though, since they don't actually do any expensive crypto. But perhaps the event bus messes with the timing, or just the overhead of marshaling / unmarshaling the records? I've had these tests fail before, so maybe it's unrelated, but they definitely spew out a lot more errors now, thanks to the bogus key thing.

Anyway, hopefully this helps with the peer exchange stuff you're working on - you should be able to listen for `EvtLocalRoutingStateUpdated` from anywhere and get the latest records.

I'm off next week, but you can ping me if you're blocked on this. I'm going to switch this PR to "ready for review", but we'll need to merge the -core and -peerstore PRs first and update gomod here before we can merge this one, since it's pinned to specific commits from those branches at the moment.

vyzo:
Is `Push` still called somehow? We need this to propagate our address changes to our peers.

yusefnapora:
@raulk, @vyzo, etc

I realized that I may be aiming a gun at our foot with respect to LAN-local addresses.

As it is now in the PR, the default behavior is to filter LAN and loopback addresses out of `PeerRecord`s before we send them out in identify, etc. There's a flag that can change this behavior so that you can include local addresses, but at the moment it's only on the `IDService`; I was about to bubble the option up to the `libp2p.New` options when I stopped to think a moment.

There are two behaviors that are problematic in tandem:
- The `Peerstore` will replace any existing addrs for a peer with the addrs from a `PeerRecord`
- There's currently nothing stopping us from sending out empty `PeerRecord`s. A peer with _only_ local addrs will still send out a record with an empty `Addrs` list after filtering the local addrs out. The receiver will then drop the existing local addrs it knows for the peer and replace with the empty set from the `PeerRecord`.

This effectively means that if you don't set the option to include LAN addresses in `PeerRecord`s, anyone that happens to find your LAN address out-of-band will forget it as soon as they talk to you :( 

I'd rather not make interacting with LAN peers worse by default, which I think this does. I still need to write a test case to confirm that this is happening, btw. But I started to suspect that this was a problem based on some odd behavior in earlier tests, so I don't think I'm chasing a ghost.

I'm going to write that test, but in the meantime if anyone has ideas for addressing this let me know.

The simplest option that I can think of is to change the default, so that you have to opt-out of including local addrs in `PeerRecord`s. That has us putting a lot of irrelevant addresses on the public DHT though...

Alternatively, we could prevent you from sending signed `PeerRecord`s in identify with no addrs in them, and change the peerstore so that it ignores records with no addrs instead of removing existing addrs. That should force everyone to fallback to the unsigned addrs that are still being sent in identify alongside the peer record, and which aren't filtered.

A more involved and annoying solution would be to have multiple kinds of `PeerRecord`, one with local addrs and one for public sharing on the DHT. But we'd have to revisit some assumptions to make that work.

Other ideas?

yusefnapora:
hmm, looks like I may have introduced a race condition, and/or my tests are buggy. will investigate :)

yusefnapora:
Regarding the problem with LAN addresses above, @raulk and I discussed in a sync chat just now and agreed that changing the default (so that LAN addresses are included in peer records) is the best option for now. This essentially gives us the status quo, but with signatures, since we're exchanging LAN addrs in identify now.

In the future, we can add specialized peer records with different scope (e.g. LAN-only, public DHT, etc), so that we can keep LAN addrs out of the records that get published globally.

p2p/protocol/identify/id.go
```diff
@@ -316,6 +344,18 @@ func (ids *IDService) populateMessage(mes *pb.Identify, c network.Conn) {
 	}
 	log.Debugf("%s sent listen addrs to %s: %s", c.LocalPeer(), c.RemotePeer(), laddrs)
 
+	// Generate a signed peer record containing our listen addresses and
+	// send it along with the unsigned addrs
+	signedRecord := ids.peerRecordManager.LatestRecord()
```
raulk:
```suggestion
	record := ids.peerRecordManager.LatestRecord()
```

p2p/protocol/identify/id.go
```diff
@@ -316,6 +344,18 @@ func (ids *IDService) populateMessage(mes *pb.Identify, c network.Conn) {
...
+	if ids.useSignedAddrs && signedRecord != nil {
```
raulk: We'll need separate protocol versions for identify and identify push. This solution duplicates our addresses, in unsigned and signed forms; that could be made more efficient at the expense of 1 RTT to support legacy peers.

* identify:
  * `/ipfs/id/1.0.0`: deals with unsigned addresses.
  * `/p2p/id/1.1.0`: deals with signed addresses.

* identify push:
  * `/ipfs/id/push/1.0.0`: deals with unsigned addresses.
  * `/p2p/id/push/1.1.0`: deals with signed addresses.

When opening outbound streams, we'd negotiate 1.1.0 and 1.0.0, in that order. If settle on 1.1.0, we know the peer understands peer records and we can omit the unsigned counterparts; if we settle on 1.0.0, it would've cost us 1 extra RTT (which should disappear with ms2.0), but as more of the network upgrades, this cost will gradually decrease network-level.

p2p/protocol/identify/peer_record.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	m.lock.Lock()
+	m.latest = envelope
+	m.lock.Unlock()
+	m.emitLatest()
+}
+
+func (m *peerRecordManager) emitLatest() {
+	m.lock.RLock()
+	stateEvt := event.EvtLocalPeerRecordUpdated{Record: m.latest}
```
raulk: I think `Record` here should be a value, so as to force a copy, not a pointer. Otherwise, we can have spooky action at a distance (an event subscriber can modify our record!).

raulk:
Reminder: we need to make a go-libp2p-core release and update the go.mod here, before we merge this. Let's:

- [ ] Fix the `EvtLocalPeerRecordUpdated` event.
- [ ] Obtain sign-off from @aschmahmann and/or @Stebalien that all this work, in the current shape and form, is indeed easily integratable into go-libp2p-kad-dht.

p2p/protocol/identify/pb/identify.pb.go
```diff
@@ -20,7 +19,7 @@ var _ = math.Inf
 // is compatible with the proto package it is being compiled against.
 // A compilation error at this line likely means your copy of the
 // proto package needs to be updated.
-const _ = proto.GoGoProtoPackageIsVersion3 // please upgrade the proto package
+const _ = proto.GoGoProtoPackageIsVersion2 // please upgrade the proto package
```
Stebalien: We shouldn't downgrade this. Please upgrade your gogo version and regenerate these.

p2p/protocol/identify/peer_record.go
```diff
@@ -0,0 +1,172 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/libp2p/go-eventbus"
+	"github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/event"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/record"
+	"github.com/multiformats/go-multiaddr"
+	"sync"
```
Stebalien: nit: imports grouping (stdlib imports, then a blank line, then everything else).

p2p/protocol/identify/id.go
```diff
@@ -3,6 +3,7 @@ package identify
 import (
 	"context"
 	"fmt"
+	"github.com/libp2p/go-libp2p-core/record"
```
Stebalien: ditto. import grouping.

p2p/protocol/identify/id.go
```diff
@@ -113,27 +123,40 @@ func NewIDService(ctx context.Context, h host.Host, opts ...Option) *IDService {
...
+	} else {
+		go s.handleEvents(s.subscriptions.localPeerRecordUpdated, s.handlePeerRecordUpdated)
 	}
 
 	s.emitters.evtPeerProtocolsUpdated, err = h.EventBus().Emitter(&event.EvtPeerProtocolsUpdated{})
 	if err != nil {
 		log.Warningf("identify service not emitting peer protocol updates; err: %s", err)
 	}
 
+	hostKey := h.Peerstore().PrivKey(h.ID())
```
Stebalien: You sure this is always non-nil?

p2p/protocol/identify/id.go
```diff
@@ -113,27 +123,40 @@ func NewIDService(ctx context.Context, h host.Host, opts ...Option) *IDService {
...
+	s.peerRecordManager, err = NewPeerRecordManager(ctx, h.EventBus(), hostKey, h.Addrs())
```
Stebalien: Instead of passing the addrs, shouldn't we trigger an addrs change event up-front?

p2p/host/basic/basic_host.go
```diff
@@ -329,32 +336,44 @@ func (h *BasicHost) background(p goprocess.Process) {
 	for {
 		select {
 		case <-ticker.C:
```
Stebalien: The timeout here is too long given that we literally won't announce new addresses to anyone, even new peers, until the timeout elapses. For now, we should probably just shrink the timeout to 5-10 seconds at the very most. Ideally we'd make this entirely event driven with multiple event "sources" but that's kind of tricky.

p2p/protocol/identify/id.go
```diff
@@ -147,13 +170,22 @@ func (ids *IDService) handleEvents() {
...
 			return
 		}
 	}
 }
 
+func (ids *IDService) handleProtosChanged(evt interface{}) {
+	ids.fireProtocolDelta(evt.(event.EvtLocalProtocolsUpdated))
+}
+
+func (ids *IDService) handlePeerRecordUpdated(evt interface{}) {
```
Stebalien: nit: signal that this isn't used by naming `evt` `_` (or `_evt`).

p2p/protocol/identify/id.go
```diff
@@ -308,13 +346,30 @@ func (ids *IDService) populateMessage(mes *pb.Identify, c network.Conn) {
...
-	for i, addr := range laddrs {
-		mes.ListenAddrs[i] = addr.Bytes()
+	if usePeerRecords {
+		// Generate a signed peer record containing our listen addresses and
+		// send it instead of unsigned addrs
+		rec := ids.peerRecordManager.LatestRecord()
+		if rec == nil {
+			log.Errorf("latest PeerRecord does not exist. identify message incomplete!")
+		} else {
+			recBytes, err := rec.Marshal()
```
Stebalien: Hm. Can we pre-marshal this?

p2p/protocol/identify/id.go
```diff
@@ -374,18 +429,37 @@ func (ids *IDService) consumeMessage(mes *pb.Identify, c network.Conn) {
...
+	if ids.Host.Network().Connectedness(p) == network.Connected {
+		ttl = peerstore.ConnectedAddrTTL
+	}
+
+	// invalidate previous addrs -- we use a transient ttl instead of 0 to ensure there
+	// is no period of having no good addrs whatsoever
+	ids.Host.Peerstore().UpdateAddrs(p, peerstore.ConnectedAddrTTL, transientTTL)
+
+	// add signed addrs if we have them and the peerstore supports them
+	cab, ok := peerstore.GetCertifiedAddrBook(ids.Host.Peerstore())
```
Stebalien: Not blocking this PR but this feels like an overkill abstraction. If certified addresses are becoming the norm, I'd expect all peerstores to handle them natively.

p2p/protocol/identify/id.go
```diff
@@ -374,18 +429,37 @@ func (ids *IDService) consumeMessage(mes *pb.Identify, c network.Conn) {
...
+	if ok && signedPeerRecord != nil {
+		_, addErr := cab.ConsumePeerRecord(signedPeerRecord, ttl)
+		if addErr != nil {
+			log.Errorf("error adding signed addrs to peerstore: %v", addErr)
```
Stebalien: Is this a programmer error? If not, we should probably log this as an info.

p2p/protocol/identify/peer_record.go
```diff
@@ -0,0 +1,172 @@
...
+		case <-m.ctx.Done():
+			return
+		}
+	}
+}
+
+func (m *peerRecordManager) update(evt event.EvtLocalAddressesUpdated) {
+	envelope, err := m.makeSignedPeerRecord(addrsFromEvent(evt))
+	if err != nil {
+		log.Warnf("error creating signed peer record: %v", err)
```
Stebalien: This is an error.

p2p/protocol/identify/peer_record.go
```diff
@@ -0,0 +1,172 @@
...
+	m.latest = envelope
+	m.lock.Unlock()
+	m.emitLatest()
+}
+
+func (m *peerRecordManager) emitLatest() {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+	if m.latest == nil {
+		log.Warnf("emitLatest called, but no latest record exists")
```
Stebalien: This is also an error, right?

p2p/protocol/identify/peer_record.go
```diff
@@ -0,0 +1,172 @@
...
+		return
+	}
+
+	stateEvt := event.EvtLocalPeerRecordUpdated{Record: m.latest}
+	err := m.emitters.evtLocalPeerRecordUpdated.Emit(stateEvt)
+	if err != nil {
+		log.Warnf("error emitting event for updated peer record: %v", err)
```
Stebalien: Is this an error or can it happen on shutdown? If so, it should be an info.

p2p/protocol/identify/peer_record.go
```diff
@@ -0,0 +1,172 @@
...
+	return m.latest
+}
+
+func (m *peerRecordManager) handleEvents() {
+	sub := m.subscriptions.localAddrsUpdated
+	defer func() {
+		_ = sub.Close()
+		// drain the channel.
+		for range sub.Out() {
+		}
```
Stebalien: @raulk Is this required by the API? That's kind of dangerous.

p2p/protocol/identify/peer_record.go
```diff
@@ -0,0 +1,172 @@
...
+func (m *peerRecordManager) emitLatest() {
+	m.lock.RLock()
+	defer m.lock.RUnlock()
+	if m.latest == nil {
+		log.Warnf("emitLatest called, but no latest record exists")
+		return
+	}
+
+	stateEvt := event.EvtLocalPeerRecordUpdated{Record: m.latest}
+	err := m.emitters.evtLocalPeerRecordUpdated.Emit(stateEvt)
```
Stebalien: Do we _want_ to be holding the lock here?

p2p/protocol/identify/peer_record.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// LatestRecord returns the most recently constructed signed PeerRecord.
+// If you have a direct reference to the peerRecordManager, this method
+// gives you a simple way to access the latest record without pulling
+// from the event bus.
+func (m *peerRecordManager) LatestRecord() *record.Envelope {
+	if m == nil {
+		return nil
+	}
```
raulk: When can m be nil? This seems weird.

p2p/protocol/identify/peer_record.go
raulk: It's superfluous here as go-eventbus already drains the channel on close:

https://github.com/libp2p/go-eventbus/blob/master/basic.go#L110-L115

We should document this behaviour in the `Subscription` interface.

p2p/protocol/identify/peer_record.go
raulk: Yup, also a good practice to state the consequence (dropping the update) in the error, e.g. `failed to create signed peer record; dropping update: %w`.

p2p/protocol/identify/peer_record.go
Stebalien: https://github.com/libp2p/go-libp2p/pull/747/files/416a878390ed172bf5c6285b1a47740c4777d4c1#diff-98be6dba669aa4123b7ef190fe7113e8R147

We can continue even if we fail to construct it.

p2p/protocol/identify/id.go
Stebalien: Non-blocking: It also seems odd that we're constructing this inside the identify service (a specific protocol for exchanging peer information). I'd expect this to live at the top level inside the host (or possibly live inside the peerstore?).

Bug, given that it does live here and it's _private_, it shouldn't be hard to extract it (especially when we eventually get a dependency injection system.

p2p/protocol/identify/peer_record.go
Stebalien: Ah, that makes more sense. Then we should probably remove this.

p2p/protocol/identify/id_delta.go
```diff
@@ -57,7 +57,7 @@ func (ids *IDService) fireProtocolDelta(evt event.EvtLocalProtocolsUpdated) {
 		}
 		log.Debugf("%s sent delta update to %s: %s", IDDelta, c.RemotePeer(), c.RemoteMultiaddr())
 	}
-	ids.broadcast(IDDelta, deltaWriter)
+	ids.broadcast([]protocol.ID{IDDelta}, deltaWriter)
```
aarshkshah1992: @yusefnapora Why no legacy Delta & a new Delta protocol here ?

yusefnapora:
@aarshkshah1992 @Stebalien @raulk 

I made some changes on Friday that address most of the review comments, but I wanted to ask for advice on something.

First, here's what changed:

- merged with master & updated dependencies
- updated gogofast & regenerated protos (thanks @Stebalien)
- made  the `peerRecordManager` stateless. It's now just  a mapper from `EvtLocalAddressesUpdated` to `EvtLocalPeerRecordUpdated`
- moved the responsibility for tracking the latest peer record to the `IDService`, which caches the latest record received over the  bus.
- moved the `peerRecordManager` from the `IDService`  to the `BasicHost`, since I agree with @Stebalien that the `IDService` is kind of a weird place for  it.

One complication is that moving the `peerRecordManager`  to `BasicHost` means that the `BlankHost` no longer gets peer record updates. However, since `BlankHost` doesn't generate `EvtLocalPeerRecordUpdated` events, it's kind of a non-issue, since the `peerRecordManager` won't have anything to do on a `BlankHost` anyway. 

I did have to change the `IDService` tests though, since previously the initial `EvtLocalPeerRecordUpdated` event would be sent when the `IDService` initialized, even for `BlankHost`. So I added a test helper to create a peer record and put it on the bus in the `IDService` tests.

The thing that bothers me is that the `IDService` can't reliably know whether the `Host` supports peer records until after it receives the first peer record over the event bus. Before, we could at least tell whether creating the `peerRecordManager` succeeded, although to be fair, we weren't handling that case well.

So far, I'm just having the `IDService` wait to register the new peer-record-capable protocol handlers until after we receive the first peer record. Until then, we listen for `EvtLocalAddressesUpdated` events and trigger the old identify/push protocol based on that event. Once we receive the first `EvtLocalPeerRecordUpdated`, we add the new protocol handlers and ignore the `EvtLocalAddressesUpdated` events, so we trigger identify/push on peer record change events instead.

This is awkward and kind of racy - the first peer record event is going to be immediately preceded on the bus by an `EvtLocalAddressesUpdated`, so we're probably going to end up sending two push updates one after the other. Also, adding the new protocols causes an identify/delta event, so this approach is overly chatty.

It seems better for the `IDService` to "know" upfront whether the Host supports peer records. We could either add an option to the `IDService` constructor, or maybe have a new interface that `BasicHost` and friends could conform to, e.g.

```go
// PeerRecordCapableHost is a pretty awkward name... ideas?
type PeerRecordCapableHost interface {
  GetPeerRecord() *record.Envelope
}
```

That would make the peer record accessible via type casting to everyone with a reference to the `Host`, without requiring an event bus subscription. We'd also need to move the tracking of the latest peer record into the `BasicHost` instead of the `IDService`, but that seems like a reasonable place to put it.

One minor thing: with the `peerRecordManager` living on the `BasicHost`, maybe it makes more sense to just get rid of the `peerRecordManager` and have the `BasicHost` emit both `EvtLocalAddressesUpdated` and `EvtLocalPeerRecordUpdated` events at the same time?

p2p/protocol/identify/id.go
```diff
@@ -116,55 +129,92 @@ func NewIDService(ctx context.Context, h host.Host, opts ...Option) *IDService {
...
+		log.Warnf("identify service not subscribed to local peer record changes; err: %s", err)
 	} else {
-		go s.handleEvents()
+		go s.handleEvents(s.subscriptions.localPeerRecordUpdated, s.handlePeerRecordUpdated)
+	}
+	s.subscriptions.localAddrsUpdated, err = h.EventBus().Subscribe(&event.EvtLocalAddressesUpdated{}, eventbus.BufSize(128))
+	if err != nil {
+		log.Warnf("identify service not subscribed to address changes. err: %s", err)
+	} else {
+		go s.handleEvents(s.subscriptions.localAddrsUpdated, s.handleLocalAddrsUpdated)
```
aarshkshah1992: @yusefnapora Wont this cause us to send a `Push` twice since we also do a Push when we get a `EvtLocalPeerRecordUpdated ` which in turn is also derived from `EvtLocalAddressesUpdated `?

aarshkshah1992: Nevermind. This has already been raised in https://github.com/libp2p/go-libp2p/pull/747#issuecomment-596537472

p2p/host/basic/basic_host.go
Stebalien: The tricky part is that we have other event sources:

* AutoNAT knows about our external address (although, I don't think we properly feed this through to the host yet).
* We also have _observed_ addresses. If you take a look inside the identify service in go-libp2p, we keep track of the addresses that our peers observe us using and, when enough peers observe the same external address, we advertise it.

aarshkshah1992:
@yusefnapora 

@raulk suggested that this is a  general problem of subscribers wanting to know if there are any emitters for an event which makes a lot of sense. Rather than solving this using the `PeerRecordCapableHost ` interface(which would solve this problem ONLY once), we can solve the broader problem doing something like:

https://github.com/libp2p/go-eventbus/issues/29

p2p/protocol/identify/id.go
aarshkshah1992: The peerRecordManager has now been moved to the Host.

p2p/host/basic/basic_host.go
aarshkshah1992: We've changed the timeout.

aarshkshah1992:
ping @yusefnapora @raulk .

p2p/host/basic/basic_host.go
```diff
@@ -343,15 +360,65 @@ func makeUpdatedAddrEvent(prev, current []ma.Multiaddr) *event.EvtLocalAddresses
...
+	addrs := make([]multiaddr.Multiaddr, 0, len(current))
+	for _, a := range current {
+		if a == nil {
+			continue
+		}
+		addrs = append(addrs, a)
+	}
+
+	rec := peer.NewPeerRecord()
+	rec.PeerID = h.network.LocalPeer()
```
raulk: host.ID()

p2p/host/basic/basic_host_test.go
```diff
@@ -91,16 +96,35 @@ func TestProtocolHandlerEvents(t *testing.T) {
...
+		for {
+			select {
+			case evt := <-sub.Out():
+				next := evt.(event.EvtLocalProtocolsUpdated)
+				if isIdentify(next) {
+					continue
+				}
+				return next
+			case <-time.After(5 * time.Second):
+				t.Fatal("event not received in 5 seconds")
```
raulk: Fatal must be called from the main goroutine: https://golang.org/pkg/testing/#T

p2p/host/basic/basic_host_test.go
```diff
@@ -605,6 +679,14 @@ func waitForAddrChangeEvent(ctx context.Context, sub event.Subscription, t *test
 	}
 }
 
+type sortedMultiaddrs []ma.Multiaddr
+
+func (sma sortedMultiaddrs) Len() int      { return len(sma) }
+func (sma sortedMultiaddrs) Swap(i, j int) { sma[i], sma[j] = sma[j], sma[i] }
+func (sma sortedMultiaddrs) Less(i, j int) bool {
+	return bytes.Compare(sma[i].Bytes(), sma[j].Bytes()) == 1
```
raulk: Shouldn’t this be == -1?

p2p/host/basic/basic_host.go
```diff
@@ -385,6 +360,80 @@ func makeUpdatedAddrEvent(prev, current []ma.Multiaddr) *event.EvtLocalAddresses
...
+	rec.Addrs = addrs
+	return record.Seal(rec, h.signKey)
+}
+
+func (h *BasicHost) background(p goprocess.Process) {
+	var lastAddrs []ma.Multiaddr
+
+	emitAddrChange := func(currentAddrs []ma.Multiaddr) {
+		changeEvt := makeUpdatedAddrEvent(lastAddrs, currentAddrs)
+		if changeEvt != nil {
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

p2p/host/basic/basic_host_test.go
```diff
@@ -3,14 +3,16 @@ package basichost
...
-	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
 	"io"
 	"reflect"
 	"sort"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/libp2p/go-libp2p-core/peerstore"
+	"github.com/libp2p/go-libp2p-core/record"
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

p2p/host/basic/basic_host_test.go
```diff
@@ -19,10 +21,12 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/protocol"
 	"github.com/libp2p/go-libp2p-core/test"
+	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

go.mod
```diff
@@ -5,21 +5,21 @@ require (
 	github.com/ipfs/go-cid v0.0.5
 	github.com/ipfs/go-detect-race v0.0.1
 	github.com/ipfs/go-ipfs-util v0.0.1
-	github.com/ipfs/go-log v1.0.2
+	github.com/ipfs/go-log v1.0.3
 	github.com/jbenet/go-cienv v0.1.0
 	github.com/jbenet/goprocess v0.1.4
 	github.com/libp2p/go-conn-security-multistream v0.1.0
-	github.com/libp2p/go-eventbus v0.1.0
+	github.com/libp2p/go-eventbus v0.1.1-0.20200326052355-c30c9409b9a4
```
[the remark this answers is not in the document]
aarshkshah1992: No, we do not. This was the remanent of a previous eventbus change I made that we do not need anymore. 

This has been fixed.

p2p/host/basic/basic_host.go
aarshkshah1992: Done.

p2p/host/basic/basic_host_test.go
aarshkshah1992: This code is being executed on the main goroutine, not on a separate one.

p2p/host/basic/basic_host_test.go
aarshkshah1992: Completely got rid of this code as we don't really need it.

p2p/host/basic/basic_host.go
```diff
@@ -385,6 +360,80 @@ func makeUpdatedAddrEvent(prev, current []ma.Multiaddr) *event.EvtLocalAddresses
...
+			}
+		}
+	}
+
+	// immediately emit the first address change event if we have a listen address
+	if h.Addrs() != nil {
+		emitAddrChange(h.Addrs())
+	}
+	// init lastAddrs
+	lastAddrs = h.Addrs()
```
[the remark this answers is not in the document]
aarshkshah1992: @raulk Even though this code is executed in a single dedicated go-routine, have refactored this to make it easier to reason about.

p2p/host/basic/basic_host.go
```diff
@@ -385,6 +360,80 @@ func makeUpdatedAddrEvent(prev, current []ma.Multiaddr) *event.EvtLocalAddresses
...
+		case <-ticker.C:
+		case <-h.addrChangeChan:
+		case <-p.Closing():
+			return
+		}
+
+		//  emit an EvtLocalAddressesUpdatedEvent event if our listen addresses have changed.
+		emitAddrChange(h.Addrs())
+		// update last seen addrs
+		lastAddrs = h.Addrs()
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

aarshkshah1992:
@Stebalien Please can I get a review of this so we can merge it ?

p2p/host/basic/basic_host.go
```diff
@@ -142,10 +148,21 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 	if h.emitters.evtLocalProtocolsUpdated, err = h.eventbus.Emitter(&event.EvtLocalProtocolsUpdated{}); err != nil {
 		return nil, err
 	}
-	if h.emitters.evtLocalAddrsUpdated, err = h.eventbus.Emitter(&event.EvtLocalAddressesUpdated{}); err != nil {
+	if h.emitters.evtLocalAddrsUpdated, err = h.eventbus.Emitter(&event.EvtLocalAddressesUpdated{}, eventbus.Stateful); err != nil {
```
Stebalien, line 151: I'm not sure if this should be stateful. Are we doing this to get around the fact that we don't have some form of `GetPeerRecord()` method as described in https://github.com/libp2p/go-libp2p/pull/747#issuecomment-596537472? If so, we should just add that method.

p2p/host/basic/basic_host.go
```diff
@@ -343,37 +360,77 @@ func makeUpdatedAddrEvent(prev, current []ma.Multiaddr) *event.EvtLocalAddresses
 	return &evt
 }
 
+func (h *BasicHost) makeSignedPeerRecord(evt *event.EvtLocalAddressesUpdated) (*record.Envelope, error) {
+	current := make([]multiaddr.Multiaddr, len(evt.Current))
```
Stebalien, line 364:
```suggestion
	current := make([]multiaddr.Multiaddr, 0, len(evt.Current))
```

p2p/host/basic/basic_host.go
```diff
@@ -343,37 +360,77 @@ func makeUpdatedAddrEvent(prev, current []ma.Multiaddr) *event.EvtLocalAddresses
...
+		current = append(current, a.Address)
+	}
+
+	addrs := make([]multiaddr.Multiaddr, 0, len(current))
+	for _, a := range current {
+		if a == nil {
+			continue
+		}
+		addrs = append(addrs, a)
+	}
```
Stebalien, line 375: Why are we copying twice?

p2p/host/basic/basic_host.go
```diff
@@ -343,37 +360,77 @@ func makeUpdatedAddrEvent(prev, current []ma.Multiaddr) *event.EvtLocalAddresses
...
+
+	rec := peer.NewPeerRecord()
+	rec.PeerID = h.ID()
+	rec.Addrs = addrs
```
Stebalien, line 379: This is a very unusual API. @yusefnapora what was the motivation for it? Why not `NewPeerRecord(id, addrs)`? Or just `&PeerRecord{ID: ..., Addrs: ..., Seq: TimestampSeq()}`?

p2p/host/basic/basic_host.go
aarshkshah1992: @Stebalien We have a `GetPeerRecord(p peer.ID)` on the peerstore now.

However, calling that when you get a `EvtLocalAddressChanged` event could be racy as the address you see on the event could be different fro the signed record you get from the peerstore.

This avoids that race.

p2p/host/basic/basic_host.go
aarshkshah1992: Done.

p2p/host/basic/basic_host.go
aarshkshah1992: Have fixed this.

p2p/host/basic/basic_host.go
aarshkshah1992: @Stebalien There is a `PeerRecordFromAddrInfo(peer.AddrInfi)` API which we weren't using before. Have fixed this.

aarshkshah1992:
@Stebalien This now has the master changes and all the review comments on this PR so far have been addressed. Please can you do a last thorough review when you get the time so we can ensure we haven't missed anything.

p2p/protocol/identify/id.go
```diff
@@ -344,18 +394,35 @@ func (ids *IDService) populateMessage(mes *pb.Identify, c network.Conn) {
...
+			} else {
+				mes.SignedPeerRecord = recBytes
+				log.Debugf("%s sent peer record to %s", c.LocalPeer(), c.RemotePeer())
+			}
+		}
+	} else {
+		// set listen addrs, get our latest addrs from Host.
+		laddrs := ids.Host.Addrs()
+		// Note: LocalMultiaddr is sometimes 0.0.0.0
+		viaLoopback := manet.IsIPLoopback(c.LocalMultiaddr()) || manet.IsIPLoopback(c.RemoteMultiaddr())
```
aarshkshah1992, line 417: @Stebalien 

Hmm.. we probably want to imitate this logic for the signed addresses ?

willscott: it's probably also worth filtering `manet.IsIPUnspecified()` records in both cases.

aarshkshah1992:
ping @Stebalien .

aarshkshah1992:
> I'm not sure if this should be stateful. Are we doing this to get around the fact that we don't have some form of GetPeerRecord() method as described in #747 (comment)? If so, we should just add that method.

Based on the new design we discussed, I think we can add /use this method because we plan to ignore the contents on the event for Identify now.

However, I'd like to merge this as is and do that change in the next PR if that's okay with you. Please can you give this one last review ?

p2p/protocol/identify/pb/identify.proto
```diff
@@ -36,4 +36,11 @@ message Identify {
 
   // a delta update is incompatible with everything else. If this field is included, none of the others can appear.
   optional Delta delta = 7;
+
+  // signedPeerRecord contains a serialized SignedEnvelope containing a PeerRecord,
+  // signed by the sending node. It contains the same addresses as the listenAddrs field, but
+  // in a form that lets us share authenticated addrs with other peers.
+  // see github.com/libp2p/go-libp2p-core/record/pb/envelope.proto and
+  // github.com/libp2p/go-libp2p-core/peer/pb/peer_record.proto for message definitions.
+  optional bytes signedPeerRecord = 8;
```
Stebalien, line 45: Can we make this repeated? That way, we can have one record per _scope_ (and we can add scope info to the signed peer record itself).

Stebalien: Note: we can also do this in a new protocol version.

p2p/protocol/identify/id.go
```diff
@@ -85,12 +89,18 @@ type IDService struct {
...
 	// our own observed addresses.
 	// TODO: instead of expiring, remove these when we disconnect
 	observedAddrs *ObservedAddrSet
 
-	subscription event.Subscription
-	emitters     struct {
+	subscriptions struct {
+		localProtocolsUpdated event.Subscription
+		localAddrsUpdated     event.Subscription
+	}
```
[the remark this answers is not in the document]
aarshkshah1992: I agree. Have made the change.

p2p/protocol/identify/id.go
```diff
@@ -226,25 +263,37 @@ func (ids *IDService) IdentifyConn(c network.Conn) {
...
 
-	s.SetProtocol(ID)
+	ids.peerrecMu.RLock()
+	havePeerRec := ids.peerrec != nil
+	ids.peerrecMu.RUnlock()
+
+	protocolIDs := []string{ID, LegacyID}
+	if !havePeerRec {
+		protocolIDs = []string{LegacyID}
+	}
```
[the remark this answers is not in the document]
aarshkshah1992: This is a great point. I will first use the signed peer records protocol and if you don't support it, I will use the unsigned one. For the former, I will try to extract a signed records from your response, for the latter, I will use the unsigned addresses.

When you try to speak to me, you will discover that I speak the signed records protocol if you do. If you don't, I also speak the old protocol and can send you my unsigned addresses.

Have made the change.

p2p/protocol/identify/id.go
aarshkshah1992: Sure. I'll do this in the next PR once we have a `GetRecord` method.

aarshkshah1992: https://github.com/libp2p/go-libp2p/issues/897

p2p/protocol/identify/id.go
aarshkshah1992: @Stebalien Could you point me to an issue/discussion/PR to better understand the idea and scope of the address scoping work ?

p2p/protocol/identify/id.go
aarshkshah1992: I agree. This can ONLY happen if we get an invalid record. Have changed this to Debug.

p2p/protocol/identify/pb/identify.proto
aarshkshah1992: @Stebalien 

If we do this in a new protocol version, will we make a "breaking" protocol change for it or add a new field for the multiple scoped records ?

aarshkshah1992:
@Stebalien

 I've addressed your comments. Just have some questions to better understand the address scoping work and the protocol changes we will need though they probably don't block this PR.


> My one concern is that we might want to change the protocol slightly when we add address scopes. I'm wondering if we should put this protocol behind a simple ExperimentSignedIdentify global variable flag so we can merge this and start testing it without accidentally deploying it on the network before we're ready.

I'm not sure I understand what you mean. Do we want to block on the address scoping work before doing a full fledged release ?

p2p/protocol/identify/id.go
willscott: @aarshkshah1992 I would imagine the intention would be to split record/addresses along the same policy distinction as in https://github.com/libp2p/go-libp2p-kad-dht/blob/master/dht_filters.go - a public scope for addresses the peer thinks might be valid for remote peers and is willing to drop on the public DHT, and separately the set of addresses valid for LAN peers.

p2p/protocol/identify/pb/identify.proto
Stebalien: I'd make a breaking protocol change.

Stebalien:
> I'm not sure I understand what you mean. Do we want to block on the address scoping work before doing a full fledged release ?

Eh, not really. I guess I just wanted a way to play around with this first before we made it the default.

But honestly, it's fine. If we find we don't like something, we can always introduce a new protocol version.

p2p/protocol/identify/id.go
```diff
@@ -30,7 +30,11 @@ import (
 var log = logging.Logger("net/identify")
 
 // ID is the protocol.ID of the Identify Service.
-const ID = "/ipfs/id/1.0.0"
+const ID = "/p2p/id/1.1.0"
+
+// LegacyID is the protocol.ID of version 1.0.0 of the identify
+// service, which does not support signed peer records.
+const LegacyID = "/ipfs/id/1.0.0"
```
Stebalien, line 37: Naming:

```
ID1_1_0
ID1_0_0
```
?

Stebalien: Because we're going to add more in the future.

p2p/protocol/identify/id.go
```diff
@@ -85,12 +89,14 @@ type IDService struct {
 
 	addrMu sync.Mutex
 
+	peerrec   *record.Envelope
+	peerrecMu sync.RWMutex
```
Stebalien, line 93: Nit: mutex-hat (lock goes above the record)

p2p/protocol/identify/id.go
Stebalien: https://github.com/libp2p/go-libp2p/issues/793

Stebalien:
Thank you for your patience! LGTM!

p2p/protocol/identify/id.go
aarshkshah1992: @Stebalien We will anyways get rid of this once we have the `GetRecord` function in the next PR.

raulk:
This does not compile, and master is broken:

```
⟩ go build ./...
go: downloading github.com/libp2p/go-libp2p-core v0.5.2
go: downloading github.com/miekg/dns v1.1.28
# github.com/libp2p/go-libp2p/p2p/protocol/identify
p2p/protocol/identify/id.go:207:14: cannot use &rec (type **record.Envelope) as type *record.Envelope in assignment
```

Issue #744, opened 2019-11-20
Recursive dnsaddr resolution.


Stebalien:
1. Resolution is not recursive. We need to repeatedly call `Resolve` until nothing happens (or we stop recursing).
2. Resolving from within the "routed host" is error-prone. That means we'll _only_ resolve addresses when the user calls `Connect(ctx, peerinfo)`. We should, instead, resolve these addresses on-demand within the swarm.

PR #764, opened 2019-12-16
feat(host): recursively resolve addresses

Stebalien:
We currently only resolve /dnsaddr addresses once. This patch resolves up to 32 dnsaddr addresses per dial.

fixes https://github.com/libp2p/go-libp2p/issues/744

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -31,6 +31,10 @@ import (
 	msmux "github.com/multiformats/go-multistream"
 )
 
+// The maximum number of address resolution steps we'll perform for a single
+// peer (for all addresses).
+const maxAddressResolution = 32
+
 var log = logging.Logger("basichost")
 
 var (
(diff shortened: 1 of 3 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -522,28 +526,59 @@ func (h *BasicHost) resolveAddrs(ctx context.Context, pi peer.AddrInfo) ([]ma.Mu
...
+		// We've resolved too many addresses. We can keep all the fully
+		// resolved addresses but we'll need to skip the rest.
+		if resolveSteps >= maxAddressResolution {
+			log.Warningf(
+				"peer %s asked us to resolve too many addresses: %s/%s",
+				pi.ID,
+				resolveSteps,
+				maxAddressResolution,
+			)
+			continue
```
vyzo: shouldn't this be a break?

Stebalien: I want to finish going through the addresses, adding any already resolved addresses. I just don't want to resolve any more.

vyzo: ok, fair enough; just concerned about actually terminating the loop.

Stebalien: There's a test case for that.

(being able to say that is really a nice feeling)

PR #770, opened 2020-01-14
identify: emit events; filter out loopback addresses

raulk:
_**Merge strategy: squash and merge.**_

With this PR, go-libp2p:

1. starts emitting identification complete/failed events when identify runs on a peer.
2. filters loopback addresses from identify.

I'm not 100% sold on the latter change, and I'm seeking validation/input. See original discussion on #657; I only merged that patch to stabilize, but now that it's percolating up to master, it's worthy of reconsideration.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -23,6 +23,7 @@ import (
 	logging "github.com/ipfs/go-log"
 
 	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr-net"
 	msmux "github.com/multiformats/go-multistream"
 )
 
@@ -85,7 +86,9 @@ type IDService struct {
 
 	subscription event.Subscription
 	emitters     struct {
-		evtPeerProtocolsUpdated event.Emitter
+		evtPeerProtocolsUpdated        event.Emitter
+		evtPeerIdentificationCompleted event.Emitter
+		evtPeerIdentificationFailed    event.Emitter
 	}
 }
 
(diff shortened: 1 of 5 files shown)


p2p/protocol/identify/id.go
```diff
@@ -193,7 +216,7 @@ func (ids *IDService) IdentifyConn(c network.Conn) {
 	s.SetProtocol(ID)
 
 	// ok give the response to our handler.
-	if err := msmux.SelectProtoOrFail(ID, s); err != nil {
+	if err = msmux.SelectProtoOrFail(ID, s); err != nil {
```
Stebalien: hm. I'm usually a `if err :=` or break it into two lines kind of guy. Side-effects from an if statement make me unhappy.

p2p/protocol/identify/id.go
```diff
@@ -310,9 +333,14 @@ func (ids *IDService) populateMessage(mes *pb.Identify, c network.Conn) {
...
-	for i, addr := range laddrs {
-		mes.ListenAddrs[i] = addr.Bytes()
+	// Note: LocalMultiaddr is sometimes 0.0.0.0
+	viaLoopback := manet.IsIPLoopback(c.LocalMultiaddr()) || manet.IsIPLoopback(c.RemoteMultiaddr())
+	mes.ListenAddrs = make([][]byte, 0, len(laddrs))
+	for _, addr := range laddrs {
+		if !viaLoopback && manet.IsIPLoopback(addr) {
+			continue
+		}
+		mes.ListenAddrs = append(mes.ListenAddrs, addr.Bytes())
```
Stebalien: This is probably for the best. We need a better localhost discovery mechanism, if any.

Stebalien:
Squashed.

@raulk I'm going to merge this so we can start moving off stabilize. I've filed https://github.com/libp2p/go-libp2p/issues/799 about reconsidering the localhost filtering (I assume that signed peer routing records may also affect this).

PR #774, opened 2020-01-22
BasicHost changes for introspection

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/775

Allows the BasicHost to be "introspectable".


p2p/host/basic/basic_host.go
```diff
@@ -191,12 +194,54 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 		h.pings = ping.NewPingService(h)
 	}
 
+	if h.introspector == nil {
+		h.introspector = introspection.NewDefaultIntrospector()
+	}
```
aarshkshah1992: This duplicates the code in config.NewNode() [here](https://github.com/libp2p/go-libp2p/blob/b5a2ce94353a163de8f20b8e58767f711c722af3/config/config.go#L124)

We can safely assume here that an introspector exists. However, do we want to allow users to be able to disable introspection ? That decision will affect this code flow.

p2p/host/basic/basic_host.go
```diff
@@ -191,12 +194,54 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
...
 	net.SetStreamHandler(h.newStreamHandler)
 
+	// register runtime provider
+	if err := h.introspector.RegisterProviders(&coreit.ProvidersTree{Runtime: h.runtimeDataProvider()}); err != nil {
+		log.Errorf("failed to register a runtime provider, err=%s", err)
+	}
+
+	// start introspection server
+	// TODO What happens if address is not configured or not available  ?
+	// TODO How do we configure a "default address"
```
aarshkshah1992: We need to cleanly manage the address on which the introspection server will run. It will need some thought.

p2p/host/basic/basic_host.go
```diff
@@ -191,12 +194,54 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
...
+		}
+
+		return h.Network().Close()
+	})
+
 	return h, nil
 }
 
+func (h *BasicHost) runtimeDataProvider() *coreit.RuntimeProviders {
+	// TODO What is the version here ?
```
aarshkshah1992: What does version mean ?

p2p/host/basic/basic_host.go
```diff
@@ -783,6 +828,13 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
 	return dedupAddrs(finalAddrs)
 }
 
+// TODO Do this for routed & relay hosts as well
```
aarshkshah1992: Need to give this a thought.

p2p/host/basic/basic_host.go
```diff
@@ -89,6 +93,9 @@ type BasicHost struct {
 	emitters  struct {
 		evtLocalProtocolsUpdated event.Emitter
 	}
+
+	// introspector is the host introspector used to register subsystem metric providers & fetch the current system state when asked
```
aarshkshah1992: This comment can be removed.

aarshkshah1992:
### **TODO**
- [ ] Release go-libp2p-core & configure the version here
- [ ] Address the TODOs in code

config/config.go
```diff
@@ -77,6 +79,9 @@ type Config struct {
 
 	EnableAutoRelay bool
 	StaticRelays    []peer.AddrInfo
+
+	Introspector            coreit.Introspector
+	IntrospectionServerAddr string
```
raulk: I don't think we should allow the user to pass in an address. For UX, I would recommend the following:

* We add an `ListenAddress() string` method to the `Introspector` interface, so we can query it later.
* The `Introspector` _implementation_ has a `DefaultAddress`, and may use fallback logic if the default listen address is taken. For example, the default address could be: `ws://localhost:8096`, and if the port is occupied, we can fall back to trying 8096 + [1,100], trying one by one until we find an available port, or fail entirely. See https://github.com/libp2p/go-libp2p-daemon/blob/b95e77dbfcd186ccf817f51e95f73f9fd5982600/p2pd/main.go#L30-L60 for an example of this logic.
* We expose the final listen address via the `ListenAddress() string` method.
* Pushing this concern to the implementation is better, because alternate implementations may use different protocols (e.g. Unix sockets) and heuristics.

For a user that wants to set a concrete address, they would construct the concrete implementation via its constructor (which would take a configuration object, that includes the listen addr), and inject that via the `Introspector` option.

config/config.go
```diff
@@ -7,13 +7,15 @@ import (
 	"github.com/libp2p/go-libp2p-core/connmgr"
 	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/host"
+	coreit "github.com/libp2p/go-libp2p-core/introspection"
```
raulk: Let's shorten this package name to `introspect`. And you can remove the `coreit` alias.

options.go
```diff
@@ -5,6 +5,7 @@ package libp2p
 
 import (
 	"fmt"
+	coreit "github.com/libp2p/go-libp2p-core/introspection"
```
raulk: Import grouping.

p2p/host/basic/basic_host.go
```diff
@@ -783,6 +828,13 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
 	return dedupAddrs(finalAddrs)
 }
 
+// TODO Do this for routed & relay hosts as well
+// RegisterProvider allows the host to be an IntrospectorRegistry. Please look at the docs for the IntrospectorRegistry
+// interface in go-libp2p-core for more details.
+func (h *BasicHost) RegisterProviders(p *coreit.ProvidersTree) error {
```
raulk: I wouldn't add this on the `BasicHost` directly, it muddies the namespace and the term "Provider" is not contextualised. Let's add another layer of indirection in between.

```go
package host

type IntrospectableHost interface {
    Introspector() inspect.Introspector
}
```

Then a consumer can do:

```go
if ins, ok := h.(host.IntrospectableHost); ok {
    ins.Introspector().FetchCurrentState()
}
```

aarshkshah1992: I see what you mean. In this case, we can remove the `IntrospectorRegistry` interface defined [here](https://github.com/libp2p/go-libp2p-core/blob/d0dd06b3894f427b8438ea21c73bd724220d565b/introspect/introspect.go#L7) as modules don't need to check if a host is a `Registry` or not anymore. They can simply check if it has an `Introspector`.

This would have the side effect of exposing the `FetchCurrentState` to modules whose primary job is to register providers but I think we can live with that.

config/config.go
aarshkshah1992: Done

p2p/host/basic/basic_host.go
aarshkshah1992: Done

options.go
aarshkshah1992: Done

config/config.go
aarshkshah1992: @raulk 

I am not sure I agree. Why shouldn't we allow to user to pass in a server address while creating a libp2p node ?

The address is a property of the introspection WS server which is owned by the Host. 

The `Introspector` is only  responsible for proving a registry for sub-systems & resolving the state when it is asked to. It is not/shouldn't be responsible for serving that state over the wire & hence dosen't need to be bothered with finding a free port which is anyways not a trivial operation & muddles the construction logic of the `Introspector`.

raulk: I think I was unclear. We should _totally_ allow passing in a listen address, but not as a dedicated host option. For two reasons:

1. Creating such granular field-level options muddies the namespace.
2. Listen addresses, along with all potential customisations of the introspection endpoint, are implementation-dependent.

Users wanting to customise the listen addr of an introspection server, should be able to do that through the endpoint's constructor:

```go
ep, err := introspection.Endpoint(&introspection.EndpointOpts{
    ListenAddress: "...",
}
h, err := host.New(host.Introspector(ep))
```

introspect_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"time"
+)
+
+func TestIntrospector(t *testing.T) {
+	msg1 := []byte("1")
+	msg2 := []byte("12")
+	msg3 := []byte("111")
+	msg4 := []byte("0000")
+
+	iaddr := "0.0.0.0:9999"
```
raulk: Tests should not listen on 0.0.0.0 (that's insecure) nor on a fixed port (it might be occupied). Instead, we should listen on 127.0.0.1:0, and allow the listener to allocate a port. `ListenAddrs()` should then return the _actual_ addresses, not the user-supplied ones.

introspect_test.go
```diff
@@ -0,0 +1,190 @@
...
+	})
+	s1, err := h1.NewStream(ctx, h3.ID(), p1)
+	require.NoError(t, err)
+	wg.Add(1)
+	_, err = s1.Write(msg1)
+	require.NoError(t, err)
+	bz1 := make([]byte, len(msg2))
+	wg.Wait()
+	_, err = s1.Read(bz1)
+	require.NoError(t, err)
```
raulk: You can do `require := require.New(t)` at the top, and omit the first `t` parameter from all assertions.

introspect_test.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// host2 -> OPENS Stream 2 -> host1 , writes a message & reads the response
+	p2 := protocol.ID("h2h1")
+	h1.SetStreamHandler(p2, func(s network.Stream) {
+		bz := make([]byte, len(msg3))
```
raulk: Let's use `buf`. It shadows previous usages, and it's idiomatic.

introspect_test.go
```diff
@@ -0,0 +1,190 @@
...
+		wg.Done()
+	})
+
+	s2, err := h2.NewStream(ctx, h1.ID(), p2)
+	require.NoError(t, err)
+	wg.Add(1)
+	_, err = s2.Write(msg3)
+	require.NoError(t, err)
+	bz2 := make([]byte, len(msg4))
+	wg.Wait()
```
raulk: No need for waitgroups. This code is racy, and we can use `io.ReadFull` instead.

introspect_test.go
```diff
@@ -0,0 +1,190 @@
...
+	_, err = s2.Read(bz2)
+	require.NoError(t, err)
+
+	// call introspection server & fetch state
+	u := url.URL{Scheme: "ws", Host: iaddr, Path: "/introspect"}
+
+	// wait till connection is established
+	i := 0
+	var connection *websocket.Conn
+	for {
```
raulk: I think this loop declaration can be made more elegant, by moving the init, test, increment up here.

raulk: Or you can use https://godoc.org/github.com/stretchr/testify/require#Assertions.Eventually.

aarshkshah1992:
@raulk  Thanks for the review. I can make the changes myself if you like. Let me know  & I'll fix the PR.

PR #787, opened 2020-02-05
chore(ci): fix a flaky test

Stebalien:
CI is slow so we should increase our tolerance. This is just a test for our _mock_ network anyways.


Stebalien:
I've tried changing tolerances but this still fails in CI and works locally.

PR #807, opened 2020-02-27
Trigger Autorelay on NAT events

willscott:
Make better use of the event bus between these two.


p2p/host/relay/autorelay.go
```diff
@@ -84,40 +85,43 @@ func (ar *AutoRelay) hostAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
...
-	case <-time.After(autonat.AutoNATBootDelay + BootDelay):
-	case <-ctx.Done():
-		return
-	}
+	subNatPublic, _ := ar.host.EventBus().Subscribe(new(event.EvtLocalRoutabilityPublic))
+	pubChan := subNatPublic.Out()
+	subNatPrivate, _ := ar.host.EventBus().Subscribe(new(event.EvtLocalRoutabilityPrivate))
+	priChan := subNatPrivate.Out()
+	subNatUnknown, _ := ar.host.EventBus().Subscribe(new(event.EvtLocalRoutabilityUnknown))
+	unkChan := subNatUnknown.Out()
```
Stebalien: Defer a close of the subscriptions.

Stebalien:
Is this ready for a review?

willscott:
yep. versions of autonat aren't there, so won't merge until there's an autonat to go.mod and CI against, but logic is ready

config/config.go
```diff
@@ -201,6 +203,13 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 		}
 	}
 
+	if cfg.EnableAutoRelay || cfg.AutoNATOpts != nil {
```
Stebalien: What sets AutoNATOpts?

willscott: IPFS?
Should there be a libp2p translation of the nat options?

config/config.go
```diff
@@ -235,11 +249,32 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
...
+		})}
+		if hop {
+			dialerStore := pstoremem.NewPeerstore()
+			dialerHost, err := cfg.newHost(ctx, dialerStore)
+			if err != nil {
+				h.Close()
+				return nil, err
+			}
+
+			autonatOpts = append(autonatOpts, autonat.EnableService(dialerHost.Network(), false))
```
Stebalien: Will the host get closed when the service is closed?

Stebalien: How do I force AutoNAT server mode?

Stebalien: Can I force AutoNAT client mode?

Stebalien: Maybe we should have an `EnableAutoNAT(mode)` option.

p2p/host/relay/autorelay.go
```diff
@@ -66,58 +65,55 @@ func NewAutoRelay(ctx context.Context, bhost *basic.BasicHost, discover discover
...
+				update = ar.findRelays(ctx)
 			}
-			ar.status = autonat.NATStatusPublic
-			ar.mx.Unlock()
 
-		case autonat.NATStatusPrivate:
-			update := ar.findRelays(ctx)
 			ar.mx.Lock()
-			if update || ar.status != autonat.NATStatusPrivate {
+			if ar.status != evt.Reachability && evt.Reachability != network.ReachabilityUnknown {
```
Stebalien: Nit: having two cases here is confusing. I'd just write:

```go
if update || (ar.status != evt.Reachability && evt.Reachability != network.ReachabilityUnknown) { 
```

PR #810, opened 2020-02-28
EnableAutoRelay should work without ContentRouting if there are StaticRelays defined

aschmahmann:
Given that if static relays are passed in to the `AutoRelay` constructor we should be able to construct an `AutoRelay` even when there are no `ContentRouting` implementations passed in.

@willscott ran into this issue when trying to construct testground integration tests of autorelay. We can hack around it by passing in a `Null ContentRouter` implementation, but that seems incorrect.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -207,19 +207,6 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 			return nil, fmt.Errorf("cannot enable autorelay; relay is not enabled")
 		}
 
-		if router == nil {
-			h.Close()
-			return nil, fmt.Errorf("cannot enable autorelay; no routing for discovery")
-		}
-
-		crouter, ok := router.(routing.ContentRouting)
-		if !ok {
-			h.Close()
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -228,11 +215,28 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
...
 		}
 
-		if hop {
-			// advertise ourselves
-			relay.Advertise(ctx, discovery)
+		if !hop && len(cfg.StaticRelays) > 0 {
+			_ = relay.NewAutoRelay(swrm.Context(), h, nil, router, cfg.StaticRelays)
 		} else {
-			_ = relay.NewAutoRelay(swrm.Context(), h, discovery, router, cfg.StaticRelays)
+			if router == nil {
```
willscott: potentially combined with previous line into `} else if router == nil {`

hsanjuan:
Please update the (already very confusing) docs: https://godoc.org/github.com/libp2p/go-libp2p#EnableAutoRelay p

Issue #801, opened 2020-02-25
Migrate the Swarm notifications to the eventbus


aarshkshah1992:
We already emit events such as `Identify Completed`, `PeerProtocols Updated` etc on the Eventbus.  It will be good if we also migrate Swarm notifications to the host Eventbus. This will enable subsystems to be notified of Network events such as Connections, Disconnections etc etc. without having to depend on the Host/Network. It will also remove the need for subsystems to have dual subscriptions(Swarm & Eventbus) to be notified of all the events in libp2p. 

However, we must ensure backward compatibility by not breaking the existing Swarm notification mechanism.

Some work for this has been done at:
1. https://github.com/libp2p/go-libp2p-core/pull/57
2. https://github.com/libp2p/go-libp2p-swarm/pull/140

However, it needs to be refined and taken to completion.

PR #813, opened 2020-02-28
Emit EvtPeerConnectednessChanged on the eventbus

aarshkshah1992:
@Stebalien 

For https://github.com/libp2p/go-libp2p/issues/801.


config/config.go
```diff
@@ -114,13 +115,19 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
...
+	swrm, err := swarm.NewSwarm(ctx, pid, cfg.Peerstore, cfg.Reporter, eb)
+	if err != nil {
+		return nil, err
+	}
 	if cfg.Filters != nil {
 		swrm.Filters = cfg.Filters
 	}
 
-	h, err := bhost.NewHost(ctx, swrm, &bhost.HostOpts{
+	h, err := bhost.NewHost(ctx, swrm, eb, &bhost.HostOpts{
```
Stebalien: Can we make this a host option (and let the host construct it if it's nil?). Slightly less breaking that way.

aarshkshah1992: @Stebalien  _For my own understanding_

How do breaking changes like these affect us/our users considering we aren't really breaking an interface ?

aarshkshah1992:
@Stebalien Have made the change. Please take a look.

p2p/host/basic/basic_host_notifiee.go
```diff
@@ -0,0 +1,41 @@
...
+type basicHostNotifiee BasicHost
+
+func (b *basicHostNotifiee) basicHost() *BasicHost {
+	return (*BasicHost)(b)
+}
+
+func (b *basicHostNotifiee) Connected(n network.Network, v network.Conn) {
+	// Is this our first connection with the peer ?
+	p := v.RemotePeer()
+	if len(n.ConnsToPeer(p)) == 1 {
```
Stebalien: `n.Connectedness` should be faster.

We also need to take some kind of lock (ideally striped per peer?). Otherwise, we could have the following ordering:

* Thread 1: Receive disconnect event.
* Thread 1: Determine that we are disconnected.
* Thread 2: Receive connect event.
* Thread 2: Determine that we have a connection.
* Thread 2: Fire connect event.
* Thread 1: Fire disconnect event.

This is why we take locks in notification handlers in bitswap, the dht, etc. This is _also_ why I'm excited to finally have this event bus. We can solve this issue once _here_ and downstream callers can know that events will be completely ordered.

aarshkshah1992: @Stebalien I don't think this problem can happen here because both the `BasicHost.Connected` & `BasicHost.Disconnected` callbacks are writing to the same emitter.

1. Swarm takes a lock before calling the callbacks for an event & releases it ONLY after all callbacks have returned.

```
var wg sync.WaitGroup

	s.notifs.RLock()
	wg.Add(len(s.notifs.m))
	for f := range s.notifs.m {
		go func(f network.Notifiee) {
			defer wg.Done()
			notify(f)
		}(f)
	}

	wg.Wait()
	s.notifs.RUnlock()
```

2. For the scenario that you highlight, the `BasicHost.Disconnected` callback will return ONLY after the event has been written to the `EvePeerStateChange` emitter. So, can't we be assured that the subsequent Connect will be written to the same emitter ONLY after the disconnect event has been written & this will be received in order by the subscriber ? 

I might be missing something here. Let me know.

Stebalien: >     Swarm takes a lock before calling the callbacks for an event & releases it ONLY after all callbacks have returned.

That's a read lock.

---

Note: for a single connection, the disconnect event will always come after the connect event. However, for separate connections to the same peer, there are no ordering guarantees.

config/config.go
Stebalien: Technically, users might be manually constructing their hosts (without using the constructor). That used to the the correct way to do this until we introduced the constructor. I don't _think_ anyone is, but I haven't checked thoroughly.

But really, either way is probably fine in this case.

raulk:
I'd like to review this before merging.

p2p/host/basic/basic_host_notifiee.go
aarshkshah1992: I got what you mean. Making the change.

aarshkshah1992: @Stebalien I am not using `n.Connectedness ` as we ONLY want to fire an event if this is our first connection with the peer.

Stebalien: > I am not using n.Connectedness as we ONLY want to fire an event if this is our first connection with the peer.

Ah, I didn't see that. Unfortunately, this is unreliable as we might create two connections before firing the events.

aarshkshah1992:
@Stebalien @raulk Please take a look.

p2p/host/basic/basic_host_notifiee.go
aarshkshah1992: This is a way trickier beast than I'd imagined.

aarshkshah1992:
Hey @Stebalien and @raulk 

This is ripe for a review. Core PR is at https://github.com/libp2p/go-libp2p-core/pull/120.

p2p/host/basic/basic_host_notifiee.go
```diff
@@ -0,0 +1,54 @@
...
+func (b *basicHostNotifiee) Connected(n network.Network, v network.Conn) {
+	p := v.RemotePeer()
+
+	// use the last byte of the peer as the key for the striped lock
+	indexForLk := len(p) - 1
+	lk := &b.stripedConnNotifLocks[p[indexForLk]]
+	lk.Lock()
+	defer lk.Unlock()
+
+	if n.Connectedness(p) == network.Connected {
```
Stebalien: Trying to write some documentation for this... I realized this isn't quite correct. We want:

good


But this code will give us:

bad

---

We used code like this elsewhere because transitioning within the current state wasn't actually an issue. However, if we're going to make this a public interface, we shouldn't do this.

Your original code tried to avoid this by checking "are we transitioning from/to 0" but the asynchronous nature of these events meant that didn't work either. Unfortunately... I think we need to remember the last event emitted for the peer to fix this.

aarshkshah1992:
@Stebalien @raulk Have made the changes. Please take a look when you can.

p2p/host/basic/basic_host_notifiee.go
```diff
@@ -0,0 +1,85 @@
...
+	// what's the last known connectedness for this peer ?
+	b.observedConnectednessLk.RLock()
+	prevConn := b.observedConnectedness[p]
+	b.observedConnectednessLk.RUnlock()
+
+	// If our last observed connectedness for the peer is "Connected" & we now have no connections to the peer,
+	// we should emit an event.
+	if prevConn == network.Connected && n.Connectedness(p) == network.NotConnected {
+		b.observedConnectednessLk.Lock()
+		b.observedConnectedness[p] = network.NotConnected
```
Stebalien: This will leak. When we switch to not connected (the "default state"), we should just forget about this peer.

p2p/host/basic/basic_host_notifiee.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// what's the last known connectedness for this peer ?
+	b.observedConnectednessLk.RLock()
+	prevConn := b.observedConnectedness[p]
+	b.observedConnectednessLk.RUnlock()
+
+	// If our last observed connectedness for the peer is "Connected" & we now have no connections to the peer,
+	// we should emit an event.
+	if prevConn == network.Connected && n.Connectedness(p) == network.NotConnected {
+		b.observedConnectednessLk.Lock()
```
Stebalien: This defeats the point of the striped locks. What if we stored the `observedConnectedness` state inside `stripedConnNotifLocks` itself? That is:

* Rename `stripedConnNotifLocks` to something like `stripedPeerConnectivity`.
* Change the datastructure to `struct { sync.Mutex; m map[peer.ID]network.Connectedness }`.

aarshkshah1992: This is a really cool & useful pattern for shading locks.

p2p/host/basic/basic_host_notifiee.go
```diff
@@ -0,0 +1,71 @@
...
+	// This striped lock ensures that both "Connected" & "Disconnected" notifications synchronize on
+	// "change in connectedness" between two peers & also protects the corresponding `observedConnectedness` state.
+	indexForLk := len(p) - 1
+	stripe := &b.stripedPeerConnectivity[p[indexForLk]]
+	stripe.Lock()
+	defer stripe.Unlock()
+
+	// If our last observed connectedness for the peer is "NotConnected" & we are now connected to the peer,
+	// we should emit an event.
+	if stripe.observedConnectedness[p] == network.NotConnected && n.Connectedness(p) == network.Connected {
```
Stebalien: nit: This _technically_ works because `network.NotConnected == 0`. But I'd prefer to be safer and check `... != network.Connected`.

p2p/host/basic/basic_host_notifiee_test.go
```diff
@@ -0,0 +1,117 @@
...
+	s, err := h1.EventBus().Subscribe(&event.EvtPeerConnectednessChanged{})
+	defer s.Close()
+	require.NoError(t, err)
+
+	// now make simultaneous connect<->disconnect from both sides
+	var wg sync.WaitGroup
+	for i := 0; i < 50; i++ {
+		wg.Add(3)
+		go func() {
+			require.NoError(t, h1.Connect(ctx, peer.AddrInfo{h2.ID(), h2.Addrs()}))
```
Stebalien: It turns out this calls `t.Fatal`. Unfortunately, that can't be called from within a goroutine (causes a data race).

We should either call `t.Error` or feed these errors out of the goroutine and call `require.NoError` on the main thread.

Stebalien: Note: The fact that this race showed up in CI means something else went wrong as well.

p2p/host/basic/basic_host.go
```diff
@@ -194,6 +204,14 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 	net.SetConnHandler(h.newConnHandler)
 	net.SetStreamHandler(h.newStreamHandler)
 
+	// initialize the data structures for peer connectedness notifications
+	for i := 0; i < 256; i++ {
+		h.stripedPeerConnectivity[i].observedConnectedness = make(map[peer.ID]network.Connectedness)
+	}
```
raulk:
```suggestion
	for i := range h.stripedPeerConnectivity {
		h.stripedPeerConnectivity[i].observedConnectedness = make(map[peer.ID]network.Connectedness)
	}
```

p2p/host/basic/basic_host.go
```diff
@@ -87,8 +87,14 @@ type BasicHost struct {
...
 	lastAddrs []ma.Multiaddr
 	emitters  struct {
+		evtPeerStateChange       event.Emitter
 		evtLocalProtocolsUpdated event.Emitter
 	}
+
+	stripedPeerConnectivity [256]struct {
+		sync.Mutex
+		observedConnectedness map[peer.ID]network.Connectedness
+	}
```
raulk: Can we push this state down into the _notifee => eventbus shim_? It doesn't really belong here. IUC, it's only used to enforce non-idempotent firing.

p2p/host/basic/basic_host_notifiee.go
```diff
@@ -0,0 +1,71 @@
...
+	// we should emit an event.
+	if stripe.observedConnectedness[p] != network.Connected && n.Connectedness(p) == network.Connected {
+		// change our observed connectedness for the peer to Connected.
+		stripe.observedConnectedness[p] = network.Connected
+
+		// emit the event on the bus
+		evt := event.EvtPeerConnectednessChanged{p, network.Connected}
+		b.basicHost().emitters.evtPeerStateChange.Emit(evt)
+	}
+}
```
raulk: Don't we want to fire and store iff the current value is _not_ equal to the previous value? If so, we could simplify this comparison, and make it less brittle to actual enum values (what if we introduce new enum values in the future?). This really seems like an atomic compare and swap.

p2p/host/basic/basic_host_notifiee_test.go
```diff
@@ -0,0 +1,117 @@
...
+		evt, ok := e.(event.EvtPeerConnectednessChanged)
+		require.True(t, ok)
+		require.Equal(t, network.Connected, evt.Connectedness)
+		require.Equal(t, h2.ID(), evt.Peer)
+	case <-time.After(1 * time.Second):
+		t.Fatal("did not get notification")
+	}
+
+	// connected -> connected does not result in a notification
+	(*basicHostNotifiee)(h1).Connected(h1.network, h1.network.ConnsToPeer(h2.ID())[0])
```
raulk: Why would this trigger anything that's relevant?

Stebalien:
> The connectedness state map seems like a workaround for not making this change in the swarm. Why don't we do it there?

We'd need to feed the event bus down into the swarm and we _thought_ it was going to be easy here. But you're right, we could save some state by doing it there.

p2p/host/basic/basic_host_notifiee.go
Stebalien: We already know that the new state will need to be connected, in this case. We could simplify this but it'll have a minor cost as we'll have to call `n.Connectedness(p)` (takes a lock) _before_ we check the local value.

aarshkshah1992:
Closing in favour of #840 as we have decided to emit the events from inside the Swarm.

PR #822, opened 2020-03-03
fix(identify): aggregate protocol updates in the background instead of blocking

### LINKED ISSUE #821: Identify delta blocks
1. It doesn't combine multiple events.
2. It blocks if the buffer fills up.

Stebalien:
If we block the event bus, bad things happen.

fixes #821


Stebalien:
Nvm, I misunderstood the issue. We'll need this, but not yet.

Stebalien:
Actually, we should probably do this per-peer.

PR #836, opened 2020-03-10
feat(mock): reliable notifications

Stebalien:
* Export StreamComplement/ConnComplement convenience functions.
* Make the TestNotifications test pass reliably, even when we have a bunch of streams (identify, etc.).
* Make the mock net order disconnect events after connect events.
* Make closing one side of a connection actually close both sides.
* Make it possible to extract a mock stream's complement.
* Fire remote events at the same time as the local events.

I'm hoping this will put an end to spurious test failures due to mocknet issues.
Just kidding, that'll never happen.

### DIFF
--- a/p2p/net/mock/complement.go
+++ b/p2p/net/mock/complement.go
@@ -1,0 +1,17 @@
+package mocknet
+
+import (
+	"github.com/libp2p/go-libp2p-core/network"
+)
+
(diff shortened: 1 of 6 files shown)


Stebalien:
All the bitswap related tests that use this are passing. I'm going to just merge this as it just improves testing code and it's not worth anyone's time to review.

PR #849, opened 2020-03-22
fix: minimal autonat dialer

Stebalien:
Currently, the AutoNAT dialer spins up a full host with identify, ping, etc. this switches to a blankhost.

This also avoids running a relay, listening on transports, etc (regression).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -13,6 +13,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/peerstore"
 	"github.com/libp2p/go-libp2p-core/pnet"
 	"github.com/libp2p/go-libp2p-core/routing"
+	"github.com/libp2p/go-libp2p-core/transport"
 	"github.com/libp2p/go-libp2p-peerstore/pstoremem"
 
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -261,13 +283,35 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
...
+			Peerstore: pstoremem.NewPeerstore(),
+		}
+		dialer, err := autoNatCfg.makeSwarm(ctx)
 		if err != nil {
 			h.Close()
 			return nil, err
 		}
-
+		dialerHost := blankhost.NewBlankHost(dialer)
+		err = autoNatCfg.addTransports(ctx, dialerHost)
```
willscott: addTransports looks like it could plausibly be refactored to take a network directly, in which case we could just pass the dialer directly without ever creating the blank host. would that be more efficient?

Stebalien: Unfortunately, _transports_ can currently accept a `Host` through reflection magic and we can't really switch that to `Network` because some transports might need to register a stream listener (only available on host).

PR #850, opened 2020-03-23
fix multiple issues in the mock tests

Stebalien:
They were OOMing on my machine and I ended up digging through multiple bugs.


p2p/net/mock/mock_test.go
```diff
@@ -598,7 +598,9 @@ func TestFuzzManyPeers(t *testing.T) {
 		peerCount = 100
 	}
 	for i := 0; i < peerCount; i++ {
-		_, err := FullMeshConnected(context.Background(), 2)
+		ctx, cancel := context.WithCancel(context.Background())
+		_, err := FullMeshConnected(ctx, 2)
+		cancel()
```
willscott: the need for this seems to indicate there are still tasks attached to the context by this  process that aren't cleaned up as we might hope. canceling the context solves this problem and is good to be able to do, but we should probably mark the underlying laziness for cleanup as well.

Stebalien: `FullMessConnected` returns a fully connected network. Unfortunately, the "correct" way to close that network is to cancel the context. There's no explicit close function.

Issue #848, opened 2020-03-20
Change CheckForAddressChanges to SignalAddressChange


Stebalien:
And:

* Do the update in the background thread.
* Have `SignalAddressChange` signal the background thread by setting a flag on a size-one channel.

That way, services can just call this function to signal that an address has changed without waiting on each other.

PR #851, opened 2020-03-24
Signal address change

aarshkshah1992:
@Stebalien 

For #848.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -359,39 +359,29 @@ func (h *BasicHost) background(p goprocess.Process) {
 	}
 	h.mx.Unlock()
 
-	//  emit an EvtLocalAddressesUpdatedEvent & a Push Identify if our listen addresses have changed.
-	go func() {
-		for {
-			select {
-			case <-h.addrChangeChan:
-				h.mx.Lock()
-				addrs := h.Addrs()
-				changeEvt := makeUpdatedAddrEvent(h.lastAddrs, addrs)
-				if changeEvt != nil {
-					h.lastAddrs = addrs
-				}
-				h.mx.Unlock()
-
-				if changeEvt != nil {
-					err := h.emitters.evtLocalAddrsUpdated.Emit(*changeEvt)
-					if err != nil {
-						log.Warnf("error emitting event for updated addrs: %s", err)
-					}
-					h.ids.Push()
-				}
-			case <-p.Closing():
-				return
-			}
-		}
-	}()
-
 	for {
 		select {
 		case <-ticker.C:
-			h.SignalAddressChange()
-
+		case <-h.addrChangeChan:
 		case <-p.Closing():
 			return
+		}
+
+		//  emit an EvtLocalAddressesUpdatedEvent & a Push Identify if our listen addresses have changed.
+		h.mx.Lock()
+		addrs := h.Addrs()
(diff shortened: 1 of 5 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -366,10 +359,36 @@ func (h *BasicHost) background(p goprocess.Process) {
 	}
 	h.mx.Unlock()
 
+	//  emit an EvtLocalAddressesUpdatedEvent & a Push Identify if our listen addresses have changed.
+	go func() {
+		for {
+			select {
+			case <-h.addrChangeChan:
```
willscott: can you not incorporate this into the below for() loop on the background goroutine, rather than adding the additional event loop for translating this event?

p2p/host/basic/basic_host.go
```diff
@@ -230,6 +233,7 @@ func New(net network.Network, opts ...interface{}) *BasicHost {
 	}
 
 	h, err := NewHost(context.Background(), net, hostopts)
+	h.Start()
```
willscott: this seems like a functional change. are we confident this is safe?

aarshkshah1992: Hey, thus function is ONLY used in tests.

p2p/host/basic/basic_host.go
aarshkshah1992: Good idea, have made the change.

p2p/host/basic/basic_host.go
willscott: yep. just worth checking where it's currently used to make sure the tests making this downstream aren't either manually starting it as well, or not expecting the changed behavior.

aarshkshah1992: @willscott Good point.

So, outside this `go-libp2p` package, it only seems to be used in IPFS for one test(it passes) and in DHT tests(no problem there).

p2p/host/basic/basic_host.go
```diff
@@ -369,11 +362,27 @@ func (h *BasicHost) background(p goprocess.Process) {
...
 		case <-ticker.C:
-			h.CheckForAddressChanges()
-
+		case <-h.addrChangeChan:
 		case <-p.Closing():
 			return
 		}
+
+		//  emit an EvtLocalAddressesUpdatedEvent & a Push Identify if our listen addresses have changed.
+		h.mx.Lock()
```
willscott: mutex can be removed fully now that it's replaced with a channel.

p2p/host/basic/basic_host.go
Stebalien: Yeah, this should be fine. We should also look into removing it eventually...

p2p/host/basic/basic_host.go
```diff
@@ -360,41 +350,31 @@ func (h *BasicHost) background(p goprocess.Process) {
...
+		if changeEvt != nil {
+			lastAddrs = addrs
+		}
 
-	for _, addr := range a {
-		_, ok := bmap[string(addr.Bytes())]
-		if !ok {
-			return false
+		if changeEvt != nil {
+			err := h.emitters.evtLocalAddrsUpdated.Emit(*changeEvt)
```
Stebalien: nit: we don't even need to store this emitter in the host, we can just create it and close it within this loop.

PR #853, opened 2020-03-24
fix: set the private key when constructing the autonat service

Stebalien:
Otherwise, it the node fails to build.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


Stebalien:
```
--- FAIL: TestLocalhostAddrFiltering (0.06s)

    id_test.go:268: expected one addr, found [/ip4/1.2.3.4/tcp/1234 /ip4/127.0.0.1/tcp/2345]
```

Issue #860, opened 2020-03-30
Fire `EvtPeerProtocolsUpdated` when identify completes.


Stebalien:
Users shouldn't need to listen for the identify event unless they need to know something like the remote peer's user agent. Having to listen on two different events to get protocol updates for peers is annoying.

aarshkshah1992:
@willscott Lmk if I can come up with a PR as I might have some cycles.

aarshkshah1992:
PR at https://github.com/libp2p/go-libp2p/pull/864.

raulk:
I actually disagree. It seems like a pursuit of overfitting for a very particular use case, and trading off correctness and pragmatism for a one-off convenience, along the way.

* `EvtPeerIdentificationComplete` fires when you establish a connection.
* `EvtPeerProtocolsUpdated` fires when, **during a connection**, we receive an identify delta or identify push event announcing a dynamic protocol change.

Why not use multi-type subscription instead? See godocs here: https://godoc.org/github.com/libp2p/go-libp2p-core/event#Bus

Issue #802, opened 2020-02-25
Event driven peerstore


aarshkshah1992:
See https://github.com/libp2p/go-libp2p-swarm/pull/140 for a great writeup on why we want to do this.

Rather than callers listening to Network events/notifications & changing the state of the data(addresses, keys etc.) in the peerstore via it's API, the peerstore itself should listen to events such as `Connected`,`Disconnected`, `Identify Completed` on the eventbus & mutate it's internal data accordingly. 

https://github.com/libp2p/go-libp2p/issues/801 should be finished before this.

burdiyan:
Since https://github.com/libp2p/specs/issues/164 is closed now, I'm posting [my concern](https://github.com/libp2p/specs/issues/164#issuecomment-1342659682) here as well:

> How large is all this peer-related information? If the amount of data per peer is small, and even for hundreds of peers it is still in the order of megabytes, maybe a simpler cleanup approach could suffice? Like cleaning up less recently used peer information once in a while, instead of trying to do it in a more real-time fashion?

So, the question is, is it really necessary to constantly trying to purge the address book, creating a lot of writes and potentially saturating the datastore? If peer data is small enough, and there's not a lot of churn of peers, maybe it's worth keeping the peer data for longer?

PR #864, opened 2020-04-01
Emit Peer Protocol updated when Identify completes

aarshkshah1992:
@Stebalien @willscott 

For https://github.com/libp2p/go-libp2p/issues/860.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -146,6 +146,7 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 		return nil, err
 	}
 
+
 	h.proc = goprocessctx.WithContextAndTeardown(ctx, func() error {
 		if h.natmgr != nil {
 			h.natmgr.Close()
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -7,7 +7,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/libp2p/go-eventbus"
 	ic "github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/event"
 	"github.com/libp2p/go-libp2p-core/helpers"
@@ -17,6 +16,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/peerstore"
 	"github.com/libp2p/go-libp2p-core/protocol"
 
+	"github.com/libp2p/go-eventbus"
 	pb "github.com/libp2p/go-libp2p/p2p/protocol/identify/pb"
 
 	ggio "github.com/gogo/protobuf/io"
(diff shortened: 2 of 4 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -146,6 +146,7 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 		return nil, err
 	}
 
+
```
Stebalien, line 149: extra line?

p2p/protocol/identify/id.go
```diff
@@ -197,11 +197,29 @@ func (ids *IDService) IdentifyConn(c network.Conn) {
...
+				log.Errorf("failed to fetch protocols from the peerstore for peer=%s, err=%s", rp.String(), err)
+				return
+			}
+
+			// TODO This can race with the ID Delta if the peer immediately changes it's protocols after connecting
+			// The long term solution is to fix https://github.com/libp2p/go-libp2p/issues/823
+			// But, in the long term, we are all dead.
+			ids.emitters.evtPeerProtocolsUpdated.Emit(event.EvtPeerProtocolsUpdated{
+				Peer:  c.RemotePeer(),
+				Added: protocol.ConvertFromStrings(peerProtocols),
```
Stebalien, line 218: Hm. I'm not sure if this is accurate, especially if we have multiple concurrent connections and/or if we're doing an identify push. In that case, we'll be removing/adding protocols.

Really, I wonder if tracking the added/removed protocols is more trouble than it's worth. Accurately tracking this with no race conditions seems horribly difficult unless we synchronously emit updates from the peerstore itself.

Thoughts @raulk? My vote would be to simply remove Added/Removed and expect subscribers to just check the peerstore for the latest values.

aarshkshah1992: +1 for subscribers checking the peerstore.

jacobheun: fwiw, in JS we emit all protocol and address change events from the peerstore itself to keep a single source of truth for peer records. We also just emit the full, current list of protocols/multiaddrs of the peer to keep the logic simple.

aarshkshah1992: @jacobheun 's comment is also a +1 for getting in https://github.com/libp2p/go-libp2p/issues/802.

raulk: Single source of truth is what we're shooting for. Unfortunately, the peerstore was poorly architected/built/integrated/used from the outset in go-libp2p, and we have to deal with that heritage until v2.

raulk: > Really, I wonder if tracking the added/removed protocols is more trouble than it's worth.

It is, when this event is used improperly. See https://github.com/libp2p/go-libp2p/issues/860#issuecomment-607203490.

raulk: To summarise:

- The current source of truth for "peer protocols updated" event is **identify delta**.
- Via identify delta, the peer reports _exactly_ which protocols were added and removed. Those fields are mapped onto the event.
- Whether we emit it from identify delta, or in a future v2 peerstore, the behaviour will be the same.
- This problem arises because this PR emits the event from an improper place.
- This event is an application of CDC (Change Data Capture): https://en.wikipedia.org/wiki/Change_data_capture.

aarshkshah1992:
@Stebalien I think @raulk has a point here: 

https://github.com/libp2p/go-libp2p/issues/860#issuecomment-607203490

In light of that, I'm inclined to close this PR unless you have any concerns.

MarcoPolo:

PR #867, opened 2020-04-02
disable write coalescing

willscott:
Bump go-libp2p-yamux and go-libp2p-mplex to point releases that have disabled write coalescing until more testing of that change occurs.


bonedaddy:
out of curiosity what's the reasoning for this?

Stebalien:
This code was very liberal in its use of timers. In go 1.14.1, this triggered https://github.com/golang/go/issues/38119. However, more than that, the frequent timer firing showed up heavily in CPU profiles.

We're going to spend some time profiling our approach.

PR #869, opened 2020-04-03
chore: update deps again

Stebalien:
One final time.


willscott:
```
# github.com/libp2p/go-libp2p/config

config/config.go:252:11: invalid operation: (func(*"github.com/libp2p/go-libp2p-circuit".Relay) error)(opt) == "github.com/libp2p/go-libp2p-circuit".OptHop (func can only be compared to nil)
```

willscott:
looks like you're fixing in https://github.com/libp2p/go-libp2p-circuit/pull/103

PR #878, opened 2020-04-07
Clearer naming of nat override options

### LINKED ISSUE #874: Rename WithReachability option
This option _did not_ pass the "wtf is this thing" test with @daviddias. We should consider calling it something like `ForceReachability` or even `AssumePublic`/`AssumePrivate`, ... I'm not sure. Just something else.

Thoughts @willscott?

willscott:
fix #874

### DIFF
--- a/options.go
+++ b/options.go
@@ -279,15 +279,20 @@ func DefaultStaticRelays() Option {
 	}
 }
 
-// WithReachability overrides automatic reachability detection to force the local node
-// to believe it is either unreachable or reachable externally.
-func WithReachability(reachable bool) Option {
+// ForceReachabilityPublic overrides automatic reachability detection in the AutoNAT subsystem,
+// forcing the local node to believe it is reachable externally.
+func ForceReachabilityPublic() Option {
 	return func(cfg *Config) error {
-		if reachable {
-			cfg.Reachability = network.ReachabilityPublic
-		} else {
-			cfg.Reachability = network.ReachabilityPrivate
-		}
+		cfg.Reachability = network.ReachabilityPublic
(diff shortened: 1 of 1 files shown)


options.go
```diff
@@ -279,15 +279,20 @@ func DefaultStaticRelays() Option {
 	}
 }
 
-// WithReachability overrides automatic reachability detection to force the local node
-// to believe it is either unreachable or reachable externally.
-func WithReachability(reachable bool) Option {
+// WithNATPublicReachability overrides automatic reachability detection to force the local node
+// to believe it is reachable externally.
+func WithNATPublicReachability() Option {
```
Stebalien, line 284: * NAT isn't _actually_ relevant. AutoNAT is poorly named...
* `With` is also funky given that we're not using that idiom for other libp2p options.
* Given the description, I fell like this should use the word force somewhere.

Maybe:

* `ForceReachabilityPublic`, `ForceReachabilityPrivate`?

Stebalien:
@daviddias better?

daviddias:
Not 100% sure, will sleep on it. I think the thing that makes it look foreign is that even for a Networks person, "reachability" as a property is not directly associated with NAT'ed vs. Public IPs. So it is a stepping stone to learn about, not huge though and the docs do help :)

Stebalien:
> I think the thing that makes it look foreign is that even for a Networks person, "reachability" as a property is not directly associated with NAT'ed vs. Public IPs.

That's actually what this option is trying to capture. The "AutoNAT" service is really trying to determine "am I directly reachable from the public internet", not "am I behind a NAT".

Issue #872, opened 2020-04-06
Connection Gating


aarshkshah1992:
See https://github.com/libp2p/go-libp2p-core/pull/99 for the motivation for this issue.

### TODO (PRs to be merged)

- [x] The `ConnectionGater` interface in core **(status: released)**:
        https://github.com/libp2p/go-libp2p-core/pull/139
       
- [x] Use the ConnectionGater in the transport upgrader **(status: released)**:
       https://github.com/libp2p/go-libp2p-transport-upgrader/pull/55 

- [x] Use the ConnectionGater in Swarm **(status: released)**:
       https://github.com/libp2p/go-libp2p-swarm/pull/201

- [x] Shim the Filter to a ConnectionGater **(status: done in https://github.com/libp2p/go-libp2p/pull/881)**:
      https://github.com/libp2p/go-maddr-filter/pull/22

- [x] Deprecate Filters with ConnectionGater in libp2p and inject it to Swarm and Upgrader. Also, reflection magic to inject gaters into transports (should they need them) **(status: merged)**: https://github.com/libp2p/go-libp2p/pull/881

### TODO (Work to be done)

- [x] Default Gater implementations.
    * (=> https://github.com/libp2p/go-libp2p/issues/930)
- [x] Use the Gater in non-upgrade transports like QUIC etc.
    * (=> https://github.com/libp2p/go-libp2p/issues/931)
- [x] Move the `MockConnGater` in Swarm to "go-libp2p/p2p/net/mock ". See https://github.com/libp2p/go-libp2p-swarm/pull/201#discussion_r406306779.
    * (=> https://github.com/libp2p/go-libp2p/issues/930)

willscott:
A timestamp in the `GoAway` message might not be a bad idea. The UUID is useful to prevent replays, but if the message ends up delayed on a queue / otherwise shows up much in the future, the node receiving it may want to be able to ignore it as out-of-date.

aarshkshah1992:
As discussed with @raulk offline, we are going to call it the `Disconnect` protocol instead of the `GoAway` protocol and designing it will need a separate focused async discussion.

For now, the first win is to get in the connection gating wok.

raulk:
@willscott I don't think I heard about the UUID idea yet -- where can I find more info?

Some notes from @aarshkshah1992 and my discussion.

* Is Disconnect an ordinary protocol? Do we mount it as any other protocols? Or is Disconnect a message in a more elemental _libp2p wire protocol_?
[...]

PR #881, opened 2020-04-09
implement connection gating at the top level

aarshkshah1992:
For #872.

* This PR deprecates the public `Filter` and `FilterAddress` API in favour of the `ConnectionGater` API.
* Users can use either a `Filter` and by extension `FilterAddress` OR the `ConnectionGater` but NOT both.
* `Filters` is adapted to a `ConnectionGater` by means of `FilterConnectionGater`
* Adds the `ConenctionGater` to the "reflection magic" so it is available to Transports that need it.
  * While transports that need upgrading like TCP & WS will have their job done by the Upgrader, transports like QUIC (which do not need upgrading) should use this injected gater to gate connections.


options_filter.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// FiltersConnectionGater is an adapter that turns multiaddr.Filter into a
+// connmgr.ConnectionGater. It's not intended to be used directly by the user,
+// but it can.
+type FiltersConnectionGater ma.Filters
```
Stebalien, line 15: @raulk shouldn't this live somewhere else? Or is there a better implementation I should use in go-ipfs?

Stebalien: (at the moment, this lives in the go-libp2p root package)

raulk: Yeah, there's no reason to make this public: https://github.com/libp2p/go-libp2p/pull/946

config/reflection_magic.go
```diff
@@ -76,7 +78,7 @@ func callConstructor(c reflect.Value, args []reflect.Value) (interface{}, error)
 	return val, err
 }
 
-type constructor func(h host.Host, u *tptu.Upgrader) interface{}
+type constructor func(h host.Host, u *tptu.Upgrader, cg connmgr.ConnectionGater) interface{}
```
Stebalien, line 81: @aarshkshah1992 

What was the motivation to change this function signature? It's not really an issue, but it's redundant with the `ConnGater` field of the upgrader.

raulk: Yes, I had the same thought but given the types are internal, I thought it wouldn't harm us to rectify later. Wanna tackle that as a follow-up, @aarshkshah1992?

aarshkshah1992: @Stebalien @raulk 

I think this was an oversight. I'll create a follow up PR to get rid of this. #951 .

Stebalien: Thanks! Note: this isn't a super high priority.

PR #882, opened 2020-04-10
Expose option for setting autonat throttling

willscott:
Exposes the underlying autonat options for throttling at the libp2p level.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -48,12 +48,11 @@ type RoutingC func(host.Host) (routing.PeerRouting, error)
 
 // autoNATConfig defines the AutoNAT behavior for the libp2p host.
 type AutoNATConfig struct {
-	ForceReachabilityPublic  bool
-	ForceReachabilityPrivate bool
-	EnableService            bool
-	ThrottleGlobalLimit      int
-	ThrottlePeerLimit        int
-	ThrottleInterval         time.Duration
+	ForceReachability   *network.Reachability
+	EnableService       bool
+	ThrottleGlobalLimit int
+	ThrottlePeerLimit   int
+	ThrottleInterval    time.Duration
 }
 
 // Config describes a set of settings for a libp2p node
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -83,10 +84,12 @@ type Config struct {
...
 
-	EnableAutoRelay bool
-	Reachability    network.Reachability
-	AutoNATService  bool
-	StaticRelays    []peer.AddrInfo
+	EnableAutoRelay   bool
+	Reachability      network.Reachability
+	AutoNATService    bool
+	AutoNATThrottling bool
+	AutoNATThrottles  [2]int
```
Stebalien, line 91: Let's use a struct instead of an array. Really, we should have one struct for all `AutoNATService` options as well.

config/config.go
```diff
@@ -280,9 +283,16 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
...
-		return addrF(h.AllAddrs())
-	})}
+	autonatOpts := []autonat.Option{
+		autonat.UsingAddresses(func() []ma.Multiaddr {
+			return addrF(h.AllAddrs())
+		}),
+	}
+	if cfg.AutoNATThrottling {
+		autonatOpts = append(autonatOpts,
+			autonat.WithThrottling(cfg.AutoNATThrottles[0], 1*time.Minute),
```
Stebalien, line 293: Lets make the time configurable as well.

options.go
```diff
@@ -307,6 +307,19 @@ func EnableNATService() Option {
 	}
 }
 
+// OverrideNATServiceThrottling changes the default rate limiting configured in helping
+// other peers determine their reachability status. When set, the host will limit
+// the number of requests it responds to in each 60 second period to the set
+// numbers. A value of '0' disables throttling.
+func OverrideNATServiceThrottling(globalThrottle, perPeerThrottle int) Option {
+	return func(cfg *Config) error {
+		cfg.AutoNATThrottling = true
```
Stebalien, line 316: Should default to _on_, no?

options.go
```diff
@@ -307,6 +307,19 @@ func EnableNATService() Option {
 	}
 }
 
+// OverrideNATServiceThrottling changes the default rate limiting configured in helping
+// other peers determine their reachability status. When set, the host will limit
+// the number of requests it responds to in each 60 second period to the set
+// numbers. A value of '0' disables throttling.
+func OverrideNATServiceThrottling(globalThrottle, perPeerThrottle int) Option {
```
Stebalien, line 314: `AutoNATServiceRateLimit(global, perPeer, interval)`?

options.go
willscott: throttling is on by default - but is over over-ridden when this option is set. Updating to a struct should make the libp2p-level config less confusing.

config/config.go
```diff
@@ -45,7 +46,15 @@ type NATManagerC func(network.Network) bhost.NATManager
 
 type RoutingC func(host.Host) (routing.PeerRouting, error)
 
-// AutoNATMode defines the AutoNAT behavior for the libp2p host.
+// autoNATConfig defines the AutoNAT behavior for the libp2p host.
+type AutoNATConfig struct {
+	ForceReachabilityPublic  bool
+	ForceReachabilityPrivate bool
```
Stebalien, line 52: Just use an enum.

PR #887, opened 2020-04-14
deprioritize unspecified addresses in mock connections

willscott:
slightly less naive address selection.


willscott:
The underlying issue is that `.Addrs()` has an overloaded meaning: if i want to know where i'm listening / what my potential source addresses are, the unspecified address is sometimes relevant. if it's in this context / in terms of addresses i'm announcing to others, than it stop being meaningful.

Stebalien:
If you want to know where you're listening, you usually use `host.Network().ListenAddresses()`. `peerstore.Addrs(peer)` should return dialable addresses.

PR #898, opened 2020-04-25
fix: refactor logic for identifying connections

Stebalien:
The DHT now assumes that the protocols will be set in the peerstore once the identify event has fired. However, if we ended up identifying the peer multiple times (multiple calls to Connect at the same time, multiple connections, etc.), we'd end up unsetting the protocols temporarily.

This change:

1. Avoids ever unsetting the protocols, ever.
2. Avoids running the identify protocol on the same connection multiple times, ever.
3. Always waits for identify to complete (or fail) before using a connection.

Chang 2 should not cause any performance degradation:

1. Before this change, if the connection had not yet been identified, we'd wait at least one round-trip per new stream to negotiate the protocol.
2. After the change, we now wait exactly one round trip to complete the identify handshake.

(so it's actually _faster_ when we're trying to open a new stream with multiple possible protocol choices)

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -188,7 +188,6 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 		h.pings = ping.NewPingService(h)
 	}
 
-	net.SetConnHandler(h.newConnHandler)
 	net.SetStreamHandler(h.newStreamHandler)
 
 	return h, nil
@@ -236,14 +235,6 @@ func New(net network.Network, opts ...interface{}) *BasicHost {
 func (h *BasicHost) Start() {
 	h.refCount.Add(1)
 	go h.background()
-}
-
-// newConnHandler is the remote-opened conn handler for inet.Network
-func (h *BasicHost) newConnHandler(c network.Conn) {
-	// Clear protocols on connecting to new peer to avoid issues caused
-	// by misremembering protocols between reconnects
-	h.Peerstore().SetProtocols(c.RemotePeer())
-	h.ids.IdentifyConn(c)
 }
 
 // newStreamHandler is the remote-opened stream handler for network.Network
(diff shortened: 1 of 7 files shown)


p2p/host/basic/basic_host_test.go
```diff
@@ -199,6 +199,12 @@ func TestHostProtoPreference(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// force the lazy negotiation to complete
+	_, err = s.Write(nil)
```
Stebalien, line 203: This is now necessary because, unlike before, we now wait for identify to complete. This shouldn't slow anything down as we'd have to wait one round trip either way. Really, it should be _faster_ because identify will wait exactly one round-trip while a protocol negotiation could take multiple.

p2p/net/mock/mock_test.go
```diff
@@ -191,8 +191,18 @@ func TestNetworkSetup(t *testing.T) {
...
-	if len(n2.Conns()) != 1 || len(n3.Conns()) != 1 {
-		t.Errorf("should have (1,1) conn. Got: (%d, %d)", len(n2.Conns()), len(n3.Conns()))
+	// should immediately have a conn on peer 1
+	if len(n2.Conns()) != 1 {
+		t.Errorf("should have 1 conn on initiator. Got: %d)", len(n2.Conns()))
+	}
+
+	// wait for reciever to see the conn.
+	for i := 0; i < 10 && len(n3.Conns()) == 0; i++ {
+		time.Sleep(time.Duration(10*i) * time.Millisecond)
```
Stebalien, line 201: timing changes

p2p/protocol/identify/id.go
```diff
@@ -187,28 +186,58 @@ func (ids *IDService) ObservedAddrsFor(local ma.Multiaddr) []ma.Multiaddr {
...
+// caller, this call will wait. If the connection has already been identified,
+// it will return immediately.
 func (ids *IDService) IdentifyConn(c network.Conn) {
+	<-ids.IdentifyWait(c)
+}
+
+// IdentifyWait triggers an identify (if the connection has not already been
+// identified) and returns a channel that is closed when the identify protocol
+// completes.
+func (ids *IDService) IdentifyWait(c network.Conn) <-chan struct{} {
```
Stebalien, line 200: The name of this function is now a bit funky but this is the least-breaking way I could do it. Before, `IdentifyWait` would never trigger an identify operation. Now it does to be consistent everywhere.

p2p/protocol/identify/id.go
```diff
@@ -280,21 +314,14 @@ func (ids *IDService) broadcast(proto protocol.ID, payloadWriter func(s network.
 		go func(p peer.ID, conns []network.Conn) {
 			defer wg.Done()
 
-			// if we're in the process of identifying the connection, let's wait.
-			// we don't use ids.IdentifyWait() to avoid unnecessary channel creation.
```
Stebalien, line 284: This is no longer a problem. If the connection has not been identified, we'll identify it. Otherwise, the channel will already exist.

p2p/host/basic/basic_host_test.go
willscott: what are the semantics of writing `nil` on streams?

whyrusleeping: IIUC, nothing gets written over the wire, this just triggers everything down to that point.

whyrusleeping: this relies on the specific behaviour of the msmux lazy stream to not nop on zero length writes

p2p/protocol/identify/id.go
```diff
@@ -187,28 +186,58 @@ func (ids *IDService) ObservedAddrsFor(local ma.Multiaddr) []ma.Multiaddr {
...
+// it will return immediately.
 func (ids *IDService) IdentifyConn(c network.Conn) {
+	<-ids.IdentifyWait(c)
+}
+
+// IdentifyWait triggers an identify (if the connection has not already been
+// identified) and returns a channel that is closed when the identify protocol
+// completes.
+func (ids *IDService) IdentifyWait(c network.Conn) <-chan struct{} {
+	ids.connsMu.RLock()
```
whyrusleeping, line 201: the reason for two sets of locking here (when you could really just get away with one) is for perf?

Stebalien: Yes. We call this for every stream. It may not be critical, but I'd rather not find out later and this was already a rw lock.

Stebalien:
> what are the semantics of writing nil on streams?

It's just an empty write. Really, it's a hack. The _write_ way to do this is to either:

1. Write a message.
2. Read a message.

However, we're doing something funny here were we don't actually want to send/receive any data on the stream.

Stebalien:
> Holding the extra mapping of all identified connections may eventually be something worth de-duplicating, since we can implicitly learn that if we have protocols stored for the peer as well, right?

Right now identify is a bit funny. It's per-connection _but_ the state is per-peer. We'll have to reconcile that at some point.

p2p/host/basic/basic_host.go
```diff
@@ -605,20 +586,13 @@ func (h *BasicHost) dialPeer(ctx context.Context, p peer.ID) error {
...
-	// identify the connection before returning.
-	done := make(chan struct{})
-	go func() {
-		h.ids.IdentifyConn(c)
-		close(done)
-	}()
-
-	// respect don contexteone
+	// TODO: Consider removing this? On one hand, it's nice because we can
+	// assume that things like the agent version are usually set when this
```
aarshkshah1992, line 590: @Stebalien Didn't we discuss that it's nice to block on `Identify` for outgoing connections so we can then directly check the peerstore for supported remote protocols, etc. ?

Why this TODO ?

p2p/protocol/identify/id.go
```diff
@@ -280,21 +314,14 @@ func (ids *IDService) broadcast(proto protocol.ID, payloadWriter func(s network.
...
-				if wait, ok := ids.currid[c]; ok {
-					ids.currmu.RUnlock()
-					select {
-					case <-wait:
-						break Loop
-					case <-ctx.Done():
-						return
-					}
+				select {
+				case <-ids.IdentifyWait(c):
```
aarshkshah1992, line 321: @Stebalien Why block at all for an incoming `Identify` before sending an outgoing Push/Delta ?

p2p/host/basic/basic_host.go
Stebalien: I did say that. I left this TODO because we may want to consider _not_ doing that.

PR #899, opened 2020-04-25
fix: keep observed addrs alive as long as their associated connections are alive

Stebalien:
Otherwise, if we're not creating new connections, we'll eventually forget them.

Up until now, this wasn't _too_ much of an issue because our peers would still remember our addresses. However, we now _tell_ our connected peers when our addresses change. That means we'll tell our peers to forget where we are, preventing anyone from finding us.


p2p/protocol/identify/id.go
```diff
@@ -584,31 +583,7 @@ func (ids *IDService) consumeObservedAddress(observed []byte, c network.Conn) {
...
-	}
-
-	if !HasConsistentTransport(maddr, ids.Host.Addrs()) {
-		log.Debugf("ignoring observed multiaddr that doesn't match the transports of any addresses we're announcing", c.RemoteMultiaddr())
-		return
-	}
-
-	// ok! we have the observed version of one of our ListenAddresses!
-	log.Debugf("added own observed listen addr: %s --> %s", c.LocalMultiaddr(), maddr)
-	ids.observedAddrs.Add(maddr, c.LocalMultiaddr(), c.RemoteMultiaddr(),
```
Stebalien, line 610: Moved into the observed addr manager.

p2p/protocol/identify/obsaddr.go
```diff
@@ -185,20 +230,92 @@ func (oas *ObservedAddrSet) gc() {
...
+	// We should reject the connection if the observation doesn't match the
+	// transports of one of our advertised addresses.
+	if !HasConsistentTransport(observed, oas.host.Addrs()) {
+		log.Debugw(
+			"observed multiaddr doesn't match the transports of any announced addresses",
+			"from", conn.RemoteMultiaddr(),
+			"observed", observed,
+		)
+		return
+	}
```
Stebalien, line 298: Moved from identify.

p2p/protocol/identify/obsaddr.go
```diff
@@ -52,40 +53,51 @@ func (oa *ObservedAddr) activated(ttl time.Duration) bool {
...
+// NewObservedAddrManager returns a new address manager using
+// peerstore.OwnObservedAddressTTL as the TTL.
+func NewObservedAddrManager(host host.Host) *ObservedAddrManager {
+	oas := &ObservedAddrManager{
+		addrs:        make(map[string][]*ObservedAddr),
+		ttl:          peerstore.OwnObservedAddrTTL,
+		wch:          make(chan newObservation, observedAddrManagerWorkerChannelSize),
+		host:         host,
+		activeConns:  make(map[network.Conn]ma.Multiaddr),
+		refreshTimer: time.NewTimer(peerstore.OwnObservedAddrTTL / 2),
```
willscott, line 89: why `/2`?

p2p/protocol/identify/obsaddr.go
```diff
@@ -127,43 +139,76 @@ func (oas *ObservedAddrSet) Addrs() (addrs []ma.Multiaddr) {
...
+	oas.mu.Unlock()
+}
+
+func (oas *ObservedAddrManager) worker() {
+	defer oas.teardown()
+
 	ticker := time.NewTicker(GCInterval)
 	defer ticker.Stop()
 
+	closingCh := oas.host.Network().Process().Closing()
```
willscott, line 171: this feels more awkward than taking in a context to close on

p2p/protocol/identify/obsaddr.go
```diff
@@ -127,43 +139,76 @@ func (oas *ObservedAddrSet) Addrs() (addrs []ma.Multiaddr) {
...
+		})
+	}
+	oas.activeConnsMu.Unlock()
+
+	oas.mu.Lock()
+	defer oas.mu.Unlock()
+	for _, obs := range recycledObservations {
+		oas.recordObservationUnlocked(obs.conn, obs.observed)
+	}
+	oas.refreshTimer.Reset(oas.ttl / 2)
```
willscott, line 202: why /2?

p2p/protocol/identify/obsaddr.go
```diff
@@ -185,20 +230,92 @@ func (oas *ObservedAddrSet) gc() {
...
+	oas.activeConnsMu.Lock()
+	defer oas.activeConnsMu.Unlock()
+
+	// We need to make sure we haven't received a disconnect event for this
+	// connection yet. The only way to do that right now is to make sure the
+	// swarm still has the connection.
+	//
+	// Doing this under a lock that we _also_ take in a disconnect event
+	// handler ensures everything happens in the right order.
+	for _, c := range oas.host.Network().ConnsToPeer(conn.RemotePeer()) {
```
willscott, line 243: warn / error if the conn isn't matched?

p2p/protocol/identify/obsaddr.go
```diff
@@ -185,20 +230,92 @@ func (oas *ObservedAddrSet) gc() {
...
+
+func (oas *ObservedAddrManager) removeConn(conn network.Conn) {
+	// DO NOT remove this lock.
+	// This ensures we don't call addConn at the same time:
+	// 1. see that we have a connection and pause inside addConn right before recording it.
+	// 2. process a disconnect event.
+	// 3. record the connection (leaking it).
+
+	oas.activeConnsMu.Lock()
+	delete(oas.activeConns, conn)
```
willscott, line 259: log / warn if conn isn't in oas.activeConns ?

p2p/protocol/identify/obsaddr.go
Stebalien: We need to pass the host in anyways... but I can just use both, I guess.

p2p/protocol/identify/obsaddr.go
Stebalien: That's an expected edge-case. It won't be matched if we disconnected while processing the address observation. This check ensures that we don't store the connection in that case, and leak memory.

p2p/protocol/identify/obsaddr.go
Stebalien: We call removeConn for every connection, but not all of them will be in this set.

Stebalien: That is, not all connections will be associated with a valid observation.

p2p/protocol/identify/obsaddr.go
Stebalien: I'll add a comment. If we wait until the address TTL expires, the addresses will expire. So we wait half the expiration period before refreshing them.

p2p/protocol/identify/obsaddr.go
```diff
@@ -52,40 +54,52 @@ func (oa *ObservedAddr) activated(ttl time.Duration) bool {
...
+type ObservedAddrManager struct {
+	host host.Host
 
+	// latest observation from active connections
+	// we'll "re-observe" these when we gc
+	activeConnsMu sync.Mutex
+	// active connection -> most recent observation
+	activeConns map[network.Conn]ma.Multiaddr
+
+	mu sync.RWMutex
```
vyzo, line 71: why do we need two mutexes here?

Issue #903, opened 2020-04-30
Pipelined identify protocol "negotiation"


Stebalien:
Now that we're going to introduce multiple identify protocols, we'll need to negotiate to pick the right one. This could cost us multiple round-trips.

However, there's a simple hack to work around this: just send all protocols we speak up-front!

That is, instead of:

```
> /multistream/1.0.0\n
> /p2p/id/1.1.0
< /multistream/1.0.0\n
< na
> /ipfs/id/1.0.0\n
< /ipfs/id/1.0.0\n
< Identify Message
```

Send all protocols we speak together:

```
> /multistream/1.0.0\n
> /p2p/id/1.1.0
> /ipfs/id/1.0.0\n
< /multistream/1.0.0\n
< na
< /ipfs/id/1.0.0\n
< Identify Message
```

The trick here is that the receiver will _ignore_ any data from the initiator of the identify request.

raulk:
Can't we just make the identify message for `/p2p/id/1.1.0` backwards compatible? Unless I'm missing something (which I probably am), it should be easily doable with Protobuf.

Stebalien:
We can. But in that case, we'd need to keep the same protocol name.

Stebalien:
(i.e., /ipfs/id/1.0.0)

raulk:
Exactly, that’s precisely what I mean. Versioning protocols by outer protocol identifiers is _one_ way to version protocol. It’s practicable when it’s accompanied by an efficient protocol negotiation scheme. Multistream-select 1.0 is not that, it’s pretty wasteful, and working around it with a bespoke flow is something I’d like to avoid. Let’s version the protocol from _inside_ the protocol in this case. When we introduce a HELLO message in ms2.0, all this will become much better.

Stebalien:
This is no longer an issue.

Issue #823, opened 2020-03-03
The Identify/push/delta protocols are async and racy


Stebalien:
We broadcast updates async so the other side could end up processing them in any order. We need to serialize updates per-peer.

Additionally, if we're going to support both identify delta and identify push, we should use identify push to send protocol update events when identify delta isn't supported. To do this, we should:

1. Have a single loop subscribing to address + protocol updates.
2. Send at most one update to a peer at a time (e.g., by having one on-demand worker per peer).
3. If we receive one or more update events while sending an update to a peer, we should aggregate these updates.

aarshkshah1992:
For posterity

> We broadcast updates async so the other side could end up processing them in any order. We need to serialize updates per-peer.

**The race is as follows:**
* We detect an address change(by polling the host addresses every minute) and a local protocol update (by listening to `EvtLocalProtocolsUpdated` on the event bus) almost "simultaneously".
* The address change makes us push a snapshot of our current Identify state on the `IDPush` protocol.
* The protocol update makes us push a list of protocols we have added/removed on the `IDDelta` protocol.
* The handler for the `IDPush` protocol on the reciving peer overwrites the protocols for the sending peer by calling `Peerstore().SetProtocols`. 
* If it so happens that the Identify state on the `IDPush` protocol contains the "stale" protocols (the one before the `EvtLocalProtocolsUpdated` was fired) but are processed/written to the peerstore by the receiving peer AFTER the protocols sent by `IDDelta`,  the receiving peer will end up with a stale state of the protocols that the sender supports.

aarshkshah1992:
This change needs to be made agains https://github.com/libp2p/go-libp2p/pull/747 because of the dependency on the `EveLocalAddressChanged` event.

aarshkshah1992:
@Stebalien Let's get this on for 0.6 ?

aarshkshah1992:
@jacobheun Give that both we now plan to support Signed records in the DHT, getting this in as part of 0.6 is important.

aarshkshah1992:
### Proposal to solve races between Identify, Push and Delta.
---
### Phase 1: Solve race between Push and Delta:
  * We introduce per-peer state which includes the "last Identify message" sent to the peer and a buffered channel that we can write address and protocol change events to.
* We have an event-loop/mono-routine that guards the above mentioned state and reacts to events coming on the channel for that peer.
* After reading a message from the channel:
     * If it's an address change event:
[...]

PR #907, opened 2020-05-04
Remove race between ID, Push & Delta

aarshkshah1992:
For #823 & #903 .

Blankhost PR at https://github.com/libp2p/go-libp2p-blankhost/pull/42.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -215,10 +215,10 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 	rec := peer.PeerRecordFromAddrInfo(peer.AddrInfo{h.ID(), h.Addrs()})
 	ev, err := record.Seal(rec, h.signKey)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create signed record for self, err=%s", err)
+		return nil, fmt.Errorf("failed to create signed record for self: %w", err)
 	}
 	if _, err := cab.ConsumePeerRecord(ev, peerstore.PermanentAddrTTL); err != nil {
-		return nil, fmt.Errorf("failed to persist signed record to peerstore, err=%s", err)
+		return nil, fmt.Errorf("failed to persist signed record to peerstore: %w", err)
 	}
 
 	return h, nil
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -37,6 +37,9 @@ import (
 // peer (for all addresses).
 const maxAddressResolution = 32
 
+// addrChangeTickrInterval is the interval between two address change ticks.
+var addrChangeTickrInterval = 5 * time.Second
+
 var log = logging.Logger("basichost")
 
 var (
@@ -403,7 +406,7 @@ func (h *BasicHost) background() {
 
 	// periodically schedules an IdentifyPush to update our peers for changes
 	// in our address set (if needed)
-	ticker := time.NewTicker(5 * time.Second)
+	ticker := time.NewTicker(addrChangeTickrInterval)
 	defer ticker.Stop()
 
 	for {
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -3,6 +3,7 @@ package basichost
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"sync"
@@ -156,7 +157,7 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 	if h.emitters.evtLocalProtocolsUpdated, err = h.eventbus.Emitter(&event.EvtLocalProtocolsUpdated{}); err != nil {
 		return nil, err
 	}
-	if h.emitters.evtLocalAddrsUpdated, err = h.eventbus.Emitter(&event.EvtLocalAddressesUpdated{}, eventbus.Stateful); err != nil {
+	if h.emitters.evtLocalAddrsUpdated, err = h.eventbus.Emitter(&event.EvtLocalAddressesUpdated{}); err != nil {
 		return nil, err
 	}
 
@@ -206,6 +207,16 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 	}
 
 	net.SetStreamHandler(h.newStreamHandler)
+
+	// persist a signed peer record for self to the peerstore.
+	rec := peer.PeerRecordFromAddrInfo(peer.AddrInfo{h.ID(), h.Addrs()})
+	ev, err := record.Seal(rec, h.signKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create signed record for self, err=%s", err)
+	}
+	if _, err := cab.ConsumePeerRecord(ev, peerstore.PermanentAddrTTL); err != nil {
+		return nil, fmt.Errorf("failed to persist signed record to peerstore, err=%s", err)
+	}
 
 	return h, nil
 }
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -14,7 +14,6 @@ import (
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/peerstore"
-	"github.com/libp2p/go-libp2p-core/protocol"
 	"github.com/libp2p/go-libp2p-core/record"
 
 	"github.com/libp2p/go-eventbus"
@@ -29,12 +28,9 @@ import (
 
 var log = logging.Logger("net/identify")
 
-// ID is the protocol.ID of the Identify Service.
-const ID = "/p2p/id/1.1.0"
-
-// LegacyID is the protocol.ID of version 1.0.0 of the identify
-// service, which does not support signed peer records.
-const LegacyID = "/ipfs/id/1.0.0"
+// ID is the protocol.ID of version 1.0.0 of the identify
+// service.
+const ID = "/ipfs/id/1.0.0"
 
 // LibP2PVersion holds the current protocol version for a client running this code
 // TODO(jbenet): fix the versioning mess.
@@ -158,10 +154,6 @@ func NewIDService(h host.Host, opts ...Option) *IDService {
 
 	// register protocols that do not depend on peer records.
 	h.SetStreamHandler(IDDelta, s.deltaHandler)
-	h.SetStreamHandler(LegacyID, s.sendIdentifyResp)
-	h.SetStreamHandler(LegacyIDPush, s.pushHandler)
-
-	// register protocols that depend on peer records.
 	h.SetStreamHandler(ID, s.sendIdentifyResp)
 	h.SetStreamHandler(IDPush, s.pushHandler)
 
@@ -201,7 +193,7 @@ func (ids *IDService) loop() {
 
 			if ids.Host.Network().Connectedness(rp) == network.Connected {
 				mes := &pb.Identify{}
-				ids.populateMessage(mes, addReq.s.Conn(), protoSupportsPeerRecords(addReq.s.Protocol()))
+				ids.populateMessage(mes, addReq.s.Conn())
 				ph = newPeerHandler(rp, ids, mes)
 				phs[rp] = ph
 				addReq.resp <- ph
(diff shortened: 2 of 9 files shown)


p2p/protocol/identify/peer_loop.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/protocol"
+
+	pb "github.com/libp2p/go-libp2p/p2p/protocol/identify/pb"
+
+	ggio "github.com/gogo/protobuf/io"
+)
+
+var errProtocolNotSupported = errors.New("protocol not supported")
```
willscott, line 19: is this error not already declared somewhere generic?

p2p/protocol/identify/peer_loop.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+
+			mes := &pb.Identify{}
+			ph.ids.populateMessage(mes, dp.Conn())
+
+			ph.msgMu.Lock()
+			ph.idMsgSnapshot = mes
+			ph.msgMu.Unlock()
+
+			ph.sendMessage(dp, mes)
```
willscott, line 85: errors silently dropped here and in delta below

p2p/protocol/identify/id.go
```diff
@@ -29,12 +28,9 @@ import (
...
 
-// ID is the protocol.ID of the Identify Service.
-const ID = "/p2p/id/1.1.0"
-
-// LegacyID is the protocol.ID of version 1.0.0 of the identify
-// service, which does not support signed peer records.
-const LegacyID = "/ipfs/id/1.0.0"
+// ID is the protocol.ID of version 1.0.0 of the identify
+// service.
+const ID = "/ipfs/id/1.0.0"
```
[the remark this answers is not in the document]
aarshkshah1992: @willscott Yes, we are now going ahead with using the same protocol and using a backward compatible protobuf to determine if the other peer sent us a signed records. If not, we read in the unsigned records from the response.

See discussion on https://github.com/libp2p/go-libp2p/issues/903.

p2p/protocol/identify/id.go
```diff
@@ -146,40 +154,96 @@ func NewIDService(h host.Host, opts ...Option) *IDService {
...
 		&event.EvtLocalAddressesUpdated{}}, eventbus.BufSize(256))
 	if err != nil {
 		log.Errorf("failed to subscribe to events on the bus, err=%s", err)
 		return
 	}
 
-	defer sub.Close()
+	defer func() {
+		sub.Close()
+		for pid := range phs {
```
Stebalien, line 177: nit:

```go
for _, ph := range phs {
  ph.close()
}
```

p2p/protocol/identify/id.go
```diff
@@ -146,40 +154,96 @@ func NewIDService(h host.Host, opts ...Option) *IDService {
...
+						case <-ids.ctx.Done():
+							return
+						case phClosedCh <- req.p:
+						}
+					}
+				}(rmReq, phs[rp])
+			}
+
+		case p := <-phClosedCh:
+			delete(phs, p)
```
Stebalien, line 223: I think we need to check if we're connected to the peer at this point. If we are, we need to restart this peer handler.

p2p/protocol/identify/peer_loop.go
```diff
@@ -0,0 +1,209 @@
...
+	oldProtos := make(map[string]struct{}, len(old))
+	currProtos := make(map[string]struct{}, len(curr))
+
+	for i := range old {
+		oldProtos[old[i]] = struct{}{}
+	}
+
+	for i := range curr {
+		currProtos[curr[i]] = struct{}{}
+	}
```
Stebalien, line 175: nit:

```suggestion
	for _, proto := range old {
		oldProtos[proto] = struct{}{}
	}

	for _, proto := range curr {
		currProtos[proto] = struct{}{}
	}
```

p2p/protocol/identify/peer_loop.go
```diff
@@ -0,0 +1,209 @@
...
+		AddedProtocols: added,
+		RmProtocols:    removed,
+	}
+}
+
+func (ph *peerHandler) sendMessage(s network.Stream, mes *pb.Identify) error {
+	defer helpers.FullClose(s)
+	c := s.Conn()
+	if err := ggio.NewDelimitedWriter(s).WriteMsg(mes); err != nil {
+		log.Warnf("error while sending %s update to %s: err=%s", s.Protocol(), c.RemotePeer(), err)
```
Stebalien, line 204: nit: prefer `log.Warnw("error while sending identify update", "protocol", s.Protocol(), "peer", c.RemotePeer(), "error", err)`. This gives us nice structured logs we can index and query.

(we _used_ to use Warnf and friends but we're switching to a new logging backend)

p2p/host/basic/basic_host.go
```diff
@@ -207,6 +211,16 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 
 	net.SetStreamHandler(h.newStreamHandler)
 
+	// persist a signed peer record for self to the peerstore.
+	rec := peer.PeerRecordFromAddrInfo(peer.AddrInfo{h.ID(), h.Addrs()})
+	ev, err := record.Seal(rec, h.signKey)
```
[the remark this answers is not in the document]
aarshkshah1992: @willscott Because it's an envelope. A sealed envelope. Do you have a preference here ?

p2p/protocol/identify/id_push.go
```diff
@@ -2,28 +2,16 @@ package identify
...
-// which does not support exchanging signed addresses in PeerRecords.
-// It is still supported for backwards compatibility if a remote peer does not support
-// the current version.
-const LegacyIDPush = "/ipfs/id/push/1.0.0"
-
-// Push pushes a full identify message to all peers containing the current state.
-func (ids *IDService) Push() {
-	ids.broadcast([]protocol.ID{IDPush, LegacyIDPush}, ids.requestHandler)
-}
+const IDPush = "/p2p/id/push/1.0.0"
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien Good catch !

p2p/protocol/identify/peer_loop.go
aarshkshah1992: @Stebalien Do we such errors declared anywhere ? I know of none.

Stebalien: https://github.com/multiformats/go-multistream/blob/bd94c78329f9d3cd394f09e53af6138dd5b4db89/client.go#L15

p2p/protocol/identify/peer_loop.go
aarshkshah1992: Done.

p2p/protocol/identify/peer_loop.go
aarshkshah1992: Doing this everywhere now.

p2p/protocol/identify/peer_loop.go
aarshkshah1992: This has been fixed by logging them here.

p2p/protocol/identify/peer_loop.go
```diff
@@ -0,0 +1,209 @@
...
+	ph.cancel()
+	ph.wg.Wait()
+	return nil
+}
+
+// per peer loop for pushing updates
+func (ph *peerHandler) loop() {
+	defer ph.wg.Done()
+
+	for {
```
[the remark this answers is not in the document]
aarshkshah1992: This is done.

p2p/protocol/identify/peer_loop.go
aarshkshah1992: This has been fixed.

p2p/protocol/identify/id.go
```diff
@@ -99,6 +101,9 @@ type IDService struct {
 		evtPeerIdentificationCompleted event.Emitter
 		evtPeerIdentificationFailed    event.Emitter
 	}
+
+	addPeerHandlerCh chan *addPeerHandlerReq
+	rmPeerHandlerCh  chan *rmPeerHandlerReq
```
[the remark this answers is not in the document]
aarshkshah1992: This is done.

p2p/protocol/identify/id.go
aarshkshah1992: @Stebalien I really like the current one better because I've run into golang problems when iterating over a map where the value type is a pointer(though that isn't the case here, I now prefer that style).

p2p/protocol/identify/id.go
```diff
@@ -146,40 +154,96 @@ func NewIDService(h host.Host, opts ...Option) *IDService {
...
+		case rmReq := <-ids.rmPeerHandlerCh:
+			rp := rmReq.p
+			if ids.Host.Network().Connectedness(rp) != network.Connected {
+				// before we remove the peerhandler, we should ensure that it will not send any
+				// more messages. Otherwise, we might create a new handler and the Identify response
+				// synchronized with the new handler might be overwritten by a message sent by this "old" handler.
+				ids.refCount.Add(1)
+				go func(req *rmPeerHandlerReq, ph *peerHandler) {
+					defer ids.refCount.Done()
+					if ph != nil {
```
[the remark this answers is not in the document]
aarshkshah1992: This is done.

p2p/protocol/identify/id.go
aarshkshah1992: @Stebalien Great find ! This is done.

p2p/protocol/identify/id.go
```diff
@@ -293,36 +343,50 @@ func (ids *IDService) identifyConn(c network.Conn, signal chan struct{}) {
...
+		return
+	}
+
+	select {
+	case ph = <-phCh:
+	case <-ids.ctx.Done():
+		return
+	}
+
+	ph.msgMu.RLock()
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien You mean block when another thread has **already** taken a write-lock, right ? Otherwise, dosen't go prefer granting read locks over waiting write locks ? Can you explain this in a bit more details and what the rare case you are talking about is ?

I'll then create an issue to track this.

p2p/protocol/identify/id.go
```diff
@@ -63,6 +59,15 @@ func init() {
 // transientTTL is a short ttl for invalidated previously connected addrs
 const transientTTL = 10 * time.Second
 
+type addPeerHandlerReq struct {
+	s    network.Stream
```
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien I do understand the intuition behind not wanting to send streams to channels but can you explain with a bit more detail why we wouldn't to pass streams across channels.

I am making the change.

p2p/protocol/identify/peer_loop.go
aarshkshah1992: Don't want a dependency on multi-stream for this. Copying for now.

aarshkshah1992:
@Stebalien Have addresses your review comments with some counter questions. Please take a look when you can !

p2p/protocol/identify/id.go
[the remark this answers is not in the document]
Stebalien: I mean:

1. Thread 1 starts responding to an identify request, taking the read lock.
2. Thread 2 tries to perform an identify push, requesting the write lock.
3. Thread 3 tries to respond to an identify request and blocks trying to take the read lock.

3 will happen because the attempt to take a write lock will prevent any new read locks.

p2p/protocol/identify/id.go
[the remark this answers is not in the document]
Stebalien: Streams on channels is fine. The issue is really:

* The identify loop doesn't actually care about the stream, we're handling the response outside of the loop. Really, we're just sending the stream so we can pull off the remote address so we can initialize an identify message that we're not even sending from the loop.
* The observed address will always be the addressed observed during the first identify request until we finally disconnect.

p2p/protocol/identify/peer_loop.go
```diff
@@ -0,0 +1,265 @@
...
+
+		idMsgSnapshot: initState,
+
+		pushCh:  make(chan struct{}, 1),
+		deltaCh: make(chan struct{}, 1),
+	}
+
+	if isTesting {
+		ph.evalTestCh = make(chan func())
+	}
```
Stebalien, line 57: (ok, maybe I was too optimization eager. I won't object if you want to revert this.

p2p/protocol/identify/id.go
Stebalien: If we do that, we may need to also "forward" any pending push/delta signals.

On the other hand, we could restart it by:

1. Not starting inside `newPeerHandler`, adding a `start` method instead.
2. Having the `start` function initialize the ctx + cancel fields and start the loop.

But that's not necessary.

p2p/protocol/identify/id.go
Stebalien: Eh, I don't mind much either way. I just find it cleaner to iterate by value (also likely slightly faster but that _really_ doesn't matter here).

What was your bad experience?

p2p/protocol/identify/id.go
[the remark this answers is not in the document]
aarshkshah1992: @Stebalien I see what you mean. Have made the change.

aarshkshah1992:
Hey @Stebalien 

Have made all the changes. This PR is ready for one last look. Please take a look when you can !

p2p/protocol/identify/peer_loop.go
aarshkshah1992: @Stebalien I'm good with this change. It's one channel alloc for every peer we are connected to.

p2p/protocol/identify/id.go
aarshkshah1992: @Stebalien After thinking through this, I think your approach of having a `start` function that we can use to re-start the handler is better as it means we don't have to take a read lock inside the event loop to copy the state. 

I have made the change.

Stebalien:
(rebased)

raulk:
This PR was merged and released, but it points to a commit of go-libp2p-blankhost instead of a release:

https://github.com/libp2p/go-libp2p/commit/973933ad7d8caf1b661d97260cb6242ac18b7aef#diff-37aff102a57d3d7b797f152915a6dc16R14

raulk:
Looks like it wasn't released, so we have time to fix this on master. @aarshkshah1992?

PR #909, opened 2020-05-07
Update id.go

cc14514:
IDService.peerrec and evt.SignedPeerRecord both are `*record.Envelope` type , cannot compile with pointer assignment

```go
func (ids *IDService) handleLocalAddrsUpdated(evt event.EvtLocalAddressesUpdated) {
	ids.peerrecMu.Lock()
	rec := evt.SignedPeerRecord
	ids.peerrec = &rec // ???
	ids.peerrecMu.Unlock()

	log.Debug("triggering push based on updated local PeerRecord")
	ids.Push()
}
```


Stebalien:
This should e fixed by #907.

PR #914, opened 2020-05-10
fix: SignedPeerRecord now expects pointer to envelope

iand:
go-libp2p-core changed EvtLocalAddressesUpdated.SignedPeerRecord into
a pointer to record.Envelope to avoid copying a mutex. This broke
the build of go-libp2p.

Fixed by not dereferencing pointers when assigning.

Also fix TestCompatibilityWithPeersThatDoNotSupportSignedAddrs test
which was failing due to a race between update and test.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -376,7 +376,8 @@ func (h *BasicHost) background() {
 			log.Errorf("error creating a signed peer record from the set of current addresses, err=%s", err)
 			return
 		}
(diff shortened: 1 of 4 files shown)


aarshkshah1992:
Hey @iand  

Thanks a lot for the PR. However, this will be fixed and refactored by https://github.com/libp2p/go-libp2p/pull/907 which should land this week.

Stebalien:
This has been fixed in master.

Issue #910, opened 2020-05-07
Filter advertised observed addresses


Stebalien:
Currently, we advertise all observed addresses that have enough votes. Instead, for each IP & transport set, we should advertise a few with the most observations.

Algorithm:

* Given `/ipX/.../{tcp,udp}/PORT/.....`, group observations by the key `/ipX/.../{tcp,udp}/0/.....`.
* For each group of observations:
  * Advertise the top 1-2 addresses observed by inbound connections.
  * Advertise the top 1-2 observations from outbound connections.

Why advertise both inbound/outbound? Port mappings may have changed.

Ideally, instead of just advertising 1-2, we'd look at standard deviations and advertise everything within 1 stdev of the most observed address (with a cutt-off of 3?).

aarshkshah1992:
@Stebalien I went though how the observed addr manager works and understand what we are trying to accomplish here. Can you please just explain this one comment of yours in a bit more detail ?

> Why advertise both inbound/outbound? Port mappings may have changed.

Stebalien:
Well, my thinking was that my NAT may have mapped me to one outbound port then switched that mapping to a new port.

If I continue advertising the inbound port only but am mostly making outbound requests (e.g., I'm a DHT client), I may not be advertising the correct address.

On the other hand, this seems like a _really_ unlikely case. It's probably safe to just announce addresses discovered from inbound connections.

aarshkshah1992:
Merged in https://github.com/libp2p/go-libp2p/pull/917.

PR #917, opened 2020-05-12
Filter observed addresses

aarshkshah1992:
@Stebalien 

For #910 .

### DIFF
--- a/p2p/protocol/identify/obsaddr.go
+++ b/p2p/protocol/identify/obsaddr.go
@@ -2,6 +2,7 @@ package identify
 
 import (
 	"context"
+	"sort"
 	"sync"
 	"time"
 
@@ -10,6 +11,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/peerstore"
 
 	ma "github.com/multiformats/go-multiaddr"
+	mafmt "github.com/multiformats/go-multiaddr-fmt"
 	manet "github.com/multiformats/go-multiaddr-net"
 )
 
@@ -30,6 +32,24 @@ var GCInterval = 10 * time.Minute
 // for adding to an ObservedAddrManager.
 var observedAddrManagerWorkerChannelSize = 16
 
+// maxObservedAddrsPerIPAndTransport is the maximum number of observed addresses
+// we will return for each (IPx/TCP or UDP) group.
+var maxObservedAddrsPerIPAndTransport = 2
+
+var (
+	ip4 = mafmt.Base(ma.P_IP4)
+	ip6 = mafmt.Base(ma.P_IP6)
+	udp = mafmt.Base(ma.P_UDP)
+	tcp = mafmt.Base(ma.P_TCP)
+
+	addrPatterns = []mafmt.Pattern{
+		mafmt.And(ip4, udp),
+		mafmt.And(ip4, tcp),
+		mafmt.And(ip6, udp),
+		mafmt.And(ip6, tcp),
+	}
+)
+
 type observation struct {
 	seenTime      time.Time
 	connDirection network.Direction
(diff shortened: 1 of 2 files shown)


p2p/protocol/identify/obsaddr.go
```diff
@@ -46,13 +66,36 @@ type ObservedAddr struct {
...
-func (oa *ObservedAddr) activated(ttl time.Duration) bool {
+func (oa *ObservedAddr) activated() bool {
+
 	// We only activate if other peers observed the same address
 	// of ours at least 4 times. SeenBy peers are removed by GC if
 	// they say the address more than ttl*ActivationThresh
 	return len(oa.SeenBy) >= ActivationThresh
 }
 
+func (oa *ObservedAddr) nInbound() int {
```
Stebalien, line 77:
```suggestion
func (oa *ObservedAddr) numInbound() int {
```

p2p/protocol/identify/obsaddr.go
```diff
@@ -46,13 +66,36 @@ type ObservedAddr struct {
...
+func (oa *ObservedAddr) GroupKey() string {
+	for i := range addrPatterns {
+		pat := addrPatterns[i]
+		if pat.Matches(oa.Addr) {
+			return pat.String()
+		}
+	}
+
+	return ""
+}
```
Stebalien, line 97: We also need to match things like `/ws`, `/quic,` etc. We should probably just call `oa.Addr.Protocols()` and concatenate the protocol codes to get the group key.

p2p/protocol/identify/obsaddr.go
Stebalien: (nit because it took me a second to realize what `n` meant in this context)

aarshkshah1992:
Will wait on https://github.com/libp2p/go-libp2p/pull/934 to ensure tests aren't flaky.

p2p/protocol/identify/obsaddr.go
aarshkshah1992: Good point. This is done.

aarshkshah1992:
@Stebalien There's definitely some flakiness in the tests

```
aarshshah@Aarshs-MacBook-Pro identify % go test -race -count=20 -run TestObservedAddrFiltering

--- FAIL: TestObservedAddrFiltering (4.52s)
    obsaddr_test.go:327: 
        	Error Trace:	obsaddr_test.go:327
        	Error:      	"[]" should have 4 item(s), but has 0
        	Test:       	TestObservedAddrFiltering
--- FAIL: TestObservedAddrFiltering (4.49s)
    obsaddr_test.go:327: 
        	Error Trace:	obsaddr_test.go:327
        	Error:      	"[]" should have 4 item(s), but has 0
        	Test:       	TestObservedAddrFiltering
FAIL
exit status 1

```

I'm looking into it.

p2p/protocol/identify/obsaddr.go
```diff
@@ -46,13 +51,36 @@ type ObservedAddr struct {
...
+
+	return count
+}
+
+func (oa *ObservedAddr) GroupKey() string {
+	key := ""
+	protos := oa.Addr.Protocols()
+
+	for i := range protos {
+		key = key + protos[i].Name
```
Stebalien, line 78: This is fine but kind of sloppy. We should either use a separator (`/ip4/udp/quic`) or concatenate `protocol.VCode`s (these are prefix-free codes so we don't need separators).

p2p/protocol/identify/obsaddr.go
```diff
@@ -111,33 +139,65 @@ func (oas *ObservedAddrManager) AddrsFor(addr ma.Multiaddr) (addrs []ma.Multiadd
...
+
+	for i := range observedAddrs {
+		a := observedAddrs[i]
+		if now.Sub(a.LastSeen) <= oas.ttl && a.activated() {
+			// group addresses by their IPX/Transport Protocol(TCP or UDP) pattern.
+			pat := a.GroupKey()
+			if len(pat) != 0 {
+				pmap[pat] = append(pmap[pat], a)
+			} else {
+				log.Debugw("unable to group observed addr into IPx/(TCP or UDP) patterm", "address",
```
Stebalien, line 173: is this possible?

p2p/protocol/identify/obsaddr.go
aarshkshah1992: Using a separator now.

p2p/protocol/identify/obsaddr.go
aarshkshah1992: Umm... it can "theoretically" happen if the address dosen't have any protocols but it probably wouldn't make it so far if that were that case. Removing it to reduce clutter.

PR #919, opened 2020-05-12
fix a compilation error introduced in 077a818.

raulk:


vyzo:
Still failing to compile:
```
*** go build github.com/libp2p/go-libp2p

# github.com/libp2p/go-libp2p/p2p/host/basic

p2p/host/basic/basic_host.go:379:30: cannot use *sr (type record.Envelope) as type *record.Envelope in assignment

```

raulk:
Tests fail because of the races that will be fixed in https://github.com/libp2p/go-libp2p/pull/907.

PR #922, opened 2020-05-13
various identify fixes and nits

Stebalien:
My fingers got itchy to write some code.

* Don't cache the identify message, cache the state and re-generate the identify message each time relative to the connection being identified:
  * Always send the correct observed addr for the connection being identified.
    * (this is what I was getting at in https://github.com/libp2p/go-libp2p/pull/907/files#r421818108)
  * Always send the correct addresses for the connection being identified by filtering.
    * I'm not sure if this is the _right_ approach, but it matches our current approach.
* Delta changes
  * Avoid extra work when the remote peer doesn't support identify delta.
  * Instead of applying delta locally, just update the snapshot with the new protocols.
  * Always access the snapshot under a lock and get rid of the test channel. This doesn't cost us in terms of performance and makes the system easier to reason about.


aarshkshah1992:
(rebased).

aarshkshah1992:
@Stebalien Good stuff. LGTM !

p2p/protocol/identify/id.go
```diff
@@ -194,9 +192,7 @@ func (ids *IDService) loop() {
 			}
 
 			if ids.Host.Network().Connectedness(rp) == network.Connected {
-				mes := &pb.Identify{}
-				ids.populateMessage(mes, rp, addReq.localConnAddr, addReq.remoteConnAddr)
-				ph = newPeerHandler(rp, ids, mes)
+				ph = newPeerHandler(rp, ids)
 				ph.start()
 				phs[rp] = ph
 				addReq.resp <- ph
```
raulk, line 198: If the peer disconnects between the time when the other goroutine sent into this one and now, we will never respond on the resp channel and we'll have a goroutine leak.

@Stebalien

Stebalien: Hm. You're right. We need to send _something_ back then reset the stream, I guess.

Stebalien: Fixed in https://github.com/libp2p/go-libp2p/pull/942

Issue #929, opened 2020-05-15
Flaky tests on master


aarshkshah1992:
Something is really weird:

https://travis-ci.com/github/libp2p/go-libp2p/builds/165988310

```
=== RUN   TestNotifications
2020-05-14T11:55:38.727Z	ERROR	boguskey	go-libp2p-netutil@v0.1.0/key.go:63	TestBogusPrivateKey.Sign -- this better be a test!
2020-05-14T11:55:38.729Z	ERROR	boguskey	go-libp2p-netutil@v0.1.0/key.go:63	TestBogusPrivateKey.Sign -- this better be a test!
2020-05-14T11:55:38.731Z	ERROR	boguskey	go-libp2p-netutil@v0.1.0/key.go:63	TestBogusPrivateKey.Sign -- this better be a test!
2020-05-14T11:55:38.732Z	ERROR	boguskey	go-libp2p-netutil@v0.1.0/key.go:63	TestBogusPrivateKey.Sign -- this better be a test!
2020-05-14T11:55:38.734Z	ERROR	boguskey	go-libp2p-netutil@v0.1.0/key.go:63	TestBogusPrivateKey.Sign -- this better be a test!
2020-05-14T11:55:38.739Z	ERROR	net/identify	identify/id.go:533	error getting peer record from Identify message: failed when unmarshalling the envelope: asn1: structure error: tags don't match (16 vs {class:0 tag:1 length:33 isCompound:false}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} publicKeyInfo @2
2020-05-14T11:55:38.750Z	ERROR	net/identify	identify/id.go:533	error getting peer record from Identify message: failed when unmarshalling the envelope: asn1: structure error: tags don't match (16 vs {class:0 tag:1 length:33 isCompound:false}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} publicKeyInfo @2
2020-05-14T11:55:38.751Z	ERROR	net/identify	identify/id.go:533	error getting peer record from Identify message: failed when unmarshalling the envelope: asn1: structure error: tags don't match (16 vs {class:2 tag:27 length:107 isCompound:false}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} publicKeyInfo @2
```

[...]

PR #934, opened 2020-05-18
Disable generation of Signed peer record for Mockenets

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/929.

Mocknets use Bogus private keys which shouldn't be used for signing records. However, this leads to the peer sending malformed signed peer records which floods up our test logs. 

This PR exposes an Option on the Host to disable generation of signed peer records. It also ensures that Identify responses we send ONLY have unsigned addresses. 
ONLY Mocknets uses this Option.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -101,8 +101,9 @@ type BasicHost struct {
 
 	addrChangeChan chan struct{}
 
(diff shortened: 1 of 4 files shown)


aarshkshah1992:
https://github.com/libp2p/go-libp2p/pull/917 is waiting on this.

Issue #911, opened 2020-05-07
Filter advertised interface addresses


Stebalien:
Currently, we advertise addresses of all interfaces. This will continue to be an issue until we fix #436 so we should at least _prioritize_ better addresses.

We can do this by extending the default address filter to filter interface addresses such that we advertise fewer interface addresses:

1. Always advertise localhost. This will be filtered by identify if necessary.
2. Find the "default" route and advertise that address (all transports).
3. Advertise addresses in the public address range (again, all transports).

Really, that's all we need to advertise. Given that we'll _also_ advertise observed addresses, if we happen to be connecting to other libp2p nodes, those addresses will show up in our observations so we don't need ti include them _here_.

aarshkshah1992:
> Really, that's all we need to advertise. Given that we'll also advertise observed addresses, if we happen to be connecting to other libp2p nodes, those addresses will show up in our observations so we don't need ti include them here.

I am not sure what you mean. The only place we use/include the observed addresses is in the `h.Addrs()` call. So, we shouldn't we also filter them here ? Or do you mean #910 will take care of them ?

Stebalien:
> Do you mean the LAN/private address ?

Yes. Specifically, I mean use https://github.com/libp2p/go-netroute/ to ask for the route to `0.0.0.0`. That _should_ give us our default local IP address.

> I am not sure what you mean.

I think I'm just being confusing.

Basically, in [`BasicHost.AllAddrs`](https://github.com/libp2p/go-libp2p/blob/077a81814f4357379d11dc93d40f48788e515fa0/p2p/host/basic/basic_host.go#L676), we :

1. Call `InterfaceListenAddresses` to get all of the addresses from our local interfaces.
2. Get the addresses from our NAT device.
3. Get our observed addresses.

I'm saying we should filter `InterfaceListenAddresses` to only include:

* localhost
* public addresses
* the _default_ address (as determined by go-netroute).

WRT observed addresses, I'm saying that even _if_ we filter out most of our interface addresses in the previous step, if any of them are actually being used, we should see them show up in our address observations so it shouldn't be a problem.

aarshkshah1992:
@Stebalien I understand wha you mean and I can get started on this.

However, can you give some examples of the kinds of "bad address" that creep in from `InterfaceListenAddresses ` that we want to filter here ?

Stebalien:
[...]

PR #936, opened 2020-05-18
Filter Interface Addresses

aarshkshah1992:
@Stebalien @willscott 

For #911 .

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -19,6 +19,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/protocol"
 	"github.com/libp2p/go-libp2p-core/record"
 
+	addrutil "github.com/libp2p/go-addr-util"
 	"github.com/libp2p/go-eventbus"
 	inat "github.com/libp2p/go-libp2p-nat"
 	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
@@ -40,6 +41,8 @@ const maxAddressResolution = 32
 
 // addrChangeTickrInterval is the interval between two address change ticks.
 var addrChangeTickrInterval = 5 * time.Second
+
+var localIpChangeTickrInterval = 5 * time.Minute
 
 var log = logging.Logger("basichost")
 
@@ -102,8 +105,10 @@ type BasicHost struct {
 
 	addrChangeChan chan struct{}
 
-	localIPv4Addr net.IP
-	localIPv6Addr net.IP
+	localIPChangeCh chan struct{}
+	lipMu           sync.RWMutex
+	localIPv4Addr   ma.Multiaddr
+	localIPv6Addr   ma.Multiaddr
 
 	disableSignedPeerRecord bool
 	signKey                 crypto.PrivKey
@@ -160,26 +165,13 @@ func NewHost(ctx context.Context, n network.Network, opts *HostOpts) (*BasicHost
 		maResolver:              madns.DefaultResolver,
 		eventbus:                eventbus.NewBus(),
 		addrChangeChan:          make(chan struct{}, 1),
+		localIPChangeCh:         make(chan struct{}, 1),
 		ctx:                     hostCtx,
 		ctxCancel:               cancel,
 		disableSignedPeerRecord: opts.DisableSignedPeerRecord,
 	}
 
-	if r, err := netroute.New(); err != nil {
-		log.Debugw("failed to build Router for kernel's routing table", "err", err)
-	} else {
-		if _, _, localIPv4Addr, err := r.Route(net.IPv4zero); err != nil {
-			log.Debugw("failed to fetch local IPv4 address", "err", err)
-		} else {
-			h.localIPv4Addr = localIPv4Addr
-		}
-
-		if _, _, localIpv6, err := r.Route(net.IPv6unspecified); err != nil {
-			log.Debugw("failed to fetch local IPv6 address", "err", err)
-		} else {
-			h.localIPv6Addr = localIpv6
-		}
-	}
+	h.updateLocalIpAddr()
 
 	var err error
 	if h.emitters.evtLocalProtocolsUpdated, err = h.eventbus.Emitter(&event.EvtLocalProtocolsUpdated{}); err != nil {
@@ -255,6 +247,33 @@ func NewHost(ctx context.Context, n network.Network, opts *HostOpts) (*BasicHost
 	return h, nil
 }
 
+func (h *BasicHost) updateLocalIpAddr() {
+	h.lipMu.Lock()
+	defer h.lipMu.Unlock()
+
+	if r, err := netroute.New(); err != nil {
+		log.Debugw("failed to build Router for kernel's routing table", "err", err)
+	} else {
+		if _, _, localIPv4, err := r.Route(net.IPv4zero); err != nil {
+			log.Debugw("failed to fetch local IPv4 address", "err", err)
+		} else {
+			maddr, err := manet.FromIP(localIPv4)
+			if err == nil {
+				h.localIPv4Addr = maddr
+			}
+		}
+
+		if _, _, localIpv6, err := r.Route(net.IPv6unspecified); err != nil {
+			log.Debugw("failed to fetch local IPv6 address", "err", err)
+		} else {
+			maddr, err := manet.FromIP(localIpv6)
+			if err == nil {
+				h.localIPv6Addr = maddr
+			}
+		}
+	}
+}
+
 // New constructs and sets up a new *BasicHost with given Network and options.
 // The following options can be passed:
 // * NATPortMap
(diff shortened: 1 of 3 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -141,11 +145,11 @@ type HostOpts struct {
 }
 
 // NewHost constructs a new *BasicHost and activates it by attaching its stream and connection handlers to the given inet.Network.
-func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHost, error) {
+func NewHost(ctx context.Context, network network.Network, opts *HostOpts) (*BasicHost, error) {
```
willscott, line 148: having the variable overload the import name so you can't refer to `network` anymore seems like it can add to confusion. `net` isn't great either, but ideally something non-overloaded?

p2p/host/basic/basic_host.go
```diff
@@ -224,6 +243,19 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 	return h, nil
 }
 
+func getLocalIPAddrFor(ip net.IP) (net.IP, error) {
+	r, err := netroute.New()
```
willscott, line 247: should be cached. as used this'll build up the full OS routing table twice. ideally make one netroute object for both the v4 and v6 query.

p2p/host/basic/basic_host.go
```diff
@@ -157,14 +161,29 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 	}
 
 	var err error
+
+	localIpv4, err := getLocalIPAddrFor(net.IPv4(0, 0, 0, 0))
+	if err != nil {
+		log.Debugw("failed to fetch local IPv4 address", "err", err)
+	} else {
+		h.localIPv4Addr = localIpv4
```
Stebalien, line 169: This can change at runtime. We can cache it, but we need to re-evaluate it semi frequently.

p2p/host/basic/basic_host.go
```diff
@@ -700,9 +732,39 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
 		natMappings = h.natmgr.NAT().Mappings()
 	}
 
-	finalAddrs := listenAddrs
-	if len(natMappings) > 0 {
+	var finalAddrs []ma.Multiaddr
+
+	for _, a := range listenAddrs {
```
Stebalien, line 737: Thinking through this, we could also just call `ListenAddresses` instead of `InterfaceListenAddresses` and manually replace all instances of the "zero" address with `h.localIPv4Addr`. That way, if the user _explicitly_ listens on a specific address (not `0.0.0.0`), we preserve it instead of filtering it out.

p2p/host/basic/basic_host.go
aarshkshah1992: This is done.

p2p/host/basic/basic_host.go
```diff
@@ -157,14 +161,29 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 	}
 
 	var err error
+
+	localIpv4, err := getLocalIPAddrFor(net.IPv4(0, 0, 0, 0))
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

p2p/host/basic/basic_host.go
```diff
@@ -157,14 +161,29 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
...
+	if err != nil {
+		log.Debugw("failed to fetch local IPv4 address", "err", err)
+	} else {
+		h.localIPv4Addr = localIpv4
+	}
+
+	localIpv6, err := getLocalIPAddrFor(net.ParseIP("::"))
```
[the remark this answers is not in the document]
aarshkshah1992: Done.

p2p/host/basic/basic_host.go
aarshkshah1992: This is done.

p2p/host/basic/basic_host.go
aarshkshah1992: @Stebalien Going through the code for those functions, I think `InterfaceListenAddresses` ALSO returns the addresses the user explicitly listens on. I wrote a small test to confirm this:

```
func TestSomething(t *testing.T) {
	ctx := context.Background()
	host1 := New(swarmt.GenSwarm(t, ctx, swarmt.OptDialOnly))

	require.NoError(t,host1.Network().Listen(ma.StringCast("/ip4/192.168.1.11/tcp/0")))
	fmt.Println("Host 1 All", host1.Addrs())
	fmt.Println("Host 1 Listen", host1.Network().ListenAddresses())

	host2 :=  New(swarmt.GenSwarm(t, ctx, swarmt.OptDialOnly))
	require.NoError(t, host2.Network().Listen(ma.StringCast("/ip4/0.0.0.0/tcp/0")))
	fmt.Println("Host 2 All", host2.Addrs())
	fmt.Println("Host 2 Listen", host2.Network().ListenAddresses())
}
```

```
=== RUN   TestSomething
Host 1 All [/ip4/192.168.1.11/tcp/52375]
Host 1 Listen [/ip4/192.168.1.11/tcp/52375]
Host 2 All [/ip4/127.0.0.1/tcp/52376 /ip4/192.168.1.11/tcp/52376]
Host 2 Listen [/ip4/0.0.0.0/tcp/52376]
```

However, if the "explicit address" is a private address that is NOT the default route, it will get filtered. Using `ListenAddresses` address here will ALSO not solve the problem.  Do we need to make any change here ?

p2p/host/basic/basic_host.go
aarshkshah1992: Can you explain how/when it would change at runtime ? Would it happen when the Router assigns a new private IP to the machine running this libp2p node ?

p2p/host/basic/basic_host.go
aarshkshah1992: This is done.

aarshkshah1992:
@willscott 

> probably need to listen for 'local addresses changed' events to invalidate the cached default source addresses.

The local address change event will happen when the libp2p node starts listening on a new address. 

Hence, your suggestion is that we change the cached default source addresses when a libp2p node starts listening on a new address. Please can you explain why the default source address we get from the `netroute` library would change if libp2p listens on a new address ?

willscott:
I think the point is that if you get a new interface (connect to wifi, plug in an ethernet cable) it'll affect both the OS routing table / default route, and the addresses libp2p believes it is listening on. It seems reasonable to want this view of 'our current source address' to reflect the OS's view of its default route.

p2p/host/basic/basic_host.go
Stebalien: Yes. E.g., we change networks.

p2p/host/basic/basic_host.go
Stebalien: > However, if the "explicit address" is a private address that is NOT the default route, it will get filtered.

That's my concern. I'm saying that we _shouldn't_ filter. At the moment, we're:

1. Resolving `/ip4/0.0.0.0/tcp/4001` to `/ip4/192.168.0.99/tcp/4001`, `/ip4/10.0.1.20/tcp/4001`, etc. in the swarm.
2. Retroactively filtering all non-localhost interface addresses _except_ the default source address.

Instead, we could just:

1. Call `ListenAddresses` to get the _exact_ listen addresses that the user used.
2. Resolve any instances of the "all" address (`0.0.0.0`) to the default source address.

This will:

1. Avoid doing a bunch of unnecessary work to _undo_ the work we're doing in `InterfaceListenAddresses`.
2. Ensure that, if the user _explicitly_ listens on an interface that _isn't_ the default one, that address will still be properly advertised.

aarshkshah1992: @Stebalien This really makes sense. Have made the change.

aarshkshah1992:
@Stebalien @willscott I think it looks much better now. Please take a look.

p2p/host/basic/basic_host.go
```diff
@@ -276,7 +315,28 @@ func New(net network.Network, opts ...interface{}) *BasicHost {
...
 func (h *BasicHost) Start() {
 	h.refCount.Add(1)
-	go h.background()
+	go h.listenAddrChangeLoop()
+
+	h.refCount.Add(1)
+	go h.localIpAddrChangeLoop()
+}
+
+func (h *BasicHost) localIpAddrChangeLoop() {
```
willscott, line 324: how painful would it be to combine this into the select loop for the addrChangeTicker and save the extra goroutine?

aarshkshah1992: @willscott The code will look a bit messy. This keeps things simple to read which I like.

aarshkshah1992:
@willscott 

> I wonder if we can get away with just running `updateLocalIPAddr` when `emitAddrChange` occurs. Presumably the irregular poll of `h.Addrs()`, which should pick up changed external addresses, will also vary in the same 5 minute interval as local addresses changed, so that might be a good enough signal directly, and remove the need for the extra event loop complexity.

Let me reason through this so I can clear my head as well:

The `emitAddrChange` event is emitted when `h.Addrs()` returns an address we haven't seen before. This can happen in two cases:

   * The user has explicitly started listening on a new address by calling `host.Network.Listen()`.
      * If the user pro-actively starts listening on the "new" local addresses, we are good here.
   * Our local IP addresses have changed and we were listening on an unspecified IP address(0.0.0.0) before. In this case `h.Addrs()` will replace the unspecified address with the new local address instead of the old one and we will detect an address change. 
     * However, you see, there is a catch-22 here. `h.Addrs()` needs to know the new local address to return them so that we can emit an addr change event. Hence, changing our local address when we emit an addr change event wont work here if the user dosen't explicitly call `Listen` for them.

aarshkshah1992:
ping @Stebalien for one last review.

p2p/host/basic/basic_host.go
Stebalien: I have to disagree. Chasing down the channel to figure out what it triggers is harder to read.

Can we just re-use the same ticker interval? Once every five seconds shouldn't be too inefficient. We can just call `updateLocalIpAddr` right before https://github.com/libp2p/go-libp2p/pull/936/files#diff-439e1c54aa40f8126fab7c895858fae3R492.

aarshkshah1992:
@Stebalien Have made the change.

Issue #939, opened 2020-05-19
Panic with signed peer records in go-lib2p v0.9.0


vyzo:
We upgraded to the latest go-libp2p in the pubsub testplans in `libp2p/test-plans` (see https://github.com/libp2p/test-plans/pull/7) and now identify no longer works.

The smoking gun is this panic:
```
2020-05-19T16:32:23.759Z	ERROR	net/identify	identify/id.go:432	latest peer record does not exist. identify message incomplete!
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x6dfb97]

goroutine 234 [running]:
github.com/libp2p/go-libp2p-core/record.(*Envelope).Marshal(0x0, 0xc0005b68e0, 0x8, 0x8, 0x203000, 0x2)
	/go/pkg/mod/github.com/libp2p/go-libp2p-core@v0.5.6/record/envelope.go:196 +0x37
github.com/libp2p/go-libp2p/p2p/protocol/identify.(*IDService).populateMessage(0xc0001f1040, 0xc0007780b0, 0xf83440, 0xc00048a720, 0xc000354380)
	/go/pkg/mod/github.com/libp2p/go-libp2p@v0.9.0/p2p/protocol/identify/id.go:469 +0x4ca
github.com/libp2p/go-libp2p/p2p/protocol/identify.(*IDService).sendIdentifyResp(0xc0001f1040, 0xf83f40, 0xc0003129c0)
	/go/pkg/mod/github.com/libp2p/go-libp2p@v0.9.0/p2p/protocol/identify/id.go:402 +0x39a
github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).SetStreamHandler.func1(0xc0001fe280, 0xe, 0x7f0c590e5198, 0xc0003129c0, 0x1, 0x0)
	/go/pkg/mod/github.com/libp2p/go-libp2p@v0.9.0/p2p/host/basic/basic_host.go:476 +0x9d
created by github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).newStreamHandler
	/go/pkg/mod/github.com/libp2p/go-libp2p@v0.9.0/p2p/host/basic/basic_host.go:327 +0x63c
```

yusefnapora:
[...]

PR #941, opened 2020-05-19
add test to demo missing peer records after listen

yusefnapora:
This shows off what I think is causing https://github.com/libp2p/go-libp2p/issues/939

The problem is that we're starting our hosts without listen addrs, and apparently the peerstore isn't keeping our signed records since they don't have any addrs. Then, when we call `.Listen`, it takes ~ 5 seconds for the change to be detected and the new record to be generated.

cc @aarshkshah1992 @raulk @vyzo

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -232,6 +232,16 @@ func NewHost(ctx context.Context, net network.Network, opts *HostOpts) (*BasicHo
 
 	net.SetStreamHandler(h.newStreamHandler)
 
+	// register to be notified when the network's listen addrs change,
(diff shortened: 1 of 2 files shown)


yusefnapora:
The last commits register `BasicHost` as a `network.Notifee` so it can get informed when we start or stop listening on an address and call `SignalAddressChanges`, which causes a new signed record to be created and added to the peerstore right away instead of having to wait for the 5 second ticker to detect the change.

It's still a little racy - on my laptop I only have to sleep for a millisecond after listening before I can get the new record out of the peerstore, but that's not quite long enough on Travis, so I bumped it to 20ms in the test. Still better than five seconds though :)

PR #948, opened 2020-05-20
fix: fix connection gater in transport constructor

Stebalien:


config/constructor_types.go
```diff
@@ -28,12 +28,12 @@ var (
 	privKeyType   = reflect.TypeOf((*crypto.PrivKey)(nil)).Elem()
 	pubKeyType    = reflect.TypeOf((*crypto.PubKey)(nil)).Elem()
 	pstoreType    = reflect.TypeOf((*peerstore.Peerstore)(nil)).Elem()
+	connGaterType = reflect.TypeOf((*connmgr.ConnectionGater)(nil)).Elem()
```
raulk, line 31: Note for myself.

The type you really want is:

```go
connGaterType = reflect.TypeOf((connmgr.ConnectionGater)(nil))  // dropping the pointer star
```

but you can’t do that because it’s syntactically illegal (can’t cast to an interface). So you create a pointer to the interface and then take the `Elem()` of the pointer value.

PR #949, opened 2020-05-20
fix: group observations by zeroing port

Stebalien:
In #917, we started dropping additional address observations if we had multiple for the same transport set. However, on further consideration, this isn't quite correct. We _want_ to keep additional observations for multiple IP addresses. The real issue is many observations for different ports.

So this patch simply changes the key with which we group observations from "address protocols" to "address without the port" (well, with the port set to 0).

### DIFF
--- a/p2p/protocol/identify/obsaddr.go
+++ b/p2p/protocol/identify/obsaddr.go
@@ -70,15 +70,24 @@ func (oa *ObservedAddr) numInbound() int {
 	return count
 }
 
+// GroupKey returns the group in which this observation belongs. Currently, an
+// observed address's group is just the address with all ports set to 0. This
+// means we can advertise the most commonly observed external ports without
+// advertising _every_ observed port.
 func (oa *ObservedAddr) GroupKey() string {
-	key := ""
-	protos := oa.Addr.Protocols()
+	key := make([]byte, 0, len(oa.Addr.Bytes()))
+	ma.ForEach(oa.Addr, func(c ma.Component) bool {
+		switch proto := c.Protocol(); proto.Code {
+		case ma.P_TCP, ma.P_UDP:
+			key = append(key, proto.VCode...)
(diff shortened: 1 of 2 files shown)


p2p/protocol/identify/obsaddr.go
```diff
@@ -70,15 +70,24 @@ func (oa *ObservedAddr) numInbound() int {
...
-
-	for i := range protos {
-		key = key + "/" + protos[i].Name
-	}
+	key := make([]byte, 0, len(oa.Addr.Bytes()))
+	ma.ForEach(oa.Addr, func(c ma.Component) bool {
+		switch proto := c.Protocol(); proto.Code {
+		case ma.P_TCP, ma.P_UDP:
+			key = append(key, proto.VCode...)
+			key = append(key, 0, 0) // zero in two bytes
```
aschmahmann, line 83: nit: Why are we adding in the zero bytes here instead of just skipping them? Is this just a defensive measure to keep them as valid multiaddrs?

Stebalien: You're right, we could just skip it. But yeah, I wanted these to be valid multiaddrs.

aarshkshah1992:
@Stebalien In what scenario would a peer have observed addresses with different IP addresses ?

p2p/protocol/identify/obsaddr.go
```diff
@@ -70,15 +70,24 @@ func (oa *ObservedAddr) numInbound() int {
...
+		default:
+			key = append(key, c.Bytes()...)
+		}
```
aarshkshah1992, line 86: @Stebalien 

Out of curiosity:

Does `string(c.Bytes)` save bytes compared to `c.String()` ?

Stebalien:
If they have multiple interfaces and/or multiple IP addresses. For example, they might be behind multiple NATs at the same time (I think this is the case for AWS).

Issue #950, opened 2020-05-20
Flaky TestIdentifyPushOnAddrChange


Stebalien:
```
--- FAIL: TestIdentifyPushOnAddrChange (0.03s)
    id_test.go:591: 
        	Error Trace:	id_test.go:591
        	Error:      	"[]" should have 1 item(s), but has 0
        	Test:       	TestIdentifyPushOnAddrChange
```

So, it looks like we're finishing the call to `Connect` before the other side sees the connection. However, that should be impossible at this point given that:

1. We wait for identify.
2. The other side won't respond to the identify request until after it has recorded the new connection.

So I'm not sure what's going on here.

Stebalien:
@aarshkshah1992 pointed out that we're using the blank host here, which _doesn't_ have these guarantees.

aarshkshah1992:
@Stebalien PR at https://github.com/libp2p/go-libp2p/pull/952.

aarshkshah1992:
Fixed in #952 .

PR #952, opened 2020-05-21
Fix flaky TestIdentifyPushOnAddrChange

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/950.


aarshkshah1992:
@Stebalien has approved.

PR #953, opened 2020-05-27
fix: remove old addresses in identify immediately

Stebalien:
Previously, we'd keep addresses discovered through the DHT for up to 2 minutes (temporary TTL) and previously seen addresses (recently connected) for up to 10 minutes (the TTL).

1. Make sure to downgrade both connected and recently connected addresses to the "temporary" ttl before adding new addresses.
2. Finally, downgrade addresses with the temporary TTL to 0.

This could be more efficient with a better peerstore abstraction, but this is better than nothing.

### DIFF
--- a/p2p/protocol/identify/id_test.go
+++ b/p2p/protocol/identify/id_test.go
@@ -58,8 +58,8 @@ func subtestIDService(t *testing.T) {
 	testKnowsAddrs(t, h2, h1p, []ma.Multiaddr{}) // nothing
 
 	// the forgetMe addr represents an address for h1 that h2 has learned out of band
-	// (not via identify protocol). Shortly after the identify exchange, it will be
-	// forgotten and replaced by the addrs h1 sends during identify
+	// (not via identify protocol). During the identify exchange, it will be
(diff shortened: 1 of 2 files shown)


p2p/protocol/identify/id_test.go
```diff
@@ -80,7 +80,7 @@ func subtestIDService(t *testing.T) {
 	// the IDService should be opened automatically, by the network.
 	// what we should see now is that both peers know about each others listen addresses.
 	t.Log("test peer1 has peer2 addrs correctly")
-	testKnowsAddrs(t, h1, h2p, h2.Peerstore().Addrs(h2p))        // has them
+	testKnowsAddrs(t, h1, h2p, h2.Addrs())                       // has them
 	testHasCertifiedAddrs(t, h1, h2p, h2.Peerstore().Addrs(h2p)) // should have signed addrs also
```
aschmahmann, line 84: I'm not sure I understand all the test changes. Why are we sometimes using `h.Addrs()` and sometimes using `h.Peerstore().Addrs(h.ID)`?

Issue #954, opened 2020-05-28
Identify message too big


Stebalien:
When users use 4096 bit RSA keys (like we do on one of our bootstrappers), the identify message exceeds 2048 bytes (by about 20 bytes) when the signed envelop is added into the mix. That is:

* 555 bytes for the key.
* ~275 bytes for the protocol list.
* ~12 bytes for the addresses
* 1024+ bytes for the signed record
  * ~555 bytes for the key again
  * ~512 bytes for the signature
  * ~12 bytes for the actual addresses.

We're missing a few bytes in this math but this paints the picture.

---

We should fix this in all of the following ways:

1. Quadruple the maximum message for all versions of identify to 8192 to be safe.
2. Switch to ed25519 by default ASAP because RSA keys are just way too big.
3. Split the identify protocol back into a version 1 and a version 2. Version 2 will:
    1. Support up to 8192 byte messages (guaranteed).
    2. Not duplicate the addresses.
    3. Critically, not duplicate the key. This last part will save us 555 bytes in this case.

Additionally, we should _not_ use version 2 until we need it as negotiating version 2, then falling back on version 1 adds a round-trip to new connections. We could also revisit #903 but that's _really_ hacky.

aarshkshah1992:
PR at https://github.com/libp2p/go-libp2p/pull/955.

vyzo:
This was discussed today. I think a reasonable solution that doesn't require a protocol upgrade would be to only send the signed peer record if the key is _not too big_ in the current protocol version. This will protect lotus/drand who depend on signed peer records for Peer Exchange through gossipsub.

Note that I am not arguing against protocol upgrade; that's probably desirable for other reasons. But the existing protocol can be mostly fixed and still provide signer peer records, except for large RSA keys.

Stebalien:
TL;DR:

Identify 1.0: Send signed record if small enough.
Identify 1.1: Always send signed record.

Peers would use identify 1 unless they absolutely _need_ the signed record. In that case, they'd use identify 1.1.

---

Alternatively, we could just keep the identify 1.0/1.1 split and add an option for lotus nodes to try 1.1 before 1.0. This would fix your concern @vyzo, right?

vyzo:
Yes, it would fix the concern. 
But really, we should _try_ to send the signed peer record if it fits in identify/1.0 so that users can get the benefit of signed peer records without having to do anything or end up not sending sprs to older peers, since the testnet is live and it will be a mixed deployment.

[...]

PR #955, opened 2020-05-29
Upgrade identify protocols with smaller size for Signed records

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/954.

@Stebalien 
Why don't we combine ID & ID Push while we are doing this ?


p2p/protocol/identify/id.go
```diff
@@ -352,14 +369,19 @@ func (ids *IDService) identifyConn(c network.Conn, signal chan struct{}) {
 		ids.removeConn(c)
 		return
 	}
-	s.SetProtocol(ID)
 
-	// ok give the response to our handler.
-	if err = msmux.SelectProtoOrFail(ID, s); err != nil {
+	// TODO This might cause us an extra roundtrip if the remote peer does not not support
```
aarshkshah1992, line 373: @Stebalien Either we "disable" `ID_1_1_0 ` for now(since we don't really need signed records anywhere) or we pay this cost. I don't there's any other clean solution till we have multistream2.0.

Lmk what you think.

Stebalien:
>  Why don't we combine ID & ID Push while we are doing this ?

ID -> who are you?
ID Push -> Hi, I'm X.

p2p/protocol/identify/id.go
vyzo: > (since we don't really need signed records anywhere) 

That's decisively not true. Both lotus and (especially) drand rely on signed peer records to bootstrap pubsub through Peer Exchange without a DHT.

aarshkshah1992: @vyzo I get what you mean. I'll make the configuration changed discussed in #954 .

Stebalien:
Superseeded in #958. This will be a non-issue when we switch to ed25519.

PR #958, opened 2020-06-02
Chunking large Identify responses with Signed Records

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/954#issuecomment-636789416.


p2p/protocol/identify/id.go
```diff
@@ -432,11 +443,34 @@ func (ids *IDService) getSnapshot() *identifySnapshot {
...
-	mes *pb.Identify,
+func (ids *IDService) writeChunkedIdentifyMsg(c network.Conn, snapshot *identifySnapshot, s network.Stream) error {
+	mes := ids.getIdentifyMsgUnsigned(c, snapshot)
+	sr := ids.getSignedRecord(snapshot)
+	mes.SignedPeerRecord = sr
+	writer := ggio.NewDelimitedWriter(s)
+
+	if sr == nil || proto.Size(mes) <= legacyIDSize {
+		return writer.WriteMsg(mes)
+	} else {
```
raulk, line 454: else superfluous

p2p/protocol/identify/id.go
```diff
@@ -432,11 +443,34 @@ func (ids *IDService) getSnapshot() *identifySnapshot {
...
+		mes.SignedPeerRecord = nil
+		mes.SignedRecordInNextMessage = true
```
raulk, line 456: Can we generalise this field to `more`? This is overly specific.

p2p/protocol/identify/id.go
```diff
@@ -432,11 +443,34 @@ func (ids *IDService) getSnapshot() *identifySnapshot {
 	return snapshot
 }
 
-func (ids *IDService) populateMessage(
-	mes *pb.Identify,
+func (ids *IDService) writeChunkedIdentifyMsg(c network.Conn, snapshot *identifySnapshot, s network.Stream) error {
+	mes := ids.getIdentifyMsgUnsigned(c, snapshot)
```
raulk, line 447: What is an _unsigned_ identify message? We don't sign identify messages...

raulk: Maybe a better name would be `createBaseIdentifyResponse`?

p2p/protocol/identify/pb/identify.proto
```diff
@@ -37,10 +39,15 @@ message Identify {
   // a delta update is incompatible with everything else. If this field is included, none of the others can appear.
   optional Delta delta = 7;
 
+  // signedRecordInNextMessage is set to true by a peer to indicate to the receiver that the signed record is absent in the current message
+  // and the peer should read the next message from the stream to fetch it.
+  // This is done because messages with signed records can exceed the legacy message size of 2K bytes.
+  optional bool signedRecordInNextMessage = 9 [(gogoproto.nullable) = false];
```
raulk, line 45: Let's call this `more` please!

p2p/protocol/identify/pb/identify.proto
```diff
@@ -37,10 +39,15 @@ message Identify {
...
+
   // signedPeerRecord contains a serialized SignedEnvelope containing a PeerRecord,
   // signed by the sending node. It contains the same addresses as the listenAddrs field, but
   // in a form that lets us share authenticated addrs with other peers.
   // see github.com/libp2p/go-libp2p-core/record/pb/envelope.proto and
   // github.com/libp2p/go-libp2p-core/peer/pb/peer_record.proto for message definitions.
   optional bytes signedPeerRecord = 8;
-}
+}
```
raulk, line 53: trailing newline removed

p2p/protocol/identify/id.go
```diff
@@ -395,28 +399,35 @@ func (ids *IDService) sendIdentifyResp(s network.Stream) {
...
-
+	ids.writeChunkedIdentifyMsg(c, ph.snapshot, s)
 	log.Debugf("%s sent message to %s %s", ID, c.RemotePeer(), c.RemoteMultiaddr())
 }
 
 func (ids *IDService) handleIdentifyResponse(s network.Stream) {
 	c := s.Conn()
 
-	r := ggio.NewDelimitedReader(s, 2048)
+	r := ggio.NewDelimitedReader(s, legacyIDSize)
```
raulk, line 409: This is weird. You should not create one delimited reader per message. One is enough, with buffer size `signedIDSize`. The second argument is the _capacity_ of the buffer.

p2p/protocol/identify/id.go
```diff
@@ -432,11 +443,34 @@ func (ids *IDService) getSnapshot() *identifySnapshot {
...
+		return writer.WriteMsg(mes)
+	} else {
+		mes.SignedPeerRecord = nil
+		mes.SignedRecordInNextMessage = true
+		if err := writer.WriteMsg(mes); err != nil {
+			return err
+		}
+
+		// then write just the signed record
+		m := &pb.Identify{SignedPeerRecord: sr}
```
raulk, line 462: Can't the signed peer record be `nil`?

aarshkshah1992: @raulk We already check for that at https://github.com/libp2p/go-libp2p/blob/3e1b364fefea98ba241e8774529bf5bb49a3cea7/p2p/protocol/identify/id.go#L452.

p2p/protocol/identify/id.go
```diff
@@ -395,28 +399,35 @@ func (ids *IDService) sendIdentifyResp(s network.Stream) {
...
+	r := ggio.NewDelimitedReader(s, signedIDSize)
 	mes := pb.Identify{}
 	if err := r.ReadMsg(&mes); err != nil {
 		log.Warning("error reading identify message: ", err)
 		s.Reset()
 		return
 	}
 
+	if mes.More {
+		m := &pb.Identify{}
```
vyzo, line 418: I don't think we need to make a new message here, just reuse the existing one and merge in the new fields.

p2p/protocol/identify/pb/identify.proto
```diff
@@ -37,6 +39,11 @@ message Identify {
   // a delta update is incompatible with everything else. If this field is included, none of the others can appear.
   optional Delta delta = 7;
 
+  // more is set to true by a peer to indicate to the receiver that the signed record is absent in the current message
+  // and the peer should read the next message from the stream to fetch it.
+  // This is done because messages with signed records can exceed the legacy message size of 2K bytes.
+  optional bool more = 9 [(gogoproto.nullable) = false];
```
vyzo, line 45: What's this nullable thing? I would advise against using gogo specific stuff in the protobuf decls, as there are other implementations in different languages that want to use the same protobuf file.

Stebalien: This is just an annotation to tell gogo how to generate the go code for this protobuf. Other protobuf implementations will ignore annotations they don't understand.

p2p/protocol/identify/id.go
Stebalien: This will also generalize this code to merge protocols, addresses, etc. We can write this as:

```go
func mergeMessages(r ggio.Reader, m proto.Message) error {
	for i := 0; i < maxMessages; i++ {
	  	switch err := r.ReadMsg(mes); err {
	    default:
		case io.EOF:
        	return nil
	    case nil:
			log.Warning("error reading identify message: ", err)
			s.Reset()
			return
		}
	}
    return fmt.Errorf("too many parts")
}
```

p2p/protocol/identify/id.go
```diff
@@ -395,28 +399,35 @@ func (ids *IDService) sendIdentifyResp(s network.Stream) {
...
-	r := ggio.NewDelimitedReader(s, 2048)
+	r := ggio.NewDelimitedReader(s, signedIDSize)
 	mes := pb.Identify{}
 	if err := r.ReadMsg(&mes); err != nil {
 		log.Warning("error reading identify message: ", err)
 		s.Reset()
 		return
 	}
 
+	if mes.More {
```
Stebalien, line 417: I'm pretty sure we don't need "more". We can just read till the end.

p2p/protocol/identify/id.go
aarshkshah1992: @Stebalien 

Had to use `proto.Merge(dst, src)` here after calling the goog reader because  the gogo varint reader uses `proto.Unmarshal` under the hood which resets(clears) the existing proto message before reading the new one into it.

Other than that, works well.

p2p/protocol/identify/id.go
aarshkshah1992: This has been removed now since merging the messages works well.

aarshkshah1992:
@Stebalien @vyzo Have made the changes. Please take a look.

vyzo:
flaky tests on travis is nothing new unfortunately, I restarted the build to see if it persists.

go.mod
```diff
@@ -25,7 +25,7 @@ require (
 	github.com/libp2p/go-libp2p-netutil v0.1.0
 	github.com/libp2p/go-libp2p-peerstore v0.2.4
 	github.com/libp2p/go-libp2p-secio v0.2.2
-	github.com/libp2p/go-libp2p-swarm v0.2.4
+	github.com/libp2p/go-libp2p-swarm v0.2.6-0.20200602101559-2b4925e207f2
```
vyzo, line 28: we'll need to update to release swarm before merging this.

p2p/protocol/identify/id.go
Stebalien: We can remove the more field now, correct? I've gone ahead and done that but I'd like to double check with you (we can revert and add it back if you left it there for a reason).

aarshkshah1992: @Stebalien Yes, we can remove it. Sorry, I forgot this bit.

Issue #961, opened 2020-06-04
Panic, seems that peerHandler is nil


ridenaio:
We got `go-libp2p v0.9.2` after updating  to `go-ipfs v0.5.2-0.20200601220152-44df41cb08bd`. Now we are receiving the panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x14acd81]

goroutine 83 [running]:
github.com/libp2p/go-libp2p/p2p/protocol/identify.(*peerHandler).start(0x0)
        /home/mikhail/go/pkg/mod/github.com/libp2p/go-libp2p@v0.9.2/p2p/protocol/identify/peer_loop.go:62 +0x51
github.com/libp2p/go-libp2p/p2p/protocol/identify.(*IDService).loop(0xc0004eba00)
        /home/mikhail/go/pkg/mod/github.com/libp2p/go-libp2p@v0.9.2/p2p/protocol/identify/id.go:232 +0xa7b
created by github.com/libp2p/go-libp2p/p2p/protocol/identify.NewIDService
        /home/mikhail/go/pkg/mod/github.com/libp2p/go-libp2p@v0.9.2/p2p/protocol/identify/id.go:144 +0x2c3
```

<details>
<summary>Version Information</summary>
<pre>
bazil.org/fuse v0.0.0-20200117225306-7b5117fecadc
cloud.google.com/go v0.53.0
cloud.google.com/go/bigquery v1.3.0
cloud.google.com/go/datastore v1.0.0
cloud.google.com/go/pubsub v1.1.0
cloud.google.com/go/storage v1.5.0
dmitri.shuralyov.com/app/changes v0.0.0-20180602232624-0a106ad413e3
dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9
dmitri.shuralyov.com/html/belt v0.0.0-20180602232347-f7d459c86be0
dmitri.shuralyov.com/service/change v0.0.0-20181023043359-a85b471d5412
dmitri.shuralyov.com/state v0.0.0-20180228185332-28bcc343414c
git.apache.org/thrift.git v0.0.0-20180902110319-2566ecd5d999
github.com/AndreasBriese/bbloom v0.0.0-20190825152654-46b345b51c96
github.com/BurntSushi/toml v0.3.1
github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802
[...]

PR #963, opened 2020-06-04
Fix peer handler panic bug caused by connection races

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/961.

@Stebalien  Let me know if this makes sense or if I'm missing something:

**a.** p1 has two connections to p2.
**b.** p1 sees both disconnections "simultaneously".
   **c.** Peerhandler removal requests for both get written to `rmPeerHandlerCh` and we start two go-routines to close the same handler, one for each request.
   **d.** When we process the request of the first go-routine written to `phClosedCh`, we are not connected to the peer and so delete the handler.
      **e**. However, when we process the request of the second go-routine in `phClosedCh`, we are connected to the peer(we got a new third connection, but there's handler for it _yet_) and so we call `start` on a nil handler.

The correct solution here is to NOT create/start the handler if it's nil. When the Identify request for that third connection is eventually "processed", we will have a new handler for it.

Using the same logic, `peerhandler.start()` should be idempotent as we could end up calling `start` on the same handler twice without closing it in between the `start()` calls.

### DIFF
--- a/p2p/protocol/identify/peer_loop.go
+++ b/p2p/protocol/identify/peer_loop.go
@@ -17,6 +17,7 @@ import (
 
 	ggio "github.com/gogo/protobuf/io"
 	ma "github.com/multiformats/go-multiaddr"
+	"github.com/tevino/abool"
 )
 
 var errProtocolNotSupported = errors.New("protocol not supported")
@@ -42,7 +43,7 @@ type peerHandler struct {
 	pushCh  chan struct{}
 	deltaCh chan struct{}
 
-	isRunning bool
+	isRunningAtomic *abool.AtomicBool
 }
 
 func newPeerHandler(pid peer.ID, ids *IDService) *peerHandler {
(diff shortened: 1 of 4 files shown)


Stebalien:
You're analysis looks correct. I have no idea how you figured this.

aarshkshah1992:
@Stebalien I think I've addressed your concern. Lmk.

p2p/protocol/identify/peer_loop.go
```diff
@@ -52,23 +55,31 @@ func newPeerHandler(pid peer.ID, ids *IDService) *peerHandler {
...
 
 func (ph *peerHandler) close() error {
-	ph.cancel()
-	ph.wg.Wait()
+	if ph.isRunningAtomic.SetToIf(true, false) {
+		ph.cancel()
+		ph.wg.Wait()
+	}
 	return nil
 }
```
Stebalien, line 84: We still need to _wait_ before writing to the "closed" channel. Otherwise, the second close will skip the wait.

aarshkshah1992:
Closing in favour of https://github.com/libp2p/go-libp2p/pull/965.

PR #990, opened 2020-08-11
Added parsing of IPv6 addresses for incoming mDNS requests

fouge:
I want to have P2P communication fully based on IPv6, with local discovery using mDNS.
I am a beginner with libp2p so don't hesitate to bring your insights.

### DIFF
--- a/p2p/discovery/mdns.go
+++ b/p2p/discovery/mdns.go
@@ -157,7 +157,7 @@ func (m *mdnsService) pollForEntries(ctx context.Context) {
 }
 
 func (m *mdnsService) handleEntry(e *mdns.ServiceEntry) {
-	log.Debugf("Handling MDNS entry: %s:%d %s", e.AddrV4, e.Port, e.Info)
+	log.Debugf("Handling MDNS entry: [IPv4 %s][IPv6 %s]:%d %s", e.AddrV4, e.AddrV6, e.Port, e.Info)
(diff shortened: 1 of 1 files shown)


lidel:
The idea SGTM, the new `libp2p/specs/discovery/mdns` (which afaik is not yet implemented in GO libs, but is in JS) lists both IPv4 and IPv6: https://github.com/libp2p/specs/blob/master/discovery/mdns.md#find-all-response

Someone familiar with the codebase needs to review this tho.

p2p/discovery/mdns.go
```diff
@@ -169,12 +169,27 @@ func (m *mdnsService) handleEntry(e *mdns.ServiceEntry) {
...
+	} else if e.AddrV6 != nil {
+		maddr, err = manet.FromNetAddr(&net.TCPAddr{
+			IP:   e.AddrV6,
+			Port: e.Port,
+		})
+		if err != nil {
+			log.Warning("Error parsing multiaddr from mdns entry: ", err)
+			return
+		}
+	} else {
```
Stebalien, line 191: We can deduplicate some of this code.

```go
var addr net.IP
if e.AddrV4 != nil {
    addr = e.AddrV4
} else if e.AddrV6 != nil {
    addr = e.AddrV6
} else {
    return some error
}

// construct the multiaddr
```

PR #995, opened 2020-08-19
fix tests

Stebalien:
1. Fix tests on go 1.15 by avoiding `string(i)`.
2. Fix the autorelay test by making relays announce fake "public" addresses. I'm not sure how this worked before.


Stebalien:
Merging as a test fix.

Issue #1002, opened 2020-09-10
filtersConnectionGater not processing inbound connections


aschmahmann:
https://github.com/libp2p/go-libp2p/blob/d4d6adff6e3260792cb4514c27368059f2558530/options_filter.go#L18-L36 returns true for everything except for outbound dials. This is a regression from previous behavior of the AddrFilters.

^^ @aarshkshah1992 @raulk

PR #1004, opened 2020-09-11
Fixed bug for inbound connections gated by the deprecated filter option

jalextowle:
Partially fixes: #1002

### DIFF
--- a/options_filter.go
+++ b/options_filter.go
@@ -23,8 +23,8 @@ func (f *filtersConnectionGater) InterceptPeerDial(p peer.ID) (allow bool) {
 	return true
 }
 
-func (f *filtersConnectionGater) InterceptAccept(_ network.ConnMultiaddrs) (allow bool) {
-	return true
+func (f *filtersConnectionGater) InterceptAccept(connAddr network.ConnMultiaddrs) (allow bool) {
+	return !(*ma.Filters)(f).AddrBlocked(connAddr.RemoteMultiaddr())
 }
 
 func (f *filtersConnectionGater) InterceptSecured(_ network.Direction, _ peer.ID, _ network.ConnMultiaddrs) (allow bool) {
--- a/options_filter.go
+++ b/options_filter.go
@@ -27,8 +27,8 @@ func (f *filtersConnectionGater) InterceptAccept(connAddr network.ConnMultiaddrs
 	return !(*ma.Filters)(f).AddrBlocked(connAddr.RemoteMultiaddr())
 }
 
-func (f *filtersConnectionGater) InterceptSecured(_ network.Direction, _ peer.ID, _ network.ConnMultiaddrs) (allow bool) {
-	return true
+func (f *filtersConnectionGater) InterceptSecured(_ network.Direction, _ peer.ID, connAddr network.ConnMultiaddrs) (allow bool) {
+	return !(*ma.Filters)(f).AddrBlocked(connAddr.RemoteMultiaddr())
 }
 
 func (f *filtersConnectionGater) InterceptUpgraded(_ network.Conn) (allow bool, reason control.DisconnectReason) {
--- a/options_test.go
+++ b/options_test.go
@@ -2,7 +2,6 @@ package libp2p
 
 import (
 	"context"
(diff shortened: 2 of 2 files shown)


options_test.go
```diff
@@ -70,3 +107,61 @@ func TestCannotSetFiltersAndConnGater(t *testing.T) {
 	require.Error(err)
 	require.Contains(err.Error(), "cannot configure multiple connection gaters")
 }
+
+/// NOTE(jalextowle): This was taken directly from https://github.com/0x-mesh/p2p/banner/banner.go
```
aarshkshah1992, line 111: @jalextowle 

You can use this instead:

https://github.com/multiformats/go-multiaddr/blob/d18c05e0e1635f8941c93f266beecaedd4245b9f/net/convert.go#L103

options_test.go
```diff
@@ -31,11 +32,47 @@ func TestDeprecatedFiltersOptions(t *testing.T) {
 	addr, _ := ma.NewMultiaddr("/ip4/127.0.0.1/tcp/0/p2p/" + id.Pretty())
 	ai, _ := peer.AddrInfoFromP2pAddr(addr)
 
-	err = host.Connect(ctx, *ai)
+	err = host0.Connect(ctx, *ai)
 	require.Error(err)
 	require.Contains(err.Error(), "no good addresses")
 }
 
+func TestDeprecatedFiltersOptionsInbound(t *testing.T) {
```
aarshkshah1992, line 40: @jalextowle Here's a reworked test:

```
func TestDeprecatedFiltersOptionsInbound(t *testing.T) {
	require := require.New(t)

	host0, err := New(context.TODO())
	require.NoError(err)
	require.NotNil(host0)

	f := ma.NewFilters()
	for _, addr := range host0.Addrs() {
		ip, err := manet.ToIP(addr)
		require.NoError(err)
		require.NotNil(t, ip)

		var mask net.IPMask
		if ip.To4() != nil {
			mask = ip4FullMask
		} else {
			mask = ip6FullMask
		}

		ipnet := net.IPNet{IP: ip, Mask: mask}
		f.AddFilter(ipnet, ma.ActionDeny)
	}
	host1, err := New(context.TODO(), Filters(f))
	require.NoError(err)
	require.NotNil(host1)

	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
	defer cancel()

	peerInfo := peer.AddrInfo{
		ID:    host1.ID(),
		Addrs: host1.Addrs(),
	}
	err = host0.Connect(ctx, peerInfo)
	require.Error(err)
}
```

aarshkshah1992:
@jalextowle LGTM modulo test refactor.

options_filter.go
```diff
@@ -23,8 +23,8 @@ func (f *filtersConnectionGater) InterceptPeerDial(p peer.ID) (allow bool) {
 	return true
 }
 
-func (f *filtersConnectionGater) InterceptAccept(_ network.ConnMultiaddrs) (allow bool) {
-	return true
+func (f *filtersConnectionGater) InterceptAccept(connAddr network.ConnMultiaddrs) (allow bool) {
+	return !(*ma.Filters)(f).AddrBlocked(connAddr.RemoteMultiaddr())
```
aschmahmann, line 27: @aarshkshah1992 is this the only place to do this? Don't we need something for at least InterceptSecured if not InterceptUpgraded also?

aarshkshah1992: @aschmahmann 

An `InterceptUpgraded` wont do match for us here as all those rejections will happen in `InterceptAccepted` also since we didn't have the notion of rejecting connections at different "lifecycle states" when we had this old filter.

However, there is no harm in doing so and I am okay either ways.

aarshkshah1992: Actually @aschmahmann , we might need this for QUIC. Thanks for catching this.

@jalextowle Please can you add a similar behaviour to `InterceptSecured` as well ?

jalextowle: @aarshkshah1992 Yep, I added that behavior.

aarshkshah1992 (approved):

PR #1009, opened 2020-10-26
Use of specific network interface for mDNS server

fouge:
I created this PR because I had to adapt the mDNS service to be used on specific interfaces.
This can be useful for others as well.

The usage of a specific network interface is not mandatory, `nil` can be passed as well.

Let me know what you think or how can this PR can be improved :)


aschmahmann:
This looks good to me. @aarshkshah1992 you've been looking at these things recently, any thoughts/concerns?

p2p/discovery/mdns.go
```diff
@@ -131,18 +155,34 @@ func (m *mdnsService) pollForEntries(ctx context.Context) {
 			}
 		}()
 
-		log.Debug("starting mdns query")
+		// get interface from name, if any, and verify the interface is up before sending a new query
+		var iface *net.Interface
+		var err error
+		if m.ifaceName != "" {
+			iface, err = net.InterfaceByName(m.ifaceName)
```
aarshkshah1992, line 162: Shouldn't we do this at the top of this function and get `NewMdnsService ` to return an error if there is no such interface on the host ?

aarshkshah1992 (commented):
@fouge Thanks for the PR.

I've posted my review.

p2p/discovery/mdns.go
```diff
@@ -13,7 +14,7 @@ import (
 
 	logging "github.com/ipfs/go-log"
 	ma "github.com/multiformats/go-multiaddr"
-	manet "github.com/multiformats/go-multiaddr-net"
+	manet "github.com/multiformats/go-multiaddr/net"
```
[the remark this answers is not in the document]
aschmahmann: The older repo was deprecated and the code was moved into multiaddr/net. Rebasing on go-libp2p master should take care of this, but this change is fine.

p2p/discovery/mdns.go
```diff
@@ -69,12 +88,11 @@ func getDialableListenAddrs(ph host.Host) ([]*net.TCPAddr, error) {
 	return out, nil
 }
 
-func NewMdnsService(ctx context.Context, peerhost host.Host, interval time.Duration, serviceTag string) (Service, error) {
-
+func NewMdnsService(ctx context.Context, peerhost host.Host, interval time.Duration, serviceTag string, iface *net.Interface) (Service, error) {
```
aschmahmann, line 91: This is a breaking change. Seems like we could avoid this using options varargs (https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis). WDYT @aarshkshah1992

aschmahmann: Ideally if we went the option varargs route the interval would not be mandatory so if we're going to be doing a breaking change we should probably do that one.

aarshkshah1992: @aschmahmann Agreed on both points.

p2p/discovery/mdns.go
```diff
@@ -58,9 +61,25 @@ func getDialableListenAddrs(ph host.Host) ([]*net.TCPAddr, error) {
...
-		tcp, ok := na.(*net.TCPAddr)
-		if ok {
-			out = append(out, tcp)
+
+		// if `iface` is specified, use address from `iface`
+		// otherwise get any TCP-capable address.
+		if iface != nil {
+			ips, _ := iface.Addrs()
+			for _, i := range ips {
+				if strings.Contains(na.String(), strings.Split(i.String(), "/")[0]) {
```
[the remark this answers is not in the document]
fouge: Actually, `i` is an `IPNet`. 
Here is the doc: 
```go
// String returns the CIDR notation of n like "192.0.2.0/24"
// or "2001:db8::/48" as defined in RFC 4632 and RFC 4291.
// If the mask is not in the canonical form, it returns the
// string which consists of an IP address, followed by a slash
// character and a mask expressed as hexadecimal form with no
// punctuation like "198.51.100.0/c000ff00".
func (n *IPNet) String() string
``` 
I am not sure I understand the goal of checking the Network as the tcp address is appended anyway?
Regarding the network part, when I run on my workstation, `i.Network()` returns `"ip+net"`, while `na.Network` returns `"tcp"`.

p2p/discovery/mdns.go
fouge: I thought about this but then I wondered how it would behave if the interface is turned down during execution...
That's why I decided to get the interface in the loop to make sure the `iface` is always up to date.

What do you think? 

Regarding checking the interface name from `NewMdnsService`, I assume the interface exists when the function is called as a pointer is passed: `iface *net.Interface`. But that can indeed be added.

p2p/discovery/mdns.go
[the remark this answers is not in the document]
aarshkshah1992: `iface ` is a `net.Interface`. The Addrs() call on it looks like:

```
func (ifi *Interface) Addrs() ([]Addr, error) {
```

Where `Addr` is

```
type Addr interface {
	Network() string // name of the network (for example, "tcp", "udp")
	String() string  // string form of address (for example, "192.0.2.1:25", "[2001:db8::1]:80")
}
```

Am I missing something here ?

Given this API, I'm saying we should compare the networks because `na` could be a UDP address for example and `i.String()` could return the same host:port but say for a TCP transport in which case, we shouldn't be using it.

fouge:
Thanks guys for taking the time to review this PR. I am quite new to Go so even if the language is simple to learn, those comments are really appreciated to level up my dev skills in Go :smiley: 
I am going to consider @aschmahmann solution to keep the `NewMdnsService` signature as before.

p2p/discovery/mdns.go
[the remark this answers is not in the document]
fouge: :thinking: 
`i` is an `net.IPNet` which doesn't seem to respect the Addr interface documentation, while `na` is `net.TCPAddr`.
```go
log.Info("i string: ", i.(net.Addr).String())
log.Info("i network: ", i.(net.Addr).Network())
log.Info("na string: ", na.(net.Addr).String())
log.Info("na network: ", na.(net.Addr).Network())
```
```
2020-11-16T16:25:24.760+0100    INFO    mdns    i string: 192.168.1.4/24
2020-11-16T16:25:24.760+0100    INFO    mdns    i network: ip+net
2020-11-16T16:25:24.760+0100    INFO    mdns    na string: [::1]:43931
2020-11-16T16:25:24.760+0100    INFO    mdns    na network: tcp
```

fouge:
@aschmahmann @aarshkshah1992 I implemented the optional functions to set mDNS service parameters.
Let me know what you think

p2p/discovery/mdns.go
aarshkshah1992: @fouge 

IMO, when `NewMdnsService ` is called with a non nil interface and no such interface exists on the machine, we should probably return an error. 

Are you thinking of a scenario where a new interface would be added to the machine after MDNS starts ? That dosen't sound like it'd happen often.

fouge: I agree about that point: when the service starts, the interface should exist. I have added a check when the service is created (in `NewMdnsService`) using the interface name, as it's now an interface name that is passed instead of a pointer to the interface (since I changed with optional functions to 

Should we keep the `iface` instantiation in the loop? 
I think this is more resilient.

aarshkshah1992 (commented):
@aschmahmann 

LGTM !

fouge:
@aschmahmann aarshkshah1992
Anything I should do to merge this PR?

PR #1029, opened 2020-12-08
fix: obey new stream timeout

Stebalien:


p2p/host/basic/basic_host.go
```diff
@@ -631,10 +631,24 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
...
 
-	selected, err := msmux.SelectOneOf(pidStrings, s)
-	if err != nil {
+	// Negotiate the protocol in the background, obeying the context.
+	var selected string
+	errCh := make(chan error, 1)
+	go func() {
+		selected, err = msmux.SelectOneOf(pidStrings, s)
+		errCh <- err
+	}()
```
raulk, line 640: Don't like the extra goroutine, but it's gotta be done.

p2p/host/basic/basic_host_test.go
```diff
@@ -777,6 +778,49 @@ func TestHostAddrChangeDetection(t *testing.T) {
...
+		t.Fatal(err)
+	case <-time.After(10 * time.Millisecond):
+		// ok, hung.
+	}
+	cancel2()
+
+	select {
+	case err := <-errCh:
+		require.Equal(t, err, context.Canceled)
+	case <-time.After(10 * time.Millisecond):
```
raulk, line 818: Can we increase the timeout to 500ms to be on the safe side? The cancellation happens asynchronously, and 10ms might be a bit tight in some cases.

raulk (approved):

p2p/host/basic/basic_host.go
Stebalien: Ideally, we'd pass the context down. But we'd need to goroutine at _some_ level, eventually.

p2p/host/basic/basic_host_test.go
Stebalien: 10ms is a _long_ time for local operations. I'd err on the side of fixing this later if it's an issue, and keeping the tests fast for now. (but I don't feel strongly either way)

raulk: SGTM, but flaky tests cause frustration. I'd rather have an (almost) imperceptible lag and never have to worry about it, but fine to keep it as is too and going your way. It's really a nit.

Stebalien: Ok, I'll split the difference.

Stebalien: (100ms)

Stebalien: Oh, wait. Nvm. 500ms is the _timeout_ (it won't take 500ms in the happy path). Changed.

PR #1032, opened 2020-12-17
Timeout all Identify stream reads

aarshkshah1992:


p2p/protocol/identify/id_push.go
```diff
@@ -13,5 +14,7 @@ const IDPush = "/ipfs/id/push/1.0.0"
 
 // pushHandler handles incoming identify push streams. The behaviour is identical to the ordinary identify protocol.
 func (ids *IDService) pushHandler(s network.Stream) {
+	s.SetReadDeadline(time.Now().Add(StreamReadTimeout))
+
 	ids.handleIdentifyResponse(s)
-}
+}
```
Stebalien, line 20: ^^

p2p/protocol/identify/id_push.go
```diff
@@ -13,5 +14,7 @@ const IDPush = "/ipfs/id/push/1.0.0"
 
 // pushHandler handles incoming identify push streams. The behaviour is identical to the ordinary identify protocol.
 func (ids *IDService) pushHandler(s network.Stream) {
+	s.SetReadDeadline(time.Now().Add(StreamReadTimeout))
+
 	ids.handleIdentifyResponse(s)
```
Stebalien, line 19: Don't we already set a read deadline in this function?

p2p/protocol/identify/id_delta.go
```diff
@@ -15,6 +16,8 @@ const IDDelta = "/p2p/id/delta/1.0.0"
 
 // deltaHandler handles incoming delta updates from peers.
 func (ids *IDService) deltaHandler(s network.Stream) {
+	s.SetReadDeadline(time.Now().Add(StreamReadTimeout))
```
Stebalien, line 19: I'd prefer to explicitly ignore the error.

p2p/protocol/identify/id.go
```diff
@@ -408,7 +414,9 @@ func (ids *IDService) sendIdentifyResp(s network.Stream) {
 	log.Debugf("%s sent message to %s %s", ID, c.RemotePeer(), c.RemoteMultiaddr())
 }
 
-func (ids *IDService) handleIdentifyResponse(s network.Stream) {
+func (ids *IDService) handleIdentifyResponse(s network.Stream) error {
```
Stebalien, line 417: why are we returning an error now? There isn't much we can do about it and we appear to always ignore it.

p2p/protocol/identify/id.go
```diff
@@ -369,10 +372,13 @@ func (ids *IDService) identifyConn(c network.Conn, signal chan struct{}) {
 		s.Reset()
 		return
 	}
-	ids.handleIdentifyResponse(s)
+
+	err = ids.handleIdentifyResponse(s)
 }
 
 func (ids *IDService) sendIdentifyResp(s network.Stream) {
+	s.SetDeadline(time.Now().Add(StreamReadTimeout))
```
Stebalien, line 380: nit: using the read timeout here is a bit strange, this mostly applies to writes.

marten-seemann (commented):
Depending on which PR gets merged first (this one or #1033), we might need to set the context for `OpenStream` here.

p2p/protocol/identify/id.go
aarshkshah1992: @Stebalien 

This is to ensure that we publish an `IdentifyCompleted` only if the stream read was successful. If the stream read fails, say because of a timeout, we want to publish an `EvtPeerIdentificationFailed` event.

p2p/protocol/identify/id.go
aarshkshah1992: I agree, had put in just in case. Have removed it.

p2p/protocol/identify/id_delta.go
aarshkshah1992: This is done.

p2p/protocol/identify/id_push.go
aarshkshah1992: Have fixed this.

aarshkshah1992:
@Stebalien Have pushed the changes.

p2p/protocol/identify/id.go
Stebalien: Ah, I see. We should be returning an error from consumeMessage in that case.

Stebalien (approved):

Issue #1030, opened 2020-12-16
Disconnected context in the identify protocol


MichaelMure:
Unless I'm missing something, the identify protocol should inherit from the host's context, which would then ensure that its main loop end when the host is terminated.

At the moment this main loop is left running after the host's context get cancelled.

See https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/identify/id.go#L130

MichaelMure:
Alternatively and similar to https://github.com/libp2p/go-libp2p/issues/1031, `Close()` is not called properly when the host is terminated. What I found is that this goroutine is left hanging after the host's context get cancelled.

aarshkshah1992:
@MichaelMure We shut down the `Identify` service when the Host is shut down over here:

https://github.com/libp2p/go-libp2p/blob/3155ff3089c0ab5ce2af49b2501600014b6c4816/p2p/host/basic/basic_host.go#L1003

The call to `Identify.Close()` will cancel the context that Identify's loop waits for:
https://github.com/libp2p/go-libp2p/blob/3155ff3089c0ab5ce2af49b2501600014b6c4816/p2p/protocol/identify/id.go#L276

Let me know if you have concerns otherwise  otherwise I can close this.

MichaelMure:
Here is what I see:

```go
func TestFoo(t *testing.T) {
	defer goleak.VerifyNone(t)

	ctx, cancel := context.WithCancel(context.Background())

	_, err := libp2p.New(ctx,
		libp2p.ListenAddrStrings("/ip4/0.0.0.0/tcp/1234"),
	)
	require.NoError(t, err)

	time.Sleep(10 * time.Second)

	cancel()

	time.Sleep(10 * time.Second)
}
```

```
=== RUN   TestFoo
    leaks.go:78: found unexpected goroutines:
        [Goroutine 21 in state select, with github.com/ipfs/go-log/writer.(*MirrorWriter).logRoutine on top of the stack:
        goroutine 21 [select]:
        github.com/ipfs/go-log/writer.(*MirrorWriter).logRoutine(0xc0000b5e30)
        	/home/michael/go/pkg/mod/github.com/ipfs/go-log@v1.0.4/writer/writer.go:71 +0x128
        created by github.com/ipfs/go-log/writer.NewMirrorWriter
        	/home/michael/go/pkg/mod/github.com/ipfs/go-log@v1.0.4/writer/writer.go:36 +0xb9
        
[...]

PR #1037, opened 2021-01-11
Close peerstore and document Host Close

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/1030.

### DIFF
--- a/libp2p.go
+++ b/libp2p.go
@@ -52,7 +52,7 @@ func ChainOptions(opts ...Option) Option {
 // - If no peerstore is provided, the host is initialized with an empty
 // peerstore.
 //
-// Canceling the passed context will stop the returned libp2p node.
+// Canceling the passed context followed by calling `Close` on the returned Host will stop the returned libp2p node.
 func New(ctx context.Context, opts ...Option) (host.Host, error) {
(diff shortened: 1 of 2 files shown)


MichaelMure:
LGTM :+1:

aarshkshah1992:
ping @Stebalien

libp2p.go
```diff
@@ -52,7 +52,7 @@ func ChainOptions(opts ...Option) Option {
 // - If no peerstore is provided, the host is initialized with an empty
 // peerstore.
 //
-// Canceling the passed context will stop the returned libp2p node.
```
raulk, line 55: Is this the only correct order to call these actions? What happens if we call Close before we cancel the context?

aarshkshah1992: @raulk The order is NOT important. I've updated the PR.

aarshkshah1992:
ping @Stebalien @raulk  for approval.

Stebalien (approved):
I seem to remember that we did it this way because we don't "own" the peerstore, it may be shared. However, closing this is probably strictly better than what we currently do.

Nit: I'd use:

```go
if pstore := h.Peerstore(); pstore != nil {
    pstore.Close()
}

Issue #1039, opened 2021-01-21
Project Flare(decentralised Hole Punching) Phase1 Meta Issue


aarshkshah1992:
This issue is to aggregate all the work and PRs we have open for taking Hole Punching to completion.

Done criteria: we recommend key consumers like go-ipfs and lotus enable it by default because the functionality works, is well tested, and has quality assertions in place to prevent regressions.

---
### Solve Simultaneous connect in multi-stream for TCP hole Punching (Ready for Review)

`go-multistream` suffers from the simultaneous connect problem wherein it fails when two peers try to connect each other at the same time. This blocks hole punching. This is ONLY a problem for TCP and NOT for QUIC because QUIC does NOT use the multistream protocol for negotiating the security and stream negotiation protocols. The PRs below solve this problem using the method specd out in https://github.com/libp2p/specs/pull/196.

- [x] https://github.com/multiformats/go-multistream/pull/42. (released)
- [x] https://github.com/libp2p/go-libp2p-core/pull/180 (released)
- [x] https://github.com/libp2p/go-conn-security-multistream/pull/14. (released)
- [x] https://github.com/libp2p/go-libp2p-transport-upgrader/pull/25. (released)
- [x] https://github.com/libp2p/go-libp2p/pull/1041. (**merged**)
---
### Swarm & Transport changes for Hole-Punching
- [x] https://github.com/libp2p/go-libp2p-core/pull/175 (support for transient Conns) 
- [x] https://github.com/libp2p/go-libp2p-swarm/pull/236 (transient conns in Swarm) 
- [x] https://github.com/libp2p/go-libp2p-transport-upgrader/pull/71  (upgrader changes for transient conns)
- [x] https://github.com/libp2p/go-libp2p-core/pull/181 (context to force a direct dial) (**merged NOT released**)
- [x] https://github.com/libp2p/go-libp2p-swarm/pull/233  (force and prefer direct conns)
---
### Emit event for NAT device type
- [x] https://github.com/libp2p/go-libp2p-core/pull/173 (define event)
- [x] https://github.com/libp2p/go-libp2p/pull/1042 (emit event)

---
### Limited Relay Protocol
- [x] Circuit v2 Spec https://github.com/libp2p/specs/pull/325
- [x] https://github.com/libp2p/go-libp2p-circuit/pull/125
- [x] https://github.com/vyzo/libp2p-relay
---
### Hole-Punching Co-ordination via Relay Server
[...]

PR #1042, opened 2021-01-27
Emit event for User's NAT Type i.e. Hard NAT or Easy NAT

aarshkshah1992:

We should emit an event informing the user of the user's NAT type for better user experience for hole punching/testing Hole punching in the wild.

### TODO
- [ ] Merge https://github.com/libp2p/go-libp2p-core/pull/173 AND Update deps.


vyzo (commented):
I am a little sceptical about the activated address heuristic; if we have one, then yes we are obviously behind an easy NAT.
But if we don't we might still be behind a penetrable NAT, especially with UDP.

aarshkshah1992:
@vyzo 

We don't determine the NAT to be  hard simply if we don't have an activated address. We determine the NAT to be hard when four different peers report for different observed addresses on outbound connections. I think this should be good enough to conclude that the NAT is bad, no ?

We can augment the code to report NAT types separately for UDP & non UDP if that's what you mean.

vyzo:
Yeah, we should probably do that.

p2p/protocol/identify/obsaddr.go
```diff
@@ -432,27 +434,89 @@ func (oas *ObservedAddrManager) recordObservationUnlocked(conn network.Conn, obs
...
 	var allObserved []*observedAddr
 	for k := range oas.addrs {
 		allObserved = append(allObserved, oas.addrs[k]...)
 	}
 
+	oas.emitTCPNATTypeUnlocked(allObserved)
+	oas.emitUDPTypeUnlocked(allObserved)
+}
+
+func (oas *ObservedAddrManager) emitTCPNATTypeUnlocked(addrs []*observedAddr) {
```
willscott, line 453: this and the subsequent `emitUDPNATTypeUnlocked` are doing basically the same thing. can you factor out a common method that takes an additional parameter (`ma.P_TCP` vs `ma.P_UDP`)?

p2p/protocol/identify/obsaddr.go
```diff
@@ -432,27 +434,89 @@ func (oas *ObservedAddrManager) recordObservationUnlocked(conn network.Conn, obs
 	oas.addrs[localString] = append(oas.addrs[localString], oa)
 }
 
-// 1. If we have an activated address, we are behind an easy NAT.
+// For a given transport Protocol (TCP/UDP):
+//
+// 1. If we have an activated address, we are behind an Cone NAT.
```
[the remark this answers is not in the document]
aarshkshah1992: @willscott 

As per the taxonomy, it can be a Full-Cone NAT, Port Restricted Cone NAT or a Restricted Cone NAT:

https://dh2i.com/kbs/kbs-2961448-understanding-different-nat-types-and-hole-punching/

I've combined all three into a "Cone NAT".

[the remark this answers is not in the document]
willscott: haven't heard that term before, and found it unclear. expand comment to list all 3 possibilities?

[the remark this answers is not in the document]
aarshkshah1992: @willscott 

They're defined in RFC 3489. I've updated the docs here with more details and have listed all 3 possibilities.

p2p/protocol/identify/obsaddr.go
aarshkshah1992: This is done.

aarshkshah1992:
@willscott I've addressed both your comments. Thanks !

p2p/protocol/identify/obsaddr.go
```diff
@@ -418,6 +434,87 @@ func (oas *ObservedAddrManager) recordObservationUnlocked(conn network.Conn, obs
...
+// With regards to RFC 3489, this could be either a Full Cone NAT, a Restricted Cone NAT or a
+// Port Restricted Cone NAT. However, we do NOT differentiate between them here and simply classify all such NATs as a Cone NAT.
+//
+// 2. If four different peers observe a different address for us on outbound connections, we
+// are MOST probably behind a Symmetric NAT.
+//
+// Please see the documentation on the enumerations for `network.NATDeviceType` for more details about these NAT Device types
+// and how they relate to NAT traversal via Hole Punching.
+func (oas *ObservedAddrManager) emitNATTypeUnlocked() {
```
willscott, line 448: why is this method named `Unlocked`?  maybe this is `observeAllNATTypes` and the next one is `observeSpecificNATType`? the naming of the two feels not quite representative of what they do.

willscott (approved):
one minor nit, and you need to get the dependencies merged and go.mod pulled up to versions before merging, but this seems ready.

aarshkshah1992:
cc @vyzo

p2p/protocol/identify/obsaddr.go
vyzo: yeah what's with the name? Is there a locked variant?

p2p/protocol/identify/obsaddr.go
```diff
@@ -418,6 +434,87 @@ func (oas *ObservedAddrManager) recordObservationUnlocked(conn network.Conn, obs
...
+
+	hasChanged, natType = oas.emitNATType(allObserved, ma.P_UDP, event.NATTransportUDP, oas.currentUDPNATDeviceType)
+	if hasChanged {
+		oas.currentUDPNATDeviceType = natType
+	}
+}
+
+// returns true along with the new NAT device type if the NAT device type for the given protocol has changed.
+// returns false otherwise.
+func (oas *ObservedAddrManager) emitNATType(addrs []*observedAddr, protoCode int, transportProto event.NATTransportProtocol,
```
vyzo, line 467: This could range from a little to _very_ expensive. 
How often do we call it?
Can we skip it altogether if we are not NATted?

p2p/protocol/identify/obsaddr.go
aarshkshah1992: Have fixed this.

p2p/protocol/identify/obsaddr.go
aarshkshah1992: @vyzo 


1. What makes this expensive ?
2. We call it every-time we have a new observed address but this function will return as soon as it sees any activated address.

The problem with skipping it altogether if we are not NATT'd is that AutoNAT dosen't take into account TCP vs UDP whereas we want to emit separate events for both.

Do you mean to say that we should emit this event for both TCP & UDP ONLY if reachability is private ?

vyzo: Imagine a big public peer with thousands of connections, this could become problematic.

And yes, this event is only relevant if we have private reachability and can/should be skipped altogether if we are public.

aarshkshah1992: done.

aarshkshah1992:
@vyzo 

1. Addressed your review.
2. Rebased & updated deps.

Please can you take a look ?

p2p/protocol/identify/obsaddr.go
```diff
@@ -232,10 +256,19 @@ func (oas *ObservedAddrManager) worker(ctx context.Context) {
...
 	hostClosing := oas.host.Network().Process().Closing()
+	subChan := oas.reachabilitySub.Out()
 	for {
 		select {
+		case evt, ok := <-subChan:
+			if !ok {
+				subChan = nil
+			}
+			ev := evt.(event.EvtLocalReachabilityChanged)
+			oas.reachability = ev.Reachability
```
vyzo, line 267: this will panic when the channel gets closed, as evt will be nil.

p2p/protocol/identify/obsaddr.go
```diff
@@ -418,6 +454,87 @@ func (oas *ObservedAddrManager) recordObservationUnlocked(conn network.Conn, obs
...
+// 2. If four different peers observe a different address for us on outbound connections, we
+// are MOST probably behind a Symmetric NAT.
+//
+// Please see the documentation on the enumerations for `network.NATDeviceType` for more details about these NAT Device types
+// and how they relate to NAT traversal via Hole Punching.
+func (oas *ObservedAddrManager) emitAllNATTypes() {
+	var allObserved []*observedAddr
+	for k := range oas.addrs {
+		allObserved = append(allObserved, oas.addrs[k]...)
+	}
```
vyzo, line 472: Index again when you mean to iterate.... It's more idiomatic to write this as
```
for _, addrs := range oas.addrs {
  allObserved = append(allObserved, addrs...)
}
```
```

Comment (vyzo) [reply p2p/protocol/identify/obsaddr.go:472]: More pertinent however, do we really need this copy?

File: p2p/protocol/identify/obsaddr.go [hunk 7]
```diff
@@ -418,6 +454,87 @@ func (oas *ObservedAddrManager) recordObservationUnlocked(conn network.Conn, obs
...
+// returns true along with the new NAT device type if the NAT device type for the given protocol has changed.
+// returns false otherwise.
+func (oas *ObservedAddrManager) emitSpecificNATType(addrs []*observedAddr, protoCode int, transportProto network.NATTransportProtocol,
+	currentNATType network.NATDeviceType) (bool, network.NATDeviceType) {
+	now := time.Now()
+	seenBy := make(map[string]struct{})
+	cnt := 0
+
+	for i := range addrs {
+		oa := addrs[i]
```
vyzo, line 494: and here again:
```
for _, oa := range addrs {
...
}
```

p2p/protocol/identify/obsaddr.go
```diff
@@ -418,6 +454,87 @@ func (oas *ObservedAddrManager) recordObservationUnlocked(conn network.Conn, obs
...
+			}
+
+			// our current NAT Device Type is already CONE, nothing to do here.
+			return false, 0
+		}
+
+		// An observed address on an outbound connection that has ONLY been seen by one peer
+		if now.Sub(oa.lastSeen) <= oas.ttl && oa.numInbound == 0 && len(oa.seenBy) == 1 {
+			cnt++
+			for s, _ := range oa.seenBy {
```
vyzo, line 517: you don't need the `, _`.

vyzo (commented):
One bug and a bunch of stylistic issues...

p2p/protocol/identify/obsaddr.go
aarshkshah1992: done.

p2p/protocol/identify/obsaddr.go
aarshkshah1992: done.

p2p/protocol/identify/obsaddr.go
aarshkshah1992: oops :/ Fixed.

p2p/protocol/identify/obsaddr.go
aarshkshah1992: well, it's probably a micro-optimization to fix the copy.  Have fixed the looping .

aarshkshah1992 (commented):
@vyzo 

Have made the changes.

go.mod
```diff
@@ -3,7 +3,9 @@ module github.com/libp2p/go-libp2p
 go 1.12
 
 require (
-	github.com/gogo/protobuf v1.3.2
+	github.com/gogo/protobuf v1.3.1
+	github.com/google/gopacket v1.1.18 // indirect
+	github.com/google/uuid v1.1.2 // indirect
```
vyzo, line 8: what's that crap popping up? Do a `go mod tidy`.

p2p/protocol/identify/obsaddr.go
vyzo: no you haven't, it's still there.

vyzo (commented):
needs a `go mod tidy`; also, fix dat loop!

go.mod
aarshkshah1992: Even a go mod tidy dosen't get rid of it.

```
aarshshah@Aarshs-MacBook-Pro go-libp2p % go mod why -m github.com/google/uuid
# github.com/google/uuid
github.com/libp2p/go-libp2p/p2p/host/routed
github.com/libp2p/go-libp2p-loggables
github.com/google/uuid
aarshshah@Aarshs-MacBook-Pro go-libp2p % go mod why -m github.com/google/gopacket
# github.com/google/gopacket
github.com/libp2p/go-libp2p/p2p/host/basic
github.com/libp2p/go-netroute
github.com/google/gopacket/routing
```

aarshkshah1992:
@vyzo This is ready.

vyzo (approved):
alright, let's merge this.

PR #1046, opened 2021-02-03
Event Driven Hole Punching

aarshkshah1992:

If either peer is behind a Symmetric NAT, there's no point in wasting a hole punch.


willscott:
Isn't the logic that if both peers are behind a symmetric nat then hole punching won't work?  as long as one of the two is behind a cone NAT, the dial from the other should get in and be able to turn into a direct connection, unless i'm not understanding something

aarshkshah1992:
@willscott 

The problem is that the side with the Cone NAT dosen't know which which address to send the packet to. The address the STUN server gives it for the remote peer wont work as that address will not be used by the Symmetric NAT peer when it punches a hole for the Easy NAT peer.

willscott:
> @willscott
> 
> The problem is that the side with the Cone NAT dosen't know which which address to send the packet to. The address the STUN server gives it for the Symmetric NAT peer wont work as that address will not be used by the Symmetric NAT peer when it punches a hole for the Easy NAT peer(five tuple failure).

and what about the other direction? (the symmetric NAT peer dialing the Cone NAT Peer)

aarshkshah1992:
@willscott 

The hole that the Cone NAT punches for the Symmetric NAT off the address it gets from the STUN server S wont work because when the Symmetric NAT's packet comes in, the source address on it will be different than the address the STUN server told it about (different destination = different source address on Symmetric NATs).

willscott:
> @willscott
> 
> The hole that the Cone NAT punches for the Symmetric NAT off the address it gets from the STUN server wont work because when the Symmetric NAT's packet comes in, the source address on it will be different than the address the STUN server told the Cone NAT (different destination = different source address on Symmetric NATs).

if it's full cone or restricted cone it'll still work though, right? it's only if it's port-restricted or symmetric that the different source port won't be valid for the inbound packet from the symmetric nat peer.

aarshkshah1992:
Based on an offline discussion with @willscott 

The fact that we dial a peer if it has any public addresses at all before giving up on Hole Punching due to NAT types will take care of Full Cone NATs.

It will break for address restricted Cones. So, we need to look at how our current code performs in the wild (with some stats around failures rates) to decide if we need to remove the restriction this PR introduces.

godcong:
but after my experiment, some operators can make reverse connections through the opened ports.
I just don’t know how many operators support it now

aarshkshah1992:
@godcong Yeah, that's a FULL Cone NAT. This PR will dial public addresses before giving up because of the NAT type.

p2p/protocol/identify/id.go
```diff
@@ -41,6 +41,13 @@ const ID = "/ipfs/id/1.0.0"
 // 0.4.17 which asserted an exact version match.
 const LibP2PVersion = "ipfs/0.1.0"
 
+const (
+	// UDPNATDeviceTypeKey is the key with which we will persist a peer's UDP NAT Device Type to the peerstore.
+	UDPNATDeviceTypeKey = "UdpNATDeviceType"
+	// TCPNATDeviceTypeKey is the key with which we will persist a peer's TCP NAT Device Type to the peerstore.
+	TCPNATDeviceTypeKey = "TcpNATDeviceType"
```
vyzo, line 48: funky capitalization in the strings; it's `UDP` not `Udp` and same for `TCP`.

p2p/protocol/identify/pb/identify.proto
```diff
@@ -11,6 +11,12 @@ message Delta {
 
 message Identify {
 
+  enum NATDeviceType {
+    UNKNOWN = 100;
+    CONE = 200;
+    SYMMETRIC = 300;
```
vyzo, line 17: 0,1,2...

p2p/protocol/identify/pb/identify.proto
```diff
@@ -43,4 +49,7 @@ message Identify {
   // see github.com/libp2p/go-libp2p-core/record/pb/envelope.proto and
   // github.com/libp2p/go-libp2p-core/peer/pb/peer_record.proto for message definitions.
   optional bytes signedPeerRecord = 8;
-}
+
+  optional NATDeviceType tcpNATDeviceType = 10;
+  optional NATDeviceType udpNATDeviceType = 11;
+}
```
vyzo, line 55: newline.

vyzo (commented):
couple of nits, but overall looks good.

p2p/protocol/identify/id.go
aarshkshah1992: Done.

p2p/protocol/identify/pb/identify.proto
aarshkshah1992: Done.

p2p/protocol/identify/pb/identify.proto
aarshkshah1992: Done.

p2p/protocol/identify/id.go
aarshkshah1992: Done.

p2p/protocol/holepunch/coordination.go
```diff
@@ -49,11 +50,53 @@ func NewHolePunchService(h host.Host, ids *identify.IDService) (*HolePunchServic
...
 	return hs, nil
 }
 
+func (hs *HolePunchService) loop(sub event.Subscription) {
+	defer hs.refCount.Done()
+	defer sub.Close()
+
+	for {
+		select {
+		case _, ok := <-sub.Out():
```
vyzo, line 73: shouldn't we look at the actual event? why are we ignoring it?

p2p/protocol/identify/id.go
```diff
@@ -542,10 +549,37 @@ func (ids *IDService) createBaseIdentifyResponse(
...
+	udpNAT, tcpNAT := ids.observedAddrs.getNATDeviceTypes()
+	udpNATPb := toPbNATDeviceTyp(udpNAT)
+	tcpNATPb := toPbNATDeviceTyp(tcpNAT)
+	mes.UdpNATDeviceType = &udpNATPb
+	mes.TcpNATDeviceType = &tcpNATPb
 
 	return mes
 }
 
+func toPbNATDeviceTyp(typ network.NATDeviceType) pb.Identify_NATDeviceType {
```
vyzo, line 561: typo in method name; shouldn't this be Type?
Also, typ is a very funky variable name, better call it just `t`.

p2p/protocol/holepunch/coordination.go
vyzo: Also, where do we set the support meta-variables into the peerstore?
Shouldn't we be doing it here? Or is it handled anywhere?

I think we need a comment if this is correct as is.

vyzo (commented):
few more comments.

p2p/protocol/holepunch/coordination.go
```diff
@@ -49,11 +50,53 @@ func NewHolePunchService(h host.Host, ids *identify.IDService) (*HolePunchServic
...
+			if !ok {
+				return
+			}
+
+			if hs.peerSupportsHolePunching(hs.host.ID(), hs.host.Addrs()) {
+				hs.host.SetStreamHandler(protocol, hs.handleNewStream)
+			} else {
+				hs.host.RemoveStreamHandler(protocol)
```
vyzo, line 81: wait, this is a BUG -- we lose the ability to dial back connect to a public node that has dialed us.

vyzo (changes requested):
removing the handler is a BUG.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: @vyzo 

1. Our local vars are set in the peerstore in `basic_host.go` during initialisation of the Host & updated in the obs addr manager.

2. Remote vars are set in the peerstore by Identify.

Both changes are included in this PR. I've added a comment.

p2p/protocol/identify/id.go
aarshkshah1992: done.

p2p/protocol/identify/obsaddr.go
```diff
@@ -200,6 +207,21 @@ func (oas *ObservedAddrManager) filter(observedAddrs []*observedAddr) []ma.Multi
 		}
 	}
 
+	// For certain use cases such as hole punching, it's better to advertise even unactivated observed addresses rather than none at all
```
aarshkshah1992, line 210: @vyzo This needs your attention. I discussed this with @Stebalien and he was okay with it.

aarshkshah1992 (commented):
@vyzo 

Rebased and addressed review.

introspection3:
@aarshkshah1992 @BigLep @vyzo @marten-seemann 
"Don't hole punch if either peer is behind a Symmetric NAT" is an evil proposal, as
Symmetric can connect Full Cone,IP restrict.
Symmetric can also connect Port restrict use
Use the birthday paradox

PR #1055, opened 2021-02-17
update deps

vyzo:
so that master doesn't depend on defunct branches.


vyzo:
Merging as this is just a dep update to released version so as to make sure master is not broken.

PR #711, opened 2019-08-26
Hole Punching Protocol

vyzo:
For https://github.com/libp2p/go-libp2p/issues/1039.

Implements the Hole Punching co-ordination protocol and upgrades dependencies to solve multi-stream simultaneous connect.

### TODO
- [ ] Deps.
- [ ] Allow the Hole Punching Stream to open on a transient connection.
- [ ] Unit Tests (once we are happy with the general direction we are taking here).

PR #1057, opened 2021-02-19
Signalling for Hole Punching 

aarshkshah1992:

@vyzo 
Replaces https://github.com/libp2p/go-libp2p/pull/711 to make it easier to review.

Based on https://github.com/libp2p/specs/pull/173/files.

### TODO
- [ ] Tests
- [ ] Deps.


p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// attempts to make a direct connection with the remote peer of `relayConn` by co-ordinating a hole punch over
+// the given relay connection `relayConn`.
+func (hs *HolePunchService) holePunch(relayConn network.Conn) {
+	rp := relayConn.RemotePeer()
+
+	// short-circuit hole punching if a direct dial works.
+	// attempt a direct connection ONLY if we have a public address for the remote peer
+	for _, a := range hs.host.Peerstore().Addrs(rp) {
```
aarshkshah1992, line 74: @vyzo 

`Identify` will be completed here as we wait on `Identify` in the `Connected` notification below before calling `holePunch`.

config/config.go
```diff
@@ -185,12 +188,17 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 		return nil, err
 	}
 
+	if cfg.EnableHolePunching && !cfg.Relay {
+		return nil, errors.New("cannot enable hole punching; relay is not enabled")
+	}
+
```
vyzo, line 194: I think we should enable this by default; but maybe that's for later, after we've done phase 1 testing.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,302 @@
...
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+// TODO Should we have options for these ?
+const (
+	protocol         = "/libp2p/holepunch/1.0.0"
+	maxMsgSize       = 4 * 1024 // 4K
+	holePunchTimeout = 2 * time.Minute
+	dialTimeout      = 60 * time.Second
```
vyzo, line 25: that's too long I think, it's a direct connection so if it takes more than 10s the dial won't succeed.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,302 @@
...
+	maxRetries       = 4
+)
+
+var (
+	log = logging.Logger("p2p/holepunch")
```
vyzo, line 30: use a dash, this is inconsistent with the other logger names we use across the stack.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,302 @@
...
+	// attempt a direct connection ONLY if we have a public address for the remote peer
+	for _, a := range hs.host.Peerstore().Addrs(rp) {
+		if manet.IsPublicAddr(a) && !isRelayAddress(a) {
+			forceDirectConnCtx := network.WithForceDirectDial(hs.ctx, "hole-punching")
+			dialCtx, cancel := context.WithTimeout(forceDirectConnCtx, dialTimeout)
+			defer cancel()
+			if err := hs.host.Connect(dialCtx, peer.AddrInfo{ID: rp}); err == nil {
+				log.Debugf("direct connection to peer %s successful, no need for a hole punch", rp.Pretty())
+				return
+			}
```
vyzo, line 82: we should log the error if we fail; also, the dialTimeout needs to be dialed way down.

p2p/protocol/holepunch/coordination.go
vyzo: I would say 5s in fact.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,302 @@
...
+			break
+		}
+	}
+
+	// hole punch
+	hpCtx := network.WithUseTransient(hs.ctx, "hole-punch")
```
vyzo, line 88: also add `WithNoDial`.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,302 @@
...
+	s, err := hs.host.NewStream(hpCtx, rp, protocol)
+	if err != nil {
+		return
```
vyzo, line 91: log this error.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/libp2p/go-msgio/protoio"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+// TODO Should we have options for these ?
+const (
+	protocol         = "/libp2p/holepunch/1.0.0"
+	maxMsgSize       = 4 * 1024 // 4K
+	holePunchTimeout = 2 * time.Minute
```
vyzo, line 24: this also needs to be dialed down me thinks.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,302 @@
...
+	// before giving up and declaring the hole punch a failure.
+	b := &backoff.Backoff{
+		Jitter: true,
+		Min:    1 * time.Second,
+		Max:    5 * time.Second,
+		Factor: 2,
+	}
+	for b.Attempt() < maxRetries {
+		time.Sleep(b.Duration())
+		err = hs.host.Connect(dialCtx, pi)
```
vyzo, line 225: We might need to make a new context here, as `WithTimeout` might convert internally to an absolute deadline.
Not sure, but let's be on the safe side.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,302 @@
...
+	"context"
+	"sync"
+	"time"
+
+	logging "github.com/ipfs/go-log"
+	"github.com/jpillora/backoff"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	holepunch_pb "github.com/libp2p/go-libp2p/p2p/protocol/holepunch/pb"
```
vyzo, line 13: just `pb`, the holepunch prefix is implied by the context.

p2p/protocol/holepunch/coordination.go
requilence: it would be nice to have these vars exported so users can play around with different timeouts

p2p/protocol/holepunch/coordination.go
aarshkshah1992: Keeping this to 1 minute for now. Can always fix after the alpha testing.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: done.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: done.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: I don't think logging the error is a good idea here. Remember, we also dial the other peer if either of us is behind an `Unknown` NAT.  If it turns out that either of us is infact behind a Symmetric NAT, this will cause unnecessary noise in the logs.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: done.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: done.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: have changed this 10s and have changed the dial loop to create a fresh context on every retry.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: done.

aarshkshah1992 (commented):
@vyzo Addressed feedback and the PR is now open against a Staging branch.

p2p/protocol/holepunch/coordination.go
vyzo: ok, fair enough.

p2p/protocol/holepunch/coordination.go
vyzo: wait, Error is too much! This should probably be Debug.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,308 @@
...
+	"context"
+	"sync"
+	"time"
+
+	logging "github.com/ipfs/go-log"
+	"github.com/jpillora/backoff"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch/pb"
```
vyzo, line 13: Can you actually name this `pb`? So that the code that uses just does `pb.Foo` instead of `holepuch_pb.Foo`.
It's consistent with how we do things across the stack.

p2p/protocol/holepunch/pb/holepunch.proto
```diff
@@ -0,0 +1,19 @@
...
+
+  optional Type type=1;
+
+  // For hole punching, we'll send some additional observed addresses to the remote peer
+  // that could have been filtered by the Host address factory (for example: AutoRelay removes all public addresses if peer has private reachability).
+  // This is a hack!
+  // We plan to have a better address discovery and advertisement mechanism in the future.
+  // See https://github.com/libp2p/go-libp2p-autonat/pull/98
+  repeated bytes ObsAddrs = 2;
+}
```
vyzo, line 19: newline.

aarshkshah1992: done.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: done.

aarshkshah1992:
@Stebalien 

If you can find the time, it would be great to have your review on this one and it's not a big PR.

aarshkshah1992:
@vyzo Looks like we are ready to go. Not merging though till I finish writing some solid unit tests.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,308 @@
...
+	err := hs.host.Connect(dialCtx, pi)
+	if err == nil {
+		log.Infof("hole punch with peer %s successful, direct conns to peer are:", pi.ID.Pretty())
+		for _, c := range hs.host.Network().ConnsToPeer(pi.ID) {
+			if !isRelayAddress(c.RemoteMultiaddr()) {
+				log.Info(c)
+			}
+		}
+		return
+	}
```
vyzo, line 214: We need to alter this logic for QUIC hole punching with the simultaneous connect; if you are the side sending garbage packets, the connect will fail but you'll get a connection from the listen.

So we should check the error, and then check to see if we have a connection anyway and be happy if that's the case.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,308 @@
...
+		if err == nil {
+			log.Infof("hole punch with peer %s successful after retry, direct conns to peer are:", pi.ID.Pretty())
+			for _, c := range hs.host.Network().ConnsToPeer(pi.ID) {
+				if !isRelayAddress(c.RemoteMultiaddr()) {
+					log.Info(c)
+				}
+			}
+			return
+		}
+	}
```
vyzo, line 241: same here for the QUIC simultaneous connect logic.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,308 @@
...
+	// Hole punch now by forcing a connect
+	pi := peer.AddrInfo{
+		ID:    rp,
+		Addrs: obsDial,
+	}
+	hs.holePunchConnectWithBackoff(pi)
+}
+
+func (hs *HolePunchService) holePunchConnectWithBackoff(pi peer.AddrInfo) {
+	forceDirectConnCtx := network.WithForceDirectDial(hs.ctx, "hole-punching")
```
vyzo, line 202: we also need `WithSimultaneousConnect`

vyzo (commented):
An important detail re: the implementation of QUIC hole punching with the use of the SimultaneousConnect option.

I am working on that, but the code here will need to change.

aarshkshah1992:
@vyzo I'll block on this PR till you have the QUIC changes merged and we can then revisit this.

vyzo:
they don't need to merge, just have an open pr for testing, ETA shortly.

p2p/protocol/holepunch/coordination.go
vyzo: Actually, this should be completely unnecessary if the inbound connection succeeds because we'll have a connection in the swarm.
The timeout for hole punching is 1s, which should be sufficient to establish the connection.

p2p/protocol/holepunch/coordination.go
vyzo: see above.

vyzo:
@aarshkshah1992 we are ready with the QUIC pr; let's test to see whether it works, and then we can merge that.

p2p/protocol/holepunch/coordination.go
aarshkshah1992: done.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,314 @@
...
+	hs := &HolePunchService{ctx: ctx, ctxCancel: cancel, host: h, ids: ids}
+
+	h.SetStreamHandler(protocol, hs.handleNewStream)
+	h.Network().Notify((*netNotifiee)(hs))
+	return hs, nil
+}
+
+// Close closes the Hole Punch Service.
+func (hs *HolePunchService) Close() error {
+	hs.closeSync.Do(func() {
```
marten-seemann, line 59: The cancel func can be called multiple times. I believe that allows you to get rid of this Once here.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,314 @@
...
+	msg.Reset()
+	msg.Type = pb.HolePunch_SYNC.Enum()
+	if err := w.WriteMsg(msg); err != nil {
+		s.Reset()
+		log.Errorf("failed to send SYNC message for hole punching, err: %s", err)
+		return
+	}
+	defer s.Close()
+
+	synTime := time.Duration(rtt.Milliseconds()/2) * time.Millisecond
```
marten-seemann, line 136: rtt/2?

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,314 @@
...
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+// TODO Should we have options for these ?
+const (
+	protocol         = "/libp2p/holepunch/1.0.0"
+	maxMsgSize       = 4 * 1024 // 4K
+	holePunchTimeout = 1 * time.Minute
+	dialTimeout      = 10 * time.Second
```
vyzo, line 24: this needs to be 5s to match the hole punch timeout in QUIC; otherwise our peers might de-synchronize.

config/config.go
vyzo: So this check gets in the way for testing with limited relays, as we will have the v1 relay disabled in this case.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,314 @@
...
+	// send a CONNECT and start RTT measurement
+	msg := new(pb.HolePunch)
+	msg.Type = pb.HolePunch_CONNECT.Enum()
+	msg.ObsAddrs = addrsToBytes(hs.ids.OwnObservedAddrs())
+	if err := w.WriteMsg(msg); err != nil {
+		s.Reset()
+		log.Errorf("failed to send hole punch CONNECT, err: %s", err)
+		return
+	}
+	tstart := time.Now()
```
vyzo, line 108: the start of the RTT measurement has to move before writing the message I think.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,314 @@
...
+		log.Errorf("failed to read connect message from remote peer, err: %s", err)
+		return
+	}
+	if msg.GetType() != pb.HolePunch_CONNECT {
+		s.Reset()
+		log.Debugf("expectd HolePunch_CONNECT message, got %s", msg.GetType())
+		return
+	}
+	obsRemote := addrsFromBytes(msg.ObsAddrs)
+	rtt := time.Since(tstart)
```
vyzo, line 124: and the end of the RTT measurement has to move right after reading the message.

marten-seemann:
Note: we'll have to update the QUIC Hole Punching coordination logic if / when https://github.com/libp2p/go-libp2p-quic-transport/pull/210 is merged.

vyzo:
I think it might make sense to implement it the way you suggest, with
another sync round to compute a the rtt anew.

marten-seemann:
Closing, since #1168 has been merged (into the feature branch at least).

PR #1062, opened 2021-02-27
fix autonat race

willscott:
fix https://github.com/ipfs/go-ipfs/issues/7947

move `BasicHost.AutoNat` to a private field (it has no public method and shouldn't be accessed afaik.
Instead add a setter for config that sets it while holding the address mutex to prevent reads of the
field at the same time.


mvdan (approved):
Looks like the only other field read is also safe, since it already holds the addrMu mutex.

Thanks for the quick fix! :)

Stebalien (approved):

Stebalien:
(someday we'll fix the dependency cycles here)

PR #1066, opened 2021-03-08
chore: update deps

Stebalien:


raulk:
Defer to @vyzo and @aarshkshah1992 since they're more aware of the upstream changes.

vyzo (approved):

PR #1072, opened 2021-04-06
feat- Codeql analysis

naveensrinivasan:
- Enabled codeql analysis for PR and as a cron job.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,47 @@
+name: "CodeQL"
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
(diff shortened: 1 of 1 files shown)


marten-seemann:
What's the purpose of using Codeql on a single repository? libp2p consists of *a lot* of different repositories.

naveensrinivasan:
> What's the purpose of using Codeql on a single repository? libp2p consists of _a lot_ of different repositories.

I understand there are multiple repositories. But every repository has to have codeql. There is no single codeql that can workflow that can run for all repositories with checks during PR.

marten-seemann:
If we decide to use Codeql (I have no opinion on that, I'm not aware of that tool), we'd best use our new Unified CI setup: https://github.com/protocol/.github/.

Can you please open an issue there describing why you think we should use Codeql on all of our repositories?

PR #1073, opened 2021-04-12
option for custom dns resolver

vyzo:


Stebalien (commented):
Don't we actually need to _use_ the resolver somewhere?

vyzo:
the code is already there AFAICT, we just needed to add it to the basic host opts in config. Did I miss that?

config/config.go
```diff
@@ -186,11 +189,12 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
...
-		AddrsFactory: cfg.AddrsFactory,
-		NATManager:   cfg.NATManager,
-		EnablePing:   !cfg.DisablePing,
-		UserAgent:    cfg.UserAgent,
+		ConnManager:       cfg.ConnManager,
+		AddrsFactory:      cfg.AddrsFactory,
+		NATManager:        cfg.NATManager,
+		EnablePing:        !cfg.DisablePing,
+		UserAgent:         cfg.UserAgent,
+		MultiaddrResolver: cfg.MultiaddrResolver,
```
vyzo, line 197: @stebalien we use it here

Stebalien (approved):
I guess we just never wired this up...

PR #1074, opened 2021-04-14
Integrates smart-record in libp2p host

adlrocha:
*Not ready to merge, depends on https://github.com/libp2p/go-smart-record/pull/11*

# Goal
The goal of the PR is to add an option+config to register the libp2p smart-record protocol in a host. For now, if the option is enabled the only thing that we do is to register the corresponding handlers for the `smart-record` protocol. 

In the future we may want to expose the smart record interface through the host (by adding it to the `host` interface or wrapping up the functionality in a `smartRecordHost` similar to how is done with the `routedHost`. No decision has been made yet.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -31,6 +31,8 @@ import (
 
 	logging "github.com/ipfs/go-log"
 	ma "github.com/multiformats/go-multiaddr"
+
+	smart "github.com/libp2p/go-smart-record/protocol"
 )
 
 var log = logging.Logger("p2p-config")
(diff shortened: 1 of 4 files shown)


Stebalien:
Everything attached to the host is a part of the "core" libp2p. Unless this is actually being used by libp2p, it doesn't belong here. For example, there are no options for pubsub, bitswap, the DHT, etc.

adlrocha:
I see. You are right, our protocol uses a libp2p-host but I don't think it is going to be part of the "core" libp2p. We want users to be able to use smart-record within a libp2p node to build applications, but we don't need a libp2p option for that. I don't know why I assumed the DHT was part of the "core" libp2p, but now I see the only thing defined at the core is the `Routing` interface.

Closing this for now, then. Thank you for the heads up, @Stebalien.

PR #1076, opened 2021-04-15
static check fixes

coryschwartz:
various fixes for staticcheck

For context: https://github.com/protocol/.github/issues/41

### DIFF
--- a/p2p/host/relay/autorelay.go
+++ b/p2p/host/relay/autorelay.go
@@ -7,13 +7,13 @@ import (
 	"sync"
 	"time"
 
+	"github.com/libp2p/go-libp2p-core/discovery"
 	"github.com/libp2p/go-libp2p-core/event"
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/routing"
 
 	circuit "github.com/libp2p/go-libp2p-circuit"
-	discovery "github.com/libp2p/go-libp2p-discovery"
 	basic "github.com/libp2p/go-libp2p/p2p/host/basic"
 
 	ma "github.com/multiformats/go-multiaddr"
@@ -247,7 +247,15 @@ func (ar *AutoRelay) discoverRelays(ctx context.Context) ([]peer.AddrInfo, error
 
 	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
-	return discovery.FindPeers(ctx, ar.discover, RelayRendezvous, discovery.Limit(1000))
+	var ret []peer.AddrInfo
+	ch, err := ar.discover.FindPeers(ctx, RelayRendezvous, discovery.Limit(1000))
+	if err != nil {
+		return nil, err
+	}
+	for p := range ch {
+		ret = append(ret, p)
+	}
+	return ret, nil
 }
 
 func (ar *AutoRelay) selectRelays(ctx context.Context, pis []peer.AddrInfo) []peer.AddrInfo {
--- a/p2p/host/relay/relay.go
+++ b/p2p/host/relay/relay.go
@@ -4,7 +4,8 @@ import (
 	"context"
 	"time"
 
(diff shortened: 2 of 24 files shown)


p2p/host/relay/autorelay.go
```diff
@@ -247,7 +247,15 @@ func (ar *AutoRelay) discoverRelays(ctx context.Context) ([]peer.AddrInfo, error
 
 	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
-	return discovery.FindPeers(ctx, ar.discover, RelayRendezvous, discovery.Limit(1000))
+	var ret []peer.AddrInfo
+	ch, err := ar.discover.FindPeers(ctx, RelayRendezvous, discovery.Limit(1000))
```
coryschwartz, line 251: This should be equivilent to the FindPeers util function in go-libp2p-discovery, to convert the channel into a slice.

p2p/host/relay/relay.go
```diff
@@ -20,7 +20,31 @@ func Advertise(ctx context.Context, advertise discovery.Advertiser) {
 	go func() {
 		select {
 		case <-time.After(AdvertiseBootDelay):
-			discovery.Advertise(ctx, advertise, RelayRendezvous, discovery.TTL(AdvertiseTTL))
+			go func() {
```
coryschwartz, line 23: This is brought in from go-libp2p-discovery.

If it's better to use the Util Functions from there I can swtich back to using them.

p2p/protocol/identify/id.go
```diff
@@ -368,7 +368,7 @@ func (ids *IDService) identifyConn(c network.Conn, signal chan struct{}) {
 
 	// ok give the response to our handler.
 	if err = msmux.SelectProtoOrFail(ID, s); err != nil {
-		log.Event(context.TODO(), "IdentifyOpenFailed", c.RemotePeer(), logging.Metadata{"error": err})
+		// log.Event(context.TODO(), "IdentifyOpenFailed", c.RemotePeer(), logging.Metadata{"error": err})
```
coryschwartz, line 371: Is there a metric that should be emitted here in leiu of depreciated go-log Events?

p2p/host/routed/routed.go
```diff
@@ -131,7 +130,7 @@ func logRoutingErrDifferentPeers(ctx context.Context, wanted, got peer.ID, err e
 	lm["error"] = err
 	lm["wantedPeer"] = func() interface{} { return wanted.Pretty() }
 	lm["gotPeer"] = func() interface{} { return got.Pretty() }
-	log.Event(ctx, "routingError", lm)
+	// log.Event(ctx, "routingError", lm)
```
marten-seemann, line 133: Why is this code commented?

p2p/net/mock/mock_test.go
```diff
@@ -15,18 +15,19 @@ import (
...
 )
 
-func randPeer(t *testing.T) peer.ID {
-	p, err := test.RandPeerID()
-	if err != nil {
-		t.Fatal(err)
-	}
-	return p
-}
+// Unused code:
```
marten-seemann, line 22: Can we delete this?

p2p/host/relay/relay.go
Stebalien: Well, those functions exist so we don't need to duplicate this code. What was the motivation for this change?

p2p/host/routed/routed.go
Stebalien: This entire function should be deleted and replaced with a call to `log.Errorw`.

Side note: please (almost) never comment out code unless you're actively debugging.

Stebalien (changes requested):

p2p/protocol/identify/id.go
Stebalien: Just a structured log message.

p2p/host/relay/relay.go
Stebalien: Ah... I see.

Stebalien (approved):

PR #1083, opened 2021-04-30
sync: update CI config files (and fix tests)

Stebalien:
Reverted https://github.com/libp2p/go-libp2p/pull/1082 because this really should have a review.


marten-seemann (approved):
So this was mostly test cleanup, which LGTM. I see we're using a bunch of sleeps in tests, which makes me a bit sad, but that's not the fault of this PR.

PR #1086, opened 2021-05-01
remove NEWS.md

marten-seemann:
Latest meaningful change to this file was in June 2018. That's pretty old news.


Stebalien:
Yeah, we can revive this with a new release policy, but we might as well get rid of the current file.

PR #1093, opened 2021-05-10
chore: update yamux

Stebalien:


Stebalien:
This just fixes a tiny error in how windows were computed in yamux so we don't immediately grow the window size when we open a stream.

Stebalien:
(merging anyways because it's just a dep update)

PR #1103, opened 2021-05-18
speed up the mock tests

marten-seemann:
It's probably not worth spending more than a minute on this single test case.

### DIFF
--- a/p2p/net/mock/mock_test.go
+++ b/p2p/net/mock/mock_test.go
@@ -593,7 +593,7 @@ func TestLimitedStreams(t *testing.T) {
 	}
 }
 func TestFuzzManyPeers(t *testing.T) {
-	peerCount := 50000
+	peerCount := 500
 	if detectrace.WithRace() {
 		peerCount = 100
 	}


p2p/net/mock/mock_test.go
```diff
@@ -593,14 +593,14 @@ func TestLimitedStreams(t *testing.T) {
...
 }
 func TestFuzzManyPeers(t *testing.T) {
-	peerCount := 50000
+	peerCount := 500
 	if detectrace.WithRace() {
 		peerCount = 100
 	}
 	for i := 0; i < peerCount; i++ {
 		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
```
Stebalien, line 602: We canceled immediately so the node didn't stay open for the entire test case.

Stebalien (changes requested):
Otherwise this seems fine. Note: this is testing the _mock_ network, so it's not super critical regardless.

Stebalien (approved):

PR #1104, opened 2021-05-18
remove slow ObservedAddrManager test that doesn't test anything

marten-seemann:
I might be missing something here, but it looks like this test doesn't test anything (other than the fact that the code doesn't panic). Doing 10000 operations with a 1ms `time.Sleep` is not such a good idea, as this one test case will take at least 10s.


Stebalien (approved):
It looks like this test was _actually_ written for profiling and _really_ shouldn't be an actual test (maybe a benchmark?). But it's not that useful regardless.

PR #1115, opened 2021-06-04
perf/basic_host: Don't handle address change if we hasn't anyone

snyh:
`basic_host.background` is used for handling address changed events, but the `node` may be created with `libp2p.NoListenerAddrs`.

the `h.updateLocalIpAddr` is cost a little CPU time.

But I use `NoListenerAddrs` with many (one thousand) tiny test nodes, and all these test nodes communicate with other peers by a  full feature node (by pubsub or relay) in the same process. 

The basic_host.background would use many CPU time and add 1000 useless goroutine in this situatuion.

-------------------------------------------------------------
I'm not sure whether the `network` could do `listen` after `host.background` started.  
So I just keep the background goroutine and skip one round if there hasn't any listener address.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -520,11 +520,12 @@ func (h *BasicHost) background() {
 	defer ticker.Stop()
 
 	for {
-		var curr []ma.Multiaddr
 		if len(h.network.ListenAddresses()) > 0 {
 			h.updateLocalIpAddr()
-			curr = h.Addrs()
 		}
+		// Request addresses anyways because, technically, address filters still apply.
+		// The underlying AllAddrs call is effectivley a no-op.
+		curr := h.Addrs()
 		emitAddrChange(curr, lastAddrs)
 		lastAddrs = curr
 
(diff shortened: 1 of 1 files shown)


snyh:
@Stebalien  thanks for https://github.com/libp2p/go-libp2p/pull/398

Stebalien:
Is it the call to `Addrs` or the call to update the IP address that costs CPU? Could you post a CPU profile?

Notes:

1. We could easily make `Addrs` short-circuit if we're not actually listening on any transports. That should save a fair amount.
2. We could avoid updating the local IP address... I'm just slightly concerned about stale addresses and want to see if we can just optimize it.

snyh:
[1.pb.gz](https://github.com/libp2p/go-libp2p/files/6605292/1.pb.gz)
Screenshot from 2021-06-07 09-10-04

profile with 30s  before this PR

snyh:
Screenshot from 2021-06-07 09-12-49
[2.pb.gz](https://github.com/libp2p/go-libp2p/files/6605298/2.pb.gz)

after this PR

Stebalien (approved):
After a lot of head scratching to make sure this can't go wrong... I'm convinced it can't go wrong.

Stebalien:
Well, one _small_ change. We want `Addrs` and the emitted events to be the same.

Stebalien:
And thanks for the detailed perf analysis!

PR #1116, opened 2021-06-07
Fix race in adding connections to connsByPeer

Kubuxu:
Side A could already be running Identify with side B before B had the
connection in the connsByPeer, this lead to Connectedness returning
NotConnected and Identify failing.

### DIFF
--- a/p2p/net/mock/mock_peernet.go
+++ b/p2p/net/mock/mock_peernet.go
@@ -160,9 +160,39 @@ func (pn *peernet) connect(p peer.ID) (*conn, error) {
 func (pn *peernet) openConn(r peer.ID, l *link) *conn {
 	lc, rc := l.newConnPair(pn)
 	log.Debugf("%s opening connection to %s", pn.LocalPeer(), lc.RemotePeer())
+	addConnPair(pn, rc.net, lc, rc)
+
 	go rc.net.remoteOpenedConn(rc)
 	pn.addConn(lc)
 	return lc
+}
+
+// addConnPair adds connection to both peernets at the same time
+// must be followerd by pn1.addConn(c1) and pn2.addConn(c2)
+func addConnPair(pn1, pn2 *peernet, c1, c2 *conn) {
+	pn1.Lock()
+	pn2.Lock()
+
+	add := func(pn *peernet, c *conn) {
+		_, found := pn.connsByPeer[c.RemotePeer()]
+		if !found {
+			pn.connsByPeer[c.RemotePeer()] = map[*conn]struct{}{}
+		}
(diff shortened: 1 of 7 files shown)


p2p/net/mock/mock_peernet.go
```diff
@@ -160,11 +160,41 @@ func (pn *peernet) connect(p peer.ID) (*conn, error) {
 func (pn *peernet) openConn(r peer.ID, l *link) *conn {
 	lc, rc := l.newConnPair(pn)
 	log.Debugf("%s opening connection to %s", pn.LocalPeer(), lc.RemotePeer())
+	addConnPair(pn, rc.net, lc, rc)
```
magik6k, line 163: Shouldn't this also have a rc.net.addConn below?

Kubuxu: It happens in the other goroutine in `remoteOpenConn`.

magik6k (approved):

p2p/net/mock/mock_peernet.go
```diff
@@ -160,11 +160,41 @@ func (pn *peernet) connect(p peer.ID) (*conn, error) {
...
 	go rc.net.remoteOpenedConn(rc)
 	pn.addConn(lc)
 	return lc
 }
 
+// addConnPair adds connection to both peernets at the same time
+// must be followerd by pn1.addConn(c1) and pn2.addConn(c2)
+func addConnPair(pn1, pn2 *peernet, c1, c2 *conn) {
+	pn1.Lock()
+	pn2.Lock()
```
Stebalien, line 174: We need to sort these or we'll deadlock. Same with the other locks we're taking.

p2p/net/mock/mock_peernet.go
```diff
@@ -160,11 +160,41 @@ func (pn *peernet) connect(p peer.ID) (*conn, error) {
...
+		if !found {
+			pn.connsByLink[c.link] = map[*conn]struct{}{}
+		}
+		pn.connsByLink[c.link][c] = struct{}{}
+	}
+	add(pn1, c1)
+	add(pn2, c2)
+
+	c1.notifLk.Lock()
+	c2.notifLk.Lock()
```
Stebalien, line 193: Are you sure we can do this? What if we connect from a notification handler?

Stebalien (changes requested):

p2p/net/mock/mock_peernet.go
Stebalien: Discussed. We're fine here because we just created these connections.

Stebalien (approved):

PR #1125, opened 2021-07-06
implement the new mDNS spec, run both implementations in parallel

marten-seemann:


p2p/discovery/mdns_legacy.go
```diff
@@ -0,0 +1,209 @@
...
+			out = append(out, tcp)
+		}
+	}
+	if len(out) == 0 {
+		return nil, errors.New("failed to find good external addr from peerhost")
+	}
+	return out, nil
+}
+
+func NewMdnsServiceLegacy(ctx context.Context, peerhost host.Host, interval time.Duration) (Service, error) {
```
marten-seemann, line 63: Should we make this private?

p2p/discovery/mdns.go
```diff
@@ -3,220 +3,197 @@ package discovery
...
-type mdnsService struct {
-	server  *mdns.Server
-	service *mdns.MDNSService
-	host    host.Host
-	tag     string
-
-	lk       sync.Mutex
-	notifees []Notifee
-	interval time.Duration
+func NewMdnsServiceNew(host host.Host) *mdnsService {
```
marten-seemann, line 39: ... and this one as well?

Stebalien:
So, I guess the question is, do we even need the "muxer" type? I'd consider:

1. Having both an `mdns` and an `mdns_legacy` package.
2. Instantiating _both_ in go-ipfs (separately).
3. Not having the muxer.

mxinden:
@marten-seemann what is the easiest way for me to build a Golang binary that is running the changes to test interoperability with the Rust implementation?

marten-seemann:
@Stebalien 

> So, I guess the question is, do we even need the "muxer" type? I'd consider:
> 1. Having both an mdns and an mdns_legacy package.
> 2. Instantiating both in go-ipfs (separately).
> 3. Not having the muxer.

It's definitely cleaner code-wise, but it's also more disruptive. libp2p users will have to actively change their code if they want to maintain compatibility with libp2p instances that haven't updated yet. Assuming that people will have similar rollout and adoption rates as we're seeing in the IPFS network, they won't have to worry at all about this change.

Stebalien:
> It's definitely cleaner code-wise, but it's also more disruptive. libp2p users will have to actively change their code if they want to maintain compatibility with libp2p instances that haven't updated yet. Assuming that people will have similar rollout and adoption rates as we're seeing in the IPFS network, they won't have to worry at all about this change.

We're already breaking the constructor, right? This is one of those cases where I'd rather have users make an explicit/informed choice. That way, _users_ can choose when they want to deprecate the old method.

---

Also with respect to breaking changes, I didn't look too closely before but it _looks_ like we're removing the "service" string. That could be an issue for some upstream users.

marten-seemann:
> Also with respect to breaking changes, I didn't look too closely before but it looks like we're removing the "service" string. That could be an issue for some upstream users.

Yes, that is intentional. We're now implementing the spec, and the spec *defines* what the service name is: https://github.com/libp2p/specs/blob/master/discovery/mdns.md#definitions. If there are use cases where users might want to set the service name to something different, we might want to update the spec.

Stebalien:
> we might want to update the spec.

I _think_ we might. Take a look at https://grep.app/search?q=github.com/libp2p/go-libp2p/p2p/discovery. It looks like _most_ systems set a custom tag. This is actually pretty reasonable because one usually wants to find like-services, not random libp2p nodes. If we release as-is, all of these services are going to be unhappy.

marten-seemann:
Good point. We'll update the spec. #1161 implements the solution suggested by @Stebalien.

mxinden:
@marten-seemann does the spec still need an update in regards to the _service name_?

mxinden:
> @marten-seemann does the spec still need an update in regards to the _service name_?

Friendly ping @marten-seemann.

PR #1128, opened 2021-07-16
make QUIC a default transport

marten-seemann:
This means that we'll finally be able to set the QUIC version in go-libp2p as well.

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -5,15 +5,16 @@ package libp2p
 import (
 	"crypto/rand"
 
-	crypto "github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/crypto"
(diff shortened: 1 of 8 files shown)


Stebalien:
1. Let's do a patch release with the new TCP transport first (and maybe get the yamux fix in?). Then we can put this in a major release.
2. We should probably get https://github.com/libp2p/go-libp2p-quic-transport/pull/211 in first, if possible.

marten-seemann:
> Let's do a patch release with the new TCP transport first (and maybe get the yamux fix in?). Then we can put this in a major release.

Agreed, that was the plan anyway.

> We should probably get implement a Transport.Close that waits for the reuse's GC to finish go-libp2p-quic-transport#211 in first, if possible.

CI on this PR seems even more flaky than before, I hope that this PR helps.

Stebalien (approved):

PR #1132, opened 2021-07-22
remove deprecated Filter option

marten-seemann:
These functions have been deprecated since May 2020.


Stebalien (approved):
This'll be breaking for go-ipfs, but it's about time that we fix it. Mind fixing https://github.com/ipfs/go-ipfs/blob/461f691810cd6dde3a594ea8027c6df2f7727668/core/node/libp2p/addrs.go#L13-L26?

PR #1134, opened 2021-07-22
use the standard comment to deprecate basichost.Option and basichost.New

marten-seemann:
These have been deprecated for a looong time (`New` since 2017), but the deprecation comment was in a form that wouldn't get picked up by IDEs.
Do we want to first rewrite the comment, and then remove this later, or should we do it right now?

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -58,7 +58,7 @@ type AddrsFactory func([]ma.Multiaddr) []ma.Multiaddr
 
 // Option is a type used to pass in options to the host.
 //
-// Deprecated in favor of HostOpts and NewHost.
+// Deprecated: use HostOpts instead.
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
Let's give it a proper deprecation period.

Stebalien:
And we should probably fix the tests...

Stebalien:
Actually, I change my mind. This is probably safe to get rid of. These have been deprecated for ages and nobody seems to use them.

BigLep:
@marten-seemann : are we good to get the checks addressed so we can merge?

marten-seemann:
@BigLep We'll abandon this PR, as @Stebalien suggested in https://github.com/libp2p/go-libp2p/pull/1134#issuecomment-885289894, and remove these functions. Will send a new PR.

PR #1137, opened 2021-07-23
feat: use the actual libp2p version

### LINKED ISSUE #714: Make LibP2PVersion configurable and dedup with UserAgent
See:

https://github.com/libp2p/go-libp2p/blob/c73d2b9c8ee65326fe6c1b76a05a6b6cdcf223b9/p2p/protocol/identify/id.go#L34

As explained in the following comment introduced in #702, IPFS needs this value to stay fixed, but libp2p is an independent project and many downstream users don't care about IPFS compatibility:

```
XXX: Don't change this till 2020. You'll break all go-ipfs versions prior to
0.4.17 which asserted an exact version match.
```

Proposed path forward:

1. Deduplicate ClientVersion/UserAgent and LibP2PVersion. Right now it's not clear what the purpose of each is.
2. `LibP2PVersion` signals the go-libp2p version, and it's overridable but discouraged to do so. Only used by IPFS to force the backwards compatibility they need.
3. `UserAgent` announces the name/version of the application using libp2p. It is strongly encouraged for the application to set it, and it defaults to `LibP2PVersion` if unset.

Stebalien:
fixes #714

Problem: using "ipfs/0.1.0" for the protocol version doesn't make sense in libp2p.
Solution: set it to the go libp2p version (per #714).

The question is, does this even make sense? The protocol version was _supposed_ to identify the libp2p _protocol_ version, allowing peers to determine if they should even bother speaking to each other in the first place. Given that, using the go-libp2p version doesn't make any sense.

However, I feel like systems should just do something like what Filecoin does and have a simple "hello" protocol to determine if two peers want to speak. It's often more complicated than a simple string test (Filecoin tests genesis blocks).

Should we just "rebrand" the "ProtocolVersion" to "Libp2pVersion"?

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"reflect"
 	"runtime/debug"
+	"strings"
 	"sync"
 	"time"
 
@@ -35,12 +37,6 @@ var log = logging.Logger("net/identify")
 // service.
 const ID = "/ipfs/id/1.0.0"
 
-// LibP2PVersion holds the current protocol version for a client running this code
-// TODO(jbenet): fix the versioning mess.
-// XXX: Don't change this till 2020. You'll break all go-ipfs versions prior to
-// 0.4.17 which asserted an exact version match.
-const LibP2PVersion = "ipfs/0.1.0"
-
 // StreamReadTimeout is the read timeout on all incoming Identify family streams.
 var StreamReadTimeout = 60 * time.Second
 
@@ -49,6 +45,7 @@ var (
 	signedIDSize     = 8 * 1024 // 8K
 	maxMessages      = 10
 	defaultUserAgent = "github.com/libp2p/go-libp2p"
+	libp2pVersion    = "github.com/libp2p/go-libp2p"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


vyzo:
I think we should just call it libp2p version; it's useful for debugging purposes.

mxinden:
Thanks for the ping @Stebalien!

Do I understand correctly that you are planning to set the [identify protocol version](https://github.com/libp2p/specs/tree/master/identify#protocolversion) field to the go libp2p version? Would users still be able to override the identify protocol version field, e.g. could an IPFS client still set it to `ipfs/0.1.0`?

> However, I feel like systems should just do something like what Filecoin does and have a simple "hello" protocol to determine if two peers want to speak. It's often more complicated than a simple string test (Filecoin tests genesis blocks).

Today at least one live network (Polkadot) uses the identify protocol version to differentiate e.g. production networks from test networks. Among other things the identify protocol version is used to prevent DHTs from the different networks from merging. In my eyes this is a core functionality of the identify protocol.

A remote can infer the go libp2p version based on the agent version. E.g. I can match an agent version like `go-ipfs/0.8.0/48f94e2` to a specific go libp2p version. With that in mind, what would be the benefit of including the go libp2p version in the identify protocol version field _once more_?

For what it is worth [rust-libp2p requires a user to provide the identify protocol version at setup time](https://github.com/libp2p/rust-libp2p/blob/e437c009dc80777300e56c0c06d73ff14e5449a1/protocols/identify/src/identify.rs#L126).

In case this is not go-libp2p specific, we should be consistent across implementations and thus this decision should not happen on the go-libp2p repository, but on the libp2p/specs repository instead.

Stebalien:
> Do I understand correctly that you are planning to set the identify protocol version field to the go libp2p version? Would users still be able to override the identify protocol version field, e.g. could an IPFS client still set it to ipfs/0.1.0?

Not in the current patch. Basically, we have two versions:

1. The "protocol version" (this).
2. The "agent version" (the application).

Applications, like go-ipfs, will set the _agent_ version.

> With that in mind, what would be the benefit of including the go libp2p version in the identify protocol version field once more?

The agent version doesn't always include the version (devel builds) and the agent may not be open source. Including the specific libp2p version would allow for implementation specific decisions (in case some quick comes up).

On the other hand, I generally agree.

> In case this is not go-libp2p specific, we should be consistent across implementations and thus this decision should not happen on the go-libp2p repository, but on the libp2p/specs repository instead.

You're right, we absolutely should. I'll close this and open a spec issue.

mxinden:
Thanks @Stebalien for the quick reply and moving it to libp2p/specs.

PR #1141, opened 2021-07-27
chore: update deps

marten-seemann:
This should conclude our deprecation party (for now): go-libp2p now builds with both go-libp2p-core and go-multiaddr updated (which were the two packages where we removed a bunch of deprecated types / functions).


marten-seemann:
We're in a dependency loop here. For the interop tests, we need to update go-libp2p-daemon, which depends on go-ipns, which depends on go-libp2p-kad-dht, which in turn depends on go-libp2p. Due to the deprecations, we will have to update go-libp2p in go-libp2p-kad-dht first, but we need to cut a go-libp2p for that, and we can't do that without merging this PR.

We _could_ just merge this PR and ignore CI, but this is quite an annoying situation.

Stebalien:
> We're in a dependency loop here. For the interop tests, we need to update go-libp2p-daemon, which depends on go-ipns, which depends on go-libp2p-kad-dht, which in turn depends on go-libp2p. Due to the deprecations, we will have to update go-libp2p in go-libp2p-kad-dht first, but we need to cut a go-libp2p for that, and we can't do that without merging this PR.

We use replace directives here so we _shouldn't_ need to do that.

Instead, I think we need to release versions of ipns/pubsub that don't use deprecated interfaces (those patches have been merged) and merge those into the libp2p daemon. We don't even need to release the daemon.

If that's not enough, we can introduce "next" branches (where these tests checkout "next" branches if they exist upstream). I though I did that, but I guess I reverted that change (probably to simplify for now).

marten-seemann:
> Instead, I think we need to release versions of ipns/pubsub that don't use deprecated interfaces (those patches have been merged) and merge those into the libp2p daemon.

Unfortunately, I already release [go-ipns v0.1.1](https://github.com/ipfs/go-ipns/releases/tag/v0.1.1) that also updates the dependencies.
I suggest disabling the interop workflow for now, until the loop has been resolved.

Stebalien:
The go-ipns dep needs to be updated in go-libp2p-daemon. I _really_ don't want to merge this without fixing CI. We introduced these checks so we could catch issues with large updates.

Stebalien:
So... we're going to need to _retract_ all of those breaking patch releases.

Stebalien:
Ok, we just cut new patch releases. Everything should be fixed now.

We should do a mini-retro; we don't have a foot-gun here, we have a minefield.

Stebalien (approved):

PR #1148, opened 2021-08-03
fix(basic_host): stream not closed when context done

godcong:
When i create some `NewStream`
A long time later,GetStreams() resulted many unused stream.
Finnally, I found the cause of the error


Stebalien:
The context governs _creation_ of the stream, not the lifetime of the stream.

Stebalien:
Oh, wait. Nvm. I should read code.

Stebalien (approved):

Stebalien:
Thanks for catching this!

PR #1161, opened 2021-08-17
implement the new mDNS spec, move the old mDNS implementation

### LINKED ISSUE #176: mdns is squating on host hostname
I've been debugging why jsipfs and go-ipfs stopped finding each other since we got it to work (end of 2015).

It seems that couple of things have changed in go-ipfs, starting with the `serviceTag` and the really critical change, `go-ipfs` now squats the host hostname to announce its records through mdns.

See that https://github.com/libp2p/go-libp2p/blob/master/p2p/discovery/mdns.go#L85 passes `""` and then https://github.com/whyrusleeping/mdns/blob/master/zone.go#L83-L91

This is the reason why I keep seeing my hostname change when I have an `ipfs daemon` running:
image

What happens is that (by definition of the mdns protocol), once there is another announcer using the same hostname, one of them has to change.

Unless I'm missing something, I don't see a need to take over the host hostname. 

PS: This 'fault' in the protocol is what enabled Watson to hack printers :) https://www.youtube.com/watch?v=BMWP20ZicsY

### LINKED ISSUE #253: mDNS discovery announces on the wrong interface
It seems like mDNS discovery only picks up the "first" interface that libp2p finds, i.e. usually the first one that's printed on ipfs startup. In my case, that's OpenVPN's tun0, but not the LAN's wifi0. Other interfaces aren't considered, so discovery is half broken in those cases.

We should probably apply some simple heuristics to decide which interfaces look like they're probably a LAN.

```
> IPFS_LOGGING=debug ipfs daemon |& grep mdns
```

### LINKED ISSUE #469: mDNS response
Using wireshark with ipfs version 0.4.18, I've discovered that IPFS sends the mDNS response to the query originator address NOT to `224.0.0.251`.

This means that other peers will not see the response and basically breaks the whole idea of passive monitoring.

### LINKED ISSUE #623: Implement new MDNS spec
That is, https://github.com/libp2p/specs/pull/80. We'll need to:

1. Use https://github.com/miekg/dns to handle parsing/formatting DNS records.
2. See https://github.com/grandcat/zeroconf/blob/85eadb44205c14827ae7cbb129bc7eed4488dd25/connection.go#L38-L97 for how to listen for multicast packets. However, we'll need to use https://github.com/libp2p/go-reuseport/blob/b72b23b78b80078df1a7991f6ca39942cb8cc5e1/interface.go#L47-L49 to fix multiple MDNS applications.

Unfortunately, none of the existing golang mdns solutions work for us as:

1. They all have race conditions.
2. None of them use reuseport.
3. All of them are very DNS-SD focused and go through a lot of effort to make sure we get things like A and AAAA records.
4. We need to use the system's mDNS responder when possible, as running multiple mDNS responders on the same machine is pretty broken.

We _could_ try to patch these however, we'll likely end up with something better if we just write something ourselves.

### LINKED ISSUE #894: mdns doesn't working on windows
https://github.com/hashicorp/mdns/issues/51

this project looks not maintained.
and it has many bug wasn't fixed
the mainest is its can't works fine in windows

i tested this fork
https://github.com/micro/mdns
it works fine in the windows

perhaps,your team can merged it into the project

marten-seemann:
Closes #1125. Fixes #176. Fixes #253. Fixes #623. Probably fixes #894. Probably fixes #469.


p2p/discovery/mdns.go
```diff
@@ -37,186 +36,183 @@ type Notifee interface {
...
 	notifees []Notifee
-	interval time.Duration
 }
 
-func getDialableListenAddrs(ph host.Host) ([]*net.TCPAddr, error) {
-	var out []*net.TCPAddr
-	addrs, err := ph.Network().InterfaceListenAddresses()
-	if err != nil {
-		return nil, err
+func NewMdnsService(host host.Host, serviceName string) *mdnsService {
```
Stebalien, line 54: Thank you for not doing strange things with contexts.

p2p/discovery/mdns.go
```diff
@@ -5,27 +5,26 @@ import (
 	"errors"
```
Stebalien, line 5: While we're breaking things, can we put this in an `mdns` package? I assume we'll introduce additional discovery mechanisms at some point (we've talked a lot about things like peer exchange).

I don't feel strongly about this.

p2p/discovery/mdns.go
```diff
@@ -37,186 +36,183 @@ type Notifee interface {
...
-	ticker := time.NewTicker(m.interval)
-	defer ticker.Stop()
-
-	for {
-		//execute mdns query right away at method call and then with every tick
-		entriesCh := make(chan *mdns.ServiceEntry, 16)
-		go func() {
-			for entry := range entriesCh {
-				m.handleEntry(entry)
+func (s *mdnsService) startResolver() {
```
Stebalien, line 157: nit: pass the context in as an argument instead of storing it in the struct?

p2p/discovery/mdns.go
```diff
@@ -37,186 +36,183 @@ type Notifee interface {
...
-
-		select {
-		case <-ticker.C:
-			continue
-		case <-ctx.Done():
-			log.Debug("mdns service halting")
-			return
+	}()
+	go func() {
+		defer close(s.resolverRunning)
```
Stebalien, line 191: Consider using a waitgroup. We'd be able to reuse it in for both goroutines.

p2p/discovery/mdns.go
```diff
@@ -37,186 +36,183 @@ type Notifee interface {
...
-
-	m.lk.Lock()
-	for _, n := range m.notifees {
-		go n.HandlePeerFound(pi)
-	}
-	m.lk.Unlock()
+func (s *mdnsService) RegisterNotifee(n Notifee) {
+	s.mutex.Lock()
+	s.notifees = append(s.notifees, n)
+	s.mutex.Unlock()
```
Stebalien, line 201: nit: defer, easier to read.

Stebalien: Eh, we can punt on this.

p2p/discovery/mdns.go
```diff
@@ -37,186 +36,183 @@ type Notifee interface {
...
+	ips, err := s.getIPs(addrs)
+	if err != nil {
+		return err
+	}
+
+	server, err := zeroconf.RegisterProxy(
+		s.mdnsInstance(),
+		s.serviceName,
+		mdnsDomain,
+		4001,
```
Stebalien, line 144: Can we leave a comment saying "this port doesn't actually matter, only the TXT records are used?"

Stebalien: Does this need to be lower case?

p2p/discovery_legacy/mdns.go
```diff
@@ -0,0 +1,215 @@
...
+	"time"
+
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p/p2p/discovery"
+
+	logging "github.com/ipfs/go-log/v2"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+	"github.com/whyrusleeping/mdns"
```
Stebalien, line 17: So... how hard would it be to use the _new_ zeroconf package with the _old_ protocol? The old package has a lot of race conditions, etc.

p2p/discovery_legacy/mdns.go
```diff
@@ -0,0 +1,215 @@
+package discovery_legacy
```
Stebalien, line 1: I assume this file is just copied.

p2p/discovery/mdns.go
```diff
@@ -37,186 +36,183 @@ type Notifee interface {
...
+		ips = append(ips, ip4)
 	}
-	if len(out) == 0 {
-		return nil, errors.New("failed to find good external addr from peerhost")
+	if ip6 != "" {
+		ips = append(ips, ip6)
 	}
-	return out, nil
+	if len(ips) == 0 {
+		return nil, errors.New("didn't find any IP addresses")
```
Stebalien, line 107: Do we actually want to fail in this case?

p2p/discovery/mdns.go
```diff
@@ -37,186 +36,183 @@ type Notifee interface {
...
-
-	var ipaddrs []net.IP
-	port := 4001
+func (s *mdnsService) mdnsInstance() string {
+	return string(s.host.ID())
+}
 
-	addrs, err := getDialableListenAddrs(peerhost)
+func (s *mdnsService) startServer() error {
+	interfaceAddrs, err := s.host.Network().InterfaceListenAddresses()
```
Stebalien, line 117: Can we file an issue to fix this? We should be listening to "address change" events, or polling, not advertising a static set of addresses indefinitely.

In terms of MVP, this is no worse than the current system, but we need to fix it.

Stebalien (commented):
I have some suggestions but nothing blocking.

p2p/discovery/mdns.go
marten-seemann: Done.

p2p/discovery/mdns.go
marten-seemann: Done.

p2p/discovery_legacy/mdns.go
marten-seemann: We could, but I really don't want to. This is a legacy package anyway, and I don't want to inadvertently break something here.

p2p/discovery_legacy/mdns.go
marten-seemann: Correct.

p2p/discovery/mdns.go
marten-seemann: What's the point in continuing if we don't have an IP address. Which address would we advertise?

p2p/discovery/mdns.go
marten-seemann: https://github.com/libp2p/go-libp2p/issues/1163

p2p/discovery/mdns.go
marten-seemann: Good point. Done.

p2p/discovery/mdns.go
marten-seemann: Done. Not sure what you're referring to when you ask about lower case.

Stebalien: Ah, sorry, commented on the wrong line. I was referring to the peer ID in the line below.

p2p/mdns/mdns.go
```diff
@@ -0,0 +1,216 @@
+package mdns
```
Stebalien, line 1: I was actually thinking discovery/mdns to better organize the packages. But this also works.

Stebalien (approved):
LGTM!

p2p/discovery_legacy/mdns.go
Stebalien: My concern is that it has some safety issues. But yeah, that's probably fine. You're right about it being legacy.

(maybe document this)

p2p/discovery/mdns.go
Stebalien: We might have other addresses (in some theoretical world...)? But you're right, that's probably not relevant for mdns.

p2p/mdns/mdns.go
marten-seemann: Yes, that's better. Let's do that.

Stebalien (approved):

melroy89:
Small feedback: https://github.com/libp2p/go-libp2p/issues/3232

PR #1165, opened 2021-08-18
remove secio from README, add noise

marten-seemann:
I'm aware that we're not using Travis any more, but that's not something I intend to fix in this PR.


Stebalien (approved):

Stebalien:
LGTM. Not sure what's up with macos.

PR #1168, opened 2021-08-21
signalling for hole punching

marten-seemann:
This is a copy of #1057, with a cleaned up commit history. I intend to continue working on this PR, and eventually close #1057.
Closes #1057.

Note: This PR targets the `hole-punching` branch, which is where we'll assemble all hole-punching related changes before merging into `master`.


p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,426 @@
...
+	maxRetries  = 5
+	retryWait   = 2 * time.Second
+)
+
+var (
+	log = logging.Logger("p2p-holepunch")
+)
+
+// The HolePunchService is used to make direct connections with a peer via hole-punching.
+type HolePunchService struct {
```
marten-seemann, line 42: This should probably just be named `Service`, since the package name is already `holepunch`.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,426 @@
...
+	for _, c := range hs.host.Network().ConnsToPeer(rp) {
+		if !isRelayAddress(c.RemoteMultiaddr()) {
+			return nil
+		}
+	}
+
+	// short-circuit hole punching if a direct dial works.
+	// attempt a direct connection ONLY if we have a public address for the remote peer
+	for _, a := range hs.host.Peerstore().Addrs(rp) {
+		if manet.IsPublicAddr(a) && !isRelayAddress(a) {
```
marten-seemann, line 111: This logic doesn't seem right. If the `Connect` succeeds, we `return` from the function. If it doesn't, we `break` the loop. That means we never process more than just the first address.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,426 @@
...
+// Close closes the Hole Punch Service.
+func (hs *HolePunchService) Close() error {
+	hs.ctxCancel()
+	hs.refCount.Wait()
+	return nil
+}
+
+// attempts to make a direct connection with the remote peer of `relayConn` by co-ordinating a hole punch over
+// the given relay connection `relayConn`.
+func (hs *HolePunchService) HolePunch(rp peer.ID) error {
```
marten-seemann, line 100: This probably should not be name `HolePunch`, since it doesn't always punch a hole.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,407 @@
...
+	logging "github.com/ipfs/go-log"
+	"github.com/libp2p/go-msgio/protoio"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+// TODO Should we have options for these ?
+var (
+	// Protocol is the libp2p protocol for Hole Punching.
+	Protocol protocol.ID = "/libp2p/holepunch/1.0.0"
```
mxinden, line 27:
```suggestion
	Protocol protocol.ID = "/libp2p/dcutr"
```

Note that the current specification uses the protocol identifier above.

https://github.com/libp2p/specs/blob/master/relay/DCUtR.md#the-protocol

marten-seemann: Thanks. Please note that the PR in its current state is nowhere near review-ready.

marten-seemann:
Tests (with race detector) failing due to https://github.com/libp2p/go-libp2p-autonat/pull/109. We'll need to update autonat before we can merge this PR.

marten-seemann:
Unfortunately, the changes in https://github.com/libp2p/go-libp2p-autonat/pull/109 were not enough to prevent race conditions here, as for setting up the relay connection, we need to mark all addresses as public, and then for hole punching we need to mark these addresses as private again, while the host is running. This is inherently racy.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+
+	h.SetStreamHandler(Protocol, hs.handleNewStream)
+	h.Network().Notify((*netNotifiee)(hs))
+	return hs, nil
+}
+
+// Close closes the Hole Punch Service.
+func (hs *Service) Close() error {
+	hs.ctxCancel()
+	hs.refCount.Wait()
```
Stebalien, line 93: This can panic because the count can go from N->0->M. 

I'm not sure if this service actually needs to be closable, honestly. It doesn't do anything in the background and will stop automatically when the host is closed.

But if we do want to do this, I'd protect `wg.Add` with a lock (checking to see if we've been "canceled"). This is also a good use-case for conditions (wait until the number of active jobs has reached zero, and broadcast whenever a job is finished).

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+		return
+	}
+	rtt := time.Since(tstart)
+
+	if t := msg.GetType(); t != pb.HolePunch_SYNC {
+		s.Reset()
+		hs.handlerError(rp, fmt.Errorf("expected SYNC message from initiator but got %d", t))
+		return
+	}
+	defer s.Close()
```
Stebalien, line 268: Any reason to defer this?

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+	hs.tracer.StartHolePunch(rp, obsDial, rtt)
+	log.Debugf("starting hole punch with %s", rp)
+	counter := atomic.AddInt32(&hs.counter, 1)
+	tstart = time.Now()
+	err := hs.holePunchConnect(pi, int(counter))
+	dt := time.Since(tstart)
+	hs.tracer.EndHolePunch(rp, dt, err)
+	if err != nil {
+		log.Warnf("hole punching with %s failed after %s: %s", rp, dt, err)
```
Stebalien, line 283: This is at most an info, probably a debug.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+	} else {
+		log.Infof("hole punching with %s successful after %s", rp, dt)
```
Stebalien, line 285: Definitely debug.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+		s.Reset()
+		hs.handlerError(rp, fmt.Errorf("failed to write CONNECT message to initator:: %w", err))
+		return
+	}
+
+	// Read SYNC message
+	msg.Reset()
+	if err := rd.ReadMsg(msg); err != nil {
+		s.Reset()
+		hs.handlerError(rp, fmt.Errorf("failed to read message from initator: %w", err))
```
Stebalien, line 258: Can we defer these? Even better, use a helper function and reset & log the error if the helper function fails.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+
+		if active {
+			return
+		}
+
+		log.Debugf("got inbound proxy conn from peer %s, connectionID is %s", v.RemotePeer().String(), v.ID())
+		hs.refCount.Add(1)
+		go func() {
+			defer hs.refCount.Done()
+			defer func() {
```
Stebalien, line 362: Combine?

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+			select {
+			// waiting for Identify here will allow us to access the peer's public and observed addresses
+			// that we can dial to for a hole punch.
+			case <-hs.ids.IdentifyWait(v):
+			case <-hs.ctx.Done():
+				return
+			}
+
+			if err := hs.HolePunch(v.RemotePeer()); err != nil {
+				log.Warnf("hole punching attempt with %s failed: %s", v.RemotePeer(), err)
```
Stebalien, line 376: Noise.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+	return fmt.Errorf("all retries for hole punch with peer %s failed", rp)
+}
+
+func (hs *Service) handlerError(p peer.ID, err error) {
+	hs.tracer.ProtocolError(p, err)
+	log.Warn(err)
+}
+
+func (hs *Service) handleNewStream(s network.Stream) {
```
Stebalien, line 222: We're about to dial arbitrary addresses, we should do some more checks:

1. Make sure that the connection on which we received the request is a relay.
2. Maybe make sure that it's an _outbound_ dial? It looks like we've defined the connection receiver as the side that initiates the upgrade.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+	return (*Service)(nn)
+}
+
+func (nn *netNotifiee) Connected(_ network.Network, v network.Conn) {
+	hs := nn.HolePunchService()
+	dir := v.Stat().Direction
+
+	// Hole punch if it's an inbound proxy connection.
+	// If we already have a direct connection with the remote peer, this will be a no-op.
+	if dir == network.DirInbound && isRelayAddress(v.RemoteMultiaddr()) {
```
Stebalien, line 345: We should _probably_ bail early if we're _known_ to be dialable. This would largely mitigate a lot of DoS concerns.

Also, is there some discussion on picking a side to initiate? On one hand, having the receiver initiate gives them a chance to try a "normal" dial. On the other hand, it means the receiver may do a _lot_ of work on behalf of a potentially malicious party.

Stebalien (commented):
Mostly coding things, but there are some design questions around when to initiate a hole-punch and when to bail.

p2p/protocol/holepunch/tracer.go
```diff
@@ -0,0 +1,181 @@
+package holepunch
```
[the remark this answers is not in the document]
marten-seemann: I don't think so. We'll want to have Prometheus monitoring for hole punching, and this will be a lot easier using a tracer, won't it?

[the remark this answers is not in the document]
Stebalien: Hm. I see. Yeah, probably.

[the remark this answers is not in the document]
Stebalien: Well, really, shouldn't we just be using opentelemetry meters directly?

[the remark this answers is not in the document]
marten-seemann: I think it's cleaner with a tracer, if only for separating Prometheus code from the application logic.

p2p/protocol/holepunch/pb/holepunch.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto2";
+
+package holepunch.pb;
+
+message HolePunch {
+  enum Type {
+    CONNECT = 100;
+    SYNC = 300;
```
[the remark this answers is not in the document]
marten-seemann: Let's keep the docs in the spec: https://github.com/libp2p/specs/blob/master/relay/DCUtR.md#rpc-messages.

p2p/protocol/holepunch/pb/holepunch.proto
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  optional Type type=1;
```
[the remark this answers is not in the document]
marten-seemann: This is taken straight from the spec: https://github.com/libp2p/specs/blob/master/relay/DCUtR.md#rpc-messages.

p2p/protocol/holepunch/pb/holepunch.proto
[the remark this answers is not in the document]
Stebalien: Fair enough. Can we link to it?

p2p/protocol/holepunch/pb/holepunch.proto
```diff
@@ -0,0 +1,19 @@
...
+
+  // For hole punching, we'll send some additional observed addresses to the remote peer
+  // that could have been filtered by the Host address factory (for example: AutoRelay removes all public addresses if peer has private reachability).
+  // This is a hack!
+  // We plan to have a better address discovery and advertisement mechanism in the future.
+  // See https://github.com/libp2p/go-libp2p-autonat/pull/98
+  repeated bytes ObsAddrs = 2;
```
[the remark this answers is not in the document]
marten-seemann: Yes, these are _our addresses_, that other peers have observed. I agree on the naming not being ideal, but we should probably fix this in the spec.

Agreed that this is not a hack. Explicitly coordinating the addresses you're expecting a hole punch on makes a lot of sense. I removed the comment.

p2p/protocol/holepunch/tracer.go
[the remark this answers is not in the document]
Stebalien: Fair enough.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+		if err == nil {
+			addrs = append(addrs, a)
+		}
+	}
+	return addrs
+}
+
+type netNotifiee Service
+
+func (nn *netNotifiee) HolePunchService() *Service {
```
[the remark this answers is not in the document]
marten-seemann: No need for this function at all, actually. We can just type assert in `Connected`.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+			dialCtx, cancel := context.WithTimeout(forceDirectConnCtx, dialTimeout)
+
+			tstart := time.Now()
+			err := hs.host.Connect(dialCtx, peer.AddrInfo{ID: rp})
+			dt := time.Since(tstart)
+			cancel()
+
+			if err != nil {
+				hs.tracer.DirectDialFailed(rp, dt, err)
+				break
```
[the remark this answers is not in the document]
marten-seemann: Actually, we haven't tried all addresses. I _assume_ that the reason is that we don't want to waste any more time on dialing addresses if one attempt already failed.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+			return err
+		}
+		synTime := rtt / 2
+		log.Debugf("peer RTT is %s; starting hole punch in %s", rtt, synTime)
+
+		// wait for sync to reach the other peer and then punch a hole for it in our NAT
+		// by attempting a connect to it.
+		timer := time.NewTimer(synTime)
+		select {
+		case <-timer.C:
```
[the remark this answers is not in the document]
marten-seemann: Fixed.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+	"github.com/libp2p/go-msgio/protoio"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+// Protocol is the libp2p protocol for Hole Punching.
+const Protocol protocol.ID = "/libp2p/dcutr"
+
+// StreamTimeout is the timeout for the hole punch protocol stream.
+var StreamTimeout = 1 * time.Minute
```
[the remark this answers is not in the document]
marten-seemann: Don't we do this _everywhere_ in libp2p? But I agree with you, config options are better.

p2p/protocol/holepunch/coordination.go
marten-seemann: No. Fixed.

p2p/protocol/holepunch/coordination.go
marten-seemann: We could, but I find it clearer to see what happens first and what happens last this way.

p2p/protocol/holepunch/coordination.go
marten-seemann: 1. Makes sense, I added this check.
2. It's the peer who's not behind the relay who's initiating the upgrade, so that won't work.

p2p/protocol/holepunch/coordination.go
marten-seemann: Yes, good point. I didn't like `handlerError` to begin with.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+		hs.activeMx.Unlock()
+
+		if active {
+			return
+		}
+
+		log.Debugf("got inbound proxy conn from peer %s, connectionID is %s", v.RemotePeer().String(), v.ID())
+		hs.refCount.Add(1)
+		go func() {
```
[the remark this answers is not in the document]
marten-seemann: Should `HolePunch` be exported? We're only calling it from `Connected` in this package, so we could easily un-export it until some need arises for other packages to call it.

p2p/protocol/holepunch/coordination.go
marten-seemann: > I'm not sure if this service actually needs to be closable, honestly. It doesn't do anything in the background

It does. `HolePunch` is called for a separate go routine, and it sleeps for `rtt /2`. Note that `rtt` can be large. We need to be able to cancel that sleeping period.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,388 @@
...
+	if err := w.WriteMsg(msg); err != nil {
+		str.Reset()
+		return nil, 0, fmt.Errorf("failed to send SYNC message for hole punching: %w", err)
+	}
+	return addrs, rtt, nil
+}
+
+// attempts to make a direct connection with the remote peer of `relayConn` by co-ordinating a hole punch over
+// the given relay connection `relayConn`.
+func (hs *Service) HolePunch(rp peer.ID) error {
```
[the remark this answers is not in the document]
marten-seemann: Renamed to `DirectConnect`.

p2p/protocol/holepunch/coordination.go
marten-seemann: > Also, is there some discussion on picking a side to initiate? On one hand, having the receiver initiate gives them a chance to try a "normal" dial. On the other hand, it means the receiver may do a lot of work on behalf of a potentially malicious party.

The node behind the NAT is the one dialing. The reason for that is that there's a good change that the peer is not behind a NAT, and we can just dial him directly. Only if the peer is not dialable, we have to fall back to hole punching.

p2p/protocol/holepunch/coordination.go
marten-seemann: > This is also a good use-case for conditions (wait until the number of active jobs has reached zero, and broadcast whenever a job is finished).

I just used a bool and a mutex, not sure if the condition would actually make things easier here (but maybe I'm just misunderstanding your proposal).

p2p/protocol/holepunch/coordination.go
marten-seemann: > We should probably bail early if we're known to be dialable. This would largely mitigate a lot of DoS concerns.

Then the remote address wouldn't be a relay address, would it?

marten-seemann: To clarify, we're looking at the remote address being a relay address, AND the connection being an inbound connection.
If we know to be publicly dialable, we shouldn't be behind a relay in the first place.

p2p/protocol/holepunch/coordination.go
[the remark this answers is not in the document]
marten-seemann: Unfortunately, this is really handy in the tests.

p2p/protocol/holepunch/coordination.go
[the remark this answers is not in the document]
Stebalien: I could see cases where exporting it might be useful, but we can probably just punt that till we actually need it.

p2p/protocol/holepunch/coordination.go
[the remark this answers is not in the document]
Stebalien: `Connect` will try all addresses in the peerstore. Basically, we're just checking to see if we _have_ a viable address. If we do, we ask the host to try to connect to the peer directly.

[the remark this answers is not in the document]
Stebalien: (I wanted a comment here because, as you can tell, this is _really_ non-obvious)

p2p/protocol/holepunch/coordination.go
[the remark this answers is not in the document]
Stebalien: > Don't we do this everywhere in libp2p? But I agree with you, config options are better.

We do, I'd just like to do it less often.

> Unfortunately, this is really handy in the tests.

Can we keep it private? And/or is there an example of a case where an option won't work?

p2p/protocol/holepunch/coordination.go
Stebalien: > It's the peer who's not behind the relay who's initiating the upgrade, so that won't work.

https://github.com/libp2p/go-libp2p/pull/1168/files#diff-c550e388063ff4af6c56eca2bd1e8d5f32a59ed74a481130884d8620c61a7ff8R337

That is:

1. Peer A receives an _inbound_ connection from peer B.
2. Peer A opens a new "hole punch" stream to peer B.
3. Peer B receives this stream, calling this function.
4. Peer B sees the underlying connection as an outbound connection.

(unless there are multiple connections)

p2p/protocol/holepunch/coordination.go
Stebalien: > The node behind the NAT is the one dialing. The reason for that is that there's a good change that the peer is not behind a NAT, and we can just dial him directly. Only if the peer is not dialable, we have to fall back to hole punching.

Makes sense.

>  If we know to be publicly dialable, we shouldn't be behind a relay in the first place.

That doesn't _stop_ someone from making a relayed connection. We'll accept those regardless (there are some valid use-cases for this, e.g., no common transport).

To know if we're publicly dialable, we need to check our "nat" status (subscribe to the reachability event).

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,386 @@
...
+		if !active && !closed {
+			hs.refCount.Add(1)
+			hs.active[p] = struct{}{}
+		}
+		hs.closeMx.RUnlock()
+		hs.activeMx.Unlock()
+
+		if active || closed {
+			return
+		}
```
Stebalien, line 352: Wrapping this all in a single `beginHolePunch` function may make it simpler.

p2p/protocol/holepunch/coordination.go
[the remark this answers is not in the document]
marten-seemann: > And/or is there an example of a case where an option won't work?

An option would work, but I doubt this is something anyone would like to configure.

> Can we keep it private?

Nope, unless we want to create a separate `internal` package for that. The tests need to use a separate package to avoid an import loop.

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,386 @@
...
+			_ = hs.DirectConnect(v.RemotePeer())
+		}()
+		return
+	}
+}
+
+func (nn *netNotifiee) Disconnected(_ network.Network, v network.Conn) {
+	if v.Stat().Direction == network.DirInbound && isRelayAddress(v.RemoteMultiaddr()) {
+		hs := (*Service)(nn)
+		hs.tracer.Cleanup(v.RemotePeer())
```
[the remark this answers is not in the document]
marten-seemann: I had hoped to avoid another `gc` function (with associated shutdown logic), but it seems like that's actually the cleaner way here.

p2p/protocol/holepunch/coordination.go
marten-seemann: That makes sense. Let's do that in a separate PR (this PR is targeting a feature-branch anyway).

p2p/protocol/holepunch/coordination.go
marten-seemann: Makes sense, and it requires the attacker to be at least _a little bit_ more sophisticated than just connecting to nodes and asking them to dial others.

p2p/protocol/holepunch/coordination.go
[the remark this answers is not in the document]
marten-seemann: Done.

p2p/protocol/holepunch/coordination.go
marten-seemann: Oh yes, it does!

p2p/protocol/holepunch/coordination.go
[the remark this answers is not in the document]
marten-seemann: We need to call this function from the tests (which use a separate package). I've therefore refactored the code such that the deduplication logic now lives in the holepunch service, as you suggested earlier.

p2p/protocol/holepunch/coordination.go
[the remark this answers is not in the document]
Stebalien: > An option would work, but I doubt this is something anyone would like to configure.

Yeah... I'm just worried about making it a part of the public API. But it's not too terrible.

Stebalien (approved):

p2p/protocol/holepunch/coordination.go
```diff
@@ -0,0 +1,404 @@
...
+
+	h.SetStreamHandler(Protocol, hs.handleNewStream)
+	h.Network().Notify((*netNotifiee)(hs))
+	return hs, nil
+}
+
+// Close closes the Hole Punch Service.
+func (hs *Service) Close() error {
+	hs.closeMx.Lock()
+	hs.closed = true
```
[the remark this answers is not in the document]
Stebalien: I agree. My point is that the tracer's close function could get called multiple times, and it may not be expecting that.

Of course, it should just "deal" with it, so we're probably fine.

PR #1169, opened 2021-08-24
clean up messy defer logic in IDService.sendIdentifyResp

marten-seemann:
I believe we don't need to hold the lock while closing the stream.


p2p/protocol/identify/id.go
```diff
@@ -413,6 +407,7 @@ func (ids *IDService) sendIdentifyResp(s network.Stream) {
 
 	ph.snapshotMu.RLock()
 	ids.writeChunkedIdentifyMsg(c, ph.snapshot, s)
+	ph.snapshotMu.RUnlock()
```
Stebalien, line 410: nit: defer this.

Stebalien (approved):
I really can't remember _why_ we did it this way...

Stebalien:
Actually, no, this is still wrong. We should be extracting the snapshot, then sending it, not holding the lock while we write to a stream.

PR #1058, opened 2021-02-19
Integrate Limited Relays

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/1039

PR #1174, opened 2021-09-02
move the circuit v2 code here

marten-seemann:
As discussed in https://github.com/libp2p/go-libp2p-circuit/pull/138, the circuit v2 implementation should live in go-libp2p, not in a separate package.

Commit history extracted from the original repo by running:
```bash
echo "github.com/libp2p/go-libp2p-circuit/v2==>github.com/libp2p/go-libp2p/p2p/host/circuitv2" >> replace.txt
git filter-repo --force --path v2/ --path-rename v2:p2p/host/circuitv2 --replace-text replace.txt
```
and then running
```bash
git rebase --committer-date-is-author-date origin/master --exec "go mod tidy"
```
as I couldn't figure out how to have `git filter-repo` do that in one step without writing a Python script.


marten-seemann:
After offline sync: we agreed to move . I suggest to keep the package name `circuitv2` for now. #1058 will allow us to deprecate go-libp2p-circuit, and we can rename the package to `circuit` then.

Stebalien:
This doesn't really belong in `host`, that's for `host` implementations. "relay" is there because it used to be a host wrapper (it really should move).

Maybe move _both_ to a "relay" directory? Really, the "relay" belongs in some form of "relay" package, and the "client" part belongs in a "transports" package, but we don't need to get too fancy.

marten-seemann:
Looks like I was too fast merging this PR. I opened #1177.

PR #1175, opened 2021-09-05
don't use a context for closing the ObservedAddrManager

marten-seemann:
This is part of the ongoing effort to stop using contexts to control shutdowns.
Most importantly, this PR removes a dependency on the swarm's `Process` in https://github.com/libp2p/go-libp2p/blob/fd4d2a47e21fc516dc97fa0077c2b8a87c35d2c3/p2p/protocol/identify/obsaddr.go#L257.

### DIFF
--- a/p2p/protocol/identify/obsaddr.go
+++ b/p2p/protocol/identify/obsaddr.go
@@ -109,7 +109,8 @@ type ObservedAddrManager struct {
 	// active connection -> most recent observation
 	activeConns map[network.Conn]ma.Multiaddr
 
-	mu sync.RWMutex
+	mu     sync.RWMutex
+	closed bool
 	// local(internal) address -> list of observed(external) addresses
 	addrs        map[string][]*observedAddr
 	ttl          time.Duration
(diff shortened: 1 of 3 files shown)


p2p/protocol/identify/obsaddr.go
```diff
@@ -534,6 +527,17 @@ func (oas *ObservedAddrManager) emitSpecificNATType(addrs []*observedAddr, proto
 	return false, 0
 }
 
+func (oas *ObservedAddrManager) Close() error {
+	oas.closeOnce.Do(func() {
+		oas.ctxCancel()
+		oas.refCount.Wait()
+		oas.reachabilitySub.Close()
+		oas.refreshTimer.Stop()
```
Stebalien, line 535: 1. I believe we still need to hold the lock here because `SetTTL` is a public function.
2. We should _probably_ nil out the `refreshTimer`, and check if it's nil in `SetTTL`.

Stebalien (changes requested):

p2p/protocol/identify/obsaddr.go
marten-seemann: I was assuming that you only call `Close` when you're done using the struct, and don't concurrently call `Close` with other methods.
On the other hand, locking the mutex here is cheap, so why not...

Stebalien: Ideally, but it can be hard to really guarantee that.

Stebalien (approved):

p2p/protocol/identify/obsaddr.go
```diff
@@ -534,6 +527,22 @@ func (oas *ObservedAddrManager) emitSpecificNATType(addrs []*observedAddr, proto
 	return false, 0
 }
 
+func (oas *ObservedAddrManager) Close() error {
+	oas.closeOnce.Do(func() {
+		oas.ctxCancel()
+
+		oas.mu.Lock()
+		oas.refreshTimer.Stop()
+		oas.refreshTimer = nil
```
marten-seemann, line 536: This doesn't work as the `worker` might be blocked on the timer channel, creating a race condition.

Stebalien: Eh, we could have a "shutdown" boolean. That or have the worker keep a copy of the timer and/or channel locally instead of reading `oas.refreshTimer`.

Stebalien (approved):

Issue #1177, opened 2021-09-07
move circuitv2 and relay packages


marten-seemann:
This doesn't really belong in `host`, that's for `host` implementations. "relay" is there because it used to be a host wrapper (it really should move).

Maybe move _both_ to a "relay" directory? Really, the "relay" belongs in some form of "relay" package, and the "client" part belongs in a "transports" package, but we don't need to get too fancy.

_Originally posted by @Stebalien in https://github.com/libp2p/go-libp2p/issues/1174#issuecomment-914304959_

vyzo:
We should have it in `protocol` I think, it doesn't belong in `host` and we don't need a separate package for this.

vyzo:
Note that `relay` contains the autorelay code, it should probably be renamed as such and maybe move into `protocol` as well.

marten-seemann:
This should have been resolved since v0.16.0.

PR #1183, opened 2021-09-10
Use circuitv2 code

vyzo:
This uses the (backwards compatible) circuit v2 client for handling relay connections.

Relay options are removed; relay servers should be explicitly instantiated or use a dedicated daemon (e.g. p2pd or relayd).
Also moves the `circuitv2` code into `p2p/protocol` (see #1177)


config/config.go
```diff
@@ -283,12 +274,7 @@ func (cfg *Config) NewNode(ctx context.Context) (host.Host, error) {
 
 			discovery := discovery.NewRoutingDiscovery(crouter)
 
-			if hop {
-				// advertise ourselves
-				relay.Advertise(ctx, discovery)
-			} else {
-				_ = relay.NewAutoRelay(ctx, h, discovery, router, cfg.StaticRelays)
-			}
+			_ = relay.NewAutoRelay(ctx, h, discovery, router, cfg.StaticRelays)
```
marten-seemann, line 277: I like this. This will make it a lot easier to implement a `Close` method here, and not rely on the `ctx` any more.

marten-seemann (approved):

PR #1184, opened 2021-09-10
don't use the context to shut down the relay

marten-seemann:
Not a big fan of this change. I'd have preferred to save the `AutoRelay` on the `BasicHost`, but that's not possible (import cycle).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -22,7 +22,7 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/host/relay"
 	routed "github.com/libp2p/go-libp2p/p2p/host/routed"
 	circuitv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/client"
-	holepunch "github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
+	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
 
 	autonat "github.com/libp2p/go-libp2p-autonat"
 	blankhost "github.com/libp2p/go-libp2p-blankhost"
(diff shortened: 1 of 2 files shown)


vyzo (commented):
I don't like this proliferation of host wrappers either -- can't we just save the `close` function in the host and invoke that?
We can even compose multiple close functions into one this way for all dependent services.

marten-seemann:
I thought of that, but I’m not sure if having a SetClose(closeFunc) method on the Host is nicer.

vyzo:
we dont have to expose it in the interface; the constructor can just
collect what it needs. It can also be a slice of io.Closers that we store.

vyzo:
Another, more convoluted abstraction is to track services by unique keys
and expose interface methods for managing that. This can also eliminate
atrocities like the basic host cast in lotus just to get to autonat.

But thats probably a more general discussion to be held here.

marten-seemann:
We can’t put it in the constructor, as the autonat depends on the host.

Let’s keep wrapping for now, I think that’s the more idiomatic way to handle this in Go anyway.

vyzo:
I meant in the libp2p constructor, *not* the object constructor; i'd rather
avoid the wrapping.

vyzo (approved):
actually I'll give the blessing; I am allergic to host wrapping, but on the same time I can't think of any really good reason not to wrap, so let's do it.

We should have the discussion about services at some point, so that we can eliminate wrapping altogether in the future.

Stebalien:
Can I not pass in a set of abstract "service" constructors into the host? Of the form `func(Host) (io.Closer, error)`?

(note: fx/dependency injection would also fix this, we really need a way to "manage" libp2p services).

marten-seemann:
> Can I not pass in a set of abstract "service" constructors into the host? Of the form func(Host) (io.Closer, error)?

I'd love to do this. We just need to figure out a way to pass options to that constructor.

PR #1186, opened 2021-09-14
Add standalone implementation of v1 Relay

vyzo:
This adds an implementation of a v1 Relay, extracted and slightly modified for a bit of resource management, from go-libp2p-circuit, intended to be used by standalone daemon implementations and tests.
This allows us to deprecate and completely dereference the go-libp2p-circuit repo.

TBD:
- [X] Remove references to go-libp2p-circuit from production code.
- [X] Update tests to use the in-tree v1 relay.
- [X] Update examples to use the in-tree v1 relay.


marten-seemann:
Do we have to do this? Can't we just leave the code in go-libp2p-circuit for the transition period? Then we wouldn't have to review a 1340 LOC PR, but could just use code that we know works in production.

vyzo:
The bultin implementation of go-libp2p-circuit is deficient; it cant be
used safely as only a relay and has resource management problems.
This remedies the situation, by extracting and fixing.

I would very much do this and kill the other repo. Originally I was going
to put the code in the relay daemon, but I decided to put it here for
generality and completeness.

Review should be easy, the mods are small and the code is similar to the v2
relay code. There is nothing exotic to consider. The bulk of the code is
protobuf, review is neede for some 200loc.

marten-seemann (approved):
I haven't done an in-depth review, since this is mostly code from go-libp2p-circuit.

Why is go-libp2p-circuit not removed from `go.mod` in this PR?

vyzo:
still have some open todos, see the pr description. I plan to finish it
today.

vyzo:
Turns out we need to keep a reference in go.mod for now, we have a compatiblity test which tests whether the v2 client can connect to legacy clients.
So we'll have to keep it for now, unless we want to remove that test (I find it useful however).

vyzo:
marked as ready -- @marten-seemann care for another pass before we merge?

marten-seemann (approved):

PR #1188, opened 2021-09-15
Tag relay hops in relay implementations

vyzo:
On top of #1186 
In both relay v1 and v2 implementations, with a fixed value of 2. The value is purposefully very low, serving as an indicator to distinguish idle connections from connections with hop circuits during connection manager prunes.

### DIFF
--- a/p2p/protocol/circuitv1/relay/relay.go
+++ b/p2p/protocol/circuitv1/relay/relay.go
@@ -29,8 +29,8 @@ const (
 	ConnectTimeout   = 30 * time.Second
 	HandshakeTimeout = time.Minute
 
-	RelayHopTag    = "relay-hop"
-	MaxRelayHopTag = 5
+	RelayHopTag      = "relay-v1-hop"
+	RelayHopTagValue = 2
 
 	maxMessageSize = 4096
 )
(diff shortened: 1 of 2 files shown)


p2p/protocol/circuitv1/relay/relay.go
```diff
@@ -262,6 +265,29 @@ func (r *Relay) handleHopStream(s network.Stream, msg *pb.CircuitRelay) {
 	go r.relayConn(bs, s, dest.ID, src.ID, done)
 }
 
+func (r *Relay) addConn(p peer.ID) {
+	conns := r.conns[p]
+	conns++
+	r.conns[p] = conns
```
Stebalien, line 271: Is this to set a limit on the tag value?

vyzo: yeah.

vyzo: we can also do a (very small) fixed tag value if you prefer.

vyzo: Upon further thought, I think I'll make it a fixed tag with a value 2; base utility is to distinguish idle connections from connections that have hop circuits on them.

vyzo: done.

Stebalien (approved):

marten-seemann (approved):

p2p/protocol/circuitv1/relay/relay.go
```diff
@@ -29,6 +29,9 @@ const (
 	ConnectTimeout   = 30 * time.Second
 	HandshakeTimeout = time.Minute
 
+	RelayHopTag      = "relay-v1-hop"
+	RelayHopTagValue = 2
```
marten-seemann, line 33: Nit: make these private.

vyzo:
rebased on master to resolve the (trivial) conflict; also made the hop tag constants private.

p2p/protocol/circuitv1/relay/relay.go
vyzo: done.

PR #1193, opened 2021-09-19
remove goprocess from the NATManager

marten-seemann:


marten-seemann:
libp2p/go-libp2p-nat#38 was merged, so this PR should be ready now.

Stebalien (approved):

PR #1197, opened 2021-09-23
add an option to start the relay v2

marten-seemann:
I'm not satisfied with the names for the config options, but I struggle to come up with something better. Suggestions would be highly appreciated.

I wish we could use `EnableRelay()` here, but unfortunately this is already taken to enable the *relay transport*.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -21,6 +21,7 @@ import (
 	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
 	routed "github.com/libp2p/go-libp2p/p2p/host/routed"
 	circuitv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/client"
+	relayv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
 	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
 
 	autonat "github.com/libp2p/go-libp2p-autonat"
@@ -74,7 +75,10 @@ type Config struct {
 	PSK                pnet.PSK
 
 	RelayCustom bool
-	Relay       bool
+	Relay       bool // should the relay transport be used
+
+	EnableRelayService bool // should we run a circuitv2 relay (if publicly reachable)
+	RelayServiceOpts   []relayv2.Option
 
 	ListenAddrs     []ma.Multiaddr
 	AddrsFactory    bhost.AddrsFactory
@@ -196,6 +200,8 @@ func (cfg *Config) NewNode() (host.Host, error) {
 		MultiaddrResolver:   cfg.MultiaddrResolver,
 		EnableHolePunching:  cfg.EnableHolePunching,
 		HolePunchingOptions: cfg.HolePunchingOptions,
+		EnableRelayService:  cfg.EnableRelayService,
+		RelayServiceOpts:    cfg.RelayServiceOpts,
 	})
 	if err != nil {
 		swrm.Close()
--- a/options.go
+++ b/options.go
@@ -17,9 +17,10 @@ import (
 	"github.com/libp2p/go-libp2p-core/pnet"
 
 	"github.com/libp2p/go-libp2p/config"
-	autorelay "github.com/libp2p/go-libp2p/p2p/host/autorelay"
+	"github.com/libp2p/go-libp2p/p2p/host/autorelay"
 	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
-	holepunch "github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
+	relayv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
+	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
 
 	ma "github.com/multiformats/go-multiaddr"
 	madns "github.com/multiformats/go-multiaddr-dns"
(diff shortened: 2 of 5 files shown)


config/config.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/host/relay"
 	routed "github.com/libp2p/go-libp2p/p2p/host/routed"
 	circuitv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/client"
+	v2relay "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
```
vyzo, line 24: let's call this `relayv2` for consistency.

config/config.go
```diff
@@ -74,7 +75,10 @@ type Config struct {
 	PSK                pnet.PSK
 
 	RelayCustom bool
-	Relay       bool
+	Relay       bool // should the relay transport be used
+
+	DisableRelay bool // should we run a circuitv2 relay (if publicly reachable)
+	RelayOpts    []v2relay.Option
```
vyzo, line 81: `RelayService` sounds like a better name, less ambiguous.

config/config.go
```diff
@@ -196,6 +200,8 @@ func (cfg *Config) NewNode() (host.Host, error) {
 		MultiaddrResolver:   cfg.MultiaddrResolver,
 		EnableHolePunching:  cfg.EnableHolePunching,
 		HolePunchingOptions: cfg.HolePunchingOptions,
+		EnableV2Relay:       !cfg.DisableRelay,
+		RelayOpts:           cfg.RelayOpts,
```
vyzo, line 204: `EnableRelayService` and `RelayServiceOptions` perhaps.

options.go
```diff
@@ -19,7 +19,8 @@ import (
 	"github.com/libp2p/go-libp2p/config"
 	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
 	autorelay "github.com/libp2p/go-libp2p/p2p/host/relay"
-	holepunch "github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
+	v2relay "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
```
vyzo, line 22: `relayv2`

options.go
```diff
@@ -231,6 +232,26 @@ func DisableRelay() Option {
 	}
 }
 
+// DisableRelaying configures libp2p to not run a circuit v2 relay.
+// Note that we only start the relay if we detect that we're publicly reachable.
+func DisableRelaying() Option {
```
vyzo, line 237: `DisalbeRelayService`

options.go
```diff
@@ -231,6 +232,26 @@ func DisableRelay() Option {
...
+	return func(cfg *Config) error {
+		cfg.DisableRelay = true
+		cfg.RelayOpts = nil
+		return nil
+	}
+}
+
+// EnableRelaying configures libp2p to run a circuit v2 relay,
+// if we dected that we're publicly reachable.
+func EnableRelaying(opts ...v2relay.Option) Option {
```
vyzo, line 247: `EnableRelayService`

p2p/host/basic/basic_host.go
```diff
@@ -133,6 +137,11 @@ type HostOpts struct {
 	// EnablePing indicates whether to instantiate the ping service
 	EnablePing bool
 
+	// EnableV2Relay enables the circuit v2 relay (if we're publicly reachable).
+	EnableV2Relay bool
+	// RelayOpts are options for the circuit v2 relay.
+	RelayOpts []v2relay.Option
```
vyzo, line 143: `EnableRelayService` and `RelayServiceOpts`.

vyzo (commented):
Yeah, these names are very confusing.

I have suggested some alternatives.

marten-seemann:
`RelayService` sounds good to me! Will update the PR.

config/config.go
```diff
@@ -74,7 +75,10 @@ type Config struct {
 	PSK                pnet.PSK
 
 	RelayCustom bool
-	Relay       bool
+	Relay       bool // should the relay transport be used
+
+	DisableRelayService bool // should we run a circuitv2 relay (if publicly reachable)
```
vyzo, line 80: let's heed to @stebalien's request to not enable it by default, so this should be `EnableRelayService` I guess and the logic inverted.

options.go
```diff
@@ -231,6 +232,26 @@ func DisableRelay() Option {
...
+// DisableRelayService configures libp2p to not run a circuit v2 relay.
+// Note that we only start the relay if we detect that we're publicly reachable.
+func DisableRelayService() Option {
+	return func(cfg *Config) error {
+		cfg.DisableRelayService = true
+		cfg.RelayServiceOpts = nil
+		return nil
+	}
+}
+
```
vyzo, line 244: we don't need this if it is not enabled by default.

p2p/host/relayv2/relay.go
```diff
@@ -0,0 +1,91 @@
+package relayv2
```
vyzo, line 1: let's call this package `relaysvc` or something similar? We are moving the autorelay stuff in `autorelay` anyway.

vyzo (commented):
LGTM; only thing we need is to make it disabled by default as @stebalien requested.
Also, the package name for the manager/service is rather confusing, let's call it something more appropriate.

config/config.go
marten-seemann: Makes sense.
We'll still enable this by default in IPFS / Filecoin, won't we?

Stebalien: >  We'll still enable this by default in IPFS / Filecoin, won't we?

Yes, or at least in IPFS. But libp2p is a library not an application.

vyzo (approved):

PR #1198, opened 2021-09-23
autorelay support for circuitv2 relays

vyzo:
Changes:
- Move the autorelay code to `p2p/host/autorelay`
- Move the host wrapper to the autorelay package.
- Add support for circuitv2 relays

Next steps: refactor the relay discovery logic to support the Flare strategy of an army of relays; this can happen in a subsequent pr, and we can even wait for deployment before we do it.
For now, user must configure relays statically.

### DIFF
--- a/p2p/host/autorelay/autorelay.go
+++ b/p2p/host/autorelay/autorelay.go
@@ -1,4 +1,4 @@
-package relay
+package autorelay
 
 import (
 	"context"
--- a/p2p/host/autorelay/autorelay.go
+++ b/p2p/host/autorelay/autorelay.go
@@ -5,8 +5,9 @@ import (
 	"fmt"
 	"math/rand"
 	"sync"
-	"sync/atomic"
 	"time"
+
+	"golang.org/x/sync/errgroup"
 
 	"github.com/libp2p/go-libp2p-core/discovery"
 	"github.com/libp2p/go-libp2p-core/event"
@@ -167,12 +168,13 @@ func (ar *AutoRelay) refreshReservations(ctx context.Context, now time.Time) boo
 		return true
 	}
 
-	// find reservations about to expire and refresh them in parallel
-	var (
-		wg   sync.WaitGroup
-		fail int32
-	)
+	if len(ar.relays) == 0 {
+		ar.mx.Unlock()
+		return false
+	}
 
+	// find reservations about to expire and refresh them in parallel
+	g := new(errgroup.Group)
 	for p, rsvp := range ar.relays {
 		if rsvp == nil {
 			// this is a circuitv1 relay, there is no reservation
@@ -183,18 +185,17 @@ func (ar *AutoRelay) refreshReservations(ctx context.Context, now time.Time) boo
 			continue
 		}
 
-		wg.Add(1)
-		go ar.refreshRelayReservation(ctx, p, &wg, &fail)
+		g.Go(func() error {
+			return ar.refreshRelayReservation(ctx, p)
+		})
 	}
 	ar.mx.Unlock()
 
-	wg.Wait()
-	return fail > 0
+	err := g.Wait()
+	return err != nil
 }
 
-func (ar *AutoRelay) refreshRelayReservation(ctx context.Context, p peer.ID, wg *sync.WaitGroup, fail *int32) {
-	defer wg.Done()
-
+func (ar *AutoRelay) refreshRelayReservation(ctx context.Context, p peer.ID) error {
 	rsvp, err := circuitv2.Reserve(ctx, ar.host, peer.AddrInfo{ID: p})
 
 	ar.mx.Lock()
@@ -206,12 +207,12 @@ func (ar *AutoRelay) refreshRelayReservation(ctx context.Context, p peer.ID, wg 
 		delete(ar.relays, p)
 		// unprotect the connection
 		ar.host.ConnManager().Unprotect(p, autorelayTag)
-		// increment fail counter
-		atomic.AddInt32(fail, 1)
 	} else {
 		log.Debugf("refreshed relay slot reservation with %s", p)
 		ar.relays[p] = rsvp
 	}
+
+	return err
 }
 
 func (ar *AutoRelay) findRelays(ctx context.Context) bool {
--- a/p2p/host/autorelay/autorelay.go
+++ b/p2p/host/autorelay/autorelay.go
@@ -15,6 +15,8 @@ import (
 
 	basic "github.com/libp2p/go-libp2p/p2p/host/basic"
 	relayv1 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv1/relay"
+	circuitv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/client"
+	circuitv2_proto "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/proto"
 
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr/net"
@@ -30,7 +32,7 @@ var (
 	BootDelay = 20 * time.Second
 )
 
-// These are the known PL-operated relays
+// These are the known PL-operated v1 relays; will be decommissioned in 2022.
 var DefaultRelays = []string{
 	"/ip4/147.75.80.110/tcp/4001/p2p/QmbFgm5zan8P6eWWmeyfncR5feYEMPbht5b1FW1C37aQ7y",
 	"/ip4/147.75.80.110/udp/4001/quic/p2p/QmbFgm5zan8P6eWWmeyfncR5feYEMPbht5b1FW1C37aQ7y",
@@ -55,7 +57,7 @@ type AutoRelay struct {
 	disconnect chan struct{}
 
 	mx     sync.Mutex
-	relays map[peer.ID]struct{}
+	relays map[peer.ID]*circuitv2.Reservation // rsvp will be nil if it is a v1 relay
 	status network.Reachability
 
 	cachedAddrs       []ma.Multiaddr
@@ -71,7 +73,7 @@ func NewAutoRelay(bhost *basic.BasicHost, discover discovery.Discoverer, router 
 		router:     router,
 		addrsF:     bhost.AddrsFactory,
 		static:     static,
-		relays:     make(map[peer.ID]struct{}),
+		relays:     make(map[peer.ID]*circuitv2.Reservation),
 		disconnect: make(chan struct{}, 1),
 		status:     network.ReachabilityUnknown,
 	}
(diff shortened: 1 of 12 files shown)


p2p/host/autorelay/autorelay.go
```diff
@@ -135,6 +138,75 @@ func (ar *AutoRelay) background(ctx context.Context) {
...
+			}
+
+			// find reservations about to expire
+			now := time.Now()
+			for p, rsvp := range ar.relays {
+				if rsvp == nil {
+					continue
+				}
+
+				if now.Add(time.Minute).Before(rsvp.Expiration) {
```
marten-seemann, line 169: If we're running this loop once per minute, we should probably refresh 2 minutes before the expiration, to be on the safe side. (Best to introduce a const for that logic).

p2p/host/autorelay/autorelay.go
```diff
@@ -135,6 +138,75 @@ func (ar *AutoRelay) background(ctx context.Context) {
...
+	for {
+		select {
+		case <-ticker.C:
+			var toRefresh []peer.ID
+
+			ar.mx.Lock()
+			if ar.status == network.ReachabilityPublic {
+				// we are public, forget about the relays, unprotect peers
+				for p := range ar.relays {
+					ar.host.ConnManager().Unprotect(p, "autorelay")
```
marten-seemann, line 154: Nit: The string should be a package-level `const`.

p2p/host/autorelay/autorelay.go
```diff
@@ -135,6 +138,75 @@ func (ar *AutoRelay) background(ctx context.Context) {
...
+			for p, rsvp := range ar.relays {
+				if rsvp == nil {
+					continue
+				}
+
+				if now.Add(time.Minute).Before(rsvp.Expiration) {
+					continue
+				}
+
+				toRefresh = append(toRefresh, p)
```
marten-seemann, line 173: I think we can get rid of this temporary slice here, by starting the refresh go routine right away (code-wise, this will be cleaner if refactored into a separate function, as suggested below).

p2p/host/autorelay/autorelay.go
```diff
@@ -71,14 +73,15 @@ func NewAutoRelay(bhost *basic.BasicHost, discover discovery.Discoverer, router
...
-		relays:     make(map[peer.ID]struct{}),
+		relays:     make(map[peer.ID]*circuitv2.Reservation),
 		disconnect: make(chan struct{}, 1),
 		status:     network.ReachabilityUnknown,
 	}
 	bhost.AddrsFactory = ar.hostAddrs
 	bhost.Network().Notify(ar)
 	ar.refCount.Add(1)
 	go ar.background(ctx)
+	go ar.refresh(ctx)
```
marten-seemann, line 84: Can we avoid starting a new go routine here by merging this into `background`?

p2p/host/autorelay/autorelay.go
```diff
@@ -135,6 +138,75 @@ func (ar *AutoRelay) background(ctx context.Context) {
 	}
 }
 
+func (ar *AutoRelay) refresh(ctx context.Context) {
+	ticker := time.NewTicker(time.Minute)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
```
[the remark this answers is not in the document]
vyzo: sure.

p2p/host/autorelay/autorelay.go
```diff
@@ -135,6 +138,75 @@ func (ar *AutoRelay) background(ctx context.Context) {
...
+				toRefresh = append(toRefresh, p)
+			}
+			ar.mx.Unlock()
+
+			// refresh reservations about to expire in parallel
+			var wg sync.WaitGroup
+			for _, p := range toRefresh {
+				wg.Add(1)
+
+				go func(p peer.ID) {
```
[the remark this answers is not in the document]
vyzo: yes indeed, will do. And oops!

p2p/host/autorelay/autorelay.go
```diff
@@ -204,14 +276,48 @@ func (ar *AutoRelay) tryRelay(ctx context.Context, pi peer.AddrInfo) bool {
 		return false
 	}
 
-	ok, err := relayv1.CanHop(ctx, ar.host, pi.ID)
+	protoIDv1 := string(relayv1.ProtoID)
+	protoIDv2 := string(circuitv2_proto.ProtoIDv2Hop)
```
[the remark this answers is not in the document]
vyzo: sure.

p2p/host/autorelay/autorelay.go
```diff
@@ -246,8 +356,10 @@ func (ar *AutoRelay) connect(ctx context.Context, pi peer.AddrInfo) bool {
 		return false
 	}
 
-	// tag the connection as very important
-	ar.host.ConnManager().TagPeer(pi.ID, "relay", 42)
+	// wait for identify to complete so that we can check the supported protocols
+	// TODO we should do this without a delay/sleep.
+	time.Sleep(time.Second)
```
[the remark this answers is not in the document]
vyzo: Of course not, hence the TODO!
 `IdentifyWait` is what I was looking for, we should be able to us it.

p2p/host/autorelay/autorelay.go
vyzo: good point, will do.

p2p/host/autorelay/autorelay.go
vyzo: sure.

p2p/host/autorelay/autorelay.go
vyzo: yeah, it will probably be nicer.

p2p/host/autorelay/autorelay.go
vyzo: probably.

p2p/host/autorelay/autorelay.go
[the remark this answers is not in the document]
vyzo: done.

p2p/host/autorelay/autorelay.go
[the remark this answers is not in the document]
vyzo: fixed.

p2p/host/autorelay/autorelay.go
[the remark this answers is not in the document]
vyzo: done.

p2p/host/autorelay/autorelay.go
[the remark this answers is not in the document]
vyzo: fixed.

p2p/host/autorelay/autorelay.go
vyzo: fixed.

p2p/host/autorelay/autorelay.go
vyzo: done.

p2p/host/autorelay/autorelay.go
vyzo: done.

p2p/host/autorelay/autorelay.go
vyzo: done.

vyzo:
@marten-seemann ready for another round of review.

p2p/host/autorelay/autorelay.go
```diff
@@ -135,6 +154,65 @@ func (ar *AutoRelay) background(ctx context.Context) {
...
+	}
+
+	// find reservations about to expire and refresh them in parallel
+	var (
+		wg   sync.WaitGroup
+		fail int32
+	)
+
+	for p, rsvp := range ar.relays {
+		if rsvp == nil {
```
marten-seemann, line 177: Can you add a comment that this a circuitv1?

p2p/host/autorelay/autorelay.go
```diff
@@ -135,6 +154,65 @@ func (ar *AutoRelay) background(ctx context.Context) {
...
+			continue
+		}
+
+		if now.Add(rsvpExpirationSlack).Before(rsvp.Expiration) {
+			continue
+		}
+
```
marten-seemann, line 184: I must have missed this in my earlier review, but do we need a check to garbage collect expired reservations?

p2p/host/autorelay/autorelay.go
```diff
@@ -135,6 +154,65 @@ func (ar *AutoRelay) background(ctx context.Context) {
...
+		wg.Add(1)
+		go ar.refreshRelayReservation(ctx, p, &wg, &fail)
```
marten-seemann, line 186: This would be a good use case for the `errgroup` as in this example: https://pkg.go.dev/golang.org/x/sync/errgroup#example-Group-JustErrors.

p2p/host/autorelay/autorelay.go
```diff
@@ -246,8 +360,29 @@ func (ar *AutoRelay) connect(ctx context.Context, pi peer.AddrInfo) bool {
...
+			select {
+			case <-ar.host.IDService().IdentifyWait(conn):
+				ready <- struct{}{}
+			case <-ctx.Done():
+			}
+		}(conn)
+	}
+
+	select {
+	case <-ready:
```
marten-seemann, line 381: This will leave all but the first go routine blocked on their `IdentifyWait`, doesn't it? Maybe this would a good use case for https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext?

marten-seemann (commented):
Thanks for the quick turnaround. Here's some additional stuff I noticed.

p2p/host/autorelay/autorelay.go
vyzo: of course.

p2p/host/autorelay/autorelay.go
vyzo: I don't think so, we'll try to refresh them. Do you see any way to create garbage here?

p2p/host/autorelay/autorelay.go
vyzo: I thought about it (i like this package a lot too), but it is not necessarily an error if we fail to refresh a reservation -- maybe the relay got overloaded.

p2p/host/autorelay/autorelay.go
vyzo: No, the channel is buffered for `len(conns)`, so we should be fine.

p2p/host/autorelay/autorelay.go
vyzo: having said that, i won't object if you think we should do it with an errgrp.

vyzo: actually on further thought, yeah, the code will be simpler and easier to grok, so I'll do the errgrp thing.

vyzo: done.

p2p/host/autorelay/autorelay.go
marten-seemann: You're right, we delete the reservation if refreshing fails.

marten-seemann (approved):
LGTM.

PR #1200, opened 2021-09-24
fix flaky TestFastDisconnect identify test

### LINKED ISSUE #1157: flaky FastDisconnect test
```
  === RUN   TestFastDisconnect
      id_glass_test.go:77: context deadline exceeded
  --- FAIL: TestFastDisconnect (5.00s)

### PR DESCRIPTION
**Pull request by marten-seemann**
Fixes #1157.

### DIFF
--- a/p2p/protocol/identify/id_glass_test.go
+++ b/p2p/protocol/identify/id_glass_test.go
@@ -36,9 +36,13 @@ func TestFastDisconnect(t *testing.T) {
 		}
 
 		// Kill the connection, and make sure we're completely disconnected.
-		s.Conn().Close()
 		assert.Eventually(t,
-			func() bool { return target.Network().Connectedness(s.Conn().RemotePeer()) != network.Connected },
+			func() bool {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: p2p/protocol/identify/id_glass_test.go [hunk 1]
```diff
@@ -36,10 +37,11 @@ func TestFastDisconnect(t *testing.T) {
 
 		// Kill the connection, and make sure we're completely disconnected.
 		s.Conn().Close()
```
marten-seemann, line 39: The problem here was that _sometimes_, we'd end up with two connections (for example, one TCP and one QUIC connection). Of course, calling `s.Conn().Close()` here would only close one of them, which is why the loop on the next line would never terminate.

marten-seemann: Alternative fix would be to loop over all connections and close them.

vyzo: actually, let's do that -- it's the right thing to do.

marten-seemann: ok, done

vyzo (approved):

PR #1204, opened 2021-09-26
fix: skip variadic params in constructors

Stebalien:
We use them for "options". Ideally we'd be able to forward options through, but that would require real dependency injection.


marten-seemann (approved):

marten-seemann:
> that would require real dependency injection

How would that look like?

Stebalien:
This would allow passing options in a "group". We'd have to define a `TcpTransport` module as follows:

```go
type TcpParams struct {
  fx.In

  Host host.Host
  Upgrader Upgrader
  Options []tcp.Option `group:"tcp_option"`
}

func TcpTransport(params TcpParams) Transport { ... }

type TcpOption {
  fx.Out

  Option tcp.Option `group:"tcp_option"`
}
```

But on second thought, this is even more complicated...

Stebalien:
Well, the options can be passed as: `fx.Supply(fx.Annotated{Group: "tcp_option", MyTcpOption})` and transports could probably be supplied as `Transport("tcp", constructor)`, or something like that. But this is an issue for another day.

PR #1213, opened 2021-09-30
fix structured logging in holepunch coordination

marten-seemann:


p2p/protocol/holepunch/coordination.go
```diff
@@ -174,7 +174,7 @@ func (hs *Service) beginDirectConnect(p peer.ID) error {
 // It first attempts a direct dial (if we have a public address of that peer), and then
 // coordinates a hole punch over the given relay connection.
 func (hs *Service) DirectConnect(p peer.ID) error {
-	log.Debugw("got inbound proxy conn from peer", p)
+	log.Debugw("got inbound proxy conn", "peer", p.Pretty())
```
Stebalien, line 177: No need to call `Pretty()`, the peer ID will get encoded on-demand if debug logging is enabled.

Stebalien (changes requested):

p2p/protocol/holepunch/coordination.go
marten-seemann: Fixed.

Stebalien (approved):

PR #1215, opened 2021-10-03
only start the mDNS resolver when the first notifee is registered

### LINKED ISSUE #1167: flaky mDNS test: TestOtherDiscovery
```
=== RUN   TestOtherDiscovery
      mdns_test.go:99: 
          	Error Trace:	mdns_test.go:99
          	Test:       	TestOtherDiscovery
          	Messages:   	expected peers to find each other
  --- FAIL: TestOtherDiscovery (26.31s)

### PR DESCRIPTION
**Pull request by marten-seemann**
Fixes #1167.

The reason this test was flaky is that mDNS discovery so fast sometimes happened _before_ `RegisterNotifee` was called. The zeroconf deduplication logic then kicked in and prevented the same host from being discovered before the test timed out.

This PR changes the mDNS startup: the server is booted immediately, but the resolver (the client) is only started when `RegisterNotifee` is called.

### CODE REVIEW

File: p2p/discovery/mdns/mdns.go [hunk 1]
```diff
@@ -150,9 +151,9 @@ func (s *mdnsService) startServer() error {
 	return nil
 }
 
-func (s *mdnsService) startResolver(ctx context.Context) {
+func (s *mdnsService) startResolver() {
 	s.resolverWG.Add(2)
-	entryChan := make(chan *zeroconf.ServiceEntry, 1000)
+	entryChan := make(chan *zeroconf.ServiceEntry, 16)
```
marten-seemann, line 156: Unrelated to the race condition, but 1000 was just way too large.

p2p/discovery/mdns/mdns.go
```diff
@@ -197,6 +198,7 @@ func (s *mdnsService) RegisterNotifee(n Notifee) {
 	s.mutex.Lock()
 	s.notifees = append(s.notifees, n)
 	s.mutex.Unlock()
+	s.resolveOnce.Do(s.startResolver)
```
Stebalien, line 201: What about just having a separate start function? Honestly, we should stop starting services from within constructors in the first place as it always causes problems like this.

marten-seemann: That's one option. If changing the API is on the table, the other option would be to pass a `[]Notifee` (or a `...Notifee` or a `WithNotifee(Notifee)`) to the constructor, and remove both `RegisterNotifee` and `UnregisterNotifee`.
Arguably, this is the cleaner API, as a newly registered notifee will have to restart the service, and therefore won't have to wait for cache entries to expire.

Stebalien: Passing in a set of notifiees sounds reasonable. I'd still consider a separate start method as being able to fully "construct" a libp2p node before actually _starting_ it would be nice... But not necessary.

My concern with the current patch is that it's kind of implicit and only helps with the first notifiee.

marten-seemann: Makes sense. #1220 implements this solution.

Stebalien: Yeah, I like that one...

PR #1221, opened 2021-10-06
don't call HandlePeerFound in a go routine, when finding a peer via mDNS

marten-seemann:
If we start a go routine, we must control its shutdown. Here we don't. Shutting down the mDNS service doesn't cancel or even wait for completion of these go routines.

There are two ways to fix this:
1. cancel and wait for these go routines
2. make this the responsibility of the caller

Option 2 seems the more desirable, since there are cases where you don't do anything blocking / computationally expensive when finding a new peer, so there wouldn't be a need to start a go routine in the first place.

### DIFF
--- a/examples/chat-with-mdns/main.go
+++ b/examples/chat-with-mdns/main.go
@@ -124,16 +124,16 @@ func main() {
 
 	// open a stream, this stream will be handled by handleStream other end
 	stream, err := host.NewStream(ctx, peer.ID, protocol.ID(cfg.ProtocolID))
-
 	if err != nil {
 		fmt.Println("Stream open failed", err)
-	} else {
-		rw := bufio.NewReadWriter(bufio.NewReader(stream), bufio.NewWriter(stream))
-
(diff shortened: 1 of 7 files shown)


aschmahmann:
closed as no longer needed

master255:
It's programming. There's logic and no concept of trust. This is a library and it is logical that it should not decide for the application developer whether to create a new thread. The library should give the application programmer a choice, not make that choice for him.
Especially since it creates a whole thread.
I, as an application programmer, am scared off by such bugs in the library.

master255:
@Stebalien I'm still scared

marten-seemann:
> @Stebalien I'm still scared

We've told you this before @master255, please pay attention to the way you communicate:
1. It's bad practice to comment on PRs that were closed years ago.
2. It's totally inappropriate to ping Steven again after just a single day.
3. Your message doesn't add any value whatsoever. "I'm still scared" is just pure spam.

PR #1231, opened 2021-10-25
remove peers from the peer store when they disconnect

marten-seemann:
Depends on #1230 and https://github.com/libp2p/go-libp2p-peerstore/pull/174.

Any idea how to test this?


marten-seemann:
Resolved by #1258.

PR #1237, opened 2021-11-11
feat: plumb through peerstore context changes

guseggert:


config/config.go
```diff
@@ -297,9 +297,13 @@ func (cfg *Config) NewNode() (host.Host, error) {
 			ConnectionGater:    cfg.ConnectionGater,
 			Reporter:           cfg.Reporter,
 			PeerKey:            autonatPrivKey,
+		}
 
-			Peerstore: pstoremem.NewPeerstore(),
+		ps, err := pstoremem.NewPeerstore()
```
marten-seemann, line 302: Nit: construct it before the initialization of `autoNatCfg`, as we do for `autonatPrivKey`.

p2p/net/conngater/conngater.go
```diff
@@ -51,7 +52,7 @@ func NewBasicConnectionGater(ds datastore.Datastore) (*BasicConnectionGater, err
 
 	if ds != nil {
 		cg.ds = namespace.Wrap(ds, datastore.NewKey(ns))
-		err := cg.loadRules()
+		err := cg.loadRules(context.TODO())
```
marten-seemann, line 55: I don't think we'll ever put a real context here, so maybe `context.Background()` is more appropriate?

p2p/net/conngater/conngater.go
```diff
@@ -124,7 +125,7 @@ func (cg *BasicConnectionGater) loadRules() error {
 // Note: active connections to the peer are not automatically closed.
 func (cg *BasicConnectionGater) BlockPeer(p peer.ID) error {
 	if cg.ds != nil {
-		err := cg.ds.Put(datastore.NewKey(keyPeer+p.String()), []byte(p))
+		err := cg.ds.Put(context.TODO(), datastore.NewKey(keyPeer+p.String()), []byte(p))
```
marten-seemann, line 128: Same here, and below.

marten-seemann (approved):

PR #1241, opened 2021-11-21
make the Type field required in the HolePunch protobuf

marten-seemann:
See https://github.com/libp2p/specs/pull/375.


vyzo (approved):
shouldnt we regen the pb?

vyzo (approved):
ah omitted in display, nm

mxinden (approved):

PR #1244, opened 2021-11-23
allow the ping protocol on transient connections

marten-seemann:


lidel (approved):
Thanks, this should allow us to ping peers over `/p2p-circuit/` addrs.

PR #1251, opened 2021-12-02
improve the autorelay's address factory

marten-seemann:
1. Only the use the cache if we actually have cached addresses.
2. Only cache relay addresses.


MarcoPolo:

PR #1252, opened 2021-12-03
remove check for consistent transports in identify

marten-seemann:
When using autorelay, we set an `AddrFactory` and overwrite the host's addresses. Until we receive a relay reservation, that address set will be empty, making us discard all addresses received via identify.
Furthermore, the relay might be reachable via transports that we don't speak.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -716,35 +716,6 @@ func (ids *idService) consumeReceivedPubKey(c network.Conn, kb []byte) {
 	log.Errorf("%s local key and received key for %s do not match, but match peer.ID", lp, rp)
 }
 
-// HasConsistentTransport returns true if the address 'a' shares a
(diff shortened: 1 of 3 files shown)


vyzo (approved):
what about the check of self listening addrs, do we still have/want that?

Other than that looks good.

marten-seemann:
> what about the check of self listening addrs, do we still have/want that?

I'm not sure. The check we have (on the lines just before the ones this PR removes) is checking that the `LocalMultiaddr()` of the `conn` is one of the addresses we're listening on. I don't really understand under which circumstances this check would fail.

It might make sense though to match the address against the union of the self listening addrs and the advertised addrs. I implemented this in #1255.

Ultimately, I want to get rid of this check though, accept _any_ address in identify, and test all received addresses using AutoNAT v2. That would allow us to remove these quite complicated assumptions we now have all over our code base.

PR #1257, opened 2021-12-08
use a basichost, not a blankhost when constructing autonat

marten-seemann:
Is this safe to do? Our default in `NewHost` are such that we don't construct any services unless a config option is set for those, so there shouldn't be a big difference between a `BasicHost` and a `BlankHost`.

Note that this PR is not sufficient to kill the blankhost. We still use it in tests, and refactoring those will take some more effort (import loops...).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -311,7 +311,7 @@ func (cfg *Config) NewNode() (host.Host, error) {
 			h.Close()
 			return nil, err
 		}
-		dialerHost, err := bhost.NewHost(dialerSwarm, nil)
+		dialerHost, err := bhost.NewHost(dialerSwarm, &bhost.HostOpts{DisableIdentify: true})
 		if err != nil {
 			h.Close()
 			return nil, err
--- a/config/config.go
+++ b/config/config.go
@@ -25,7 +25,6 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
 
 	autonat "github.com/libp2p/go-libp2p-autonat"
-	blankhost "github.com/libp2p/go-libp2p-blankhost"
 	discovery "github.com/libp2p/go-libp2p-discovery"
 	swarm "github.com/libp2p/go-libp2p-swarm"
 	tptu "github.com/libp2p/go-libp2p-transport-upgrader"
@@ -307,12 +306,16 @@ func (cfg *Config) NewNode() (host.Host, error) {
 			Peerstore:          ps,
 		}
 
-		dialer, err := autoNatCfg.makeSwarm()
+		dialerSwarm, err := autoNatCfg.makeSwarm()
 		if err != nil {
 			h.Close()
 			return nil, err
 		}
-		dialerHost := blankhost.NewBlankHost(dialer)
+		dialerHost, err := bhost.NewHost(dialerSwarm, nil)
+		if err != nil {
+			h.Close()
+			return nil, err
+		}
 		if err := autoNatCfg.addTransports(dialerHost); err != nil {
 			dialerHost.Close()
 			h.Close()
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -307,12 +306,16 @@ func (cfg *Config) NewNode() (host.Host, error) {
 			Peerstore:          ps,
 		}
 
-		dialer, err := autoNatCfg.makeSwarm()
+		dialerSwarm, err := autoNatCfg.makeSwarm()
+		if err != nil {
+			h.Close()
+			return nil, err
+		}
+		dialerHost, err := bhost.NewHost(dialerSwarm, nil)
```
vyzo, line 314: are we sure there is nothing listening? We need identify only and nothing else.

Also, whats this nil we are passing?

marten-seemann: Identify is always added: https://github.com/libp2p/go-libp2p/blob/0f76f17417428a9c1ead92f199264c88c0b031f7/p2p/host/basic/basic_host.go#L223-L227

The `nil` is a `HostOpts`: https://github.com/libp2p/go-libp2p/blob/0f76f17417428a9c1ead92f199264c88c0b031f7/p2p/host/basic/basic_host.go#L113-L155.

vyzo: can we also neuter the peerstore?

vyzo: we should pass a noop one, which we may have to shim.

marten-seemann: What do you mean by "neuter"? We need _some kind of_ peerstore, as autonat is first adding the peer info, and then connecting (since the swarm doesn't allow us to dial an address directly).

vyzo: ah yes, indeed.

Can we make a variant with minimal overhead that just immediately removes _everything_ when the peer is disconnected? 
We dont need a peer manager and we dont care about caching addrs.

marten-seemann: Maybe that should be part of autonat? The peerstore exposes a `RemovePeer` method, so we could call that one.

vyzo: that might be the right thing to do, esp if we move the pstoremgr responsibility to the host.

Only concern is we keep addrs for some 10min, as recently connected.

Stebalien:
What's the motivation? We intentionally use a blank host for better performance. Do we _need_ identify?

marten-seemann:
> What's the motivation? We intentionally use a blank host for better performance.

The motivation is that there shouldn't be a (notable) difference between a blank host and a basic host constructed without any options. For example, we recently had to add support the peer connectedness event to the blank host: https://github.com/libp2p/go-libp2p-blankhost/pull/58, and stuff like this will be coming up and might break us in subtle ways.
That way, we'll be able to move the peerstore manager to the host, which after discussion in that repo (see https://github.com/libp2p/go-libp2p-peerstore/pull/188) is the current plan.

Mid-term, I hope to deprecate the blankhost repo altogether.

> Do we need identify?

Good point. We don't. The blankhost doesn't have identify. I've add a `DisableIdentify` option that we can pass to the basic host.

Stebalien:
As long as the basic host does literally nothing, I guess that's fine. I'd consider introducing a new _base_ constructor.

I guess my main concerns are:

1. Will we randomly panic if some service we expect isn't there?
2. We should probably avoid calling `updateLocalIpAddr` (or move this into some singleton service). Getting routing information isn't cheap.

Note: I think the real problem here is that the host is to "fat". Ideally, we'd have a `FullHost` wrapping a `BaseHost` where the `BaseHost` managed base events, and the `FullHost` had all the services. But we don't have to do that now.

marten-seemann:
> 1. Will we randomly panic if some service we expect isn't there?

I think we will, and there seems no good way to disentangle this, other than splitting the basic host into a base and full host, as you suggest.

PR #1259, opened 2021-12-10
split a blank host out of the basic host

### LINKED ISSUE #1247: remove go-libp2p-blankhost
This should be a configuration profile for go-libp2p, maximally a `libp2p.NewMinimalHost` or `libp2p.New(libp2p.MinimalHostOpts...)`. We could avoid code duplication and make sure that go-libp2p supports a wide range of configurations.

We currently use the blankhost for AutoNAT, and probably in a bunch of tests.

marten-seemann:
Fixes #1247.

As suggested by @Stebalien in https://github.com/libp2p/go-libp2p/pull/1257#issuecomment-990616783.

We probably need to work on naming here. Should we rename the `basic` package to `full`?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,15 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/libp2p/go-libp2p/p2p/host/autonat"
+	"github.com/libp2p/go-libp2p/p2p/host/autorelay"
+	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+	"github.com/libp2p/go-libp2p/p2p/host/blank"
+	routed "github.com/libp2p/go-libp2p/p2p/host/routed"
+	circuitv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/client"
+	relayv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
+	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
+
 	"github.com/libp2p/go-libp2p-core/connmgr"
 	"github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/host"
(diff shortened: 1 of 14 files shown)


vyzo (commented):
lets punt on this for now.

marten-seemann:
> lets punt on this for now

go-libp2p v0.17.0 is out now.

I'd really like to get this into v0.18.0. Not having duplicated code for our hosts is a prerequisite for further refactors.

vyzo:
yeah, lets do it; base -vs- basic is confusing. Lets also have a blank profile to replace the blankhost.

vyzo:
or is that not useful?

vyzo (commented):
looking good, lets introduce the `FullHost` name.

vyzo:
I believ we want identify in the base host, the autonat dialer wont work without it.

p2p/host/base/base_host.go
```diff
@@ -0,0 +1,255 @@
...
+	go handle(protoID, s)
+}
+
+func (h *BaseHost) NewStream(ctx context.Context, p peer.ID, protos ...protocol.ID) (network.Stream, error) {
+	s, err := h.Network().NewStream(ctx, p)
+	if err != nil {
+		return nil, err
+	}
+
+	selected, err := msmux.SelectOneOf(protocol.ConvertToStrings(protos), s)
```
Stebalien, line 206: This won't obey the context unless we play the goroutine game (same as in BasicHost.NewStream).

p2p/host/base/base_host.go
```diff
@@ -0,0 +1,255 @@
...
+	n.SetStreamHandler(h.newStreamHandler)
+	return h, nil
+}
+
+func (h *BaseHost) initSignedRecord() error {
+	cab, ok := peerstore.GetCertifiedAddrBook(h.network.Peerstore())
+	if !ok {
+		return errors.New("peerstore does not support signed records")
+	}
+	rec := peer.PeerRecordFromAddrInfo(peer.AddrInfo{ID: h.ID(), Addrs: h.Addrs()})
```
Stebalien, line 109: What happens if we _don't_ do this? I'm asking because:

1. We're going to clobber it in the Basic/Full host.
2. The addresses are going to be wrong here.

Stebalien (approved):
This actually seems pretty clean.

Although, now that i see it... it would be really nice if we just had a truly pluggable host.

Stebalien:
> I believe we want identify in the base host, the autonat dialer wont work without it.

What does it use identify for?

vyzo:
actually maybe you are right, it doesnt really care.

marten-seemann:
Right, the blankhost used by autonat today doesn't have identify.

p2p/host/base/base_host.go
marten-seemann: Fixed.

p2p/host/base/base_host.go
marten-seemann: @Stebalien Can you elaborate? It would be nice if the base host supported signed records, wouldn't it?

The addresses are only going to wrong if an address factory is used.
Also, in the full host we generate a new record for ourselves, so we'll overwrite the one that the base host generated.

BigLep:
2022-08-26 conversation: this turned out to be more than we expected.  Core maintainers will take this on sometime in the future.

MarcoPolo:
Turns out it was a lot more work than expected. But it seems like we may be circling back on this soon. This PR though is really outdated.

PR #1274, opened 2021-12-21
Fix typo in options.go

mudler:
Fixes a typo while I was looking over docs  :+1:

### DIFF
--- a/options.go
+++ b/options.go
@@ -233,7 +233,7 @@ func DisableRelay() Option {
 }
 
 // EnableRelayService configures libp2p to run a circuit v2 relay,
-// if we dected that we're publicly reachable.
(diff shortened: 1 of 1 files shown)


options.go
```diff
@@ -233,7 +233,7 @@ func DisableRelay() Option {
 }
 
 // EnableRelayService configures libp2p to run a circuit v2 relay,
-// if we dected that we're publicly reachable.
+// if we detected that we're publicly reachable.
```
marten-seemann, line 236:
```suggestion
// if we detect that we're publicly reachable.
```

marten-seemann (changes requested):

marten-seemann (approved):

PR #1275, opened 2021-12-30
use the resource manager

marten-seemann:
Integrates the ResourceManager in go-libp2p; #1285 and #1289 have been squashed in.

TBD: 
- [ ] integration tests (see #1296)


p2p/protocol/circuitv2/client/transport.go
```diff
@@ -49,14 +51,25 @@ var _ transport.Transport = (*Client)(nil)
 var _ io.Closer = (*Client)(nil)
 
 func (c *Client) Dial(ctx context.Context, a ma.Multiaddr, p peer.ID) (transport.CapableConn, error) {
+	var connScope network.ConnectionManagementScope
+	if rcmgr := c.host.Network().ResourceManager(); rcmgr != nil {
+		var err error
+		// TODO: what's the best way to tell if we'll be using a fd here?
```
vyzo, line 57: it is a virtual connection, so i think we dont need to restrict here.

vyzo (commented):
We need to do the circuit servers as well, at the stream level -- SetService and reserve the buffers.

p2p/protocol/circuitv2/client/transport.go
marten-seemann: Fixed.

p2p/net/mock/mock_conn.go
```diff
@@ -195,3 +195,7 @@ func (c *conn) RemotePublicKey() ic.PubKey {
 func (c *conn) Stat() network.ConnStats {
 	return c.stat
 }
+
+func (c *conn) Scope() network.ConnScope {
+	return nil
```
vyzo, line 200: we should return the NullScope here

p2p/net/mock/mock_peernet.go
```diff
@@ -368,3 +368,7 @@ func (pn *peernet) notifyAll(notification func(f network.Notifiee)) {
 	pn.notifmu.Unlock()
 	wg.Wait()
 }
+
+func (pn *peernet) ResourceManager() network.ResourceManager {
+	return nil
```
vyzo, line 373: and here too.

p2p/net/mock/mock_stream.go
```diff
@@ -289,6 +289,10 @@ func (s *stream) transport() {
 	}
 }
 
+func (s *stream) Scope() network.StreamScope {
+	return nil
```
vyzo, line 293: ditto.

vyzo (commented):
will open a pr on top for adding rcmgr support in the internal services.

p2p/net/mock/mock_conn.go
vyzo: fixed in #1285

p2p/net/mock/mock_peernet.go
vyzo: fixed in #1285

p2p/net/mock/mock_stream.go
vyzo: fixed in #1285

p2p/net/mock/mock_peernet.go
vyzo: done in #1285

limits.go
```diff
@@ -0,0 +1,75 @@
...
+	circuitv1 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv1/relay"
+	circuitv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
+	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
+	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
+	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
+
+	rcmgr "github.com/libp2p/go-libp2p-resource-manager"
+)
+
+// SetDefaultServiceLimits sets the default limits for bundled libp2p services
```
raulk, line 14: Can we document what these limits are so users don't have to read the source?

p2p/protocol/circuitv1/relay/relay.go
```diff
@@ -64,6 +67,17 @@ func NewRelay(h host.Host, opts ...Option) (*Relay, error) {
...
+	err := h.Network().ResourceManager().ViewService(ServiceName,
+		func(s network.ServiceScope) error {
+			var err error
+			r.scope, err = s.BeginTransaction()
+			return err
+		})
+	if err != nil {
+		return nil, err
+	}
+
```
raulk, line 80: This block looks like it'll become boilerplate. Is there a missing API here?

p2p/protocol/circuitv1/relay/relay.go
```diff
@@ -72,19 +86,33 @@ func NewRelay(h host.Host, opts ...Option) (*Relay, error) {
...
-
 	log.Debugf("new relay stream from: %s", s.Conn().RemotePeer())
 
+	if err := s.Scope().SetService(ServiceName); err != nil {
+		log.Debugf("error attaching stream to relay service: %s", err)
+		s.Reset()
+		return
+	}
+
+	if err := s.Scope().ReserveMemory(maxMessageSize, network.ReservationPriorityAlways); err != nil {
```
raulk, line 104: Are we sure that all relay usages will have the highest priority? This seems off... I would expect nodes to offer relay services on best-effort basis, except for dedicated relay supernodes.

p2p/protocol/circuitv1/relay/relay.go
```diff
@@ -108,55 +136,74 @@ func (r *Relay) handleStream(s network.Stream) {
 }
 
 func (r *Relay) handleHopStream(s network.Stream, msg *pb.CircuitRelay) {
+	txn, err := r.scope.BeginTransaction()
```
raulk, line 139: Are we sure that "transaction" is the right term here? To me, a transaction denotes logical atomicity and isolation of some kind, which I'm not sure is the right semantics here. There is no atomicity (succeed or fail all) nor isolation (reservations are in effect immediately before the transaction is done). Looks like this thing is just a convenience API to model lifetimes/spans, right? In that case, would `Block` or `Span` fit better as a concept?

p2p/protocol/circuitv1/relay/relay.go
```diff
@@ -190,7 +238,26 @@ func (r *Relay) handleHopStream(s network.Stream, msg *pb.CircuitRelay) {
...
+	}
+
+	if err := bs.Scope().SetService(ServiceName); err != nil {
+		log.Debugf("error attaching stream to relay service: %s", err)
+		fail(pb.CircuitRelay_HOP_CANT_SPEAK_RELAY)
+		return
+	}
+
 	// stop handshake
+	if err := bs.Scope().ReserveMemory(maxMessageSize, network.ReservationPriorityAlways); err != nil {
```
raulk, line 254: Same as above. Are we sure this is the right priority?

p2p/protocol/ping/ping.go
```diff
@@ -33,7 +37,21 @@ func NewPingService(h host.Host) *PingService {
...
 
 func (p *PingService) PingHandler(s network.Stream) {
-	buf := make([]byte, PingSize)
+	if err := s.Scope().SetService(ServiceName); err != nil {
+		log.Debugf("error attaching stream to ping service: %s", err)
+		s.Reset()
+		return
+	}
+
+	if err := s.Scope().ReserveMemory(PingSize, network.ReservationPriorityAlways); err != nil {
```
raulk, line 46: Why not make priorities user configurable? Sounds like that's exactly what the user will want to do: tell the resource manager what's important for them.

vyzo: they are user configurable, it is a uint8 -- these are just some predefine levels.

p2p/protocol/circuitv1/relay/relay.go
vyzo: yeah, that's right. Maybe `Block/Span` is a better term indeed.

p2p/protocol/circuitv1/relay/relay.go
vyzo: highest priority means it wont work if you can't allocate this buffer; it is fine to fail!

p2p/protocol/circuitv1/relay/relay.go
vyzo: not really, most services will work at the stream level.

limits.go
vyzo: sure!

p2p/protocol/circuitv1/relay/relay.go
vyzo: yeah, if you can't allocate this buffer you can't even handshake so no point at proceeding.

p2p/protocol/circuitv1/relay/relay.go
vyzo: note that there are tight limits everywhere at play, so this will get throttled early at the service or peer level.
I would find it unlikely to hit the system level.

p2p/protocol/circuitv1/relay/relay.go
vyzo: on the same time, it would be very easy to define a utility method to get a txn scope for a service, and it might eliminate some boilerplate.

Where should we put it though?

p2p/protocol/ping/ping.go
```diff
@@ -130,7 +158,16 @@ func Ping(ctx context.Context, h host.Host, p peer.ID) <-chan Result {
 }
 
 func ping(s network.Stream) (time.Duration, error) {
-	buf := make([]byte, PingSize)
+	if err := s.Scope().ReserveMemory(2*PingSize, network.ReservationPriorityAlways); err != nil {
```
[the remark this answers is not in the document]
vyzo: in and out, two buffers which we need both to compare.

p2p/protocol/circuitv2/relay/relay.go
```diff
@@ -251,7 +299,26 @@ func (r *Relay) handleConnect(s network.Stream, msg *pbv2.HopMessage) {
...
+		return
+	}
+
 	// handshake
+	if err := bs.Scope().ReserveMemory(maxMessageSize, network.ReservationPriorityAlways); err != nil {
+		log.Debugf("erro reserving memory for stream: %s", err)
+		fail(pbv2.Status_RESOURCE_LIMIT_EXCEEDED)
+		return
+	}
+	defer bs.Scope().ReleaseMemory(maxMessageSize)
```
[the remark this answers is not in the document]
vyzo: yeah, it's all good -- it will be noop.

limits.go
```diff
@@ -0,0 +1,75 @@
...
+			WithMemoryLimit(1, 4<<20, 64<<20). // max 64MB service memory
+			WithStreamLimit(1024, 1024)        // max 1024 streams
+		limiter.ServicePeerLimits[circuitv1.ServiceName] = peerSvcLimit(128)
+	}
+	// relay/v2
+	if _, ok := limiter.ServiceLimits[circuitv2.ServiceName]; !ok {
+		limiter.ServiceLimits[circuitv2.ServiceName] = limiter.DefaultServiceLimits.
+			WithMemoryLimit(1, 4<<20, 64<<20). // max 64MB service memory
+			WithStreamLimit(1024, 1024)        // max 1024 streams
+		limiter.ServicePeerLimits[circuitv2.ServiceName] = peerSvcLimit(128)
```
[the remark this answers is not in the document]
vyzo: this is the server, the client we don't limit other than the default protocol limits.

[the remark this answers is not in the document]
vyzo: and yeah, they are separate and should be treated separately as one is for our own functionality while the other is a service.

limits.go
```diff
@@ -0,0 +1,75 @@
...
+			WithMemoryLimit(1, 4<<20, 64<<20). // max 64MB service memory
+			WithStreamLimit(128, 128)          // max 128 streams
+		limiter.ServicePeerLimits[autonat.ServiceName] = peerSvcLimit(2)
+	}
+	// relay/v1
+	if _, ok := limiter.ServiceLimits[circuitv1.ServiceName]; !ok {
+		limiter.ServiceLimits[circuitv1.ServiceName] = limiter.DefaultServiceLimits.
+			WithMemoryLimit(1, 4<<20, 64<<20). // max 64MB service memory
+			WithStreamLimit(1024, 1024)        // max 1024 streams
+		limiter.ServicePeerLimits[circuitv1.ServiceName] = peerSvcLimit(128)
```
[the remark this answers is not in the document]
vyzo: no, this is for the relayv1 server.

limits.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	if limiter.ServicePeerLimits == nil {
+		limiter.ServicePeerLimits = make(map[string]rcmgr.Limit)
+	}
+	// identify
+	if _, ok := limiter.ServiceLimits[identify.ServiceName]; !ok {
+		limiter.ServiceLimits[identify.ServiceName] = limiter.DefaultServiceLimits.
+			WithMemoryLimit(1, 4<<20, 64<<20). // max 64MB service memory
+			WithStreamLimit(128, 128)          // max 128 streams
+		limiter.ServicePeerLimits[identify.ServiceName] = peerSvcLimit(16)
```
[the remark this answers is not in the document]
vyzo: yeah.

limits.go
```diff
@@ -0,0 +1,75 @@
...
+		limiter.ServiceLimits = make(map[string]rcmgr.Limit)
+	}
+	if limiter.ServicePeerLimits == nil {
+		limiter.ServicePeerLimits = make(map[string]rcmgr.Limit)
+	}
+	// identify
+	if _, ok := limiter.ServiceLimits[identify.ServiceName]; !ok {
+		limiter.ServiceLimits[identify.ServiceName] = limiter.DefaultServiceLimits.
+			WithMemoryLimit(1, 4<<20, 64<<20). // max 64MB service memory
+			WithStreamLimit(128, 128)          // max 128 streams
```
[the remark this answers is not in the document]
vyzo: we have stopped closing the connection if it fails (in master) and we'll also add some code for retrying with an exponential backoff.

defaults.go
```diff
@@ -85,6 +86,14 @@ var DefaultEnableRelay = func(cfg *Config) error {
 	return cfg.Apply(EnableRelay())
 }
 
+var DefaultResourceManager = func(cfg *Config) error {
+	// Default memory limit: 1/8th of total memory, minimum 128MB, maximum 1GB
+	limiter := rcmgr.NewStaticLimiter(0.125, 128<<20, 1<<30)
```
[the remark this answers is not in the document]
vyzo: That's the max for network buffers, which is what we are accounting.
I don't think that's too low, but we can certainly raise it.

p2p/protocol/circuitv1/relay/relay.go
vyzo: changed to Span.

vyzo (approved):
ship it!

PR #1276, opened 2021-12-30
use the resource manager in the holepunch protocol

marten-seemann:


vyzo (commented):
we should also do `SetService("/libp2p/holepunch")` so that we can set the per service limit.

p2p/protocol/circuitv2/client/transport.go
```diff
@@ -49,14 +51,25 @@ var _ transport.Transport = (*Client)(nil)
 var _ io.Closer = (*Client)(nil)
 
 func (c *Client) Dial(ctx context.Context, a ma.Multiaddr, p peer.ID) (transport.CapableConn, error) {
+	var connScope network.ConnectionManagementScope
+	if rcmgr := c.host.Network().ResourceManager(); rcmgr != nil {
+		var err error
+		// TODO: what's the best way to tell if we'll be using a fd here?
+		connScope, err = rcmgr.OpenConnection(network.DirOutbound, true)
```
vyzo, line 58: it's a virtual connection, no fd is used.

vyzo:
implemented in #1285.

PR #1277, opened 2022-01-02
use the transport.Upgrader interface

marten-seemann:


config/constructor_types.go
```diff
@@ -29,29 +27,35 @@ var (
...
+	peerIDType = reflect.TypeOf((peer.ID)(""))
+	pskType    = reflect.TypeOf((pnet.PSK)(nil))
 )
 
 var argTypes = map[reflect.Type]constructor{
-	upgraderType:  func(h host.Host, u *tptu.Upgrader, cg connmgr.ConnectionGater) interface{} { return u },
-	hostType:      func(h host.Host, u *tptu.Upgrader, cg connmgr.ConnectionGater) interface{} { return h },
-	networkType:   func(h host.Host, u *tptu.Upgrader, cg connmgr.ConnectionGater) interface{} { return h.Network() },
-	muxType:       func(h host.Host, u *tptu.Upgrader, cg connmgr.ConnectionGater) interface{} { return u.Muxer },
-	securityType:  func(h host.Host, u *tptu.Upgrader, cg connmgr.ConnectionGater) interface{} { return u.Secure },
```
marten-seemann, line 44: I removed the `muxType` and the `securityType`. All tests are passing, but I'm not sure this means that this is actually safe.

vyzo: i dont see why not.

vyzo (approved):

config/constructor_types.go
Stebalien: IIRC, the idea was that some transports may not need the entire upgrader, and may just want a multiplexer etc. But honestly, we might as well give them an upgrader.

PR #1280, opened 2022-01-03
move go-libp2p-discovery to p2p/discovery/generic

marten-seemann:
I'm not really happy with the name `generic` here, but it's the best I can come up with.
We already have a `p2p/discovery` folder (which contains `mdns`). @Stebalien, any ideas for a better name?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,6 +17,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/transport"
 	"github.com/libp2p/go-libp2p-peerstore/pstoremem"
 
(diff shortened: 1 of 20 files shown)


vyzo (approved):

Stebalien:
Maybe just make them individual packages? I.e.:

1. backoff (backoff.go, backoffcache.go, backoffconnector.go)
2. routing (routing.go)
3. util (util.go). Maybe put backoff under util?

Mocks could either go in the routing package, or in a new "mocks" package.

PR #1285, opened 2022-01-07
Integrate the ResourceManager

vyzo:
Integrate the resource manager in libp2p bundled services and initializes a default resource manager.

Includes #1289 squashed in.

### DIFF
--- a/p2p/host/autonat/client.go
+++ b/p2p/host/autonat/client.go
@@ -36,12 +36,26 @@ func (c *client) DialBack(ctx context.Context, p peer.ID) (ma.Multiaddr, error) 
 	if err != nil {
 		return nil, err
 	}
+
+	if err := s.Scope().SetService(ServiceName); err != nil {
+		log.Debugf("error attaching stream to autonat service: %s", err)
+		s.Reset()
+		return nil, err
+	}
+
+	if err := s.Scope().ReserveMemory(maxMsgSize, network.ReservationPriorityAlways); err != nil {
+		log.Debugf("error reserving memory for autonat stream: %s", err)
+		s.Reset()
+		return nil, err
+	}
+	defer s.Scope().ReleaseMemory(maxMsgSize)
+
 	s.SetDeadline(time.Now().Add(streamTimeout))
 	// Might as well just reset the stream. Once we get to this point, we
 	// don't care about being nice.
 	defer s.Close()
 
-	r := protoio.NewDelimitedReader(s, network.MessageSizeMax)
+	r := protoio.NewDelimitedReader(s, maxMsgSize)
 	w := protoio.NewDelimitedWriter(s)
 
 	req := newDialMessage(peer.AddrInfo{ID: c.h.ID(), Addrs: c.addrFunc()})
--- a/p2p/host/autonat/svc.go
+++ b/p2p/host/autonat/svc.go
@@ -19,6 +19,12 @@ import (
 
 var streamTimeout = 60 * time.Second
 
+const (
+	ServiceName = "libp2p.autonat"
+
+	maxMsgSize = 4096
+)
+
 // AutoNATService provides NAT autodetection services to other peers
 type autoNATService struct {
 	instanceLock      sync.Mutex
@@ -45,13 +51,26 @@ func newAutoNATService(c *config) (*autoNATService, error) {
 }
 
 func (as *autoNATService) handleStream(s network.Stream) {
+	if err := s.Scope().SetService(ServiceName); err != nil {
+		log.Debugf("error attaching stream to autonat service: %s", err)
+		s.Reset()
+		return
+	}
+
+	if err := s.Scope().ReserveMemory(maxMsgSize, network.ReservationPriorityAlways); err != nil {
+		log.Debugf("error reserving memory for autonat stream: %s", err)
+		s.Reset()
+		return
+	}
+	defer s.Scope().ReleaseMemory(maxMsgSize)
+
 	s.SetDeadline(time.Now().Add(streamTimeout))
 	defer s.Close()
 
 	pid := s.Conn().RemotePeer()
 	log.Debugf("New stream from %s", pid.Pretty())
 
-	r := protoio.NewDelimitedReader(s, network.MessageSizeMax)
+	r := protoio.NewDelimitedReader(s, maxMsgSize)
 	w := protoio.NewDelimitedWriter(s)
 
 	var req pb.Message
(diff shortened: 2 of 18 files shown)


p2p/protocol/identify/id.go
```diff
@@ -382,10 +384,23 @@ func (ids *idService) identifyConn(c network.Conn, signal chan struct{}) {
 		return
 	}
 
+	// this is not set I believe, so we have to do it manually
+	if s.Scope().(network.StreamManagementScope).SetProtocol(ID); err != nil {
+		log.Warnf("error setting identify protocol scope: %s", err)
+		s.Reset()
+		return
+	}
+
```
vyzo, line 393: @marten-seemann is this assumption correct? Or does the `SetProtocol` above also set it in the scope?
If that's the case, we should remove this code. In fact, we should make it do so and not need this code.

p2p/protocol/holepunch/coordination.go
```diff
@@ -308,6 +324,7 @@ func (hs *Service) directConnect(rp peer.ID) error {
 }
 
 func (hs *Service) incomingHolePunch(s network.Stream) (rtt time.Duration, addrs []ma.Multiaddr, err error) {
+
```
marten-seemann, line 327:
```suggestion
```

p2p/protocol/identify/id.go
marten-seemann: It's not implemented yet, but services won't need to call `SetProtocol`. This will be done by the host's stream handler.

p2p/protocol/ping/ping.go
```diff
@@ -84,14 +102,24 @@ func (ps *PingService) Ping(ctx context.Context, p peer.ID) <-chan Result {
 // Ping pings the remote peer until the context is canceled, returning a stream
 // of RTTs or errors.
 func Ping(ctx context.Context, h host.Host, p peer.ID) <-chan Result {
-	s, err := h.NewStream(network.WithUseTransient(ctx, "ping"), p, ID)
-	if err != nil {
+	fail := func(err error) chan Result {
```
marten-seemann, line 105: Can we do this by introducing a `ping(...) error` function? I find that easier to read than a `fail` function assigned to a variable.

p2p/protocol/ping/ping.go
```diff
@@ -33,7 +37,21 @@ func NewPingService(h host.Host) *PingService {
...
+	}
+
+	if err := s.Scope().ReserveMemory(PingSize, network.ReservationPriorityAlways); err != nil {
+		log.Debugf("error reserving memory for ping stream: %s", err)
+		s.Reset()
+		return
+	}
+	defer s.Scope().ReleaseMemory(PingSize)
+
+	buf := pool.Get(PingSize)
```
marten-seemann, line 53: Nice!

p2p/protocol/identify/id.go
vyzo: yeah, identify is just special -- it opens a raw stream and does stuff.

Ok, let's keep it for now, and when it gets implemented we remove it.

p2p/host/autonat/svc.go
```diff
@@ -19,6 +19,12 @@ import (
 
 var streamTimeout = 60 * time.Second
 
+const (
+	ServiceName = "libp2p.autonat"
```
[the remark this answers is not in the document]
vyzo: Actually I think we should export it so that we can set specific service limits without having to use raw unchecked strings.

p2p/host/autonat/client.go
```diff
@@ -36,12 +36,26 @@ func (c *client) DialBack(ctx context.Context, p peer.ID) (ma.Multiaddr, error)
...
+		return nil, err
+	}
+	defer s.Scope().ReleaseMemory(maxMsgSize)
+
 	s.SetDeadline(time.Now().Add(streamTimeout))
 	// Might as well just reset the stream. Once we get to this point, we
 	// don't care about being nice.
 	defer s.Close()
 
-	r := protoio.NewDelimitedReader(s, network.MessageSizeMax)
```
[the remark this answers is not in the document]
vyzo: Well, it was one thought that it would be nice to have -- but it is obviously deleterious with the megabyte per pop.
Maybe we should just kill it?

p2p/protocol/circuitv1/relay/relay.go
```diff
@@ -25,6 +25,8 @@ var log = logging.Logger("relay")
 const (
 	ProtoID = "/libp2p/circuit/relay/0.1.0"
 
+	ServiceName = "libp2p.relay/v1"
```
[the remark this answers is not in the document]
vyzo: as above.

p2p/protocol/circuitv1/relay/relay.go
```diff
@@ -108,30 +136,50 @@ func (r *Relay) handleStream(s network.Stream) {
...
+	if err := txn.ReserveMemory(2*r.rc.BufferSize, network.ReservationPriorityHigh); err != nil {
+		log.Debugf("error reserving memory for relay: %s", err)
+		txn.Done()
+		r.handleError(s, pb.CircuitRelay_HOP_CANT_SPEAK_RELAY)
+		return
+	}
+
 	src, err := peerToPeerInfo(msg.GetSrcPeer())
 	if err != nil {
+		txn.Done()
```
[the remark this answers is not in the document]
vyzo: sure!

p2p/protocol/ping/ping.go
vyzo: well, it is supposed to return a channel so we can't avoid the ugliness -- but we can make it a standalone function.

p2p/host/autonat/client.go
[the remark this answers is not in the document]
marten-seemann: sgtm

p2p/protocol/circuitv1/relay/relay.go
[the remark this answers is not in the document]
vyzo: done!

p2p/protocol/ping/ping.go
vyzo: done.

vyzo:
rebased on the base branch.

p2p/protocol/identify/id.go
vyzo: this is fixed in #1289.

vyzo: fixed in #1289

marten-seemann (approved):

PR #1289, opened 2022-01-09
Default ResourceManager

vyzo:
On top of #1285

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -10,6 +10,7 @@ import (
 	noise "github.com/libp2p/go-libp2p-noise"
 	"github.com/libp2p/go-libp2p-peerstore/pstoremem"
 	quic "github.com/libp2p/go-libp2p-quic-transport"
+	rcmgr "github.com/libp2p/go-libp2p-resource-manager"
 	tls "github.com/libp2p/go-libp2p-tls"
 	yamux "github.com/libp2p/go-libp2p-yamux"
 	"github.com/libp2p/go-tcp-transport"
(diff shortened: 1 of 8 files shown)


defaults.go
```diff
@@ -85,6 +86,13 @@ var DefaultEnableRelay = func(cfg *Config) error {
 	return cfg.Apply(EnableRelay())
 }
 
+var DefaultResourceManager = func(cfg *Config) error {
+	limiter := rcmgr.NewStaticLimiter(0.125, 128<<20, 1<<30)
```
marten-seemann, line 90: maybe add a comment what this is in human readable units? (128 MB and 1 GB, if I'm not mistaken?)

limits.go
```diff
@@ -0,0 +1,75 @@
...
+	circuitv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
+	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
+	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
+	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
+
+	rcmgr "github.com/libp2p/go-libp2p-resource-manager"
+)
+
+// SetDefaultServiceLimits sets the default limits for bundled libp2p services
+func SetDefaultServiceLimits(limiter *rcmgr.BasicLimiter) {
```
marten-seemann, line 15: Now that we check if the value already exists, this doesn't need to be exported any more.

marten-seemann (approved):

defaults.go
vyzo: yeah, good point, will do.

limits.go
vyzo: I think we should keep it exported so that the user can call it on his own limiter.

defaults.go
vyzo: done.

PR #1293, opened 2022-01-10
don't close the connection when opening the identify stream fails

marten-seemann:
This is obviously very important once the rcmgr is used. This PR also cleans up some rather messy `defer` logic that was used for error handling.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -321,7 +321,6 @@ func (ids *idService) IdentifyWait(c network.Conn) <-chan struct{} {
 	defer ids.connsMu.Unlock()
 
 	wait, found = ids.conns[c]
-
 	if !found {
 		wait = make(chan struct{})
 		ids.conns[c] = wait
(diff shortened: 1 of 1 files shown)


p2p/protocol/identify/id.go
```diff
@@ -341,48 +347,26 @@ func (ids *idService) removeConn(c network.Conn) {
...
-		if p := c.RemotePeer(); err == nil {
-			ids.emitters.evtPeerIdentificationCompleted.Emit(event.EvtPeerIdentificationCompleted{Peer: p})
-		} else {
-			ids.emitters.evtPeerIdentificationFailed.Emit(event.EvtPeerIdentificationFailed{Peer: p, Reason: err})
-		}
-	}()
-
-	s, err = c.NewStream(network.WithUseTransient(context.TODO(), "identify"))
+func (ids *idService) identifyConn(c network.Conn) error {
+	s, err := c.NewStream(network.WithUseTransient(context.TODO(), "identify"))
```
vyzo, line 351: shouldnt we have a timeout here? Or do we set the deadline?

vyzo (commented):
I would also ask for @Stebalien 's opinion here as well, as this might have interesting effects up the stack.

p2p/protocol/identify/id.go
marten-seemann: Let's do that in a separate PR. We do set a bunch of deadlines, but only read deadlines, which can lead to funny edge cases if the peer withholds flow control credit.

marten-seemann:
Oh damn, I merged this PR accidentally. Sorry for that. I'm happy to revert / fix in a follow-up PR if necessary.

vyzo:
thats fine, i was going to approve anyway. I still want @Stebalien 's opinion on this however.

Stebalien:
Currently, this will mark the function as "identified" instead of trying again in the future.

1. It would be _nice_ if `IdentifyWait` returned an error, but that's probably not _necessary_.
2. We should definitely remove the channel from `ids.conns` if identify fails so we try again next time.

PR #1297, opened 2022-01-17
move the go-libp2p-connmgr here

marten-seemann:
This is easy, as go-libp2p doesn't yet depend on this package yet.
We will initialize a default connection manager in an upcoming PR very soon though.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/libp2p/go-libp2p
 go 1.16
 
 require (
+	github.com/benbjohnson/clock v1.1.0
 	github.com/btcsuite/btcd v0.22.0-beta // indirect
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -46,6 +47,7 @@ require (
 	github.com/multiformats/go-varint v0.0.6
 	github.com/prometheus/common v0.30.0 // indirect
 	github.com/prometheus/procfs v0.7.3 // indirect
+	github.com/raulk/go-watchdog v1.2.0
```
vyzo, line 50: that a problem?

vyzo (approved):
do it!

go.mod
marten-seemann: It's hidden behind the cgo flag (https://github.com/libp2p/go-libp2p-connmgr/pull/103).

PR #1326, opened 2022-02-08
update go-libp2p-yamux and go-libp2p-quic-transport

marten-seemann:
Anything else we need to update for rc4?


vyzo (approved):
I think we are good.

PR #1334, opened 2022-02-18
auto reconnect relay

optman:
make sure node behind NAT always connectable.


marten-seemann:
What issue is this supposed to fix?

optman:
relay only connect once, after that it will never try again.  a node may lose connection with relay, after that it will never reachable.

optman:
Where is the reconnect logic, is this connection protection? 
```
ar.host.ConnManager().Protect(pi.ID, autorelayTag)
```
But what if the current relay is offline, how to change to another relay?
 @vyzo

PR #1340, opened 2022-02-22
chore: update go-tcp-transport to v0.5.1

marten-seemann:


vyzo:
update has been pulled in #1341

PR #1343, opened 2022-02-22
Close resource manager when host closes

gammazero:
When a host is closed, it should close its resource manager.  Otherwise, goroutines are leaked.

Generally this is not a problem when a host exists for the life of a program, but can be a problem if hosts are created and closed.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -1054,6 +1054,10 @@ func (h *BasicHost) Close() error {
 		}
 
 		h.refCount.Wait()
+
+		if h.Network().ResourceManager() != nil {
+			h.Network().ResourceManager().Close()
+		}
 	})
 
 	return nil


willscott (approved):

vyzo (changes requested):
wait,no, there is a reason we havent done this. It is not necessary the case that we own the rcmgr; it could be shared between hosts in the app.

We can only safely close the default instantiation.

cc @marten-seemann

willscott:
whoops, sorry for the fast approval, i mistakenly thought this PR was up a level in application logic

vyzo:
no problem, we do want to close the default instantiation to avoid the
goroutine leak. If you can rework this to track that case, it would be
great.
Otherwise we'll sort it out with marten.

marten-seemann:
@gammazero Are you planning to fix this PR?

gammazero:
@marten-seemann I will fix it unless you plan to do so this week.

Would you prefer an additional option to tell `BasicHost` that is should assume ownership of the resource manager, or should BasicHost only close the resource manager if it is the default one?

marten-seemann:
I'm going to merge this PR. In fact, we have the same problem with the connection manager, and we also close it when the host closes.
This should get solved at some point, but it will probably require a bigger refactor of our host and services.

PR #1355, opened 2022-03-10
refactor the holepunching code

marten-seemann:
This came out of a discussion with @mxinden and @dennis-tra.

This PR splits the hole punching code into two components:
1. A `Service`, which is run on all hosts, independent of their NAT status. This service accepts DCUtR streams and responds to hole punching requests.
2. A `HolePuncher`, which is only run on hosts behind a NAT. This service initiates DCUtR streams when it detects new relayed connections.

Note: This will be merged for the v0.19.0 release.


p2p/protocol/holepunch/holepuncher.go
```diff
@@ -0,0 +1,275 @@
...
+			log.Debugw("hole punching failed", "peer", rp, "error", err)
+			hp.tracer.ProtocolError(rp, err)
+			return err
+		}
+		synTime := rtt / 2
+		log.Debugf("peer RTT is %s; starting hole punch in %s", rtt, synTime)
+
+		// wait for sync to reach the other peer and then punch a hole for it in our NAT
+		// by attempting a connect to it.
+		timer := time.NewTimer(synTime)
```
vyzo, line 147: why are we creating a timer for every iteration? This is no better than using time.After.

vyzo (commented):
this is a very welcome refactor; some questions about our favourite topic, the timer.

p2p/protocol/holepunch/holepuncher.go
marten-seemann: Because a `time.Timer` can be stopped, whereas a `time.After` cannot.

vyzo: yes but same inefficiency.
We should move the timer out of the loop, defer stop it, and reset for use.

marten-seemann: Yeah, I don't care about the inefficiency in this case:
1. If the hole punch succeeds, the code only gets executed once.
2. Even if it doesn't, it's 3 timers on the entire connection
Probably not worth getting into trouble with the Reset API.

vyzo: ok fair enough, lets avoid the foot bazooka.

vyzo (approved):

PR #1358, opened 2022-03-14
limit ping duration to 30s

vyzo:
so that ping streams cannot be squatted

### DIFF
--- a/p2p/protocol/ping/ping.go
+++ b/p2p/protocol/ping/ping.go
@@ -20,8 +20,9 @@ import (
 var log = logging.Logger("ping")
 
 const (
-	PingSize    = 32
-	pingTimeout = time.Second * 60
+	PingSize     = 32
+	pingTimeout  = 10 * time.Second
+	pingDuration = 30 * time.Second
 
 	ID = "/ipfs/ping/1.0.0"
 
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
Are we ok with the stream resets this is causing? If so, should we at least fix the logging on our side?

vyzo:
What else can we do? We cant let the other side ping ad infinum and squat the resources.

What do you recommend doing about logging?

marten-seemann:
> What else can we do? We cant let the other side ping ad infinum and squat the resources.

True. There are multiple options:
1. We could limit the client side to the same value though, so that a well-behaved client would not run into that error. That however would require making this duration a protocol constant.
2. Instead of resetting the stream, we could close the write side of the stream and reset the read side, and define stream closing as "normal shutdown". The client could then suppress the error when it receives the stream reset (on its write-side of the stream).
3. Or maybe all of this is too complex and we can just live with it?

> What do you recommend doing about logging?

We can special-case the deadline error, and either not log anything, or log "ping aborted after 30s" when it occurred.

vyzo:
want to take it over and mold it to suit your taste?

MarcoPolo:
I'll carry this forward. 🫡

MarcoPolo (approved):

MarcoPolo:
Update from the future, the small timeout change interacts poorly with rust-libp2p's defaults: https://github.com/libp2p/rust-libp2p/blob/c34668e2ca6ab6fad4e097fceeeccc1f4881b3d4/protocols/ping/src/handler.rs#L65

By default a rust node will always hit the timeout. I think changing it to 30s seems reasonable.

PR #1370, opened 2022-04-01
add a utility function to regularly question a Discoverer for peers

marten-seemann:
The idea is the following:
1. We construct an `AutoRelay` with a `peerChan <-chan peer.AddrInfo`. Peers sent on this channel are potential candidates for v2 relays.
2. We then regularly query the DHT for new peers, and send those on the channel.

This avoids a circular dependency between the DHT (which depends on the `host.Host`), and `AutoRelay`: You can now construct your host (with `AutoRelay`) first, just passing in the `peerChan`. Once the host and the DHT are constructed, you can now start filling that chan.

I'm not sure if the implementation here makes sense though. We're regularly restarting `FindPeers`. Instead, maybe we can just continue receiving values from that chan?

### DIFF
--- a/p2p/discovery/util/util.go
+++ b/p2p/discovery/util/util.go
@@ -28,6 +28,45 @@ func FindPeers(ctx context.Context, d discovery.Discoverer, ns string, opts ...d
 	return res, nil
 }
 
+func FindPeersRegularly(ctx context.Context, d discovery.Discoverer, ns string, interval time.Duration, opts ...discovery.Option) <-chan peer.AddrInfo {
+	peerChan := make(chan peer.AddrInfo, 10)
+	go func() {
+		defer close(peerChan)
+		t := time.NewTicker(interval)
+		defer t.Stop()
+
+		var (
+			fctx   context.Context
+			cancel context.CancelFunc
+		)
+		var ch <-chan peer.AddrInfo
+		for {
+			select {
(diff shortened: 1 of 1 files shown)


p2p/discovery/util/util.go
```diff
@@ -28,6 +28,45 @@ func FindPeers(ctx context.Context, d discovery.Discoverer, ns string, opts ...d
...
+			case <-t.C:
+				if cancel != nil {
+					cancel()
+				}
+				fctx, cancel = context.WithCancel(ctx)
+				var err error
+				ch, err = d.FindPeers(fctx, ns, opts...)
+				if err != nil {
+					log.Errorf("FindPeers failed: %s", err)
+					return
```
vyzo, line 62: this will short circuit in first error, probably not what we want?

vyzo (commented):
we might want to adjust the query rate so that we backoff after we find some peers.

Short circuiting in the first error may be problematic here, maybe we need some more smarts? Eg check for tmp errors.

marten-seemann:
So GitHub just closed this PR because I deleted the `discover-v2-relays` branch. And now it doesn't even allow me to change the merge target to `master`. Will have to open a new PR for this.

marten-seemann:
https://github.com/libp2p/go-libp2p/pull/1379

PR #1376, opened 2022-04-09
feat: catch panics in identify

Stebalien:
We haven't seen any panics in identify, but I'm going through and adding panic handlers to libp2p handlers because the benefits often outweigh the risk:

- Risk: We might continue with corrupted state. In this case, that means continuing with a corrupted peerstore, which likely doesn't have a huge impact on anything. If something is really wrong, we'll crash somewhere else.
- Benefit: if we panic when processing an identify response, we won't crash the entire node.

Part of #1389

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -630,9 +630,26 @@ func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn) {
 		log.Errorf("error getting peer record from Identify message: %v", err)
 	}
 
+	ids.consumeAddresses(p, lmaddrs, signedPeerRecord)
+	log.Debugf("%s received listen addrs for %s: %s", c.LocalPeer(), c.RemotePeer(), lmaddrs)
+
+	// get protocol versions
+	pv := mes.GetProtocolVersion()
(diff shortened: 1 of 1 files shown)


p2p/protocol/identify/id.go
```diff
@@ -421,16 +421,32 @@ func (ids *idService) sendIdentifyResp(s network.Stream) {
...
+
+		// Cleanup by resetting on any kind of failure.
+		if err != nil {
+			s.Reset()
+		}
+	}()
+
 	if err := s.Scope().SetService(ServiceName); err != nil {
 		log.Warnf("error attaching stream to identify service: %s", err)
-		s.Reset()
```
marten-seemann, line 427: You're creating a new `err` here that shadows the `err` return value, don't you?

Stebalien: `err` will be shadowed inside this condition, but the `err` in the "defer" will still refer to the named return value.

BigLep:
2022-09-16 conversation: we need to evaluate if this is the right place to do this or determine if our panic handlers in serialization functions are sufficient.

marten-seemann:
Closing since we've done a major refactor of Identify.

PR #1378, opened 2022-04-10
remove unneeeded build info function in Identify

marten-seemann:
I'm now sure if this is actually the correct fix, but it makes our tests pass with Go 1.18.

`debug.ReadBuildInfo` didn't work in Go 1.17 (`ok` was always `false`). In Go 1.18 is now suddenly works, and is failing the TestIDService test.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"runtime/debug"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


vyzo (approved):
this changes the user agent, anyway around it?

marten-seemann:
I don't think it does, as the `init` was never actually executed because `ReadBuildInfo` returned `false`. But then I'm not sure why it was introduced in the first place. @Stebalien?

Stebalien (changes requested):
This is meant for setting the user agent when included in a dependency of another go module (e.g., go-ipfs). For cases like that, I believe it actually works (I thought I tested it). I believe the correct fix is to check if `bi.Main.Path` is empty.

Stebalien:
(we should maybe document that...)

marten-seemann:
> This is meant for setting the user agent when included in a dependency of another go module (e.g., go-ipfs). For cases like that, I believe it actually works (I thought I tested it).

That makes sense. And even better, we can now use the vcs infos that Go 1.18 makes accessible (see https://blog.carlmjohnson.net/post/2021/golang-118-minor-features/). I'll make another PR.

marten-seemann:
There we go: https://github.com/libp2p/go-libp2p/pull/1381

PR #1379, opened 2022-04-10
add a utility function to regularly question a Discoverer for peers

marten-seemann:
The idea is the following:
1. We construct an `AutoRelay` with a `peerChan <-chan peer.AddrInfo`. Peers sent on this channel are potential candidates for v2 relays.
2. We then regularly query the DHT for new peers, and send those on the channel.

This avoids a circular dependency between the DHT (which depends on the `host.Host`), and `AutoRelay`: You can now construct your host (with `AutoRelay`) first, just passing in the `peerChan`. Once the host and the DHT are constructed, you can now start filling that chan.

I'm not sure if the implementation here makes sense though. We're regularly restarting `FindPeers`. Instead, maybe we can just continue receiving values from that chan?

### DIFF
--- a/p2p/discovery/util/util.go
+++ b/p2p/discovery/util/util.go
@@ -28,6 +28,45 @@ func FindPeers(ctx context.Context, d discovery.Discoverer, ns string, opts ...d
 	return res, nil
 }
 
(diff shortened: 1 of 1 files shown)


vyzo (approved):

MarcoPolo:
This has changed significantly. I think we have a version of this?

PR #1395, opened 2022-04-14
remove circuit v1 support from AutoRelay

marten-seemann:
This simplifies things quite a bit.


vyzo (commented):
as we discussed, lets remove by default and provide an option when using static.

PR #1413, opened 2022-04-21
Stream Migration PoC

MarcoPolo:
[Not meant to be merged]

A PoC implementation of stream migration. The example test here runs ping 20 times and migrates the underlying stream 4 times. No changes to ping except getting a reference to the stream (so we know what stream to migrate from, only for this example).

This test migrates a stream to another stream on the same connection. This is just as an example since I couldn't figure out how to open a new connection. While we are actually concerned about migrating streams to new connections the code here is unaware of were the streams live so should work just as well with a stream on another connection.

This follows https://github.com/libp2p/specs/pull/406.


marten-seemann (commented):
Cool stuff! I've added a bunch of comments.

I think we have to be a bit more careful with how we negotiate the stream migration protocol in the host:
* We want to check if the peer supports the stream migration protocol (see https://pkg.go.dev/github.com/libp2p/go-libp2p-core/peerstore#ProtoBook). We learn this from the Identify protocol. This allows to not waste one roundtrip if the peer doesn't support this protocol.
* If the peer says that it supports the protocol, but multistream negotiation of `/libp2p/stream-migration` fails, we know that something is wrong. Maybe the peer removed support for the stream migration protocol without telling us. In that case, we should probably disable stream migration for this peer altogether. Note: this is very unlikely, we don't expect this to happen in 99.9%. You can focus on the other parts of this PR first, especially since detecting this condition requires us to fix https://github.com/multiformats/go-multistream/issues/83 first.

MarcoPolo:
Closing in favor of eventually supporting QUIC's native connection migration feature.

PR #1417, opened 2022-04-21
update go-multistream, stop using deprecated NegotiateLazy

marten-seemann:


MarcoPolo (approved):
Looks good. I didn't know NegotiateLazy was deprecated. Is that documented anywhere? It's not mentioned in this interface https://github.com/libp2p/go-libp2p-core/blob/master/protocol/switch.go#L62

marten-seemann:
Good point. We should deprecate it on the interface as well. PR incoming.

We just deprecated it in go-multistream yesterday, because it actually didn't do anything that deserved the name "lazy": https://github.com/multiformats/go-multistream/pull/85.

vyzo (approved):
CI is angry...

marten-seemann:
> CI is angry...

oops, and angry in a very bad way (test timeouts). Debugging now. This _should_ not have caused any failures...

marten-seemann:
The problem was occurring in the `TestStreamsStress` in `mock/net`. I've removed this test entirely.
This might be a bit yolo, but the problem with `mock/net` is that it tries to mock an entire networking stack. Specifically, streams are modeled as `io.Pipe`s, which is not the best implementation of an actual stream, since reads and writes from an `io.Pipe` need to be synchronized.

vyzo (approved):
LGTM

CI is still complaining for some checks reason however.

Issue #1187, opened 2021-09-14
Repo Consolidation: Transports


marten-seemann:
The proposal is to move all transport-related repos to go-libp2p. 
To be clear, this consolidation does NOT mean that only these transports are supported. We will keep the interfaces and the methods to add new or custom handshakes / muxers / transports. libp2p will remain as extensible as it is now.

## Repositories

Security Handshakes:
* [TLS](https://github.com/libp2p/go-libp2p-tls/)
* [Noise](https://github.com/libp2p/go-libp2p-noise)

Muxers:
* yamux ([go-libp2p-yamux](https://github.com/libp2p/go-libp2p-yamux), NOT go-yamux)
* mplex ([go-libp2p-mplex](https://github.com/libp2p/go-libp2p-mplex), NOT go-mplex)

Transports:
* [QUIC transport](https://github.com/libp2p/go-libp2p-quic-transport) (not quic-go)
* [TCP transport](https://github.com/libp2p/go-tcp-transport)

Others:
* [Upgrader](https://github.com/libp2p/go-libp2p-transport-upgrader)
* TODO: add more

Stebalien:
I'd also move the resuseport "transport" package.

But, in short

# Do it! :tada:

marten-seemann:
What's our policy on preserving git histories? For circuitv2, we preserved the history using `git filter-tree`. We can do that as well for these repositories, but the price we'll pay is that we'll end up with a lot of commits that won't compile, as interfaces have changed over time (making it difficult to `git bisect` later).

Stebalien:
I try to preserve history wherever possible. I'm fine with commits that don't compile. Generally, you'd just browse _master's_ history, and all of those should compile.

marten-seemann:
Here's how the directory structure could look like.

## libp2p repo consolidation

### Transports

go-reuseport-transport => `p2p/transport/transport/go-reuseport-transport`
go-tcp-transport => `p2p/transport/transport/tcp`
go-libp2p-quic-transport => `p2p/transport/transport/quic`
go-ws-transport => `p2p/transport/ransport/websocket`

*`transport/transport` is kind of ugly*!

### Security

go-libp2p-tls => `p2p/transport/security/tls`
go-libp2p-noise => `p2p/transport/security/noise`

### Muxers

go-libp2p-yamux => `p2p/transport/muxer/yamux`
go-libp2p-yamux => `p2p/transport/muxer/yamux`

### Misc

[...]

PR #1422, opened 2022-04-22
move go-ws-transport here

marten-seemann:
Part of #1187.

### DIFF
--- a/p2p/transport/websocket/websocket_test.go
+++ b/p2p/transport/websocket/websocket_test.go
@@ -1,0 +1,350 @@
+package websocket
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"math/big"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/network"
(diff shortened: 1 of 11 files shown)


p2p/transport/websocket/websocket_test.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/sec"
+	"github.com/libp2p/go-libp2p-core/sec/insecure"
+	"github.com/libp2p/go-libp2p-core/test"
+	"github.com/libp2p/go-libp2p-core/transport"
+
+	csms "github.com/libp2p/go-conn-security-multistream"
+	mplex "github.com/libp2p/go-libp2p-mplex"
```
vyzo, line 27: we should probably wait for these to be moved first.

marten-seemann: Why? go-libp2p depending on external repos is ok (the other way around would not).

vyzo: sure, but we'll have to remember to fix it!

marten-seemann: Of course. I do a `sed` for the import path every time I move a repo.

vyzo: alright, fair enough.

vyzo (approved):

PR #1450, opened 2022-04-25
move go-libp2p-mplex here

marten-seemann:
Part of #1187.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/libp2p/go-libp2p-testing v0.9.2
 	github.com/libp2p/go-libp2p-tls v0.4.1
 	github.com/libp2p/go-libp2p-transport-upgrader v0.7.1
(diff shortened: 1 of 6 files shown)


Stebalien (approved):

p2p/transport/websocket/websocket_test.go
```diff
@@ -15,6 +15,8 @@ import (
 	"testing"
 	"time"
 
+	mplex "github.com/libp2p/go-libp2p/p2p/muxer/mplex"
```
Stebalien, line 18: do we need the rename?

marten-seemann: The package is named `peerstream_multiplex`, so I think we do.

Stebalien: So, can we, like, fix that. Now?

Stebalien: Probably.

PR #1453, opened 2022-04-25
swarm: don't check return value of str.Close in TestResourceManager

marten-seemann:
If the stream was canceled, QUIC will return an error when calling Close.


vyzo (approved):
ouch!

PR #1460, opened 2022-04-26
move go-conn-security-multistream here

marten-seemann:
Part of #1187.

### DIFF
--- a/config/security.go
+++ b/config/security.go
@@ -3,13 +3,13 @@ package config
 import (
 	"fmt"
 
+	csms "github.com/libp2p/go-libp2p/p2p/net/conn-security-multistream"
+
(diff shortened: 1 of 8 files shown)


config/security.go
```diff
@@ -3,13 +3,13 @@ package config
 import (
 	"fmt"
 
+	csms "github.com/libp2p/go-libp2p/p2p/net/conn-security-multistream"
```
Stebalien, line 6: Feel free to rename to something sane.

Stebalien: Maybe even have a multistream package? Or an upgraders package? Not sure.

Stebalien (approved):

PR #1462, opened 2022-04-26
move go-libp2p-noise here

marten-seemann:
Part of #1187.


Stebalien (approved):

Stebalien:
Probably need to tell something to ignore the warnings.

marten-seemann:
> Probably need to tell something to ignore the warnings.

1. These warning occur because go-libp2p includes a more recent version than go-libp2p-noise.
2. The only thing we're using from `golang.org/x/crypto/poly1305` is the `TagSize` constant. As far as I can tell, this should be equivalent to the `golang.org/x/crypto/chacha20poly1305` constant named `Overhead`.

I'll just switch over to the other package.

PR #1467, opened 2022-04-27
fix race condition and nil pointer dereference in holepunch

### LINKED ISSUE #1448: holepunch: s.holePuncher is nil on close
## Sys info
go-libp2p@v0.19.0

## Issue
When closing holepuncher it panics when it hadn't been created yet. Let me explain:

When holepuncher is closed, then the s.holePuncher is closed no matter what: 
https://github.com/libp2p/go-libp2p/blob/5151d4b4faaae04523dd83b9377881fbc717c752/p2p/protocol/holepunch/svc.go#L148-L156

It's worth noting that s.holePuncher is a pointer that is `nil` by default:
https://github.com/libp2p/go-libp2p/blob/5151d4b4faaae04523dd83b9377881fbc717c752/p2p/protocol/holepunch/svc.go#L44-L57

And that s.holePuncher is only initialized in s.watchForPublicAddr func when some conditions are met:
https://github.com/libp2p/go-libp2p/blob/5151d4b4faaae04523dd83b9377881fbc717c752/p2p/protocol/holepunch/svc.go#L141

This is a problem because if the s.holePuncher isn't initalized and we would call s.Close then s.holePuncher.Close() would panic (because it is not designed to be "zero value" closed).

## Solution:
The simplest solution is to introduce a nil check when closing holepuncher, it will check if s.holePuncher is nil or not.

watjurk:
This PR fixes #1448

### DIFF
--- a/p2p/protocol/holepunch/svc.go
+++ b/p2p/protocol/holepunch/svc.go
@@ -266,5 +266,7 @@ func (s *Service) handleNewStream(str network.Stream) {
 // TODO: find a solution for this.
 func (s *Service) DirectConnect(p peer.ID) error {
 	<-s.hasPublicAddrsChan
+	s.holePuncherMx.Lock()
+	defer s.holePuncherMx.Unlock()
 	return s.holePuncher.DirectConnect(p)
 }
--- a/p2p/protocol/holepunch/svc.go
+++ b/p2p/protocol/holepunch/svc.go
@@ -45,9 +45,11 @@ type Service struct {
 	ctx       context.Context
 	ctxCancel context.CancelFunc
 
-	host        host.Host
-	ids         identify.IDService
-	holePuncher *holePuncher
+	host host.Host
+	ids  identify.IDService
+
+	holePuncherMx sync.Mutex
+	holePuncher   *holePuncher
 
 	hasPublicAddrsChan chan struct{}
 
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
There's another use of the `holePuncher` in `DirectConnect`. This should also be protected by the mutex.

watjurk:
Are you sure this is what we want? In comment above direct connect we read:
```
// DirectConnect is only exposed for testing purposes.
// TODO: find a solution for this.
```
So this function should not be accesses by "normal" user

marten-seemann:
True, but doesn’t cost us anything to lock the mutex there as well, so let’s do it.

p2p/protocol/holepunch/svc.go
```diff
@@ -257,5 +266,7 @@ func (s *Service) handleNewStream(str network.Stream) {
 // TODO: find a solution for this.
 func (s *Service) DirectConnect(p peer.ID) error {
 	<-s.hasPublicAddrsChan
+	s.holePuncherMx.Lock()
+	defer s.holePuncherMx.Unlock()
```
marten-seemann, line 270: This locks the mutex for a **very** long time (on the order of 10s or so). Save `s.holePuncher` to a local variable instead.

marten-seemann (changes requested):

p2p/protocol/holepunch/svc.go
watjurk: Or maybe just use RMutex? And call RLock?

PR #1468, opened 2022-04-27
fix natManager to close natManager.nat

### LINKED ISSUE #1449: NATManager should close nmgr.nat
libp2p v0.19.0

When NATManager is closed it does not closes nmgr.nat:
https://github.com/libp2p/go-libp2p/blob/5151d4b4faaae04523dd83b9377881fbc717c752/p2p/host/basic/natmgr.go#L65-L69

When resolving this issue look at #1448, not to fall in the same pitfall

watjurk:
This PR fixes #1449

### DIFF
--- a/p2p/host/basic/natmgr.go
+++ b/p2p/host/basic/natmgr.go
@@ -65,7 +65,13 @@ func newNatManager(net network.Network) *natManager {
 func (nmgr *natManager) Close() error {
 	nmgr.ctxCancel()
 	nmgr.refCount.Wait()
-	return nil
+	var err error
+	nmgr.natMx.Lock()
+	if nmgr.nat != nil {
+		err = nmgr.nat.Close()
+	}
+	nmgr.natMx.Unlock()
+	return err
 }
 
 // Ready returns a channel which will be closed when the NAT has been found
--- a/p2p/host/basic/natmgr.go
+++ b/p2p/host/basic/natmgr.go
@@ -65,13 +65,7 @@ func newNatManager(net network.Network) *natManager {
 func (nmgr *natManager) Close() error {
 	nmgr.ctxCancel()
 	nmgr.refCount.Wait()
-	var err error
-	nmgr.natMx.Lock()
-	if nmgr.nat != nil {
-		err = nmgr.nat.Close()
-	}
-	nmgr.natMx.Unlock()
-	return err
+	return nil
 }
 
 // Ready returns a channel which will be closed when the NAT has been found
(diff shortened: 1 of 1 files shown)


p2p/host/basic/natmgr.go
```diff
@@ -65,7 +65,13 @@ func newNatManager(net network.Network) *natManager {
 func (nmgr *natManager) Close() error {
 	nmgr.ctxCancel()
 	nmgr.refCount.Wait()
-	return nil
+	var err error
+	nmgr.natMx.Lock()
+	if nmgr.nat != nil {
+		err = nmgr.nat.Close()
+	}
+	nmgr.natMx.Unlock()
```
Stebalien, line 73: We should do this in a defer in `background()`. That way we can be sure it only happens once.

p2p/host/basic/natmgr.go
```diff
@@ -37,7 +37,7 @@ func NewNATManager(net network.Network) NATManager {
 //  * closing the natManager closes the nat and its mappings.
 type natManager struct {
 	net   network.Network
-	natmu sync.RWMutex
+	natMx sync.RWMutex
```
Stebalien, line 40: Let's leave the name (`mu` stands for `mutex` and is a pretty common suffix in go).

Stebalien (changes requested):
Thanks!

p2p/host/basic/natmgr.go
watjurk: Are you saying that someone could call Close many times? In my opinion this would be incorrect.

p2p/host/basic/natmgr.go
watjurk: I've renamed it because in holepuncher we use "Mx".
https://github.com/libp2p/go-libp2p/blob/f383f4681f75b004ec2da0f4e9df0983bfbc65d2/p2p/protocol/holepunch/holepuncher.go#L36-L53

p2p/host/basic/natmgr.go
Stebalien: People often `defer thing.Close()`. It should usually be idempotent.

p2p/host/basic/natmgr.go
Stebalien: Ah, I see. Eh, that's fine.

p2p/host/basic/natmgr.go
watjurk: But doing this (closing it more than one) here should not be a problem, firstly `nmgr.nat.Close()` is idempotent, secondly by moving this to a deferred function in `background()` we lose the ability to return the error from `nmgr.nat.Close()`.

MarcoPolo: Generally I agree that these child objects should be closed when the parent is closed directly.

However, the current code already follows a pattern of _background worker + context cancel for cleanup_ so this change should follow the conventions of the current code rather than introducing a new pattern. If that makes sense, @watjurk, could you please move this to a defer in the background instead?

MarcoPolo (changes requested):
Please move close logic to defer in background. Thank you!

MarcoPolo (approved):
Thank you!

PR #1475, opened 2022-05-03
use real keys in tests instead of go-libp2p-testing/netutil fake keys

marten-seemann:
The performance impact of key generation should be negligible. This will allow us to deprecate go-libp2p-testing/netutil entirely. Security-wise, eliminating a way to generate fake keys is a win as well.


Stebalien (approved):
Yeah, seems reasonable. I think this was primarily an issue when we used RSA.

PR #1476, opened 2022-05-04
quic: add `DisableReuseport` option

### LINKED ISSUE #1428: quic: make the reuseport feature optional
Hi from @berty,
We recently discovered that if we enable the `reuseport` feature on the `libp2p-tcp-transport` while using a cellular connection, it could lead to network errors with some mobile operators. This is the case with the French operator Free Mobile which is massively used in France and we suspect that the problem occurs with several other operators around the world.

It seems that the QUIC transport has exactly the same issue but does not have an option to disable the `reuseport`. At the moment, this prevents us from using the QUIC transport on cellular connection even though it is particularly suitable for mobile use.

It would be nice to make the feature optional like in the `libp2p-tcp-transport`. I checked if it was possible to simply add an option to be able to toggle it off, but the feature seems too tightly bonded to the rest of the code.

gfanton:
Moved from  https://github.com/libp2p/go-libp2p-quic-transport/pull/272

---

Fixes #1428 

Hey, I just started to add the `DisableReuseport` options to quic transport.  It seems to be working as expected with some cellular operators that couldn't maintain the connection (I tried with this script https://github.com/gfanton/libp2p-reuseport-test)

Just some few things:

- It seems that there is no option/configuration mechanism, so I added one. Let me know if you are ok with this or if you want to use another method to pass the argument.
- I wrapped all the `conn_tests` with a simple table test case to test everything conn related with the reuseport_on/off
- I have to use `net.ListenUDP` instead of `net.DialUDP` for the Dial method to work, otherwise the call to the `WriteTo` method fails.

Also there are two tests that currently fail with the reuseport disabled:

**TestHolePunching**
```
=== RUN   TestHolePunching/reuseport_off
    conn_test.go:59: using a Secp256k1 key: 16Uiu2HAmPxEseg6Hti88hpHTyUB3tQk9bw7paj9yS3rPkBgPnWgM
    conn_test.go:59: using a Ed25519 key: 12D3KooWAaU8sc9j783H9hH3eG3yyJDnvz9Le1LdhCQFbYe5gFyu
    conn_test.go:581: 
        	Error Trace:	conn_test.go:581
        	           				asm_arm64.s:1133
        	Error:      	An error is expected but got nil.
        	Test:       	TestHolePunching/reuseport_off
        	Messages:   	didn't expect to accept any connections
    conn_test.go:602: 
        	Error Trace:	conn_test.go:602
        	Test:       	TestHolePunching/reuseport_off
```

**TestStatelessReset**
```
=== RUN   TestStatelessReset/reuseport_off
    conn_test.go:59: using a ECDSA key: QmRoF9AeMPnxYMrErq91y2xCvjiULp9n9cJuVwDHu9Kdao
    conn_test.go:59: using a RSA key: QmNvwzR7VDFy1wFdHE3PgV8mCztJbFWz1d5NXVL3wXYVgX
    conn_test.go:534: 
        	Error Trace:	conn_test.go:66
        	           				conn_test.go:534
        	           	listen udp4 127.0.0.1:55213: bind: address already in use
        	Test:       	TestStatelessReset/reuseport_off
```

Do these tests make sense when used with resuseport disabled?

### DIFF
--- a/p2p/transport/quic/conn_test.go
+++ b/p2p/transport/quic/conn_test.go
@@ -30,6 +30,16 @@ import (
 
 //go:generate sh -c "mockgen -package libp2pquic -destination mock_connection_gater_test.go github.com/libp2p/go-libp2p-core/connmgr ConnectionGater && goimports -w mock_connection_gater_test.go"
 
+type connTestCase struct {
+	Name    string
+	Options []Option
+}
+
+var connTestCases = []*connTestCase{
+	{"reuseport_on", []Option{}},
+	{"reuseport_off", []Option{DisableReuseport()}},
+}
+
 func createPeer(t *testing.T) (peer.ID, ic.PrivKey) {
 	var priv ic.PrivKey
 	var err error
@@ -52,20 +62,29 @@ func createPeer(t *testing.T) (peer.ID, ic.PrivKey) {
 
 func runServer(t *testing.T, tr tpt.Transport, addr string) tpt.Listener {
 	t.Helper()
+
 	ln, err := tr.Listen(ma.StringCast(addr))
 	require.NoError(t, err)
 	return ln
 }
 
 func TestHandshake(t *testing.T) {
+	for _, tc := range connTestCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			testHandshake(t, tc)
+		})
+	}
+}
+
+func testHandshake(t *testing.T, tc *connTestCase) {
 	serverID, serverKey := createPeer(t)
 	clientID, clientKey := createPeer(t)
-	serverTransport, err := NewTransport(serverKey, nil, nil, nil)
+	serverTransport, err := NewTransport(serverKey, nil, nil, nil, tc.Options...)
 	require.NoError(t, err)
 	defer serverTransport.(io.Closer).Close()
 
 	handshake := func(t *testing.T, ln tpt.Listener) {
-		clientTransport, err := NewTransport(clientKey, nil, nil, nil)
+		clientTransport, err := NewTransport(clientKey, nil, nil, nil, tc.Options...)
 		require.NoError(t, err)
 		defer clientTransport.(io.Closer).Close()
 		conn, err := clientTransport.Dial(context.Background(), ln.Multiaddr(), serverID)
@@ -100,6 +119,14 @@ func TestHandshake(t *testing.T) {
 }
 
 func TestResourceManagerSuccess(t *testing.T) {
+	for _, tc := range connTestCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			testResourceManagerSuccess(t, tc)
+		})
+	}
+}
+
+func testResourceManagerSuccess(t *testing.T, tc *connTestCase) {
 	serverID, serverKey := createPeer(t)
 	clientID, clientKey := createPeer(t)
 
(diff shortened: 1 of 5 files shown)


marten-seemann:
The QUIC tests were very flaky, and we've cleaned them up since. Could you rebase this PR on the current master?

gfanton:
I just rebased the branch on master, looks like all concerned tests are passing now!

p2p/transport/quic/conn_test.go
```diff
@@ -58,529 +66,571 @@ func runServer(t *testing.T, tr tpt.Transport, addr string) tpt.Listener {
...
-		require.True(t, conn.LocalPrivateKey().Equals(clientKey), "local private key doesn't match")
-		require.Equal(t, conn.RemotePeer(), serverID)
-		require.True(t, conn.RemotePublicKey().Equals(serverKey.GetPublic()), "remote public key doesn't match")
-
-		require.Equal(t, serverConn.LocalPeer(), serverID)
-		require.True(t, serverConn.LocalPrivateKey().Equals(serverKey), "local private key doesn't match")
-		require.Equal(t, serverConn.RemotePeer(), clientID)
-		require.True(t, serverConn.RemotePublicKey().Equals(clientKey.GetPublic()), "remote public key doesn't match")
-	}
+	for _, tc := range connTestCases {
```
marten-seemann, line 69: Please introduce a separate function, so you don't have to indent everything here.

p2p/transport/quic/conn_test.go
```diff
@@ -58,529 +66,571 @@ func runServer(t *testing.T, tr tpt.Transport, addr string) tpt.Listener {
...
+				handshake(t, ln)
+			})
+		})
+	}
 }
 
 func TestResourceManagerSuccess(t *testing.T) {
-	serverID, serverKey := createPeer(t)
-	clientID, clientKey := createPeer(t)
+	for _, tc := range connTestCases {
```
marten-seemann, line 115: Same here.

p2p/transport/quic/options.go
```diff
@@ -0,0 +1,24 @@
+package libp2pquic
+
+type Option func(opts *Config) error
+
+type Config struct {
```
marten-seemann, line 5: This doesn't need to be exported.

p2p/transport/quic/transport.go
```diff
@@ -80,20 +92,43 @@ func (c *connManager) getReuse(network string) (*reuse, error) {
...
+		}
+		return reuse.Listen(network, laddr)
+	}
+
+	conn, err := net.ListenUDP(network, laddr)
 	if err != nil {
 		return nil, err
 	}
-	return reuse.Listen(network, laddr)
+	return newNoReuseConn(conn), nil
```
marten-seemann, line 108: Nit: This doesn't need a constructor, just use `noreuseConn{Conn: conn}`.

marten-seemann (commented):
This mostly looks good. I added a few comments.

gfanton:
I've made the requested changes, but I've missed some `DisableReuseport` option in some test and still have the same issues described [above](https://github.com/libp2p/go-libp2p/pull/1476#issue-1225596723) with `TestHolePunching` & `TestStatelessReset`:
- for `TestStatelessReset` I've successfully fixed it in my last commit by closing the connection on `listener.Close` releasing the used port.
- for `TestHolePunching` I'm still facing the same issue, but I'm not really sure that this test has any interest with reuseport disabled. Do you have any clue about this ?

thanks !

marten-seemann:
Hole punching is only expected to work with reuseport enabled. You can disable that test for the non-reuseport version (please add a comment why!).

gfanton:
done 👍

gfanton:
One test is failing, but it doesn't seem to be related to my PR.

marten-seemann (approved):

PR #1478, opened 2022-05-10
holepunch: fix incorrect message type for the SYNC message

### LINKED ISSUE #1477: holepuncher not sending sync message
When doing holepuching. i got  an  **expected SYNC message from initiator but got 100** error

After some debuging with the code. I think the reason is due to the following code
in **initiateHolePunchImpl**

```golang	
func (hp *holePuncher) initiateHolePunchImpl(str network.Stream) ([]ma.Multiaddr, time.Duration, error) {

        .....
	// send a CONNECT and start RTT measurement.

	// wait for a CONNECT message from the remote peer
       ....
	if err := w.WriteMsg(&pb.HolePunch{Type: **pb.HolePunch_CONNECT**.Enum()}); err != nil {
		return nil, 0, fmt.Errorf("failed to send SYNC message for hole punching: %w", err)
	}

	return addrs, rtt, nil
}
```
 there we want to send a **SYNC** message.  but the code sends a **CONNECT** message
so the other end comply with:   **expected SYNC message from initiator but got 100**

below is the code for the opposite side, and it shows it expects a  **HolePunch_SYNC** msgType

```golang
func (s *Service) incomingHolePunch(str network.Stream) (rtt time.Duration, addrs []ma.Multiaddr, err error) {
       ...
	// Read SYNC message
	msg.Reset()
	if err := rd.ReadMsg(msg); err != nil {
		return 0, nil, fmt.Errorf("failed to read message from initator: %w", err)
	}
	if t := msg.GetType(); t != pb.HolePunch_SYNC {
		return 0, nil, fmt.Errorf("expected SYNC message from initiator but got %d", t)
	}
	return time.Since(tstart), obsDial, nil
}
```

marten-seemann:
Fixes #1477.

This is **bad**. We'll need to backport this and release v0.19.2.

### DIFF
--- a/p2p/protocol/holepunch/holepunch_test.go
+++ b/p2p/protocol/holepunch/holepunch_test.go
@@ -112,8 +112,9 @@ func TestDirectDialWorks(t *testing.T) {
 }
 
 func TestEndToEndSimConnect(t *testing.T) {
-	tr := &mockEventTracer{}
-	h1, h2, relay, _ := makeRelayedHosts(t, nil, holepunch.WithTracer(tr), true)
+	h1tr := &mockEventTracer{}
+	h2tr := &mockEventTracer{}
+	h1, h2, relay, _ := makeRelayedHosts(t, holepunch.WithTracer(h1tr), holepunch.WithTracer(h2tr), true)
 	defer h1.Close()
 	defer h2.Close()
 	defer relay.Close()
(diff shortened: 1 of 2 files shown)


vyzo (commented):
ouch, can you check the handler as well?

vyzo:
how did it work?

vyzo (changes requested):
spec is CONNECT, CONNECT, SYNC; the implementation is correct.

marten-seemann:
> spec is CONNECT, CONNECT, SYNC; the implementation is correct.

The current implementation doesn't send any SYNC. Before the refactoring, this was https://github.com/libp2p/go-libp2p/blob/v0.18.1/p2p/protocol/holepunch/coordination.go#L204-L210 (unfortunately git doesn't produce helpful diffs when you split one file into two, so this was easy to miss).

vyzo (approved):
oh boy.

PR #1481, opened 2022-05-10
Support generating custom x509 certificates

peterargue:
This PR was originally opened here https://github.com/libp2p/go-libp2p-tls/pull/99 and relates to this issue: fixes https://github.com/libp2p/go-libp2p/issues/1538

## Context

`libp2ptls.NewIdentity` generates a new `Identity` which includes an x509 certificate with a special signed extension which ties the TLS key pair to a libp2p network identity. In flow, we use this key pair to secure a gRPC interface which is accessible to nodes on the network as well as external public clients, allowing them to verify that they've connected to the correct staked node.

The current implementation of `NewIdentity` does not support customizing the certificate template, and uses default options that aren't compatible with standard TLS clients (e.g. Subject doesn't include a `cn` field).

## Proposal

This PR refactors keyToCertificate into 3 methods

* `GenerateSignedExtension` (exported), which generates the signed extension
* `certTemplate` which returns the default certificate template used to generate the self-signed certificate,
* `keyToCertificate`, which accepts an `x509.Certificate` template, adds the signed extension and returns the `tls.Certificate` object.

This PR also adds a new optional parameter to `NewIdentity`, which can provide a custom certificate template.

The changes are intended to be backwards compatible and add the ability for a caller to specify the x509 certificate fields they wish to include in the generated certificate.

### DIFF
--- a/p2p/security/tls/crypto.go
+++ b/p2p/security/tls/crypto.go
@@ -63,7 +63,7 @@ func NewIdentity(privKey ic.PrivKey, opts ...IdentityOption) (*Identity, error) 
 
 	var err error
 	if config.CertTemplate == nil {
-		config.CertTemplate, err = DefaultCertTemplate()
+		config.CertTemplate, err = certTemplate()
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


peterargue:
@marten-seemann here's the new PR

MarcoPolo (commented):
This looks good to me, but I'll let @marten-seemann submit the approval. Thanks for this and moving the PR over here!

peterargue:
@marten-seemann this look OK to you?

marten-seemann (commented):
I added a few comments. Other than that, the code looks good to me.
We should probably also have a test where we perform a handshake using a custom certificate, to make sure things work end-to-end.

peterargue:
Thanks @marten-seemann. I just pushed some changes addressing your comments.

MarcoPolo (approved):

peterargue:
Hey @marten-seemann think this PR is ok to merge?

marten-seemann:
Yes, it is. I approved it. But we won't merge it before the v0.21 release.

MarcoPolo:
Right after https://github.com/libp2p/go-libp2p/pull/1648, we’ll merge this. We wanted to make this release pretty small and we’ve started running some smoke tests for v0.21.

peterargue:
awesome. thanks!

PR #1482, opened 2022-05-11
AutoNAT: return E_DIAL_REFUSED instead of E_DIAL_ERROR of when skipping dial

elenaf9:
Discussed in https://github.com/libp2p/specs/issues/411:
When skipping a dial as a server, we should return `E_DIAL_REFUSED` instead of `E_DIAL_ERROR` because we never actually tried to dial the client.

### DIFF
--- a/p2p/host/autonat/autonat_test.go
+++ b/p2p/host/autonat/autonat_test.go
@@ -37,7 +37,7 @@ func sayPrivateStreamHandler(t *testing.T) network.StreamHandler {
 		w := protoio.NewDelimitedWriter(s)
 		res := pb.Message{
 			Type:         pb.Message_DIAL_RESPONSE.Enum(),
-			DialResponse: newDialResponseError(pb.Message_E_DIAL_ERROR, "no dialable addresses"),
+			DialResponse: newDialResponseError(pb.Message_E_DIAL_ERROR, "dial failed"),
 		}
 		w.WriteMsg(&res)
 	}
--- a/p2p/host/autonat/client.go
+++ b/p2p/host/autonat/client.go
@@ -31,6 +31,10 @@ type client struct {
 
 // DialBack asks peer p to dial us back on all addresses returned by the addrFunc.
 // It blocks until we've received a response from the peer.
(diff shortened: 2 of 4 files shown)


MarcoPolo (commented):
I think we should leave a comment here too: https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autonat/client.go#L79 since it's used there by the client.

elenaf9:
> I think we should leave a comment here too: https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autonat/client.go#L79 since it's used there by the client.

But there it only checks if there is an error, not what type of error it is. Shouldn't we then rather put that comment on the `DialBack` function itself, or here: https://github.com/libp2p/go-libp2p/blob/f383f4681f75b004ec2da0f4e9df0983bfbc65d2/p2p/host/autonat/autonat.go#L376-L396 where the type of error is checked?

MarcoPolo:
> > I think we should leave a comment here too: https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autonat/client.go#L79 since it's used there by the client.
> 
> But there it only checks if there is an error, not what type of error it is. Shouldn't we then rather put that comment on the `DialBack` function itself, or here:
> 
> https://github.com/libp2p/go-libp2p/blob/f383f4681f75b004ec2da0f4e9df0983bfbc65d2/p2p/host/autonat/autonat.go#L376-L396
> 
> where the type of error is checked?

I think that makes sense. I’d like to cover the case where someone in the future looks at this code and thinks “Hmm, if I check for dial refused I can do something clever” and doesn’t know that older peers will almost always return dial error. As long as that case is covered, I’m happy :)

marten-seemann:
@elenaf9 Are you planning to address @MarcoPolo's comments? Would be great if we could this fix into the next release!

elenaf9:
> @elenaf9 Are you planning to address @MarcoPolo's comments? Would be great if we could this fix into the next release!

Yes, will add something later today or tomorrow. Sorry for the delay.

PR #1483, opened 2022-05-15
identify: stop using mocknet in tests

marten-seemann:
The identify tests were the last consumer of the `mocknet` package (which in turn is the last consumer of `go-libp2p-testing/net` package). By rewriting those tests, we can get rid of `mocknet`, and we won't have to migrate `go-libp2p-testing/net` here.

UPDATE 2022-10-12: We don't remove the `mocknet` package, we just stop using it ourselves.

### DIFF
--- a/p2p/protocol/identify/obsaddr_test.go
+++ b/p2p/protocol/identify/obsaddr_test.go
@@ -1,120 +1,76 @@
 package identify_test
 
 import (
+	"context"
 	"crypto/rand"
+	"fmt"
+	"net"
 	"testing"
 	"time"
 
+	"github.com/libp2p/go-libp2p"
 	ic "github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/event"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/p2p/host/eventbus"
-	mocknet "github.com/libp2p/go-libp2p/p2p/net/mock"
 	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
 
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/stretchr/testify/require"
 )
 
-type harness struct {
-	t *testing.T
-
-	mocknet mocknet.Mocknet
-	host    host.Host
-
-	oas *identify.ObservedAddrManager
+type mockConn struct {
+	local, remote ma.Multiaddr
+	peer          peer.ID
+	direction     network.Direction
 }
 
-func (h *harness) add(observer ma.Multiaddr) peer.ID {
-	// create a new fake peer.
-	sk, _, err := ic.GenerateECDSAKeyPair(rand.Reader)
-	if err != nil {
-		h.t.Fatal(err)
+var _ network.Conn = &mockConn{}
+
+func newMockConn(dir network.Direction, peer peer.ID, local, remote ma.Multiaddr) network.Conn {
+	return &mockConn{
+		direction: dir,
+		peer:      peer,
+		local:     local,
+		remote:    remote,
 	}
-	h2, err := h.mocknet.AddPeer(sk, observer)
-	if err != nil {
-		h.t.Fatal(err)
-	}
-	_, err = h.mocknet.LinkPeers(h.host.ID(), h2.ID())
-	if err != nil {
-		h.t.Fatal(err)
-	}
-	return h2.ID()
 }
 
-func (h *harness) conn(observer peer.ID) network.Conn {
-	c, err := h.mocknet.ConnectPeers(h.host.ID(), observer)
-	if err != nil {
-		h.t.Fatal(err)
-	}
-	if c.Stat().Direction != network.DirOutbound {
-		h.t.Fatal("expected conn direction to be outbound")
-	}
-	return c
+func (m *mockConn) Close() error                  { panic("implement me") }
+func (m *mockConn) LocalPeer() peer.ID            { panic("implement me") }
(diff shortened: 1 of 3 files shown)


vyzo (commented):
there might be downstream dependents.

marten-seemann:
> there might be downstream dependents.

The only one I can find is `go-libp2p-kad-dht/ext_test.go`. They'll have to find a way to deal with this. This code has caused us _so much_ pain, it's time to get rid of it.

Stebalien:
I agree this library is annoying, but it's all we have for large integration tests with many nodes.

If we just "remove" it here and tell upstream to "deal" with it, I guarantee they'll "deal" with it by not upgrading libp2p.

How about just not using it here? I'd also like to note that it's not all that useful for libp2p itself, because libp2p generally has no need for large-scale mock tests (it mostly needs to test things on the transport level). However, for the dht, bitswap, etc., the ability to mock is really useful.

marten-seemann:
I’d argue that it shouldn’t live within libp2p then (or at least not go-libp2p). How about moving it to a separate repo?

Stebalien:
> I’d argue that it shouldn’t live within libp2p then (or at least not go-libp2p). How about moving it to a separate repo?

That seems reasonable. Can we just move it to go-libp2p-testing?

marten-seemann:
We can. The original plan was to deprecate go-libp2p-testing, but we can just shift packages around: all the muxer and transport tests go here, and the mocknet tests go there.

Wondertan:
`mocknet` is helpful for @celestiaorg.
We mock away OS networking in our protocols tests using it.

Moving it out seems inconsistent with the latest efforts to consolidate everything in one root go-libp2p repo. OOC, what problems was mocknet causing, and why is it annoying to the point where it needs to be invisible from people exploring the repo? Is there a plan to keep testing pkgs outside of the repo?

MarcoPolo:
I've not reviewed this code yet, but at a high level my thoughts are:

1. If this is a tool we are exposing to end users to be able to test large scale mock tests, then it makes sense that this should live in the go-libp2p repo and get tested/updated alongside everything else and for the same reasons we are consolidating everything else.
2. If we want to deprecate this tool (mocknet), then I think we need a viable alternative for folks that is tested and lives in go-libp2p.

I'm not opposed to deprecating this and providing something similar/simpler that can support the original use case. But I think something like this is useful.

p2p/protocol/identify/obsaddr_test.go
```diff
@@ -135,107 +84,146 @@ func TestObsAddrSet(t *testing.T) {
...
-	harness := newHarness(t)
-	if !addrsMatch(harness.oas.Addrs(), nil) {
-		t.Error("addrs should be empty")
-	}
+	p1 := newPeer(t)
+	oam.Record(newMockConn(network.DirOutbound, p1, local, a4), a1)
+	oam.Record(newMockConn(network.DirOutbound, p1, local, a4), a2)
+	oam.Record(newMockConn(network.DirOutbound, p1, local, a4), a3)
+	// these are all different so we should not yet get them.
+	time.Sleep(50 * time.Millisecond)
```
MarcoPolo, line 94: why do we need this sleep?

p2p/protocol/identify/obsaddr_test.go
```diff
@@ -19,110 +23,55 @@ import (
...
-			}
-		}
-		return true
-	}
+	local := ma.StringCast("/ip4/127.0.0.1/tcp/10086")
+	h, err := libp2p.New(libp2p.ListenAddrs(local))
+	require.NoError(t, err)
+	defer h.Close()
+
+	oam, err := identify.NewObservedAddrManager(h)
```
MarcoPolo, line 73: Is it possible to see some flakiness because the number of observations becomes greater than the buffer size for the [channel](https://github.com/libp2p/go-libp2p/blob/identify-test-no-more-mocknet/p2p/protocol/identify/obsaddr.go#L135)? Also I'm concerned with the dropping log being just a [debug](https://github.com/libp2p/go-libp2p/blob/identify-test-no-more-mocknet/p2p/protocol/identify/obsaddr.go#L240), maybe a warn would be more appropriate?

p2p/protocol/identify/obsaddr_test.go
```diff
@@ -135,107 +84,146 @@ func TestObsAddrSet(t *testing.T) {
...
-		"addrs should have timed out",
-	)
+	oam.SetTTL(200 * time.Millisecond)
+	// disconnect all addresses that gave us a1
+	for _, c := range conns1 {
+		notif.Disconnected(h.Network(), c)
+	}
+	require.Eventually(t, func() bool { return len(oam.Addrs()) < 2 }, time.Second, 10*time.Millisecond)
+	require.Equal(t, []ma.Multiaddr{a2}, oam.Addrs())
+	time.Sleep(time.Second)
```
MarcoPolo, line 215: Why sleep here at the end?

MarcoPolo (commented):
I think it's a good idea to remove the mocknet stuff here. It's simpler this way. I would like to see the tests avoid sleeps if possible (maybe a flush method? maybe a blocking record method?).

Seeing the complexity around mocknet and how we don't use it in go-libp2p, I'm now torn. On the one hand it would be great to have a tool like this we can provide to users, but if we aren't using it we can't realistically support it.

And if we can't support it, I don't think it makes sense to live in go-libp2p. I'm now in favor of putting mocknet into its own repo, with the hope that folks who use mocknet can contribute their and improve it to the point that we can later integrate it into go-libp2p as a first class supported tool.

BigLep:
2022-09-09 conversation: we'll get the changes in to the tests so that go-libp2p doesn't use mocknet.  It's a separate item/discussion on what to do with mocknet.

marten-seemann:
I rebased this PR and removed the commit that removes the `mocknet` package.

PR #1484, opened 2022-05-15
quic: deflake the holepunching test

marten-seemann:
Unfortunately, this uses an unexported member variable, but the only option is a long timeout.


vyzo (approved):
its fine!

PR #1487, opened 2022-05-15
quic: adjust timeout for reuse garbage collector detection in tests

### LINKED ISSUE #1455: quic: flaky reuse tests
```
=== RUN   TestReuseListenOnAllIPv4
      reuse_test.go:59: 
          	Error Trace:	reuse_test.go:59
          	Test:       	TestReuseListenOnAllIPv4
          	Messages:   	expected garbage collector to be running
  --- FAIL: TestReuseListenOnAllIPv4 (0.14s)

marten-seemann:
Fixes #1455.

### DIFF
--- a/p2p/transport/quic/reuse_test.go
+++ b/p2p/transport/quic/reuse_test.go
@@ -56,7 +56,7 @@ func cleanup(t *testing.T, reuse *reuse) {
 
 func TestReuseListenOnAllIPv4(t *testing.T) {
 	reuse := newReuse()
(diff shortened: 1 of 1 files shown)


MarcoPolo (approved):
This is crazy that it takes over 100 ms to spin up a goroutine. This PR itself is fine.

I'm going to look into if we see the same flakiness on circleci.

marten-seemann:
> This is crazy that it takes over 100 ms to spin up a goroutine.

I haven't dug into why this is a problem in this test but not the other tests, where we use 100ms without any problems. There might be more to it.

Issue #1505, opened 2022-05-18
Tracking Issue for flaky tests on CircleCI 


MarcoPolo:
This is a meta issue to track which tests we've skipped since they've caused flakiness on CircleCI.

- https://github.com/libp2p/go-libp2p/issues/1164#issuecomment-1130232184
- #1440 
- #1384

PR #1506, opened 2022-05-18
Circleci project setup

MarcoPolo:
This adds a circle ci build that runs our test matrix for {Windows, macOS, Linux}⨯{go1.17,go1.18}.

I think we should merge this so we can see how the flakiness compares with GH Actions. Then later we can decide if we want to switch or not.

Some tests were skipped in this PR to minimize noise. Those tests are tracked here: https://github.com/libp2p/go-libp2p/issues/1505

### Motivation
We've been seeing a lot of flakiness around timings in tests. Yes, generally relying on time in tests is not ideal, but there are somethings that should happen _eventually_. [For example](https://github.com/libp2p/go-libp2p/pull/1487#pullrequestreview-974406773) takes over 100ms to check if the GC is running.

If CircleCI saves us from having to deal with a lot of these flaky conditions it seems worth the switch. However if ipdx comes along with another solution I'm happy to switch to that as well. In the mean time I'd like to have us unblocked from flaky test land.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,74 @@
+version: 2.1
+
+orbs:
+  os-detect: circleci/os-detect@0.3.0
+  go: circleci/go@1.7.1
+  win: circleci/windows@4.1.1
+
+executors:
+  linux: &linux-executor
+    # This needs to be a machine image (not docker) so that ipv6 is supported.
+    machine:
+      image: ubuntu-2004:current
+  macos: &macos-executor
+    macos:
+      xcode: 13.3.1
+
+jobs:
+  test:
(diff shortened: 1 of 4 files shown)


MarcoPolo:
It would be nice to include this in linux: `sysctl -w net.core.rmem_max=2500000` to increase the udp recv buffer size. Easy change, but can do it in another PR if this gets reviewed before I make that change.

MarcoPolo:
closing this since I'm less convinced circle CI is less flaky than GH actions. A rough count of 30 runs on this branch found around 7 failed. We can reopen this if needed. For now I'm going to focus on making our tests less flaky to start with.

PR #1511, opened 2022-05-19
move go-stream-muxer-multistream here

marten-seemann:
Part of #1187. We didn't discuss the target directory in that issue, so please pay special attention to that when reviewing.


p2p/net/multistream-muxer/multistream.go
```diff
@@ -0,0 +1,75 @@
+// Package multistream implements a peerstream transport using
```
Stebalien, line 1: File name: `muxer-multistream`? It would be nice if this matched the security version.

Location: Maybe put this in `/muxer`? it's a muxer combinator, but it's also a muxer. But I don't feel strongly about that.

Package name: Not multistream (going to get a bit confusing?). But I'm not sure what to call it. Maybe `multistream_muxer`?

marten-seemann: Not a big fan, but let's go with muxer-multistream for now.

Stebalien: Yeah, all these names are long and annoying. I just want to be as consistent as possible.

marten-seemann: Is there anything else I should change? It's muxer-multistream everywhere now.

vyzo (approved):

PR #1513, opened 2022-05-20
move go-libp2p-nat here

marten-seemann:
This was also not on the list in #1187, but I can't see a reason to not consolidate this repo as well.


Stebalien (approved):
Yeah, we had three levels of indirection...

PR #1516, opened 2022-05-20
backoff: fix flaky tests in backoff cache

### LINKED ISSUE #1502: flaky TestBackoffDiscoveryMultipleBackoff
```
=== RUN   TestBackoffDiscoveryMultipleBackoff
      backoffcache_test.go:161: Was supposed to find 1, found 2 instead
  --- FAIL: TestBackoffDiscoveryMultipleBackoff (1.08s)

### PR DESCRIPTION
**Pull request by MarcoPolo**
Fixes #1502 by introducing a mocked clock. Time can be incremented in the mock clock and things check the clocks time via `.Now()`. If no clock is passed in, the default realClock uses `time.Now`

This means that these tests are not dependent on actual wall clock times, but our monotonically increasing mock clock.

This PR also fixes a subtle bug that would cause the limit of a `sendingChs` to be ignored. And a bug that prevented other `sendingChs` from getting sent data.

### DIFF
--- a/p2p/discovery/backoff/backoffcache.go
+++ b/p2p/discovery/backoff/backoffcache.go
@@ -21,6 +21,8 @@ type BackoffDiscovery struct {
 
 	parallelBufSz int
 	returnedBufSz int
+
+	clock clock
 }
 
 type BackoffDiscoveryOption func(*BackoffDiscovery) error
@@ -33,6 +35,8 @@ func NewBackoffDiscovery(disc discovery.Discovery, stratFactory BackoffFactory, 
 
 		parallelBufSz: 32,
 		returnedBufSz: 32,
+
+		clock: realClock{},
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: p2p/discovery/backoff/backoffcache.go [hunk 1]
```diff
@@ -221,13 +246,9 @@ func findPeerDispatcher(ctx context.Context, c *backoffCache, pch <-chan peer.Ad
 			c.peers[ai.ID] = sendAi
 
 			for ch, rem := range c.sendingChs {
-				ch <- sendAi
-				if rem == 1 {
-					close(ch)
-					delete(c.sendingChs, ch)
-					break
-				} else if rem > 0 {
-					rem--
```
MarcoPolo, line 230: This is broken, right? This is a copy of an int not a reference so it wouldn't affect the rem value in the map at all.

p2p/discovery/backoff/backoffcache.go
```diff
@@ -221,13 +246,9 @@ func findPeerDispatcher(ctx context.Context, c *backoffCache, pch <-chan peer.Ad
 			c.peers[ai.ID] = sendAi
 
 			for ch, rem := range c.sendingChs {
-				ch <- sendAi
-				if rem == 1 {
-					close(ch)
-					delete(c.sendingChs, ch)
```
MarcoPolo, line 227: There's no point doing this here, we can clean everything up in one go at the end.

marten-seemann (commented):
In other parts of the code, we use https://github.com/benbjohnson/clock to mock the clock. Can we reuse that package here?

MarcoPolo:
> In other parts of the code, we use https://github.com/benbjohnson/clock to mock the clock. Can we reuse that package here?

Sure thing. Taking a look at that library I noticed they used a pointer receiver though, which goes against my goal of having the real clock be zero sized. It doesn't actually matter I guess but worth pointing out.

Here's some benchmarks:
https://gist.github.com/MarcoPolo/abde3e02d8a6ae49933fe28778b38ffd

The pointer receiver version is 45ns/op vs 38ns/op.

MarcoPolo:
I can actually keep the real clock as I have it here and use this only in tests.

marten-seemann (approved):

PR #1519, opened 2022-05-23
fix: #1512 RelayFinder should trigger a new candidate selection when …

### LINKED ISSUE #1512: RelayFinder should trigger a new candidate selection when a relay lost connection
Correct me if i am wrong.

In refreshRelayReservation, if a relay lost connection, it will be removed from rf.relays  but it's not removed frome rf.candidates

In the worst case, if the rf.cabdidates number reached max candidates, new candidates from peerChan will not be added and no handleNewCandidates call will be triggered. 

For a node running behind an NAT firewall.  After all relays have lost connections, the node wouldn't be able to select any candidates. leaving the node unreachable.

Maybe we shoud find some way to trigger a new candidate selection when a relay lost connection. or just put that relay peer to backoff. making room for new peers from the peerChan

nickjfree:
fix #1512

### DIFF
--- a/p2p/host/autorelay/relay_finder.go
+++ b/p2p/host/autorelay/relay_finder.go
@@ -127,6 +127,8 @@ func (rf *relayFinder) background(ctx context.Context) {
 	for {
 		// when true, we need to identify push
 		var push bool
+		// when true, we need to select new candidates
+		var disconnected bool
 
 		select {
 		case ev, ok := <-subConnectedness.Out():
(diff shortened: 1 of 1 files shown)


marten-seemann:
As I've elaborated in https://github.com/libp2p/go-libp2p/issues/1512#issuecomment-1132763141, I'm not convinced there's anything to fix here.

BigLep:
2022-05-27 conversation: this code is going to get rewritten due to some API changes we'll be making soon.  @marten-seemann will create the issue for that, link this pr to that issue and then close.

marten-seemann:
Here's the issue: #1569.

PR #1524, opened 2022-05-24
chore: replace btcd with one without chaincfg/chainhash

galargh:
These pass with the replace:
```
go mod tidy
go mod test
```

They removed `chaincfg/chainhash` in https://github.com/btcsuite/btcd/pull/1851


galargh:
No, then it stops complaining about `v0.22.0-beta` but it starts complaining about `v0.20.1-beta`.

Looks like you found how to make it work though: https://github.com/libp2p/go-libp2p-core/issues/252#issuecomment-1136262604 🎉

PR #1558, opened 2022-05-26
update Codecov to v3.1.0

marten-seemann:
Looks like it might be time for a shiny new upload script. Because apparently providing an uploader that just works is too hard for @codecov.


marten-seemann:
Looks like this doesn't help.

Issue #1529, opened 2022-05-25
tls: flaky TestInvalidCerts


marten-seemann:
```
      --- FAIL: TestInvalidCerts/client_offending:_signature_is_invalid (0.03s)
  panic: runtime error: invalid memory address or nil pointer dereference [recovered]
  	panic: runtime error: invalid memory address or nil pointer dereference
  [signal 0xc0000005 code=0x0 addr=0x10 pc=0xeae8a3]
  
  goroutine 144 [running]:
  testing.tRunner.func1.2({0xeea760, 0x12c5398})
  	C:/hostedtoolcache/windows/go/1.17.10/x64/src/testing/testing.go:1209 +0x2ab
  testing.tRunner.func1(0x118843c0)
  	C:/hostedtoolcache/windows/go/1.17.10/x64/src/testing/testing.go:1212 +0x27d
  panic({0xeea760, 0x12c5398})
  	C:/hostedtoolcache/windows/go/1.17.10/x64/src/runtime/panic.go:1038 +0x1bc
  github.com/libp2p/go-libp2p/p2p/security/tls.TestInvalidCerts.func19(0x118843c0)
  	D:/a/go-libp2p/go-libp2p/p2p/security/tls/transport_test.go:494 +0x4f3
  testing.tRunner(0x118843c0, 0x11b4aa50)
  	C:/hostedtoolcache/windows/go/1.17.10/x64/src/testing/testing.go:1259 +0xe6
  created by testing.(*T).Run
  	C:/hostedtoolcache/windows/go/1.17.10/x64/src/testing/testing.go:1306 +0x33a

**Comment by marten-seemann** [comment]
This only fails on Windows, but with different error messages:

```
=== RUN   TestInvalidCerts/server_offending:_signature_is_malformed
    transport_test.go:60:
                Error Trace:    transport_test.go:60
                                                        transport_test.go:512
                Error:          Received unexpected error:
                                dial tcp 127.0.0.1:49534: connectex: Only one usage of each socket address (protocol/network address/port) is normally permitted.
                Test:           TestInvalidCerts/server_offending:_signature_is_malformed
--- FAIL: TestInvalidCerts (0.03s)
```

```
=== RUN   TestInvalidCerts/client_offending:_signature_is_invalid
    transport_test.go:490: unexpected error: read tcp 127.0.0.1:64919->127.0.0.1:64916: wsarecv: An existing connection was forcibly closed by the remote host.
[...]

PR #1560, opened 2022-05-26
tls: fix flaky TestInvalidCerts on Windows

marten-seemann:
Fixes https://github.com/libp2p/go-libp2p/issues/1529.

The amount of trivial things that simply don't work reliably on Windows is mind blowing. Absolutely incredible.

This PR should not work around those and make the TLS tests pass on Windows.


marten-seemann:
@MarcoPolo This is ready for review now.

MarcoPolo (approved):

PR #1562, opened 2022-05-26
swarm: send notifications synchronously

marten-seemann:
Recreating https://github.com/libp2p/go-libp2p-swarm/pull/295.

We've notified downstream (especially bitswap) a long time ago, so we'll go ahead and make this change.


MarcoPolo:
> A better solution would likely be to either:
 
>     2. Write connect/disconnect events to a channel (synchronously) and manage them in a background event loop (in-order).

I agree, but just to clarify: This should be done by the protocol implementation itself and not the library. The implementation may want to keep a buffer here or spawn workers or something else. The library shouldn't do extra allocs or work to try to predict what a protocol implementation would want.

So the patch I linked above could be expanded to write to a channel and spawn a goroutine to do the work. Ordering would be guaranteed and you'd have the same semantics as now.

Wondertan:
> Maybe we should rework our Connectedness event so that we can signal the transition from transient -> not transient?

Yes, pls. Currently, this is what I've been doing when handling Connectedness event and it would be nicer to have transient flag on the event instead.
```go
for _, c := range m.host.Network().ConnsToPeer(p) {
		if c.Stat().Transient {
			continue
		}
}
```

MarcoPolo (commented):
What do you think about changing the API to accept a new context parameter? I'd like to make this a breaking change for users of this interface because the semantics change a bit (now a notifiee can delay other notifiee for the same event).

p2p/net/swarm/swarm.go
```diff
@@ -541,18 +541,10 @@ func (s *Swarm) Backoff() *DialBackoff {
...
 // notifyAll sends a signal to all Notifiees
 func (s *Swarm) notifyAll(notify func(network.Notifiee)) {
-	var wg sync.WaitGroup
-
 	s.notifs.RLock()
-	wg.Add(len(s.notifs.m))
 	for f := range s.notifs.m {
-		go func(f network.Notifiee) {
-			defer wg.Done()
-			notify(f)
```
MarcoPolo, line 551: I think this old code is actually a bit broken wrt to ordering of events.

Since it only acquires a read lock, there's nothing stopping the notifiee from getting an event out of order since the notify call happens in async in a goroutine.

I think we should use a normal mutex here instead of RWMutex to guarantee this ordering.

Stebalien:
> Maybe we should rework our Connectedness event so that we can signal the transition from transient -> not transient?

_That_ would be great. For now, I'm playing around with some other fixes.

> I agree, but just to clarify: This should be done by the protocol implementation itself and not the library. The implementation may want to keep a buffer here or spawn workers or something else. The library shouldn't do extra allocs or work to try to predict what a protocol implementation would want.

I agree that go-libp2p shouldn't have to spawn goroutines, that's technical debt that we need to handle in the users of go-libp2p.

However, go-libp2p definitely needs to continue delivering notifications in-order.

> What do you think about changing the API to accept a new context parameter? I'd like to make this a breaking change for users of this interface because the semantics change a bit (now a notifiee can delay other notifiee for the same event).

I'm not sure what we'd use the context for?

MarcoPolo (changes requested):

marten-seemann:
> What do you think about changing the API to accept a new context parameter?

How would that context be used? Under which conditions would we cancel it?

MarcoPolo:
> How would that context be used? Under which conditions would we cancel it?

No specific thoughts here, but maybe a 1 second timeout would be good enough. It's not like we can guarantee that notifiees would use it anyways. And maybe in the future if we have a useful parent context we could pass that in.

MarcoPolo:
> I'm not sure what we'd use the context for?

My reasoning here is to purposely break the api so that downstream users notice this change.

p2p/net/swarm/swarm.go
Stebalien: That lock is protecting the notifiee list (and is global). The lock you're looking for is https://github.com/libp2p/go-libp2p/blob/c7928312d6c90635c835ac3e4bd5534b1763189a/p2p/net/swarm/swarm.go#L303-L304.

MarcoPolo: Ah okay, thank you!

MarcoPolo:
No strong opinion re context here really.

MarcoPolo (approved):

marten-seemann:
I do like the idea of breaking the API, but I'm a bit worried introducing a context that's not really used. Let's merge it as is.

Stebalien:
Bitswap patch to avoid blocking: https://github.com/ipfs/go-bitswap/pull/565.

Issue #1556, opened 2022-05-26
Repo Consolidation: Round 3


marten-seemann:
The transport consolidation (#1187) has dramatically reduced repo sprawl. We now only have a few go-libp2p-* repos left in the dependency tree of go-lipbp2p. This is a proposal to also consolidate them into go-lipbp2:

- [x] go-libp2p-resource-manager: `/p2p/host/resource-manager`
   - [x] Update links in dos-mitigation.md: https://github.com/libp2p/docs/pull/160/
- [x] go-peerstore: `/p2p/host/peerstore`
- [x] go-libp2p-core: `/core`
- [x] go-eventbus: `/p2p/host/eventbus`

Not sure what to do about go-libp2p-asn-util. This repo should probably have been named go-asn-util, it's not libp2p related in any way. I also don't feel like moving several MB of ASN here.

Anything I missed? Thoughts, @MarcoPolo @vyzo @Stebalien @raulk?

vyzo:
asn no, the rest is reasonable proposition.

vyzo:
I think we should keep core separate, it's the interface after all.

marten-seemann:
> I think we should keep core separate, it's the interface after all.

I think I disagree. core being separate is causing us a lot of trouble:
* When we want to change one of the interfaces, we first have to cut a core release before we can use them in go-libp2p. It would be nicer to make that change atomically.
* core and go-libp2p being out of sync (core already having released an update, but go-libp2p not yet) causes errors for downstream users who run `go get -u`, and then we need to tell them which versions of core go with which versions of go-libp2p

We probably should keep the GitHub Action that posts the `gocompat` output every time we make a change to `/core`, so we don't accidentally break things.

vyzo:
Yes, but it allows libraries to just depend on core and not pull the whole
go-libp2p except for trsts.

marten-seemann:
Should we be worried about this? I'd assume that if you pull in libp2p types, you're also using go-libp2p. You _could_ imagine a project suffering from repo sprawl where a library only needs the type, and go-libp2p is pulled in by their main repo, but even then the only effect is that `go get` will download slightly more data.

vyzo:
It is semantically different; i think a separate interface is good practice for something as complex as libp2p.
After all, apart from instantiation and tests, user code only cares about the interface.

vyzo:
think of it as the header file.

marten-seemann:
[...]

PR #1564, opened 2022-05-26
move go-libp2p-resource-manager here

marten-seemann:
Part of #1556.


Stebalien (approved):

vyzo (approved):
alright, lets do it.

vyzo:
So there is a bunch of activity here, with tooling that needs to be created -- see all the new issues.

I propose we delay the merge until these are resolved.

PR #1572, opened 2022-05-27
basichost: Fix flakey test on windows by scaling up timeout

MarcoPolo:
`TestSignedPeerRecordWithNoListenAddrs` Has flaked at least 10 times on Windows in the last week: https://observablehq.com/@realmarcopolo/top-flaky-tests-on-go-libp2p

This simply scales up the timeout times on CI.


marten-seemann:
Any reason not to just use a longer timeout here, as we've done for the other tests? (I think I've submitted a fix along those lines in #1559)

MarcoPolo:
No strong thoughts here.

My reasoning is to convey that CI may run 10x slower than your laptop. When writing tests locally it's fine to use numbers that make sense for you, then wrap those numbers with this function. It's a better flow I think to pick reasonable numbers that work for you locally and then wrap this scale duration function, than it is to scale numbers directly.

MarcoPolo:
Closing as already fixed

PR #1574, opened 2022-05-28
swarm: emit PeerConnectedness event from swarm instead of from hosts

### LINKED ISSUE #1571: swarm: emit PeerConnectedness event
Currently, both the blank host and the basic host run identical code to emit `PeerConnectedness` events. This logic should probably live in the swarm instead.

Potential challenge: we'll need to pass the event bus to the swarm constructor somehow.

marten-seemann:
Fixes #1571.

This was surprisingly easy. Unfortunately, I can't find a way to test the deduplication (that no event is emitted when a second connection is dialed), since the swarm is too good at preventing dials to hosts we're already connected to.


p2p/host/basic/basic_host.go
```diff
@@ -112,6 +112,9 @@ var _ host.Host = (*BasicHost)(nil)
 // HostOpts holds options that can be passed to NewHost in order to
 // customize construction of the *BasicHost.
 type HostOpts struct {
+	// EventBus sets the event bus. Will construct a new event bus if omitted.
```
vyzo, line 115: I feel uneasy about this.
Maybr we should add the EventBus method to the Network interface and forward from the host so that wr cant end up with two of them.

marten-seemann: Not really how this could happen, but we can make this change.

Can we defer this until we consolidated all the other repos as well (#1566)? It's happening in the next release.

p2p/net/swarm/swarm.go
```diff
@@ -310,6 +321,13 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
 	})
 	c.notifyLk.Unlock()
 
+	if isFirstConn {
```
MarcoPolo, line 324: Are we sure we want this outside of the `notifyLk`? That changes the behavior from what it used to be (inside the notifyAll).

p2p/net/swarm/swarm.go
```diff
@@ -588,9 +607,17 @@ func (s *Swarm) removeConn(c *Conn) {
...
 				s.conns.m[p] = cs[:len(cs)-1]
 			}
-			return
+			break
 		}
 	}
+	s.conns.Unlock()
+
+	if disconnected {
+		s.emitter.Emit(event.EvtPeerConnectednessChanged{
```
MarcoPolo, line 616: What's the reason to do this here instead of near the old place where this happened: https://github.com/libp2p/go-libp2p/blob/marco/1575/p2p/net/swarm/swarm_conn.go#L85 ?

p2p/host/basic/basic_host.go
Stebalien: Another option would be to just make this entire thing private.

p2p/net/swarm/swarm.go
Stebalien: Yeah, I think this needs to be inside. We need to make sure that the disconnect comes after the connect event.

Stebalien: Hm. Actually, I'm not sure if this lock helps us here. We need to make this guarantee across all connections.

p2p/net/swarm/swarm.go
Stebalien: Probably to ensure that we only emit the event on the last disconnect. But it does look like there's a bit of a race and we can emit a disconnect after a connect (looks like we already _have_ that race, actually).

Stebalien:
So, I think the only way to sanely do this is to write something like: https://github.com/ipfs/go-bitswap/pull/565/files. We probably don't want to block some global queue while sending events.

That is:

1. On connect/disconnect, mark the peer as connected/disconnected in some tracker. Add the peer to some "todo" queue, and tell the background process that it needs to wake up.
2. In a background event loop, fire connect/disconnect events.

Even if this background process falls behind (someone is blocking events), that's fine. We'll just always fire the "last" event. This _does_ mean we can get unmatched connected/disconnected events, but, IMO, that's _fine_.

I.e.:

1. Connect.
2. Disconnect.
3. Connect.
4. Disconnect.

May be seen as:

1. Connect.
2. Connect.
3. Disconnect.

Or

1. Disconnect.
2. Disconnect.

Or

1. Connect.
2. Disconnect.

Or

1. Disconnect.

But never:

1. Connect.
2. Disconnect.
3. Connect.

Importantly, the _final_ state is always accurate, which is what users care about. At the moment, it's possible for the _last_ event fired to be "connect", which could cause memory leaks.

marten-seemann:
Rebased this PR on top of the current master.

I think emitting the event while holding the `conns` mutex should be fine:
* It's correct since the the `swarm.conns.m[peer.ID]` is the source of truth: Once we're down to 0 connections in this list, we're officially disconnected from the peer.
* It's ok performance-wise, because emitting events is cheap, as long as the subscribers consume events quickly enough. Now we could (somewhat) protect against that by introducing a worker Go routine and another tracking data structure as suggested by @Stebalien in https://github.com/libp2p/go-libp2p/pull/1574#issuecomment-1143048078, but that data structure would need to be limited in size _somehow_, so this is no better than using a larger channel for the event bus subscriber. Using the new event bus dashboard (#2038) it should be easy to detect a misconfiguration there.

I introduced the change in the last commit. @MarcoPolo, @Stebalien does this reasoning make sense to you?

p2p/host/basic/basic_host.go
marten-seemann: I'm not a big fan of having both a `Config` and `HostOpts`. I expect us to merge these two (or refactor them) in some other way when we start using Fx for service construction (#1993). I suggest not making any big changes in this PR.

p2p/net/swarm/swarm_event_test.go
```diff
@@ -0,0 +1,66 @@
...
+		require.Equal(t, expected.Peer, evt.Peer)
+	case <-time.After(time.Second):
+		t.Fatal("didn't get PeerConnectedness event")
+	}
+
+	// check that there are no more events
+	select {
+	case <-sub.Out():
+		t.Fatal("didn't expect any more events")
+	case <-time.After(50 * time.Millisecond):
```
MarcoPolo, line 45: probably bump this to 100ms for CI. I know it'll slow things down 50ms, sorry.

MarcoPolo (approved):
Changes look good and reasoning makes sense.

> Unfortunately, I can't find a way to test the deduplication (that no event is emitted when a second connection is dialed)

You could try this:
Make a new test transport that wraps an existing transport. Put two multiaddrs for this test transport (they might have to be slightly different if we're deduping these somewhere). In the test transport have them synchronize with a waitgroup to return a dial at the same time.

p2p/net/swarm/swarm_event_test.go
marten-seemann: Done.

marten-seemann:
> Make a new test transport that wraps an existing transport. Put two multiaddrs for this test transport (they might have to be slightly different if we're deduping these somewhere). In the test transport have them synchronize with a waitgroup to return a dial at the same time.

That doesn't work either. The dial worker loop _somewhere_ deduplicates. I have intention of diving into that mess to figure out where.

Issue #1569, opened 2022-05-27
make AutoRelay ask the application for new candidates


marten-seemann:
... instead of expecting the application to pass in new candidates.

PR #1576, opened 2022-05-28
change the default key type to Ed25519

marten-seemann:
Cryptographically, there's no reason to prefer RSA over Ed25519 (quite the contrary, actually!).

The peer ID spec says that all implementations MUST support Ed25519, so this _should_ be a safe change to make.

@vyzo @Stebalien Are you aware of any backwards compatibility issues this might cause (that wouldn't be better handled by the application that requires backwards compatibility)?

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -56,7 +56,7 @@ var DefaultPeerstore Option = func(cfg *Config) error {
 
 // RandomIdentity generates a random identity. (default behaviour)
 var RandomIdentity = func(cfg *Config) error {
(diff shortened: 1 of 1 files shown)


marten-seemann:
Interestingly, this change is driving the AutoRelay `TestSingleRelay` crazy. This test is already the most flaky test by a large margin, and will get fixed once I tackle #1569. I suggest delaying the merge of this PR until that PR has landed, and CI is happy.

vyzo:
I think this is a change we should have made a while ago, RSA keys are a pita.

Stebalien (approved):
We made this switch quite a while ago in go-ipfs.

MarcoPolo (approved):

PR #1580, opened 2022-05-31
Added perPeerRateLimit env config

### LINKED ISSUE #1579: Is there a way to dynamically configure `DefaultPerPeerRateLimit`?
https://github.com/libp2p/go-libp2p/blob/5c297a87002a04a68143c59167502a9be6c73f2f/p2p/net/swarm/limiter.go#L57

millken:
Fix #1579

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -104,6 +104,9 @@ type Config struct {
 
 	EnableHolePunching  bool
 	HolePunchingOptions []holepunch.Option
+
+	FDLimit      int
+	PerPeerLimit int
 }
 
 func (cfg *Config) makeSwarm() (*swarm.Swarm, error) {
@@ -150,6 +153,12 @@ func (cfg *Config) makeSwarm() (*swarm.Swarm, error) {
 	}
 	if cfg.ResourceManager != nil {
 		opts = append(opts, swarm.WithResourceManager(cfg.ResourceManager))
+	}
+	if cfg.FDLimit != 0 {
+		opts = append(opts, swarm.WithFDLimit(cfg.FDLimit))
+	}
+	if cfg.PerPeerLimit != 0 {
+		opts = append(opts, swarm.WithPerPeerLimit(cfg.PerPeerLimit))
 	}
 	// TODO: Make the swarm implementation configurable.
 	return swarm.NewSwarm(pid, cfg.Peerstore, opts...)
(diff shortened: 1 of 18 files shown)


vyzo (approved):

marten-seemann:
Not a big fan of making this configurable via environment variable.
The best fix would be to make the swarm injectable, and then use a constructor option. The next-better fix would be to have a package-level global variable.

MarcoPolo:
> Not a big fan of making this configurable via environment variable.
> The best fix would be to make the swarm injectable, and then use a constructor option. The next-better fix would be to have a package-level global variable.

Agreed. @millken mind making the package level global  change please? It’s better since then this is exposed prominently in the pkg reference, and your code can read your env var and set the variable.

millken:
> > Not a big fan of making this configurable via environment variable.
> > The best fix would be to make the swarm injectable, and then use a constructor option. The next-better fix would be to have a package-level global variable.
> 
> Agreed. @millken mind making the package level global change please? It’s better since then this is exposed prominently in the pkg reference, and your code can read your env var and set the variable.

Updated

config/config.go
```diff
@@ -104,6 +104,9 @@ type Config struct {
 
 	EnableHolePunching  bool
 	HolePunchingOptions []holepunch.Option
+
+	FDLimit      int
+	PerPeerLimit int
```
marten-seemann, line 109: There's no way for a libp2p user to directly set values in this config.

marten-seemann (changes requested):

marten-seemann (changes requested):
I don't think we should have global config options for this. This seems overly specific.
Why don't you just add a variable, as @MarcoPolo suggested in https://github.com/libp2p/go-libp2p/pull/1580#issuecomment-1145483558.

millken:
> I don't think we should have global config options for this. This seems overly specific. Why don't you just add a variable, as @MarcoPolo suggested in [#1580 (comment)](https://github.com/libp2p/go-libp2p/pull/1580#issuecomment-1145483558).

Ok, Just change `const DefaultPerPeerRateLimit = 8` to `var DefaultPerPeerRateLimit = 8`

MarcoPolo (approved):

MarcoPolo:
Thanks!

PR #1586, opened 2022-06-01
swarm: add ListenClose

birros:
Recreating https://github.com/libp2p/go-libp2p-swarm/pull/319.

Add the missing `ListenClose` method, as it is impossible to close existing listeners.

Useful in many situations, especially on systems that no longer handle listeners after the operating system has suspended them without closing them, when applications run in the background for example (iOS).

This offers the possibility to update the network listener without completely restarting the swarm.

I think this method should also be added to the network interface.

The decision is yours

### DIFF
--- a/p2p/net/swarm/swarm_listen.go
+++ b/p2p/net/swarm/swarm_listen.go
@@ -9,6 +9,10 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+var (
+	ErrSwarmListenerAcceptError = fmt.Errorf("swarm listener accept error")
+)
+
 // Listen sets up listeners for all of the given addresses.
 // It returns as long as we successfully listen on at least *one* address.
 func (s *Swarm) Listen(addrs ...ma.Multiaddr) error {
@@ -33,6 +37,20 @@ func (s *Swarm) Listen(addrs ...ma.Multiaddr) error {
 	}
 
 	return nil
+}
+
+// ListenClose stop and delete listeners for all of the given addresses.
+func (s *Swarm) ListenClose(addrs ...ma.Multiaddr) {
+	s.listeners.Lock()
+	defer s.listeners.Unlock()
+
+	for l := range s.listeners.m {
+		if !containsMultiaddr(addrs, l.Multiaddr()) {
+			continue
+		}
+
+		l.Close()
+	}
 }
 
 // AddListenAddr tells the swarm to listen on a single address. Unlike Listen,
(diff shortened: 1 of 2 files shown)


p2p/net/swarm/swarm_listen.go
```diff
@@ -35,6 +39,20 @@ func (s *Swarm) Listen(addrs ...ma.Multiaddr) error {
...
+func (s *Swarm) ListenClose(addrs ...ma.Multiaddr) {
+	s.listeners.Lock()
+	defer s.listeners.Unlock()
+
+	for l := range s.listeners.m {
+		if !containsMultiaddr(addrs, l.Multiaddr()) {
+			continue
+		}
+
+		l.Close()
```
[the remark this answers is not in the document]
Stebalien: We can copy the listeners into a temporary slice.

p2p/net/swarm/swarm_listen.go
```diff
@@ -79,24 +97,27 @@ func (s *Swarm) AddListenAddr(a ma.Multiaddr) error {
...
+
+			// log if no more listener and the swarm is still running.
+			if len(s.listeners.m) == 0 && s.ctx.Err() == nil {
+				log.Error(ErrSwarmListenerAcceptError)
+			}
 		}()
 		for {
 			c, err := list.Accept()
 			if err != nil {
-				if s.ctx.Err() == nil {
```
Stebalien, line 96: This is here to tell the user that a listener failed for some reason, we need to keep it here.

It sounds like we may need a way to mark listeners as _intentionally_ closed. Maybe just remove them from the listeners map in `ListenClose`? That's probably the simplest way (and ensures that there's no funky race-condition while the listener is in the process of being closed). Then, in the defer, we can:

1. Close the listener if and only if it's still in the listeners map.
2. Log an error if and only if the listener was still in the map. (we clear the map on shutdown as well).

Stebalien (changes requested):

p2p/net/swarm/swarm_listen.go
birros: I agree, but we have to close the listeners in `ListenClose` because if we just delete them from the map, they will never be closed. In the logic you describe, the listener is only closed in the defer if it is still present in the map, so if it was simply deleted before in `ListenClose`, it will never be closed.

Stebalien: I'm saying that we _both_ close the listener _and_ remove it from the map in `ListenClose`.

Stebalien: That way:

1. We can tell here if the listener was intentionally closed (is it still in the map or not).
2. ListenAddresses, and friends will return the correct answers immediately instead of waiting for this function to return asynchronously.

birros: That's what I did (at least I hope so) in the last commit. I use the presence of the listener in the map to conditionally close it in the defer, while the call to `ListenClose` closes it immediately.

p2p/net/swarm/swarm_listen.go
```diff
@@ -94,8 +122,15 @@ func (s *Swarm) AddListenAddr(a ma.Multiaddr) error {
...
-					log.Errorf("swarm listener accept error: %s", err)
+					s.listeners.Lock()
+					_, ok := s.listeners.m[list]
+					s.listeners.Unlock()
+
+					if ok {
+						// only log if the swarm is still running and the
+						// listener has not been closed intentionally.
+						log.Errorf("swarm listener accept error: %s", err)
+					}
```
Stebalien, line 133: 1. We don't need to check `s.ctx` anymore.
2. We can move this into the defer. If the listener ever closes without being removed from the map, that means it wasn't intentionally closed and we should log.

Stebalien (approved):
LGTM. One nit, but I don't feel strongly about it.

p2p/net/swarm/swarm_listen.go
Stebalien: Uh, yes, yes you did.

p2p/net/swarm/swarm_listen.go
birros: Done, I also changed the log to a more explicit message: `swarm listener unintentionally closed`.

Stebalien (approved):
LGTM!

Stebalien:
@marten-seemann LGTM, but I'll leave it up to you to merge.

marten-seemann (approved):

PR #1587, opened 2022-06-01
autorelay: poll for new candidates when needed

### LINKED ISSUE #1440: autorelay: flaky TestSingleRelay
```
  === RUN   TestSingleRelay
      autorelay_test.go:154: 
          	Error Trace:	autorelay_test.go:154
          	Test:       	TestSingleRelay
  --- FAIL: TestSingleRelay (3.66s)

### LINKED ISSUE #1569: make AutoRelay ask the application for new candidates

### PR DESCRIPTION
**Pull request by marten-seemann**
Fixes #1569. Fixes #1440. Closes #1577.

Unblocks https://github.com/libp2p/go-libp2p/pull/1576 and https://github.com/libp2p/go-libp2p/pull/1615.

### DIFF
--- a/p2p/host/autorelay/relay_finder.go
+++ b/p2p/host/autorelay/relay_finder.go
@@ -66,10 +66,11 @@ type relayFinder struct {
 	ctxCancel   context.CancelFunc
 	ctxCancelMx sync.Mutex
 
-	peerChan <-chan peer.AddrInfo
+	peerSource func(int) <-chan peer.AddrInfo
 
 	candidateFound            chan struct{} // receives every time we find a new relay candidate
 	candidateMx               sync.Mutex
+	lastCandidateAdded        time.Time
 	candidates                map[peer.ID]*candidate
 	candidatesOnBackoff       []*candidateOnBackoff // this slice is always sorted by the nextConnAttempt time
 	handleNewCandidateTrigger chan struct{}         // cap: 1
@@ -83,27 +84,37 @@ type relayFinder struct {
 	cachedAddrsExpiry time.Time
 }
 
-func newRelayFinder(host *basic.BasicHost, peerChan <-chan peer.AddrInfo, conf *config) *relayFinder {
-	r := &relayFinder{
+func newRelayFinder(host *basic.BasicHost, peerSource func(int) <-chan peer.AddrInfo, conf *config) *relayFinder {
+	return &relayFinder{
 		bootTime:                  conf.clock.Now(),
 		host:                      host,
 		conf:                      conf,
-		peerChan:                  peerChan,
+		peerSource:                peerSource,
 		candidates:                make(map[peer.ID]*candidate),
 		candidateFound:            make(chan struct{}, 1),
 		handleNewCandidateTrigger: make(chan struct{}, 1),
 		relays:                    make(map[peer.ID]*circuitv2.Reservation),
 		relayUpdated:              make(chan struct{}, 1),
 	}
-	return r
 }
 
 func (rf *relayFinder) background(ctx context.Context) {
-	rf.refCount.Add(1)
-	go func() {
-		defer rf.refCount.Done()
-		rf.findNodes(ctx)
-	}()
+	relayDisconnected := make(chan struct{}, 1)
+
+	if rf.usesStaticRelay() {
+		rf.refCount.Add(1)
+		go func() {
+			defer rf.refCount.Done()
+			rf.handleStaticRelays(ctx)
+		}()
+	} else {
+		rf.refCount.Add(1)
+		go func() {
+			defer rf.refCount.Done()
+			rf.findNodes(ctx, relayDisconnected)
+		}()
+	}
+
 	rf.refCount.Add(1)
 	go func() {
 		defer rf.refCount.Done()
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by vyzo — COMMENTED** [review]
strictly an improvement i would say, but see comment for potential little bug.

File: p2p/host/autorelay/relay_finder.go [hunk 1]
```diff
@@ -83,28 +84,34 @@ type relayFinder struct {
...
 
 func (rf *relayFinder) background(ctx context.Context) {
-	rf.refCount.Add(1)
-	go func() {
-		defer rf.refCount.Done()
-		rf.findNodes(ctx)
-	}()
-	rf.refCount.Add(1)
+	relayDisconnected := make(chan struct{}, 1)
+	rf.refCount.Add(2)
```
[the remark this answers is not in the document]
marten-seemann: We start one Go routine in the if/else and one after that. I can split that up into multiple `Add` statements, if that makes things more readable.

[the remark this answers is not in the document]
vyzo: yeah, lets do that.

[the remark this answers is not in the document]
marten-seemann: Done.

vyzo (approved):

marten-seemann:
I rewrote this PR: We now don't keep candidates that we fail to connect to, we throw them out immediately. The application may return them to use on subsequent calls of `PeerSource`, and we apply a backoff to make sure we don't connect to a node too frequently.

This should now resolve all the flaky autorelay tests.

p2p/host/autorelay/relay_finder.go
```diff
@@ -178,20 +207,59 @@ func (rf *relayFinder) background(ctx context.Context) {
 // It garbage collects old entries, so that nodes doesn't overflow.
 // This makes sure that as soon as we need to find relay candidates, we have them available.
 func (rf *relayFinder) findNodes(ctx context.Context) {
+	peerChan := rf.peerSource(rf.conf.maxCandidates)
```
MarcoPolo, line 210: Don't you have to check if this func var may be nil? Or if you assume this is non nil, should we assert that in the constructor?

p2p/host/autorelay/relay_finder.go
```diff
@@ -232,10 +331,17 @@ func (rf *relayFinder) handleNewNode(ctx context.Context, pi peer.AddrInfo) {
...
+		added:           rf.conf.clock.Now(),
+		ai:              pi,
+		supportsRelayV2: supportsV2,
+	}
 	rf.candidateMx.Unlock()
 
-	rf.notifyNewCandidate()
+	// Don't notify when we're using static relays. We need to process all entries first.
+	if !rf.usesStaticRelay() {
+		rf.notifyNewCandidate()
```
MarcoPolo, line 343: It's not ideal that this function has a notion of "static relay setup".

I think it would be better if this function returned whether this node was added as a candidate, and the caller should then decide if it wants to notify new candidate. This way the `handleStaticRelays` caller can wait until the end, and the findNodes caller can notifyNewCandidate if it was added.

p2p/host/autorelay/timer.go
```diff
@@ -0,0 +1,42 @@
...
+	if !t.running {
+		return
+	}
+	if !t.timer.Stop() && !t.read {
+		<-t.timer.C
+	}
+	t.read = false
+}
+
+func (t *timer) SetRead() {
```
MarcoPolo, line 35: Unused?

p2p/host/autorelay/timer.go
```diff
@@ -0,0 +1,42 @@
+package autorelay
+
+import (
+	"time"
+
+	"github.com/benbjohnson/clock"
+)
+
+type timer struct {
```
MarcoPolo, line 9: Do we need this? It's only used in one spot and it's not hard to call .Stop() yourself

MarcoPolo (changes requested):
Looks good, just a couple nits. I wonder how flaky these tests will be. At least we aren't relying on a real clock anymore!

p2p/host/autorelay/timer.go
marten-seemann: Fixed.

p2p/host/autorelay/timer.go
marten-seemann: It's not hard to call stop, but we need to drain the channel. But only if the timer was running before, otherwise `<-t.C` will block indefinitely.

p2p/host/autorelay/relay_finder.go
marten-seemann: That would be a violation of the API contract. Not sure if we have to defend against that (panic?).

`findNodes` is called repeatedly, so we can't move it to the constructor.

p2p/host/autorelay/relay_finder.go
marten-seemann: Good suggestion! Applied in f5ea2ef58487bed41c6e94cebf8ea18423189110.

marten-seemann:
All suggestions applied. Re-requesting a review, @vyzo and @MarcoPolo 🙏

p2p/host/autorelay/timer.go
MarcoPolo: ah I see. Makes sense, thanks

MarcoPolo (approved):

Issue #1514, opened 2022-05-20
go-libp2p v0.21.0


marten-seemann:
### Monitoring and Resource Management

- [x] https://github.com/libp2p/go-libp2p-resource-manager/issues/32 and https://github.com/libp2p/go-libp2p-resource-manager/pull/48
- [x] https://github.com/libp2p/go-libp2p-resource-manager/issues/29 and https://github.com/libp2p/go-libp2p-resource-manager/pull/47
  - This includes the relevant configuration as well
- [x] https://github.com/libp2p/go-libp2p-resource-manager/issues/50 and https://github.com/libp2p/go-libp2p-resource-manager/pull/54
- [x] https://github.com/libp2p/go-libp2p/pull/1624

### Swarm Sync Notification

* [x] https://github.com/libp2p/go-libp2p/pull/1562

### Misc

* [x] update go-yamux (at least to v3.1.2), but better to a version that using go-log
* [x] https://github.com/libp2p/go-buffer-pool/issues/26
* [x] https://github.com/libp2p/go-libp2p/issues/1599
* [x] https://github.com/libp2p/go-libp2p/pull/1192

### Resource Management
* Default resource management config limits will now [autoscale](https://pkg.go.dev/github.com/libp2p/go-libp2p-resource-manager#ScalingLimitConfig.AutoScale) depending on the machine size.
* You can now specify an [allowlisted](https://pkg.go.dev/github.com/libp2p/go-libp2p-resource-manager#WithAllowlistedMultiaddrs) set of multiaddrs that can still connect to you even if you've reached your normal limits (still limited by the allowlist limits).
* Resource manager now exposes opencensus metrics and provides a default [grafana dashboard](https://github.com/libp2p/go-libp2p-resource-manager/tree/v0.5.1/obs/grafana-dashboards).

### Swarm Notifications are now synchronous

If your protocol depended on being given notifications in an async context and taking its time processing that, you will slow down your whole go-libp2p application. See https://github.com/libp2p/go-libp2p/pull/1562 for more details. This only matters if you implemented a [`network.Notifiee`](https://pkg.go.dev/github.com/libp2p/go-libp2p-core/network#Notifiee) and registered to receive notifications on swarm network events.

## Changelog

- github.com/libp2p/go-libp2p:
  - Release v0.21.0 RC (#1638) ([libp2p/go-libp2p#1638](https://github.com/libp2p/go-libp2p/pull/1638))
[...]

PR #1592, opened 2022-06-09
feat: allow dialing wss peers using DNS multiaddrs

### LINKED ISSUE #1111: Release v0.14.1
@marten-seemann will fill in.
This is followup to https://github.com/libp2p/go-libp2p/issues/1090

## 🗺 What's left for release

<List of items with remaining PRs and/or Issues to be considered for this release>

## 🔦 Highlights

< top highlights for this release notes >

## Changelog

< changelog generated by scripts/mkreleaselog >

## ✅ Release Checklist

- [ ] **Stage 0 - Finishing Touches**
    - [ ] Go through relevant libp2p repos looking for unreleased changes that should make it into the release. If you find any, cut releases.
    - [ ] Run `go get -u ./...` to see if there are any out-of-date deps that look important. If there are, bubble them. Try to avoid _directly_ updating indirect deps in go-libp2p's go.mod when possible.
    - [ ] Make sure local tests are passing.
- [ ] **Stage 1 - Upstream Testing**
  - Create testing branches in lotus & go-ipfs with the new go-libp2p release and run CI/tests. Many upstream projects are tested in CI, but lotus & go-ipfs are not.
    - [ ] [filecoin-project/lotus](https://github.com/filecoin-project/lotus)
    - [ ] [ipfs/go-ipfs](https://github.com/ipfs/go-ipfs)
  - _(someday)_ Run upstream testground tests. Unfortunately, this is too time consuming at the moment.
    - _(someday)_ Run bitswap testground tests.
    - _(someday)_ Run DHT testground tests.
- [ ] **Stage 2 - Infrastructure Testing**
  - How: Using the testing branches created above, work with the infrastructure team to deploy the new libp2p versions to our infrastructure.
  - Where:
    - [ ] A go-ipfs gateway.
      - [ ] Deploy
      - [ ] Analyze
        - [ ] Look at pprof profile dumps, especially CPU profiles and heap allocation profiles, noting any significant changes.
    - [ ] A go-ipfs bootstrapper.
      - [ ] Deploy
      - [ ] Analyze
        - [ ] Look at pprof profile dumps, especially CPU profiles and heap allocation profiles, noting any significant changes.
        - [ ] Check peers (e.g., `ipfs swarm peers`) to make sure we're connecting to peers on all transports.
        - [ ] Check advertised addresses and protocols (e.g., `ipfs id`) to make sure they're sane.
- [ ] **Stage 3 - Release**
  - [ ] Tag the release on master.
  - [ ] Publish the release through the GitHub UI, adding the release notes. Some users rely on this to receive notifications of new releases.
  - [ ] Announce the release on the [discuss.libp2p.io](https://discuss.libp2p.io).
- [ ] **Stage 4 - Update Upstream**
  - [ ] Update the upstream testing branches to the final release and create PRs.
    - [ ] [filecoin-project/lotus](https://github.com/filecoin-project/lotus)
    - [ ] [ipfs/go-bitswap](https://github.com/ipfs/go-bitswap)
    - [ ] [ipfs/go-ipfs](https://github.com/ipfs/go-ipfs)
    - [ ] [libp2p/go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht)
    - [ ] [libp2p/go-libp2p-pubsub](https://github.com/libp2p/go-libp2p-pubsub)
    - [ ] [libp2p/go-libp2p-daemon](https://github.com/libp2p/go-libp2p-daemon)
- [ ] Make required changes to the release process.

### LINKED ISSUE #1593: Allow dialing wss peers using DNS multiaddrs
# Problem

From what I can tell while we technically have WSS dialing from go-libp2p due to https://github.com/libp2p/go-ws-transport/pull/115 it's basically unusable.

Currently trying to dial some peer like `/dns4/node3.preload.ipfs.io/tcp/443/wss/p2p/QmY7JB6MQXhxHvq7dBDh4HpbH29v4yE9JRadAVpndvzySN` will result in an error like:
```
failed to dial QmY7JB6MQXhxHvq7dBDh4HpbH29v4yE9JRadAVpndvzySN:
  * [/ip4/145.40.97.77/tcp/443/wss] x509: cannot validate certificate for 145.40.97.77 because it doesn't contain any IP SANs
```

i.e. we do not make use of the DNS information in the dialing and so we run into a certificates problem because the certificate is tied to the domain name.

---

Small tool if you want to play with this and see the issue is https://github.com/aschmahmann/vole. You can run something like:

```
vole.exe bitswap check bafybeibml5uieyxa5tufngvg7fgwbkwvlsuntwbxgtskoqynbt7wlchmfm /dns4/node3.preload.ipfs.io/tcp/443/wss/p2p/QmY7JB6MQXhxHvq7dBDh4HpbH29v4yE9JRadAVpndvzySN
```

aschmahmann:
fixes #1593

This fixes an issue where the WSS dialer can only handle IP multiaddrs not DNS https://github.com/libp2p/go-libp2p/blob/d7ba37217c6f1a0d739666f78ac19a87728a7280/p2p/transport/websocket/websocket.go#L26

A couple areas to consider here for improvement (recommendations or maintainers pushing commits to this PR are welcome).
- [ ] We should use the host's DNS resolver rather than the OS resolver
- [ ] The basic host should probably not resolve the DNS addresses covered by WSS, or the dialer should at least prioritize DNS WSS dials over IP ones

### DIFF
--- a/p2p/transport/websocket/websocket.go
+++ b/p2p/transport/websocket/websocket.go
@@ -23,7 +23,7 @@ var WsFmt = mafmt.And(mafmt.TCP, mafmt.Base(ma.P_WS))
 
 // This is _not_ WsFmt because we want the transport to stick to dialing fully
 // resolved addresses.
-var dialMatcher = mafmt.And(mafmt.IP, mafmt.Base(ma.P_TCP), mafmt.Or(mafmt.Base(ma.P_WS), mafmt.Base(ma.P_WSS)))
+var dialMatcher = mafmt.And(mafmt.Or(mafmt.IP, mafmt.DNS), mafmt.Base(ma.P_TCP), mafmt.Or(mafmt.Base(ma.P_WS), mafmt.Base(ma.P_WSS)))
 
 func init() {
 	manet.RegisterFromNetAddr(ParseWebsocketNetAddr, "websocket")


marten-seemann:
Does it make sense to dial WSS in the first place? We might want to consider disabling WSS dialing altogether (except for tests), as any node offering WSS will also offer TCP and that's always preferable.

> We should use the host's DNS resolver rather than the OS resolver

Can we pass the resolver to this transport? Probably would have to teach Steven's constructor magic that there's a new parameter.

BigLep:
Is there a way to have a test to demonstrate correctness?

marten-seemann:
> Does it make sense to dial WSS in the first place? We might want to consider disabling WSS dialing altogether (except for tests), as any node offering WSS will also offer TCP and that's always preferable.

Thinking about it, this logic should probably live in the swarm.

> Is there a way to have a test to demonstrate correctness?

@BigLep, what do you mean? We have unit tests for dialing.

BigLep:
> > Is there a way to have a test to demonstrate correctness?
> 
> @BigLep, what do you mean? We have unit tests for dialing.

We have a source code change without a corresponding test.  I'm assuming we didn't have a failing test in the past that this code is now fixing.  If that is true, then is there a test we should add to show this works as intended?

marten-seemann:
Yes, I think we should:
* stop resolving wss addresses in the host
* pass the resolver to the WebSocket transport
* not dial wss addresses if we have a TCP address from the same peer

aschmahmann:
> not dial wss addresses if we have a TCP address from the same peer

Should we prefer WS over WSS as well, or the other way around?

@marten-seemann are you able to take a look at this, or is this waiting on me/someone else to push things along?

Areas where I feel most in need of help here:
- A commit that uses the custom DNS resolver with the WSS library that'd be great. I can probably figure out the dependency injection, but you likely have more experience than me messing around with the TLS configs and/or gorilla
- How we want this tested before we're happy merging

marten-seemann:
> It seems like generally either what we have here is sufficient or we shouldn't be using ranking independent of this PR.

If I understand correctly, this ranking function isn't particularly useful, since we dial 8 addresses at the same time. My point is that we shouldn't dial WebSocket _at all_ (if we have a TCP address), as there's no benefit in first establishing a TCP connection, the layering WebSocket on top of that, and then speaking libp2p. The only reason we do this is to enable browsers to connect to us. The same will apply for WebTransport and QUIC addresses.

> Should we prefer WS over WSS as well, or the other way around?

We should prefer WS, WSS doesn't buy us anything, unless we encode the libp2p peer ID into the certificate (like we do for the TLS handshake). Since we don't expect to use WSS in any other environment than browsers (that don't provide an API to access the certificate), there's not really a point in doing so.

> How we want this tested before we're happy merging

We should have unit tests in the websocket package that show that we can dial DNS multiaddrs (we might need to mock the DNS resolver for that). Given how close we are to basic Testground interop tests, I think we can defer integration tests.

MarcoPolo:
I know this isn't the ideal solution, but it does fix dialing WSS with dns multiaddrs. What are the downsides of merging this and doing the proper solution later?

aschmahmann:
> But what are the downsides of merging this and doing the proper solution later?

The only downside I can think of at the moment (aside from wasted dials, but IMO that's not as bad as failing to dial at all) is that there's potentially a security leak if someone doesn't want go-libp2p to be using the OS resolver since this will use the OS resolver rather than the one given to the libp2p host.

marten-seemann (approved):
Let's merge this for now, with the understanding that we'll work on a proper fix (#1597) soon. Unfortunately, we probably won't be able to tackle that issue as long as we're still working on rcmgr stuff, but I've put it on the release checklist for v0.21.0 (#1514).

PR #1600, opened 2022-06-15
logging: Add canonical log for misbehaving peers

MarcoPolo:
Adds logging when a peer fails our security handshake or fails to setup a muxer

### DIFF
--- a/p2p/net/upgrader/upgrader.go
+++ b/p2p/net/upgrader/upgrader.go
@@ -7,7 +7,6 @@ import (
 	"net"
 	"time"
 
-	"github.com/libp2p/go-libp2p-core/canonicallog"
 	"github.com/libp2p/go-libp2p/p2p/net/pnet"
 
 	"github.com/libp2p/go-libp2p-core/connmgr"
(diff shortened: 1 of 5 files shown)


marten-seemann (commented):
Can we somehow filter out multistream errors here? Setup can fail if there's no compatible security protocol / muxer between those two peers. It seems odd to block a peer for that, doesn't it?

MarcoPolo:
> Setup can fail if there's no compatible security protocol / muxer between those two peers. It seems odd to block a peer for that, doesn't it?

If a peer doesn't have compatible security or muxer, will it ever successfully connect? We close the connection here right after a failure.

The idea with tools like fail2ban is that they won't ban a node for misbehaving once, but many times. For example, if I mistyped my password once, I wouldn't want to ban myself for N minutes. But if a node sends 100 wrong password attempts it is probably malicious.

Likewise here. If a node is constantly hitting this it's probably doing something bad. Right? Or am I missing something?

MarcoPolo:
@marten-seemann updated

p2p/net/upgrader/upgrader.go
```diff
@@ -174,6 +175,7 @@ func (u *upgrader) upgrade(ctx context.Context, t transport.Transport, maconn ma
 	smconn, err := u.setupMuxer(ctx, sconn, server, connScope.PeerScope())
 	if err != nil {
 		sconn.Close()
+		canonicallog.LogMisbehavingPeerNetAddr(p, conn.RemoteAddr(), "muxer", err, "failed to setup muxer")
```
marten-seemann, line 178: Should we remove this one, for the same reason we moved the log to the security transport? Setting up the muxer is not really expensive, so there's no real attack vector here anyway.

marten-seemann (approved):

Issue #1535, opened 2018-06-12
quic: require address validation when under load


marten-seemann:

marten-seemann:
Adding a bit of context here:
* Address validation is done using the [`AcceptToken` callback](https://pkg.go.dev/github.com/lucas-clemente/quic-go#Config). It will be useful to look at the [`defaultAcceptToken` implementation](https://github.com/lucas-clemente/quic-go/blob/6fbc6d951a4005d7d9d086118e1572b9e8ff9851/server.go#L244-L262) in quic-go.
* We can figure out the number of QUIC handshake started by counting how many times `AcceptToken` was called with 1. no token or 2. a token that's not a Retry token
* We can't count the number of failed handshakes (quic-go will swallow them), but we can count the number of successful (incoming) connections
* There might be a _slight_ overcounting if the clients send multiple Initial packets, or if an Initial packet is duplicated. I don't think we need to be worried about that.
* We should require address validation if the number of failed handshakes in the last `x` minutes is > 50% (configurable). When implementing this logic, we need to be careful to handle the cold-boot case (a node just booted up, is receiving a bunch of incoming connections: now we have a high incoming connection count, but because the QUIC handshake takes 1 RTT, none of them have completed yet).

ErikPelli:
> Adding a bit of context here:
> * Opening a QUIC packet and starting a TLS handshake is expensive. One attack on a QUIC stack is to flood the server with ClientHello packets (potentially with spoofed UDP source addresses) to generate a lot of CPU load. To defend against this attack, QUIC introduced the [Retry mechanism](https://datatracker.ietf.org/doc/html/rfc9000#section-8.1.2).
> * Address validation is done using the [`AcceptToken` callback](https://pkg.go.dev/github.com/lucas-clemente/quic-go#Config). It will be useful to look at the [`defaultAcceptToken` implementation](https://github.com/lucas-clemente/quic-go/blob/6fbc6d951a4005d7d9d086118e1572b9e8ff9851/server.go#L244-L262) in quic-go.
> * In go-libp2p, this is wired up in https://github.com/libp2p/go-libp2p/blob/884028550c9b4d1d0b1090aaf18e087bff3fa8bf/p2p/transport/quic/transport.go#L46-L49.
> * We can figure out the number of QUIC handshake started by counting how many times `AcceptToken` was called with 1. no token or 2. a token that's not a Retry token
[...]

PR #1602, opened 2022-06-17
quic: require address validation when under load

fulviodenza:
This PR adds the address validation when under load, referred to the issue #1535


ErikPelli:
I recommend reviewers to squash various commits

p2p/transport/quic/transport.go
```diff
@@ -38,17 +39,116 @@ var quicDialContext = quic.DialContext // so we can mock it in tests
...
+	mtx              sync.Mutex
+	duration         time.Duration
+	fraction         float64
+	start            sync.Once
+	startTime        time.Time // Time of the first execution
+	lastCycle        int64     // last period id (incremental)
+	retryProbability float64   // probability for the Retry packet in the actual time window
+}
+
+func (a *accept) setOptions(duration time.Duration, fraction float64) *accept {
```
marten-seemann, line 60: Probably nicer to have a constructor instead: `newAccept(duration time.Duration, fraction float64)`.

p2p/transport/quic/transport.go
```diff
@@ -38,17 +39,116 @@ var quicDialContext = quic.DialContext // so we can mock it in tests
...
+		fraction = 0.5
+	}
+
+	a.duration = duration
+	a.fraction = fraction
+	return a
+}
+
+func (a *accept) acceptFunction(clientAddr net.Addr, token *quic.Token) bool {
+	a.start.Do(func() { a.startTime = time.Now() })
```
marten-seemann, line 75: This could also be done in the constructor, then you don't need a `sync.Once`.

p2p/transport/quic/transport.go
```diff
@@ -166,19 +299,10 @@ func NewTransport(key ic.PrivKey, psk pnet.PSK, gater connmgr.ConnectionGater, r
...
-		connManager:  connManager,
-		gater:        gater,
-		rcmgr:        rcmgr,
-		conns:        make(map[quic.Connection]*conn),
-		holePunching: make(map[holePunchKey]*activeHolePunch),
-	}
 	config.AllowConnectionWindowIncrease = tr.allowWindowIncrease
 	tr.serverConfig = config
 	tr.clientConfig = config.Clone()
+	tr.clientConfig.AcceptToken = new(accept).setOptions(tr.period, tr.fraction).acceptFunction
```
marten-seemann, line 305: This needs to be set on the server config, not on the client config ;)

MarcoPolo (commented):
I've not reviewed the code yet, but, at minimum, this needs tests. Thank you!

marten-seemann (commented):
Thank you for submitting this PR!

I have 2 high-level comments:
1. It's probably easier if you use a slice of `bufferEntry`s. That slice could then start at length 1, and we'd append every minute, until we reach the maximum length (5). As we append to that slice once per minute, so there's no need to worried about the cost of allocations.
2. The logic is not _quite_ right. We need to count the number of successful handshakes, not successful retries. The attacker could start handshakes and complete the retries, without any of the expensive crypto of a TLS handshake. We know a handshake was successful when we accept the connection from `listener.Accept`. The fraction we're interested in is then `<number of successful handshakes> / <number of Retries received with token == nil || token.IsRetryToken == false>`.

p2p/transport/quic/transport.go
ErikPelli: I did it also on the server but here there was a call to Clone(), i don't know how it's implemented but I assumed that it uses the same AccessToken instance, thus causing a conflict. Isn't that the behaviour?

p2p/transport/quic/transport.go
ErikPelli: In the first implementation I did something like that, but I don't know the time difference between the object creation and the first invocation, if the difference is negligible I can modify

p2p/transport/quic/transport.go
marten-seemann: Oh, I missed that earlier part. It's probably better to clone a server config and a client config in L288, and then only set the callback on the server config.

ErikPelli:
> It's probably easier if you use a slice of bufferEntrys. That slice could then start at length 1, and we'd append every minute, until we reach the maximum length (5). As we append to that slice once per minute, so there's no need to worried about the cost of allocations.
When you finish the minutes what do you do? Overwrite starting from the first value? Anyway i don't think that append is necessary, as you can simply do a make([]bufferEntry, minutes) in the constructor, because the allocated space is very small

ErikPelli:
>  We know a handshake was successful when we accept the connection from listener.Accept. The fraction we're interested in is then <number of successful handshakes> / <number of Retries received with token == nil || token.IsRetryToken == false>.

How can you can get the number of handshakes? from the length of conns map in transport?

marten-seemann:
>  > It's probably easier if you use a slice of bufferEntrys. That slice could then start at length 1, and we'd append every minute, until we reach the maximum length (5). As we append to that slice once per minute, so there's no need to worried about the cost of allocations.
>
> When you finish the minutes what do you do? Overwrite starting from the first value? Anyway i don't think that append is necessary, as you can simply do a make([]bufferEntry, minutes) in the constructor, because the allocated space is very small

Yes exactly, you add a new `bufferEntry` at the end. And if you reach a length of 5, you drop the first entry. This way you avoid the index calculations that you need when using a ring buffer, and you avoid special cases when you don't have 5 entries yet because you just booted the node.

ErikPelli:
5 is fixed or is the number of minutes chosen by the user?

marten-seemann:
I think it can be a constant, the user chooses the duration (and that duration could, in principle, even be less than a minute).

ErikPelli:
In the fraction, what represents token.IsRetryToken == false? From the name I think that it's a boolean that says if the user sent the correct retry token, but in that case go-quic doesn't even call the function I think.

marten-seemann:
> In the fraction, what represents token.IsRetryToken == false? From the name I think that it's a boolean that says if the user sent the correct retry token, but in that case go-quic doesn't even call the function I think.

There are two types of tokens, both used in the same header field of the Initial packet. Retry tokens, as the name says, are tokens that are echoed from a Retry packet. There's also tokens generated on previous connections (see https://datatracker.ietf.org/doc/html/rfc9000#section-8.1.3). You don't need to worry about them too much.

fulviodenza:
Hi @marten-seemann, we finished the implementation and were working on testing the new feature.
We are a bit confused about the testing part, mainly on these points:

1. Should we mock the two peers? We found `createPeer` function to instantiate the peers that need to connect with each other but we're not sure about the correctness of this approach. 
2. Should we create a mock for some component with the clock like in this function? https://github.com/libp2p/go-libp2p/blob/bedce2290de15f7f6150ea1740a59d07b78a34f4/p2p/discovery/mocks/mocks.go#L28-L33 If yes what component should we add the clock to?
3. How do we interrupt handshakes after the accept function and before the conclusion?

p2p/transport/quic/transport.go
```diff
@@ -38,17 +38,118 @@ var quicDialContext = quic.DialContext // so we can mock it in tests
...
+func (a *accept) getPosition() (bool, int) {
+	currentCycle := int(time.Since(a.startTime) / a.duration)
+	if currentCycle > a.current.cycle {
+		if len(a.buffer) < cap(a.buffer) {
+			a.buffer = append(a.buffer, bufferEntry{})
+			a.current = status{
+				cycle: currentCycle,
+				pos:   len(a.buffer) - 1,
+				check: false,
+			}
```
marten-seemann, line 88: You need to make sure the slice doesn't grow unboundedly:
```go
if len(a.current) > bufferCap {
     a.current = a.current[len(a.current)-bufferCap:]
}
```

p2p/transport/quic/transport.go
```diff
@@ -38,17 +38,118 @@ var quicDialContext = quic.DialContext // so we can mock it in tests
...
+	a.buffer = make([]bufferEntry, 1, bufferCap)
+	return a
+}
+
+// Call it from inside a mutex to avoid a data race
+func (a *accept) getPosition() (bool, int) {
+	currentCycle := int(time.Since(a.startTime) / a.duration)
+	if currentCycle > a.current.cycle {
+		if len(a.buffer) < cap(a.buffer) {
+			a.buffer = append(a.buffer, bufferEntry{})
```
marten-seemann, line 83: There's an annoying edge case here: There might be minutes during which we don't receive _any_ handshakes (imagine a node that's only receiving new connections very infrequently). We need to clean up old entries in that case.

marten-seemann (commented):
Sorry, I dropped the ball on reviewing this. I just did another round.

p2p/transport/quic/transport.go
```diff
@@ -38,17 +38,118 @@ var quicDialContext = quic.DialContext // so we can mock it in tests
...
+					total:      sumAll.total + a.buffer[i].total,
+				}
+			}
+
+			// Check Retry fail ratio
+			if sumAll.total > 0 && float64(sumAll.handshakes)/float64(sumAll.total) < a.fraction {
+				success = false
+			}
+		}
+		if success {
```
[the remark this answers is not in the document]
ErikPelli: That's to prevent double counting on the total for an incoming connection. If the fraction exceeded the fail ratio, you then receive a token != nil.
With `if success` in token== nil case, we avoid to count when we expect to receive another packet with the token verify: without this every new connection will count in the total as 2, therefore risking not to reach the right ratio to return to normality.

For all the others proposed changes, we implemented the solution.

marten-seemann:
> 1. Should we mock the two peers? We found `createPeer` function to instantiate the peers that need to connect with each other but we're not sure about the correctness of this approach.

Yes, that sounds reasonable.

> 2. Should we create a mock for some component with the clock like in this function? https://github.com/libp2p/go-libp2p/blob/bedce2290de15f7f6150ea1740a59d07b78a34f4/p2p/discovery/mocks/mocks.go#L28-L33

Yes. Having a `WithClock` `Option` is probably the best way to go.

> 3. How do we interrupt handshakes after the accept function and before the conclusion?

Probably the easiest way will be to give the client an invalid certificate, like we do in the TLS tests: https://github.com/libp2p/go-libp2p/blob/master/p2p/security/tls/transport_test.go#L281. That will then lead to a failure of the QUIC handshake.

BigLep:
2022-08-26 conversation: @marten-seemann has done some work to cleanup the quic-go API.  @marten-seemann needs to release this and then will circle back here to explain the changes so the PR can be updated.

BigLep:
Status: blocked until the quic-go issue is addressed.

PR #1604, opened 2022-06-17
fix: return the best _acceptable_ conn in NewStream

### LINKED ISSUE #1603: Connect should wait for an acceptable connection

Stebalien:
Otherwise, we can return, e.g., a transient connection that can't actually be used.

I noticed bitswap was trying to use transient connections, which shouldn't be possible. To really fix this issue, we'd also need to fix #1603 as well, but that's harder.

### DIFF
--- a/p2p/net/swarm/swarm.go
+++ b/p2p/net/swarm/swarm.go
@@ -350,7 +350,7 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
 	dials := 0
 	for {
 		// will prefer direct connections over relayed connections for opening streams
-		c := s.bestConnToPeer(p)
+		c := s.bestAcceptableConnToPeer(ctx, p)
 		if c == nil {
 			if nodial, _ := network.GetNoDial(ctx); nodial {
 				return nil, network.ErrNoConn
--- a/p2p/net/swarm/swarm.go
+++ b/p2p/net/swarm/swarm.go
@@ -350,7 +350,11 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
 	dials := 0
 	for {
 		// will prefer direct connections over relayed connections for opening streams
-		c := s.bestAcceptableConnToPeer(ctx, p)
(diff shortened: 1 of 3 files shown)


marten-seemann:
Not sure I understand the swarm code correctly. Let's assume we only have a transient conn to a peer. We will therefore call `dialPeer`: https://github.com/libp2p/go-libp2p/blob/884028550c9b4d1d0b1090aaf18e087bff3fa8bf/p2p/net/swarm/swarm.go#L354-L369

`dialPeer` will then start a new dial sync (there are currently no active dials), as we only have a transient conn: https://github.com/libp2p/go-libp2p/blob/884028550c9b4d1d0b1090aaf18e087bff3fa8bf/p2p/net/swarm/swarm_dial.go#L251-L264

Does that mean we'll end up with two transient connections now?

Stebalien:
Ah.... ok, this is all broken.

Stebalien:
Ok, it's "fixed". Horribly broken, but at least it's correct.

p2p/net/swarm/swarm.go
```diff
@@ -447,15 +451,26 @@ func (s *Swarm) bestConnToPeer(p peer.ID) *Conn {
...
-			return conn
-		}
+	if conn == nil {
+		return nil, nil
 	}
-	return nil
+
+	forceDirect, _ := network.GetForceDirectDial(ctx)
+	if forceDirect && !isDirectConn(conn) {
+		return nil, nil
```
MarcoPolo, line 465: What happens here? We may have a transient connection but we'll try to dial again?

MarcoPolo (commented):
It's a bit confusing that we have two different concepts of "transient" connections in go-libp2p. Transient as in a limited connection over a relay and transient as in a connection that hasn't been upgraded yet and doesn't have a peerid associated with it yet.

p2p/net/swarm/swarm.go
Stebalien: Yes. We need to do that for hole-punching.

MarcoPolo (approved):

marten-seemann (approved):

PR #1607, opened 2022-06-21
feat: inject DNS resolver

Stebalien:
This injects the DNS resolver into transport constructors.

### DIFF
--- a/config/constructor_types.go
+++ b/config/constructor_types.go
@@ -13,6 +13,8 @@ import (
 	"github.com/libp2p/go-libp2p/core/pnet"
 	"github.com/libp2p/go-libp2p/core/sec"
 	"github.com/libp2p/go-libp2p/core/transport"
+
+	madns "github.com/multiformats/go-multiaddr-dns"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


config/constructor_types.go
```diff
@@ -28,43 +29,47 @@ var (
 	connGaterType = reflect.TypeOf((*connmgr.ConnectionGater)(nil)).Elem()
 	upgraderType  = reflect.TypeOf((*transport.Upgrader)(nil)).Elem()
 	rcmgrType     = reflect.TypeOf((*network.ResourceManager)(nil)).Elem()
+	resolverType  = reflect.TypeOf((*madns.Resolver)(nil)).Elem()
```
marten-seemann, line 32: Would be nice to use an interface instead of a concrete type here. `madns.BasicResolver`?

Could it be that `madns.Resolver` and `madns.BasicResolver` should switch their names?

Stebalien: So, `madns.Resolver` is basically a light-weight wrapper over a set of `BasicResolver`s that implements a multiaddr handing [`Resolve`](https://github.com/multiformats/go-multiaddr-dns/blob/0bad63ddb983cf7824f45f1490f81f3b3b66fe62/resolve.go#L108) method. The pluggable component is really `madns.BasicResolver`, where the user should then construct an `madns.Resolver` from one or more `madns.BasicResolver`s.

BigLep:
2022-08-19: this is needed for 0.23 to fix address resolution issues.  There is an open issue we can link this to.

marten-seemann (approved):
LGTM.

marten-seemann:
Rebased to resolve merge conflict in imports.

PR #1609, opened 2022-06-22
muxer: expose func to create MuxedConn from backing Conn

paralin:
Expose a function to create a MuxedConn from the base type: if a user has a *mp.Multiplex or a *yamux.Session, they should be able to construct the libp2p wrapper without going through the Transport interfaces.


MarcoPolo (approved):
Looks good, thank you!

PR #1624, opened 2022-06-30
Add canonical peer status logs

MarcoPolo:
This PR logs 1% of the following events:

1. A connection is established.
2. A handshake fails.

This is useful for a couple reasons:
1. High churn from a misbehaving peer becomes obvious. If a peer churns 100 connections / second. Then it'll log roughly once a second. An operator can either manually ban this peer or hook the logs up to fail2ban so that the system automatically blocks misbehaving peers. See https://github.com/libp2p/go-libp2p-core/pull/269 for an example.
2. Handshake failures in aggregate also become obvious.
3. Keep a pulse on the system. e.g. No new connections in hours seem suspicious.
4. At a glance you can also see the distribution of quic vs the rest of the transports. This would be better as a metric of course, but still interesting.

The tests will fail since this updates go-libp2p-core which breaks go-libp2p-circuit, which would work if we updated go-libp2p-circuit's dependency on go-libp2p.

Not sure how to resolve. One idea is to depend on a branch of go-libp2p-circuit which depends on go-libp2p-master. Any other suggestions?

### DIFF
--- a/p2p/net/upgrader/upgrader.go
+++ b/p2p/net/upgrader/upgrader.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/libp2p/go-libp2p/p2p/net/pnet"
 
-	"github.com/libp2p/go-libp2p-core/canonicallog"
 	"github.com/libp2p/go-libp2p-core/connmgr"
 	"github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/peer"
@@ -177,8 +176,6 @@ func (u *upgrader) upgrade(ctx context.Context, t transport.Transport, maconn ma
 		sconn.Close()
 		return nil, fmt.Errorf("failed to negotiate stream multiplexer: %s", err)
 	}
-
-	canonicallog.LogPeerStatus(100, sconn.RemotePeer(), maconn.RemoteMultiaddr(), "connection_status", "established")
 
 	tc := &transportConn{
 		MuxedConn:      smconn,
--- a/p2p/net/upgrader/upgrader.go
+++ b/p2p/net/upgrader/upgrader.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/libp2p/go-libp2p/p2p/net/pnet"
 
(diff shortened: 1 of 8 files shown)


p2p/net/upgrader/upgrader.go
```diff
@@ -177,6 +178,8 @@ func (u *upgrader) upgrade(ctx context.Context, t transport.Transport, maconn ma
 		return nil, fmt.Errorf("failed to negotiate stream multiplexer: %s", err)
 	}
 
+	canonicallog.LogPeerStatus(100, sconn.RemotePeer(), maconn.RemoteMultiaddr(), "connection_status", "established")
```
marten-seemann, line 181: Should this happen in the swarm, so we also catch QUIC connections?

marten-seemann (commented):
We can escape dependency hell by rebasing after https://github.com/libp2p/go-libp2p/pull/1626 was merged.

p2p/net/upgrader/upgrader.go
MarcoPolo: I added a log to quic transport directly. Would you prefer I do this in the swarm?

marten-seemann: Yes, that way we'd make sure that every connection is logged, even if it doesn't use the upgrader. We'll thank ourselves for that when we add WebTransport.

You can do it in a follow-up PR, if you prefer.

p2p/transport/quic/transport.go
```diff
@@ -255,6 +256,7 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 		return nil, fmt.Errorf("secured connection gated")
 	}
 	t.addConn(qconn, c)
+	canonicallog.LogPeerStatus(100, p, remoteMultiaddr, "transport", "quic", "connection_status", "established")
```
marten-seemann, line 259: I assume this is a leftover?

marten-seemann (approved):
LGTM, modulo the leftover log in QUIC.

marten-seemann (approved):

PR #1625, opened 2022-07-01
ci: run testground:ping plan on pull requests

laurentsenta:
Add a workflow that calls the testground interop testing on every push and pull request.

## Examples

- master: https://github.com/laurentsenta/go-libp2p/runs/7148996579?check_suite_focus=true
- pull request: https://github.com/laurentsenta/go-libp2p/pull/3
  - run: https://github.com/laurentsenta/go-libp2p/runs/7149039457?check_suite_focus=true
  - note the `GitReference=d9f4b4447188fafe7c15b66b70d98b0ba6e0eeb6 GitTarget=github.com/laurentsenta/go-libp2p` parameters in "Run the composition file" job.
- pull request that breaks the ping service: https://github.com/laurentsenta/go-libp2p/pull/4
  - run: https://github.com/laurentsenta/go-libp2p/runs/7149252149?check_suite_focus=true
  - note the `protocol not supported` errors

## Known limitation

- This PR will work with local branches & forks, but it will test the latest commit of the pull request, NOT the merge commit,
- The test takes ~14 minutes for now (5 min setup, 9 min to run),

### DIFF
--- a/.github/workflows/interop-test.yml
+++ b/.github/workflows/interop-test.yml
@@ -1,0 +1,16 @@
+on: [push, pull_request]
+name: Interoperability Testing
+
+jobs:
+  run-ping-interop-cross-version:
+    uses: "libp2p/test-plans/.github/workflows/run-composition.yml@master"
+    with:
+      composition_file: "ping/_compositions/go-cross-versions.toml"
+      # NOTE: during a pull request run, github creates a merge commit referenced in `github.sha`
+      # that merge commit is not a regular commit. You won't find it with a regular `git checkout SHA` and
+      # tools like `go get repo@SHA` won't find it.
+      #
+      # As a workaround, we generate a path to the actual pull request's commit, it looks like:
+      # `github.com/external-org/go-libp2p@latest-commit-on-their-branch`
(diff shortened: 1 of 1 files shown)


mxinden:
Very cool!

>  but it will test the latest commit of the pull request, NOT the merge commit,

I can only comment for rust-libp2p. Here `master` is a protected branch and we enabled "Require branches to be up to date before merging", thus it is fine to test the commit and not the merge.

laurentsenta:
@marten-seemann I have noticed no flakiness with the current setup (a few dozen runs, 0 failures).

The one thing I'm not sure about is how much maintenance the `master` and the `current` test groups will require:

They both auto-update libp2p using a `replace` directive in the modfile. Which doesn't looks super stable with go and libp2p. 
I already had to patch the test plan (https://github.com/libp2p/test-plans/pull/25) and I hope that will be enough. From my understanding, libp2p's repo consolidation will help too.

I think it's worth a shot, especially because if we make this work, we might be able to make this work everywhere (ipfs, lotus, etc).

I'll fix the tests along if needed, and if the maintenance cost is too high we have options (automation, running the test only on releases, moving the plan into the go-libp2p repo, etc).

marten-seemann (approved):

galargh:
🚀

PR #1626, opened 2022-07-01
move go-libp2p-circuit here

marten-seemann:
go-libp2p-circuit has been deprecated for a long time, but we're still using it in our circuit v2 compatibility tests.

With the repo consolidation in v0.20 go-libp2p-circuit would now depend on go-libp2p, creating a circular dependency between the two repo (NOT between the packages, so the Go compiler is fine with it).

Assuming that we want to keep the compatibility test, merging -circuit here seems like the least bad thing to do. We're moving it to an `internal` directory, so that no user can import it.


MarcoPolo (approved):

vyzo (approved):
ugh.... but what do? so fine.

PR #1627, opened 2022-07-01
fix: deny connections to peers in the right place

Stebalien:
The previous location would prevent explicit calls to `Connect`, but not implicit connections on `NewStream`. Given that the host calls `Connect` before calling `NewStream`, it's unclear whether or not this would likely be triggered by user code, but we still need to fix it.


marten-seemann (approved):
Thank you!

PR #1634, opened 2022-07-01
chore: update deps

marten-seemann:
These are dep upgrades that need to make it into the next release (v0.21.0).


MarcoPolo (approved):

MarcoPolo:
I rebased this off master

MarcoPolo:
I think this is going to give us some problems because we updated indirect deps. I'm thinking now we should never update indirect deps. We should require the minimum acceptable version of indirect deps.

PR #1636, opened 2022-07-02
docs: point to SetDefaultServiceLimits in ResourceManager option

marten-seemann:


options.go
```diff
@@ -342,6 +342,8 @@ func ConnectionGater(cg connmgr.ConnectionGater) Option {
 }
 
 // ResourceManager configures libp2p to use the given ResourceManager.
+// When using the go-libp2p-resource-manager implementation of the ResourceManager interface,
+// it is recommended to set limits for libp2p protocol by calling SetDefaultServiceLimits.
```
MarcoPolo, line 346: (optional) What do you think about adding an example that gets rendered in the godoc here too?

https://pkg.go.dev/github.com/fluhus/godoc-tricks#Examples

MarcoPolo (approved):

options.go
marten-seemann: That's a good idea, but we should do that after we've updated the resource manager to include the config refactor.

PR #1637, opened 2022-07-02
rcmgr: use autoscaling limits

marten-seemann:
Depends on https://github.com/libp2p/go-libp2p-resource-manager/pull/48.

### DIFF
--- a/limits.go
+++ b/limits.go
@@ -15,143 +15,114 @@ import (
 )
 
 // SetDefaultServiceLimits sets the default limits for bundled libp2p services
-//
-// More specifically this sets the following limits:
-// - identify:
-//   128 streams in, 128 streams out, 256 streams total, 4MB min, 64MB max svc memory
-//   16/16/32 streams per peer
-// - ping:
-//   128 streams in, 128 sreams out, 256 streasms total, 4MB min, 64MB max svc memory
-//   2/3/4 streams per peer
-// - autonat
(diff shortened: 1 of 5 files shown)


limits.go
```diff
@@ -15,143 +15,114 @@ import (
...
 	// holepunch
-	setServiceLimits(limiter, holepunch.ServiceName,
-		limiter.DefaultServiceLimits.
-			WithMemoryLimit(1, 4<<20, 64<<20). // max 64MB service memory
-			WithStreamLimit(128, 128, 256),    // max 256 streams - symmetric
-		peerLimit(2, 2, 2))
-	setProtocolLimits(limiter, holepunch.Protocol,
-		limiter.DefaultProtocolLimits.WithMemoryLimit(1, 4<<20, 64<<20),
-		peerLimit(2, 2, 2))
+	addServiceAndProtocolLimit(config,
```
MarcoPolo, line 70: I know this was what the old code did, but why do we need to set both the service and protocol limit? Shouldn’t the service limit cover the protocol? And/or shouldn’t the protocol limit cover the service?

MarcoPolo (approved):
looks one question that’s not a blocker because it matches the old behavior

MarcoPolo:
Don’t forget to update the resource manager and this PR’s go.mod before merge

limits.go
marten-seemann: Yes, for services that use a single protocol. Identify uses multiple protocol IDs, so at least in theory, it could make sense to set separate limits.

PR #1639, opened 2022-07-05
ci: demonstration for test running in parallel

laurentsenta:
This PR illustrates how we could run go-tests in parallel with the unified ci.

This might make CI slower though: instead of waiting for 6 runners, we'll wait for 12.

Example run:
https://github.com/laurentsenta/go-libp2p/actions/runs/2617209338

### DIFF
--- a/.github/workflows/go-test.yml
+++ b/.github/workflows/go-test.yml
@@ -11,10 +11,18 @@ jobs:
       matrix:
         os: [ "ubuntu", "windows", "macos" ]
         go: [ "1.17.x", "1.18.x" ]
+        suite: [ "basic", "32-bit", "race-detector" ]
+        exclude:
+          - os: macos  # can't run 32 bit tests on OSX.
+            suite: 32-bit
+          - os: macos # speed things up. Windows and OSX VMs are slow
+            suite: race-detector
+          - os: windows # speed things up. Windows and OSX VMs are slow
+            suite: race-detector
     env:
       COVERAGES: ""
     runs-on: ${{ format('{0}-latest', matrix.os) }}
-    name: ${{ matrix.os }} (go ${{ matrix.go }})
+    name: ${{ matrix.suite }} tests - ${{ matrix.os }} (go ${{ matrix.go }})
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/go-test.yml
```diff
@@ -39,13 +47,14 @@ jobs:
...
         uses: protocol/multiple-go-modules@v1.2
+        if: ${{ matrix.suite == 'basic' }}
         with:
           # Use -coverpkg=./..., so that we include cross-package coverage.
           # If package ./A imports ./B, and ./A's tests also cover ./B,
           # this means ./B's coverage will be significantly higher than 0%.
           run: go test -v -coverprofile=module-coverage.txt -coverpkg=./... ./...
       - name: Run tests (32 bit)
-        if: ${{ matrix.os != 'macos' }} # can't run 32 bit tests on OSX.
+        if: ${{ matrix.suite == '32-bit' }} # can't run 32 bit tests on OSX.
```
marten-seemann, line 57:
```suggestion
        if: ${{ matrix.suite == '32-bit' }}
```

.github/workflows/go-test.yml
```diff
@@ -54,14 +63,16 @@ jobs:
             export "PATH=${{ env.PATH_386 }}:$PATH"
             go test -v ./...
       - name: Run tests with race detector
-        if: ${{ matrix.os == 'ubuntu' }} # speed things up. Windows and OSX VMs are slow
+        if: ${{ matrix.suite == 'race-detector' }} # speed things up. Windows and OSX VMs are slow
```
marten-seemann, line 66:
```suggestion
        if: ${{ matrix.suite == 'race-detector' }}
```

marten-seemann (commented):
Do we have a way to measure the execution of a CI run (not only of a single job)? I think we'd benefit from some hard numbers to base our decisions on.

Problem here: Jobs might be waiting in the GitHub Actions queue (because we're limited to 20 (?) runners), which they wouldn't when using self-hosted runners, assuming that we dynamically scale up the number of runners.

Another thing we could consider doing is caching the Go dependencies. I'm not sure how much we'd gain from this, now that we've drastically reduced the libp2p repo sprawl, but this would probably be worth measuring as well.

laurentsenta:
Closing, I'll revisit when we tackle https://github.com/pl-strflt/ipdx/issues/27

PR #1641, opened 2022-07-05
chore: update core dep

MarcoPolo:
* Upgrades rcmgr and core by a patch release
* Dowgrades grpc indirect dep to avoid breaking downstream users unnecessarily (see: https://github.com/weaveworks/common/issues/239)
* removed btcd dep

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -108,7 +108,7 @@ require (
 	golang.org/x/net v0.0.0-20220630215102-69896b714898 // indirect
 	golang.org/x/tools v0.1.11 // indirect
 	golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f // indirect
-	google.golang.org/grpc v1.47.0 // indirect
+	google.golang.org/grpc v1.36.0 // indirect
 	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/tomb.v1 v1.0.0-20141024135613-dd632973f1e7 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -110,7 +108,7 @@ require (
 	golang.org/x/net v0.0.0-20220630215102-69896b714898 // indirect
 	golang.org/x/tools v0.1.11 // indirect
 	golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f // indirect
-	google.golang.org/grpc v1.47.0 // indirect
+	google.golang.org/grpc v1.36.0 // indirect
```
marten-seemann, line 111: What made us update this dependency since the v0.20.0 release? In v0.20.0 we were using v1.31.1.

MarcoPolo: You updated this in this pr: https://github.com/libp2p/go-libp2p/pull/1634/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R113

I think I picked 1.36 by looking at the go mod graph.

MarcoPolo: Actually we can get rid of grpc dep completely. Need this first: https://github.com/libp2p/go-libp2p-testing/pull/63

go.mod
```diff
@@ -110,7 +108,6 @@ require (
 	golang.org/x/net v0.0.0-20220630215102-69896b714898 // indirect
 	golang.org/x/tools v0.1.11 // indirect
 	golang.org/x/xerrors v0.0.0-20220609144429-65e65417b02f // indirect
-	google.golang.org/grpc v1.47.0 // indirect
```
marten-seemann, line 113: ❤️

marten-seemann (approved):

Issue #1324, opened 2022-02-07
connmgr: flaky TestMultiplePeers


marten-seemann:
```
  === RUN   TestMultiplePeers
      decay_test.go:220: wrong value; expected = 2670; got = 2671
  --- FAIL: TestMultiplePeers (1.08s)

**Comment by schomatis** [comment]
Assigning myself.

**Comment by schomatis** [comment]
Looking into this. After changes from https://github.com/libp2p/go-libp2p/pull/1354 I can no longer reproduce the above but get this error instead:

```
=== RUN   TestMultiplePeers
    decay_test.go:151: 
          Error:        Condition never satisfied
          Test:         TestMultiplePeers
--- FAIL: TestMultiplePeers (0.65s)

```

So will start there with an eye on the `2670` check also.

schomatis:
@marten-seemann I think we can close this issue.

I haven't been able to reproduce the error and with a controlled number of elapsed decaying ticks (through the mock clock of https://github.com/libp2p/go-libp2p/pull/1354) this shouldn't happen again, and if it does it would no longer be a flake but an actual bug (this will be made even more explicit now with https://github.com/libp2p/go-libp2p/pull/1642).

PR #1642, opened 2022-07-07
chore(connmgr): clean TestMultiplePeers

schomatis:
In the context of looking into https://github.com/libp2p/go-libp2p/issues/1324, clean up `TestMultiplePeers`.

### DIFF
--- a/p2p/net/connmgr/decay_test.go
+++ b/p2p/net/connmgr/decay_test.go
@@ -1,6 +1,7 @@
 package connmgr
 
 import (
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


schomatis:
Leaving as a draft for now as I'm missing the target for `id[2]` of 40. It doesn't match the expected value of adding the bumps and subtracting the decays as it does for the other two peers. The decay is the same for each peer 330 (120, 150, 60 from the 3 tags combined), but the peers start with different initial bump values:

* ID0: 3000 - 330 = 2670
* ID1: 1500 - 330 = 1170
* ID2: 300 - 330 = -30. This is what's showing up on the test now but in the original code it was 40, not sure what's the rationale.

schomatis:
@raulk Was that the math you had in mind when you were writing the tests or am I missing something?

schomatis:
Seems I'm not correctly interpreting the logic of the negative case in this more linearized version of the test:

https://github.com/libp2p/go-libp2p/blob/66cd76df22e8f50717da0cfd1765f90d4d58f1e0/p2p/net/connmgr/decay.go#L191-L202

@raulk Would it be ok to keep this test with positive values (to simplify testing logic) and move the negative case to a separate test (with more elaborate logic accounting for the code above)?

marten-seemann:
Closing due to inactivity.

PR #1644, opened 2022-07-07
routedhost: fix decoding of relay peer ID

marten-seemann:
A circuit multiaddr contains an encoded multiaddr, so we need to decode it: https://github.com/libp2p/specs/tree/master/addressing#p2p-circuit-relay-addresses

Not 100% sure that this is correct, please review carefully.


Stebalien (approved):
This is correct but now I'm really confused as to why we didn't catch this earlier. The "value" returned from `ValueFromProtocol` should be a peer ID encoded as a string.

marten-seemann:
> This is correct but now I'm really confused as to why we didn't catch this earlier

No unit tests and very few integration tests for circuits...

PR #1652, opened 2022-07-09
examples: fix unresponsive pubsub chat example

goforbroke1006:
Update ui for example
examples/pubsub/chat/ui.go:125

Remove unneeded lock on peersList textview.
Clear method already has lock.

```golang
// Clear removes all text from the buffer.
func (t *TextView) Clear() *TextView {
	t.Lock()
	defer t.Unlock()

	t.clear()
	return t
}
```


marten-seemann (approved):
`Clear` indeed does lock a mutex, so I assume it's thread-safe.

PR #1653, opened 2022-07-10
swarm: fix selection of transport for dialing

marten-seemann:
With WebTransport's `/webtransport/certhash/xyz` addresses, the assumption that the last component of a multiaddr identifies the transport to use for dialing doesn't hold any more.

Note that WebRTC will probably also use the `certhash` multiaddr component to encode its certificate hashes.

This is a draft until we actually move the webtransport transport into this repo (#1737), but I'm asking for reviews nonetheless. Please review carefully, this has the potential to break us in painful ways.

### DIFF
--- a/p2p/net/swarm/swarm_addr_test.go
+++ b/p2p/net/swarm/swarm_addr_test.go
@@ -2,13 +2,25 @@ package swarm_test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
+	"github.com/libp2p/go-libp2p/core/peer"
+
+	ic "github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/core/test"
+	"github.com/libp2p/go-libp2p/p2p/net/swarm"
 	swarmt "github.com/libp2p/go-libp2p/p2p/net/swarm/testing"
+	circuitv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/client"
+	quic "github.com/libp2p/go-libp2p/p2p/transport/quic"
+	"github.com/libp2p/go-libp2p/p2p/transport/tcp"
+	webtransport "github.com/libp2p/go-libp2p/p2p/transport/webtransport"
 
+	"github.com/minio/sha256-simd"
 	ma "github.com/multiformats/go-multiaddr"
+	"github.com/multiformats/go-multibase"
+	"github.com/multiformats/go-multihash"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


p2p/net/swarm/swarm_addr_test.go
```diff
@@ -57,3 +67,34 @@ func TestAddressesWithoutListening(t *testing.T) {
 	require.NoError(t, err)
 	require.Empty(t, a1, "expected to be listening on no addresses")
 }
+
+func TestDialAddressSelection(t *testing.T) {
```
marten-seemann, line 71: I'm slightly concerned about the amount of dependencies that this test adds. Not that much of a problem, now that we have a mono-repo, but maybe this would be nicer if we used a mock `transport.Transport` interface?

p2p/net/swarm/swarm_transport.go
```diff
@@ -19,25 +19,23 @@ func (s *Swarm) TransportForDialing(a ma.Multiaddr) transport.Transport {
...
 	}
-
-	for _, p := range protocols {
-		transport, ok := s.transports.m[p.Code]
-		if !ok {
-			continue
-		}
-		if transport.Proxy() {
-			return transport
+	if isRelayAddr(a) {
```
marten-seemann, line 30: I don't really like this logic here. Instead, we could have transports reject `/p2p-circuit` address. wdyt?

marten-seemann:
Rebased the PR. This should be ready for review now.

p2p/net/swarm/swarm_transport.go
```diff
@@ -19,25 +19,23 @@ func (s *Swarm) TransportForDialing(a ma.Multiaddr) transport.Transport {
...
+		return s.transports.m[ma.P_CIRCUIT]
+	}
+	for _, t := range s.transports.m {
```
MarcoPolo, line 33: I think in general this needs to go the other way. We should be interpreting this right to left.

p2p/net/swarm/swarm_transport.go
```diff
@@ -19,25 +19,23 @@ func (s *Swarm) TransportForDialing(a ma.Multiaddr) transport.Transport {
...
+		if t.CanDial(a) {
+			return t
 		}
 	}
-
-	return s.transports.m[protocols[len(protocols)-1].Code]
```
MarcoPolo, line 40: What if we have a `TransportForComponent` function that takes the component and the rest of the multiaddr (and maybe the list of registered transports) and returns the transport to use. This fits with the encapsulation idea of multiaddrs.

This function could be a method of an interface and components could implement this interface. The certhash component would simply return the next transport in the multiaddr along with the certhash parameters.

MarcoPolo (changes requested):
I think this logic should change a bit after we discuss the multiaddr interpretation strategy. But at the very least I think we should be traversing the multiaddr backwards when looking for a transport. i.e. we don't want `quic` to match before `webtransport` in `.../quic/webtransport`.

p2p/net/swarm/swarm_transport.go
marten-seemann: That would be a layer violation, wouldn't it?
Components are a multiaddr concept. Multiaddr doesn't know anything about transports.

p2p/net/swarm/swarm_transport.go
marten-seemann: I don't understand this comment. We're iterating over a `map[int]transport.Transport` here. Inverting the order of iteration shouldn't have any effect, since map iterations are (pseudo-) randomized anyway.

MarcoPolo: I misunderstood. I thought we were iterating of the multiaddr components. I see now that we're iterating over all registered transports.

I think this isn't ideal because:
* we have to make sure every registered transport doesn't intersect with any other transport. It would be tricky behavior if two transports return `CanDial -> true` for the same `a` since we would randomly pick one of the two. I think simply a guaranteed ordering here would make it better.
* This means the transport owns the interpretation of the full multiaddr. This doesn't compose well. Imagine we have a `percentencode` component and a `lenprefixencode` component, we don't want every transport to have to know how to read the encoded data from the multiaddr. Instead they should be given the decoded data from something else and deal with their part of the multiaddr. This relates to https://github.com/multiformats/multiaddr/pull/140.


I'll draft a patch of what I'm thinking here and how it relates to https://github.com/multiformats/multiaddr/pull/140

MarcoPolo: Here's concretely what I mean: https://github.com/libp2p/go-libp2p/compare/fix-dial-transport-selection...marco/layered-transports?expand=1#diff-1d9e588fd7c4fedee40f7f985be2346b02ba540389251f4603cbc13e8f8fc371R94

Notice this gives us the benefit of removing the "extract certhash" logic out of webtransport completely both in the `Dial` and the `CanDial` method (I didn't actually do this refactor, since this is a demo).

MarcoPolo:
> I agree that this is a mess and should be changed, but can we focus on getting something landed to unblock WebTransport for v0.23.0?

Ok. Then maybe it's worth asking if anything here will make things worse than than they already are, or is there something that will make backwards compatibility harder? I don't think so. This has the same behavior as before if no two transports return `candial=true` for any given multiaddr (if this wasn't true, this would introduce a non-deterministic bug).

> I'm not really sure what a concrete proposal would look like.

I hope my patch clarifies this, but let me know if things are still unclear.

> we still have `/p2p-circuit` to deal with.

Yeah, this is the one multiaddr format that doesn't match up with how the rest work. We can still deal with it following the same principles but it would involve some changes (see https://github.com/multiformats/multiaddr/pull/140). Right now we deal with this with a special case. We could keep doing that too, but it would be better if we can unify the interpretation of multiaddrs.

p2p/net/swarm/swarm_transport.go
MarcoPolo: Let's not change this just yet since that would involve changing every transport. Let's revisit after discussing https://github.com/multiformats/multiaddr/pull/140

MarcoPolo (approved):
I don't think this changes the old behavior, and it unblocks webtransport. This code needs refactoring but I don't want to block webtransport. Let's revisit after discussing https://github.com/multiformats/multiaddr/pull/140

marten-seemann:
Merging this PR, with the understanding that this will get refactored very soon.

Issue #1421, opened 2022-04-22
flaky TestDialSimultaneousJoin


marten-seemann:
```
=== RUN   TestDialSimultaneousJoin
      dial_test.go:578: third dial succedded; conn: <swarm.Conn[*tcp.TcpTransport] /ip4/127.0.0.1/tcp/50014 (12D3KooWRuYVGEsecrJJhZsSoKf1UNdBVYKFCmFLNj9ucZiSQCYj) <-> /ip4/127.0.0.1/tcp/50015 (12D3KooWGEcD5sW5osB6LajkHGqiGc3W8eKfYwnJVVqfujkpLWX2)>
      dial_test.go:560: second dial succedded; conn: <swarm.Conn[*tcp.TcpTransport] /ip4/127.0.0.1/tcp/50014 (12D3KooWRuYVGEsecrJJhZsSoKf1UNdBVYKFCmFLNj9ucZiSQCYj) <-> /ip4/127.0.0.1/tcp/50015 (12D3KooWGEcD5sW5osB6LajkHGqiGc3W8eKfYwnJVVqfujkpLWX2)>
      dial_test.go:[588](https://github.com/libp2p/go-libp2p/runs/6129949662?check_suite_focus=true#step:7:588): 
          	Error Trace:	dial_test.go:588
          	Error:      	Received unexpected error:
          	            	failed to dial 12D3KooWGEcD5sW5osB6LajkHGqiGc3W8eKfYwnJVVqfujkpLWX2:
          	            	  * [/ip4/127.0.0.1/tcp/50016] failed to negotiate security protocol: context deadline exceeded
          	Test:       	TestDialSimultaneousJoin
  --- FAIL: TestDialSimultaneousJoin (0.26s)

vyzo:
I don't recall targeting a specific line, just making sure we have a test for joined dials.

schomatis:
@vyzo Ok, could you point me to 'joined dials' in the code to better understand what are we trying to test, please?

schomatis:
And particularly how are we enforcing (or approaching) the "simultaneous" part of the test.

vyzo:
It's the invariant that two concurrent dials to the same addresses are joined.

schomatis:
What I'm seeing here is the first dial timeouting before the second one has a chance to hit and I'm trying to figure out how to better guarantee that simultaneity.

vyzo:
[...]

PR #1659, opened 2022-07-13
Fix flaky TestDialSimultaneousJoin

schomatis:
Toward https://github.com/libp2p/go-libp2p/issues/1421.

@vyzo This is a blunt and incorrect first approximation for feedback.

The 1st and 3rd `DialPeer`s happen after the 2nd one succeeds. This passes but defeats the purpose of the test of leaving the 1st dial hanging while 2nd one connects. I need more guidance on how to do a better fine-grained simultaneity (without messing with dialing internals) to make sure the *1st dial hangs but does not timeout* while 2nd happens. (Messing with dialing internals is fine by me but I don't think is the point here.)

### DIFF
--- a/p2p/net/swarm/dial_test.go
+++ b/p2p/net/swarm/dial_test.go
@@ -2,6 +2,7 @@ package swarm_test
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


schomatis:
I'm trying to add a `WithCancel` context in the 1st dial to make it fail explicitly from the 2nd one after it succeeds, but the cancel error seems to go through a different flow from the timeout error and doesn't join the 2nd connection, just fails.

marten-seemann:
Haven't seen this flaking in a while, I assume that @sukunrt's smart dialing changes have resolved it.

PR #1660, opened 2022-07-23
Some deprecated function calls and bugs were fixed in the example.

MGMCN:
Hello, I am an M1 student in Japan. Recently, I was doing some research related to P2P, so I tried to use the library written by your organization, but I found some small problems when running the example. 

For example, some deprecated methods are used in the chapter “chat-with-rendezvous”. At the same time, there is a small bug in the chapter “chat-with-mdns” (well, in fact, I don't know whether this is a bug or not). When we start a peer in a terminal first, this peer will always try to establish a connection with itself first. So I modified this part of the code. I asked this peer not to try to connect to itself after discovering itself, but to continue to wait for a connection with a different ID from itself to be read from the peerchannel. 

looking forward to your reply.

### DIFF
--- a/examples/chat-with-mdns/main.go
+++ b/examples/chat-with-mdns/main.go
@@ -116,6 +116,9 @@ func main() {
 	peerChan := initMDNS(host, cfg.RendezvousString)
 
 	peer := <-peerChan // will block untill we discover a peer
+	for peer.ID == host.ID() {
+		peer = <-peerChan
+	}
 	fmt.Println("Found peer:", peer, ", connecting")
 
 	if err := host.Connect(ctx, peer); err != nil {
--- a/examples/chat-with-mdns/main.go
+++ b/examples/chat-with-mdns/main.go
@@ -116,9 +116,6 @@ func main() {
 	peerChan := initMDNS(host, cfg.RendezvousString)
 
 	peer := <-peerChan // will block untill we discover a peer
-	for peer.ID == host.ID() {
-		peer = <-peerChan
-	}
 	fmt.Println("Found peer:", peer, ", connecting")
 
 	if err := host.Connect(ctx, peer); err != nil {
--- a/examples/chat-with-rendezvous/chat.go
+++ b/examples/chat-with-rendezvous/chat.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/libp2p/go-libp2p/p2p/discovery/routing"
 	"os"
 	"sync"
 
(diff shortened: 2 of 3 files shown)


examples/chat-with-mdns/main.go
```diff
@@ -116,6 +116,9 @@ func main() {
 	peerChan := initMDNS(host, cfg.RendezvousString)
 
 	peer := <-peerChan // will block untill we discover a peer
+	for peer.ID == host.ID() {
```
marten-seemann, line 119: Why? That would mean that we only discover ourselves, or am I missing something here?

MGMCN: Hello, thank you for your reply.

My understanding is that we first take a connection from this peerchan. When the ID of this connection is the same as that of our current host, we will continue to get it from peerchan. Until we found another peer with a different ID. We jump out of the for loop.

In fact, similar codes have appeared in https://github.com/libp2p/go-libp2p/blob/master/examples/chat-with-rendezvous/chat.go （line161）you have written. See the first two lines of code in the for loop.

(ps: maybe you should run my modified code and unmodified code)

looking forward to your reply.
 
```go
go func() {
		for peer := range peerChan {
			if peer.ID == host.ID() {
				continue
			}
			fmt.Println("Found peer:", peer)

			fmt.Println("Connecting to:", peer)
			stream, err := host.NewStream(ctx, peer.ID, protocol.ID(config.ProtocolID))

			if err != nil {
				fmt.Println("Connection failed:", err)
				continue
			} else {
				rw := bufio.NewReadWriter(bufio.NewReader(stream), bufio.NewWriter(stream))

				go writeData(rw)
				go readData(rw)
			}

			fmt.Println("Connected to:", peer)
		}
	}()
 ```

Comment (MGMCN) [reply examples/chat-with-mdns/main.go:119]: image
This is a screenshot. First, I compiled the code, and then tried to run the code in two terminals. I found that both of them found themselves and tried to establish a connection with themselves.

But I don't think it should happen that I try to connect with myself for the first time. So when we find a peer with the same ID as ourselves, we should not try to connect, but continue to wait.

image
After compiling my modified code, this program can successfully create a connection with another peer every time. Then the two peers can communicate.

looking forward to your reply.

Comment (marten-seemann) [reply examples/chat-with-mdns/main.go:119]: I got the logic the wrong way round, sorry for that!

I'm wondering if this fix should live inside our mDNS code though. We _shouldn't_ discover our own node.

Comment (MGMCN) [reply examples/chat-with-mdns/main.go]:
```go
for _, info := range infos {
	if info.ID == s.host.ID() {
		continue
	}
	go s.notifee.HandlePeerFound(info)
}
```
Hello, thank you for your reply.

Do you mean here? Add such code in line 179 of [mdns.go](https://github.com/libp2p/go-libp2p/blob/master/p2p/discovery/mdns/mdns.go) file.

Or should we not find ourselves in the discovery stage?

looking forward to your reply.

MGMCN: Or do you just want me to modify mdns.go under the chat-with-mDNS file directory? 
looking forward to your reply.

marten-seemann: Can you please send a separate PR for your changes to mdns.go? It's an actual bug fix in libp2p!

MGMCN: Thank you for your reply. I have resubmitted another PR.

MGMCN:
I click wrong , sorry

PR #1661, opened 2022-07-26
mdns: don't discover ourselves

MGMCN:


marten-seemann (approved):

marten-seemann:
Looks like we had a test case `TestSelfDiscovery` asserting that we discover ourselves. I'm not why that would be useful. @Stebalien, wdyt?

MGMCN:
Thank you for your reply.
Yes, I saw the test.
Maybe I overlooked some details?
But I don't know the meaning of finding ourselves.

MGMCN:
image

Because after this code modification, both "TestSelfDiscovery" and "TestOtherDiscovery" failed. I have a bold guess. 

Before we modify the code, whether these two tests passed because peer found itself. 

But this is just my assumption.

Looking forward to your reply.

Stebalien:
> Looks like we had a test case TestSelfDiscovery asserting that we discover ourselves. I'm not why that would be useful. @Stebalien, wdyt?

I have no idea...

marten-seemann:
Then let's remove it. I don't see a use case for self-discovery. @MGMCN, do you want to modify the test cases?

MGMCN:
Thank you for your reply. 

I have removed this test function(TestSelfDiscovery).

marten-seemann:
Looks like there's another test that's failing as well.

MGMCN:
Hello, I think I probably know why our TestOtherDiscovery test failed.

```go
for _, id := range hostIDs { ... )
```

Because we traverse all hostids in line 82 of the [mdns_test.go](https://github.com/libp2p/go-libp2p/blob/master/p2p/discovery/mdns/mdns_test.go), but because we have modified the mDNS code, we make the current host unable to find itself. In other words, when we test the file, if we find that the ith host is checking whether we have found peers other than ourselves, we should skip checking whether this host is in our discovery list. But if we follow the original logic, we will also find out whether we are in this discovery list.

After I modify the logic of the test code, the test case passes.

MGMCN:
Sorry, is there anything else I need to correct about this PR?

p2p/discovery/mdns/mdns_test.go
```diff
@@ -97,13 +79,13 @@ func TestOtherDiscovery(t *testing.T) {
...
-			for _, notif := range notifs {
-				infos := notif.GetPeers()
+			for i := 0; i < n; i++ {
+				infos := notifs[i].GetPeers()
 				ids := make([]peer.ID, 0, len(infos))
 				for _, info := range infos {
 					ids = append(ids, info.ID)
 				}
-				if !containsAllHostIDs(ids) {
+				if !containsAllHostIDs(ids,i) {
```
marten-seemann, line 88: Why not just pass the peer ID into this function?

marten-seemann (changes requested):
Please don't remove the `range` loops in the test. Passing around array indices is very difficult to reason about.

p2p/discovery/mdns/mdns_test.go
MGMCN: Thank you for your suggestion. 
I've revised it. 
Do you think it's ok? 
Please contact me if there is anything else that needs to be modified.

marten-seemann (commented):
The code lgtm, but `gofmt` is unhappy.

MGMCN:
Hahaha.
Thank you for your reply. 
Maybe I need to execute "gofmt -s -w" on the [mdns_test.go](https://github.com/libp2p/go-libp2p/pull/1661/files#diff-291ec0b28f1bc4659443b71be4acd3859aa0024eb0477c2808222091b8dad31c)?

marten-seemann (approved):

marten-seemann:
Thank you @MGMCN!

PR #1663, opened 2022-08-01
Allow adding prologue to noise connections

ckousik:
Allows adding the prologue to an existing noise transport. This is for potential use in implementing the webrtc transport.
Related discussion: https://github.com/libp2p/specs/pull/412#discussion_r931163931

### DIFF
--- a/p2p/security/noise/intermediate_transport.go
+++ b/p2p/security/noise/intermediate_transport.go
@@ -1,0 +1,50 @@
+package noise
+
+import (
+	"context"
+	"net"
+
+	"github.com/libp2p/go-libp2p-core/canonicallog"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/sec"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+type NoiseOptions struct {
+	Prologue []byte
+}
+
+var _ sec.SecureTransport = &intermediateTransport{}
+
+// intermediate transport can be used
+// to provide per-connection options
+type intermediateTransport struct {
+	t *Transport
+	// options
+	prologue []byte
+}
+
+// SecureInbound runs the Noise handshake as the responder.
+// If p is empty, connections from any peer are accepted.
+func (i *intermediateTransport) SecureInbound(ctx context.Context, insecure net.Conn, p peer.ID) (sec.SecureConn, error) {
+	c, err := newSecureSession(i.t, ctx, insecure, p, i.prologue, false)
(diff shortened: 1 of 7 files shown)


p2p/security/noise/intermediate_transport.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	return c, err
+}
+
+// SecureOutbound runs the Noise handshake as the initiator.
+func (i *intermediateTransport) SecureOutbound(ctx context.Context, insecure net.Conn, p peer.ID) (sec.SecureConn, error) {
+	return newSecureSession(i.t, ctx, insecure, p, i.prologue, true)
+}
+
+func (t *Transport) WithNoiseOptions(opts NoiseOptions) sec.SecureTransport {
```
marten-seemann, line 45: Why can't this be a normal constructor option?

p2p/security/noise/intermediate_transport.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+	"net"
+
+	"github.com/libp2p/go-libp2p-core/canonicallog"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/libp2p/go-libp2p-core/sec"
+	manet "github.com/multiformats/go-multiaddr/net"
+)
+
+type NoiseOptions struct {
```
marten-seemann, line 13: Please avoid config structs and use the normal config function pattern.

marten-seemann (changes requested):

p2p/security/noise/intermediate_transport.go
ckousik: This is because the noise transport has both transport level and session level options. eg. The private key could be shared among multiple different connections, but the prologue can be unique to each connection.

ckousik: I can rename this to `WithSessionOptions`

p2p/security/noise/intermediate_transport.go
ckousik: I considered that but want to avoid exposing the intermediate transport.

marten-seemann (approved):
lgtm, module a minor nit in the test case.

ckousik:
@marten-seemann It looks like the previous commit failed the tests because `TestSingleRelay` in `p2p/host/autorelay` failed. I pulled the master branch and it also fails on master.
go-libp2p-autorelay-error

marten-seemann:
Yeah, that's a known flaky test.

PR #1672, opened 2022-08-10
examples: remove ipfs components from echo

ZuoFuhong:


examples/echo/main.go
```diff
@@ -18,19 +18,13 @@ import (
...
 )
 
 func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	// LibP2P code uses golog to log messages. They log with different
-	// string IDs (i.e. "swarm"). We can control the verbosity level for
-	// all loggers with:
-	golog.SetAllLoggers(golog.LevelInfo) // Change to INFO for extra info
```
marten-seemann, line 32: This is useful, please don't remove.

marten-seemann (commented):
This looks like a worthwhile change, but please don't remove the logging part.

examples/echo/main.go
ZuoFuhong: I think it is possible to use the standard library log.

Stebalien: We use go-log extensively in libp2p, so this doesn't really help.

ZuoFuhong: Ok，i undo

marten-seemann (approved):
LGTM, thank you!

PR #1675, opened 2022-08-12
connmgr: prefer peers with no streams when closing connections

MarcoPolo:
This prioritized closing connections with peers that have no active streams. This enables us to prune inactive connections rather than a potentially active but lower value connection.

Removes `SortByValue` and changes `SortByValueAndStreams` to work for both the normal case and the emergency memory free up mode.

### DIFF
--- a/p2p/net/connmgr/connmgr.go
+++ b/p2p/net/connmgr/connmgr.go
@@ -243,19 +243,11 @@ type peerInfo struct {
 
 type peerInfos []peerInfo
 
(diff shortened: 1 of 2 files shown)


marten-seemann (approved):
This change looks good to me.

At some point in the future, we probably should make this act on connections, not on peers. For example, if we have more than a single connection to a peer, we should prune the duplicate connections first (interesting question: how would we decide if to prune a TCP connection with more streams vs a QUIC connection with fewer streams).

PR #1683, opened 2022-08-17
move go-libp2p-core here

marten-seemann:
The last remaining part of #1556.

This PR also:
* moves the mock resource manager from `go-libp2p-testing/mocks/network` to `core/network/mocks`
* stops using the `go-libp2p-testing/net` helper function for generating peer identities. This is required because currently (until we've forwarded the type, which requires a release), the old and the new `peer.ID` are not identical, which breaks the build. These functions were not super useful anyway, it's not that hard to generate a new identity.


vyzo:
As I have outlined in the issue, I *disagree* with moving core in the main repo.

I have rather strong conviction that applications (and libraries) should not depend on go-libp2p itself except for initialization and tests; everything else only depends on the interface.

This is also going to create a ton of work for upstream users, who currently only depend on the core interfaces for application logic (other than initialization and tests).

vyzo:
The last part should br stressed out and very much considered. You are effectively creating a bifurcation point of incompatibility and forcing everyone to change threir code in tandem to accomodate your whims.

You really should put more consideration to the ecosystem as a whole and the myriad of libraries and applications that depend on libp2p.

marten-seemann:
> As I have outlined in the issue, I disagree with moving core in the main repo.
> 
> I have rather strong conviction that applications (and libraries) should not depend on go-libp2p itself except for initialization and tests; everything else only depends on the interface.

There was a good discussion about this on #1556. I really appreciate your feedback!

Unfortunately, this is a binary decision (either we consolidate, or we don't), and we have to decide one way or the other. I realize we haven't been able to convince each other of our respective opinions. 
Based on the discussion on the issue, I've made the decision to move forward with repo consolidation. The feedback from other team members shows that most people think that the pros of consolidating outweigh the cons.

I suggest keeping the discussion on that issue, and use this PR only for conversations about the specifics of the plan.

MarcoPolo (approved):
I’m really excited for this as a user and a maintainer. We can stop telling people “You need to reference the correct version of go-libp2p-core when you use go-libp2p. Which version? the one that go-libp2p uses itself…”

I won’t rehash our discussion in #1556, but we can all agree that we want go-libp2p to be easier to use and easier to maintain. I think this is a step in the right direction and in reviewing #1556 I believe the main objections have been addressed.

PR #1691, opened 2022-08-19
simplify the release checklist

marten-seemann:
If we're cutting more frequent and smaller releases, we need to cut down on the overhead a release is creating.
Most of these changes just reflect what we've been doing (or rather, what we've _not_ been doing) in the past.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.md
+++ b/.github/ISSUE_TEMPLATE/release.md
@@ -19,39 +19,13 @@ about: 'Start a new libp2p release.'
 
 - [ ] **Stage 0 - Finishing Touches**
     - [ ] Go through relevant libp2p repos looking for unreleased changes that should make it into the release. If you find any, cut releases.
-    - [ ] Run `go get -u ./...` to see if there are any out-of-date deps that look important. If there are, bubble them. Try to avoid _directly_ updating indirect deps in go-libp2p's go.mod when possible.
-    - [ ] Make sure local tests are passing.
-- [ ] **Stage 1 - Upstream Testing**
-  - Create testing branches in lotus & go-ipfs with the new go-libp2p release and run CI/tests. Many upstream projects are tested in CI, but lotus & go-ipfs are not.
-    - [ ] [filecoin-project/lotus](https://github.com/filecoin-project/lotus)
-    - [ ] [ipfs/go-ipfs](https://github.com/ipfs/go-ipfs)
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/release.md
```diff
@@ -19,39 +19,13 @@ about: 'Start a new libp2p release.'
...
   - [ ] Publish the release through the GitHub UI, adding the release notes. Some users rely on this to receive notifications of new releases.
   - [ ] Announce the release on the [discuss.libp2p.io](https://discuss.libp2p.io).
-- [ ] **Stage 4 - Update Upstream**
+- [ ] **Stage 2 - Update Upstream**
   - [ ] Update the examples to the final release
-  - [ ] Update the upstream testing branches to the final release and create PRs.
+  - [ ] Update the upstream dependencies to the final release and create PRs.
     - [ ] [filecoin-project/lotus](https://github.com/filecoin-project/lotus)
-    - [ ] [ipfs/go-bitswap](https://github.com/ipfs/go-bitswap)
     - [ ] [ipfs/go-ipfs](https://github.com/ipfs/go-ipfs)
```
BigLep, line 30:
```suggestion
    - [ ] [ipfs/kubo](https://github.com/ipfs/kubo)
```

BigLep (commented):
General comments:
1. Can we have a "breaking changes" section so that is clearly called out for users?
2. My understanding is that we cut PRs against Kubo/Lotus because their CI (e.g., Kubo sharness tests) help catch issues in libp2p before release.  Is that right?  I agree libp2p team shouldn't be doing the work to get go-libp2p features exposed in Kubo unless it's critical for libp2p's success (e.g., AutoRelay / Flare).  Anyways, I think we should document/capture the guiding principles here so it's clear for new team members.

The goal we want to have here is that no tribal knowledge should be execute it.  We should be able to drop other/new team members in and they should be able to execute it as well.

MarcoPolo:
> My understanding is that we cut PRs against Kubo/Lotus because their CI (e.g., Kubo sharness tests) help catch issues in libp2p before release.

Yeah I think that's right. But I think that's an indicator that our own testing coverage isn't good enough. I know I've caught bugs I've written because of Kubo sharness. Maybe we keep the "Kubo CI passes" pre-release check if the changes are significant (major release) until we feel more confident in our own test suite?

MarcoPolo (approved):

PR #1722, opened 2022-08-26
rcmgr: Fix threshold calculation

### LINKED ISSUE #1721: Resource manager's InfiniteLimits are not infinite
cc: @MarcoPolo

When using Resource manager with `InfiniteLimits` via:
```
rcmgr.NewResourceManager(rcmgr.NewFixedLimiter(rcmgr.InfiniteLimits))
```
We still seeing connections failing with `ErrResourceLimitExceeded`, likely due to the numeric overflow in [checkMemory](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/resource-manager/scope.go#L83)

PS. While here it would be nice to have unittest for the `InfiniteLimits`.

MarcoPolo:
Fixes #1721 

This changes the code to check for overflows and handle them.

This also adds a special case to check if the limit is MaxInt64 and return early since we assume this is an infinite limit (just to save a big int multiplication).

### DIFF
--- a/p2p/host/resource-manager/scope.go
+++ b/p2p/host/resource-manager/scope.go
@@ -2,6 +2,8 @@ package rcmgr
 
 import (
 	"fmt"
+	"math"
+	"math/big"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


mudler:
sorry to chime in here, but this PR solves quite some issues in a project of mine that it's based on libp2p. since I've tried to bump to latest I met various problems that I couldn't pin point easily in tests, hence I didn't opened any issue because I was still digging from my side. 

That being said I've pinned to this PR and (almost) all tests went back to green:
before: https://github.com/mudler/edgevpn/actions/runs/2899789247
after: https://github.com/mudler/edgevpn/runs/8051539079?check_suite_focus=true

Just wanted to let you know, here looks good :+1:

vyzo (approved):
i am surprised we actually hit this one.

PR #1724, opened 2022-08-29
Configure protocolVersion for Identify protocol

tshakalekholoane:
Allows the protocolVersion field of the Idenfity protocol to be
configured on the host. The current value is fixed for what appears to
be for backwards compatibility with IPFS which makes it difficult for
non-IPFS protocols to use the library.

References:
  - https://github.com/libp2p/go-libp2p/issues/714
  - https://github.com/libp2p/go-libp2p/pull/1137
  - https://github.com/libp2p/rust-libp2p/blob/6855ab943bd7427a2135b46ad3d08f48fbf10872/protocols/identify/src/identify.rs#L125-L127

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -64,6 +64,11 @@ type Config struct {
 	// Set it via the UserAgent option function.
 	UserAgent string
 
+	// ProtocolVersion is the protocol version that identifies the family
+	// of protocols used by the peer in the Identify protocol. It is set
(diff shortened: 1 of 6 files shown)


Stebalien (approved):
I've checked js-libp2p, python-libp2p, jvm-libp2p, and rust-libp2p. None of them assert the protocol version (as far as I can tell) so we should be good to go.

PR #1736, opened 2022-09-02
autorelay: Add a context.Context to WithPeerSource callback

Jorropo:
The main point of this is to stop the search when the libp2p stops.
I've added that this *might* be used in the future to notify when AutoRelay is satisfied just to not close any door in the future but havn't implemented anything like that.

### DIFF
--- a/p2p/host/autorelay/relay_finder.go
+++ b/p2p/host/autorelay/relay_finder.go
@@ -59,7 +59,7 @@ type relayFinder struct {
 	ctxCancel   context.CancelFunc
 	ctxCancelMx sync.Mutex
 
-	peerSource func(int) <-chan peer.AddrInfo
+	peerSource func(context.Context, int) <-chan peer.AddrInfo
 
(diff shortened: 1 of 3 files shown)


p2p/host/autorelay/relay_finder.go
```diff
@@ -59,7 +59,7 @@ type relayFinder struct {
 	ctxCancel   context.CancelFunc
 	ctxCancelMx sync.Mutex
 
-	peerSource func(int) <-chan peer.AddrInfo
+	peerSource func(context.Context, int) <-chan peer.AddrInfo
```
marten-seemann, line 62: Would it make sense to remove the `int` here, and have the caller give us new nodes until we cancel the context?

Jorropo: Maybe ?
The `int` allows to create a buffered channel of the right size for synchronous or partially synchronous implementations tho.

marten-seemann (approved):

PR #1738, opened 2022-09-04
examples: add basic pubsub-with-rendezvous example

smartin015:
I'm evaluating libp2p's pubsub implementation for coordinating a dynamic number of distributed nodes. I noticed there was a neat (but IMO rather complex and hard to quickly grok) example for pubsub using mdns, but as my nodes could span multiple networks I needed to use rendezvous instead.

This PR is a pared-down example that mashes together `pubsub/chat` and `chat-with-rendezvous`, stripping away all of the UI prettiness in order to more clearly show the most basic setup needed to get a shared pubsub topic sent to 2+ nodes.

I haven't yet tested this on multiple PCs / across NAT barriers, but it does work with multiple processes running locally.


marten-seemann:
Does it make sense to exhaust the combinatory space here, and have a pubsub with mdns, a pubsub with rendezvouz, and a chat with rendezvouz example? Every example creates maintenance burden for years to come, and we should only add it if it actually adds value over what we currently have.

smartin015:
That's fair - you're welcome to reject this PR, or suggest something else. I just wanted to share the product of a couple hours of trying to work out pubsub with rendezvous in case it was useful to anyone else.

marten-seemann:
@vyzo, what do you think? You're more familiar with our pubsub code.

vyzo (approved):
All that looks fine to me.

BigLep:
@smartin015 : CI is failing.  Can you please investigate to try and get CI to pass?

smartin015:
> @smartin015 : CI is failing. Can you please investigate to try and get CI to pass?

Can do - looks like some deprecation notices are the issue. I'll try to have a fix in by the next couple of days.

smartin015:
And yet... apparently I needed to `go mod tidy` after mucking with the import names. Apologies, that's also now checked in.

MarcoPolo (approved):

MarcoPolo:
Thanks @smartin015  :)

PR #1740, opened 2022-09-05
rcmgr: increase default connection memory limit to 32 MB

### LINKED ISSUE #1706: rcmgr: connection-level memory base limits are too small for performant QUIC transfers
Our connection-level memory base limit is 1 MB: https://github.com/libp2p/go-libp2p-resource-manager/blob/9bb1bbdc782e0f7c43768e1946e659432edbc6b0/limit_defaults.go#L507-L513

quic-go starts with a stream flow control window of 512 kB and a connection flow control window of 768 kB: https://github.com/lucas-clemente/quic-go/blob/8c0c481da1644f9934df399c50649d65967d7f22/internal/protocol/params.go#L24-L28

This means that with the base limit, we won't be allowed to ever increase the flow control limit of a QUIC connection. Throughput is then limited by flow control.

## How bad is this?

I ran some benchmark between two servers (US East to Europe) located about 80ms RTT apart. With TCP, I was able to achieve transfer speeds of roughly 33 MB/s, while QUIC was only able to achieve 6 MB/s.

The situation gets better for larger servers, as limit autoscaling will allocate more memory to the connection. **I don't think QUIC connections being significantly slower than TCP connections is tenable on any size of machine though.**

## Problem Analysis

The problem here is the following:
1. We allow 128 connections in the base configuration, and it's questionable if we can go a lot lower than this value if we want to keep libp2p usable.
2. Our system memory limit is 128 MB, so we can't really go beyond 1 MB per connection, otherwise we might use _all_ of our allowed memory for the QUIC transport, rendering the entire libp2p stack useless, as it won't be able to allocate even a small buffer for message parsing.
3. It is quite easy for a malicious node to allocate its entire flow control window: All it has to do is not send the first byte (preventing us to read from the stream), and then fill up the flow control window with data.

## Possible Solution

I have the feeling that this brings us back to libp2p/go-libp2p#1708: Memory management is really only useful for muxers on the libp2p layer.
If we limit memory accounting to muxers (the only real consumers of memory using the rcmgr), and introduce a dedicated muxer memory pool, this would give us 128 MB for QUIC connections. Each connection would start with 768 kB of connection flow control limit, consuming 96 MB of memory (worst case). With the remaining 32 MB some QUIC connections (the ones that are currently getting used for high BDP transfers) would be able to increase their flow control windows.
We might also be able to teach quic-go to decrease flow control windows once utilization of the connection goes down, so that a long-living connection that transferred a lot of data an hour ago doesn't keep occupying a lot of data.

cc @vyzo @MarcoPolo @Stebalien

marten-seemann:
Fixes #1706.

### DIFF
--- a/p2p/host/resource-manager/limit_defaults.go
+++ b/p2p/host/resource-manager/limit_defaults.go
@@ -514,7 +514,7 @@ var DefaultLimits = ScalingLimitConfig{
 		ConnsOutbound: 1,
 		Conns:         1,
 		FD:            1,
-		Memory:        1 << 20,
+		Memory:        32 << 20,
 	},
 
 	StreamBaseLimit: BaseLimit{


vyzo (approved):
16 is probably enough, but shouldnt be a problem.
You might also want to adjust the peer base limit to 16M at least.

marten-seemann:
The maximum flow control window that [quic-go uses is 15 MB](https://github.com/lucas-clemente/quic-go/blob/62b82789c0cc8a9acf48ee516ce2dd303253f395/internal/protocol/params.go#L33-L34), and we're reserving with 128 priority, so we'll need to set this to 30 MB at least. Will update the PR to use 32 MB.

PR #1746, opened 2022-09-10
noise: don't fail handshake when early data is received without handler

marten-seemann:
I just realized that the change we introduced in #1728 would preclude us from rolling out the change that @julian88110 is working on (see https://github.com/libp2p/specs/pull/446). This means that this PR must make it into v0.23.0.


julian88110 (approved):
Marten, Thanks for changing this, yes the Muxer selection feature relies on this behavior for backward compatibility.

marten-seemann:
I just realized that https://github.com/libp2p/specs/pull/446 uses the second and the third flight, so technically this would break. It's a good idea to be not fail the handshake in case we ever want to add some data to the first flight though.

MarcoPolo (approved):
Nice catch

PR #1750, opened 2022-09-13
noise: make it possible for the server to send early data

marten-seemann:
So far, only the client could send early data in its first handshake message, which the server would then receive.

This PR extends the logic, such that the server can now also send early data in its first handshake message.

Needed for https://github.com/libp2p/specs/pull/404#discussion_r968836367. Also needed for https://github.com/libp2p/specs/pull/446, if I understand correctly.

### DIFF
--- a/p2p/security/noise/handshake.go
+++ b/p2p/security/noise/handshake.go
@@ -10,8 +10,6 @@ import (
 	"runtime/debug"
 	"time"
 
-	"golang.org/x/crypto/chacha20poly1305"
-
 	"github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/p2p/security/noise/pb"
@@ -73,12 +71,8 @@ func (s *secureSession) runHandshake(ctx context.Context) (err error) {
 		}
 	}
 
-	// We can re-use this buffer for all handshake messages as its size
-	// will be the size of the maximum handshake message for the Noise XX pattern.
-	// Also, since we prefix every noise handshake message with its length, we need to account for
-	// it when we fetch the buffer from the pool
-	maxMsgSize := 2*noise.DH25519.DHLen() + 2*chacha20poly1305.Overhead + 1024 /* payload */
-	hbuf := pool.Get(maxMsgSize + LengthPrefixLength)
+	// We can re-use this buffer for all handshake messages.
+	hbuf := pool.Get(2 << 10)
 	defer pool.Put(hbuf)
 
 	if s.initiator {
--- a/p2p/security/noise/handshake.go
+++ b/p2p/security/noise/handshake.go
@@ -64,11 +64,6 @@ func (s *secureSession) runHandshake(ctx context.Context) (err error) {
 		return fmt.Errorf("error initializing handshake state: %w", err)
 	}
 
-	payload, err := s.generateHandshakePayload(kp)
-	if err != nil {
-		return err
-	}
-
 	// set a deadline to complete the handshake, if one has been supplied.
 	// clear it after we're done.
 	if deadline, ok := ctx.Deadline(); ok {
@@ -82,7 +77,7 @@ func (s *secureSession) runHandshake(ctx context.Context) (err error) {
 	// will be the size of the maximum handshake message for the Noise XX pattern.
 	// Also, since we prefix every noise handshake message with its length, we need to account for
 	// it when we fetch the buffer from the pool
-	maxMsgSize := 2*noise.DH25519.DHLen() + len(payload) + 2*chacha20poly1305.Overhead
+	maxMsgSize := 2*noise.DH25519.DHLen() + 2*chacha20poly1305.Overhead + 1024 /* payload */
 	hbuf := pool.Get(maxMsgSize + LengthPrefixLength)
 	defer pool.Put(hbuf)
 
@@ -102,12 +97,22 @@ func (s *secureSession) runHandshake(ctx context.Context) (err error) {
 		if err != nil {
 			return fmt.Errorf("error reading handshake message: %w", err)
 		}
-		if err := s.handleRemoteHandshakePayload(plaintext, hs.PeerStatic()); err != nil {
+		rcvdEd, err := s.handleRemoteHandshakePayload(plaintext, hs.PeerStatic())
+		if err != nil {
 			return err
+		}
+		if s.earlyDataHandler != nil {
+			if err := s.earlyDataHandler.Received(ctx, s.insecureConn, rcvdEd); err != nil {
+				return err
+			}
 		}
 
 		// stage 2 //
 		// Handshake Msg Len = len(DHT static key) +  MAC(static key is encrypted) + len(Payload) + MAC(payload is encrypted)
+		payload, err := s.generateHandshakePayload(kp, nil)
+		if err != nil {
+			return err
+		}
 		if err := s.sendHandshakeMessage(hs, payload, hbuf); err != nil {
 			return fmt.Errorf("error sending handshake message: %w", err)
 		}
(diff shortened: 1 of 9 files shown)


p2p/security/noise/handshake.go
```diff
@@ -102,12 +97,22 @@ func (s *secureSession) runHandshake(ctx context.Context) (err error) {
 		if err != nil {
 			return fmt.Errorf("error reading handshake message: %w", err)
 		}
-		if err := s.handleRemoteHandshakePayload(plaintext, hs.PeerStatic()); err != nil {
+		rcvdEd, err := s.handleRemoteHandshakePayload(plaintext, hs.PeerStatic())
+		if err != nil {
 			return err
 		}
+		if s.earlyDataHandler != nil {
+			if err := s.earlyDataHandler.Received(ctx, s.insecureConn, rcvdEd); err != nil {
```
julian88110, line 105: The early data handler may need to know the handshake state to drive its internal state machine, for example if this is the server or client side, at which stage is the early data received. Do you think we should pass the handshake state info to the earlyDataHandler too?  or maybe that is available in other ways already and I missed it?

p2p/security/noise/handshake.go
```diff
@@ -82,7 +77,7 @@ func (s *secureSession) runHandshake(ctx context.Context) (err error) {
 	// will be the size of the maximum handshake message for the Noise XX pattern.
 	// Also, since we prefix every noise handshake message with its length, we need to account for
 	// it when we fetch the buffer from the pool
-	maxMsgSize := 2*noise.DH25519.DHLen() + len(payload) + 2*chacha20poly1305.Overhead
+	maxMsgSize := 2*noise.DH25519.DHLen() + 2*chacha20poly1305.Overhead + 1024 /* payload */
```
julian88110, line 80: Just wondering if there is a more efficient way of setting length rather than a hard limit here, I don't have a better way readily available, just want to throw some thoughts here to explore the options.

julian88110 (commented):
Hi Marten, Thanks for making this API available!

p2p/security/noise/handshake.go
marten-seemann: For both client and server, the data is sent with their respective first flights.

I think the nicest way to resolve this is to pass separate client and server `EarlyDataHandlers` to the `EarlyData` option. I applied that change in 0b3683270e9380124b675625cf77f0113f7905e4. What do you think?

p2p/security/noise/handshake.go
marten-seemann: Changed to a fixed value of 2048 in 24459c5264da47b2d1431bc06473a13964fe09ca.

It's not even a hard limit, it's just the size we use to get a buffer from the [buffer pool](https://github.com/libp2p/go-buffer-pool). If we get a buffer that's too small, the worst thing that happens is that `append` would allocate a larger buffer.

We can just use a large enough buffer here. What exactly "large enough" means depends on the amount of early data we send, but 2048 should be enough for practically all use cases. And if not, we can live with that one allocation.

p2p/security/noise/handshake.go
julian88110: That looks great, thanks for the changes which make the Noise based muxer selection and other early data utilization smoother.

p2p/security/noise/handshake.go
julian88110: Thanks for the clarification, that makes sense. Sorry I was not aware that we can do extra allocation if the limit is exceeded.

julian88110 (approved):
LGTM

p2p/security/noise/session.go
```diff
@@ -36,22 +36,24 @@ type secureSession struct {
 	dec *noise.CipherState
 
 	// noise prologue
-	prologue         []byte
-	earlyDataHandler EarlyDataHandler
+	prologue []byte
+
+	initiatorEarlyDataHandler, receiverEarlyDataHandler EarlyDataHandler
 }
```
MarcoPolo, line 42: same suggestion for calling this `responder`

MarcoPolo (changes requested):

p2p/security/noise/transport_test.go
```diff
@@ -447,74 +447,105 @@ func (e *earlyDataHandler) Received(ctx context.Context, conn net.Conn, data []b
...
+		select {
+		case <-time.After(500 * time.Millisecond):
+			t.Fatal("timeout")
+		case err := <-errChan:
+			serverErr = err
+		}
+		return
+	}
+
+	receivingEDH := &earlyDataHandler{
```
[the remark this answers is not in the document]
marten-seemann: We use these both ways here. Sending and receiving is here the direction of the early data.

p2p/security/noise/session_transport.go
```diff
@@ -31,9 +31,10 @@ type EarlyDataHandler interface {
 	Received(context.Context, net.Conn, []byte) error
 }
 
-func EarlyData(h EarlyDataHandler) SessionOption {
+func EarlyData(initiator, receiver EarlyDataHandler) SessionOption {
```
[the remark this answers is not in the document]
marten-seemann: Done.

p2p/security/noise/session.go
marten-seemann: Done.

p2p/security/noise/handshake.go
```diff
@@ -118,15 +117,23 @@ func (s *secureSession) runHandshake(ctx context.Context) (err error) {
 		if err != nil {
 			return fmt.Errorf("error reading handshake message: %w", err)
 		}
-		if s.earlyDataHandler != nil {
-			if err := s.earlyDataHandler.Received(ctx, s.insecureConn, initialPayload); err != nil {
+		if s.receiverEarlyDataHandler != nil {
```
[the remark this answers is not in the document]
marten-seemann: Done.

marten-seemann:
@MarcoPolo Good catch, this will make the code easier to understand. I updated the variable names to match the spec.

p2p/security/noise/session_transport.go
```diff
@@ -31,9 +31,10 @@ type EarlyDataHandler interface {
 	Received(context.Context, net.Conn, []byte) error
 }
 
-func EarlyData(h EarlyDataHandler) SessionOption {
+func EarlyData(initiator, responder EarlyDataHandler) SessionOption {
```
MarcoPolo, line 34: We should have a comment here explaining when the initiator and responder sends and receives this early data. "Early data" isn't a noise concept, we made this up. We should be clear that early data sent by the initiator is sent **plaintext** in the **first** message. And the responder early data is sent in the second message. 

I got tripped up by this recently since in the js-libp2p noise implementation early data from the initiator is sent in the **third** message so that it's encrypted.

I think a comment with at least a big warning about the plaintext part and an explanation of when the data is sent would help here.

MarcoPolo (changes requested):
One small thing

p2p/security/noise/session_transport.go
MarcoPolo: oh and update the comments on EarlyDataHandler: https://github.com/libp2p/go-libp2p/blob/4f8a466b4cf3da8df34ea2cf26b4045aa435f2bb/p2p/security/noise/session_transport.go#L25

MarcoPolo: @marten-seemann I made these changes in https://github.com/libp2p/go-libp2p/pull/1750/commits/0fe5519d217b2a404a6dac55fe5ce7c8b0fbeb06 to help. Feel free to remove/revert that commit if you want to do something different.

marten-seemann: @MarcoPolo Thank you for updating the documentation. I really like your change!

MarcoPolo (approved):
Approved with the comments I added to EarlyDataHandler. Feel free to tweak, but the gist should be the same (a user should understand the caveats around this and what "early data" means in go-libp2p since this isn't a standard).

marten-seemann:
I updated the PR, such that early data is sent with the 2nd and 3rd handshake message. This means that early data is always sent encrypted.

The slight annoyance here is that if the `Received` function returns an error, we can't fail the handshake any more more, as the client already completed its handshake. Instead, it will just encounter a closed connection. Note that this is similar to what happens in TLS 1.3 when the server rejects the client certificate, so I'm not too worried about this.

p2p/security/noise/session_transport.go
```diff
@@ -22,18 +22,28 @@ func Prologue(prologue []byte) SessionOption {
 	}
 }
 
-// EarlyDataHandler allows attaching an (unencrypted) application payload to the first handshake message.
-// While unencrypted, the integrity of this early data is retroactively authenticated on completion of the handshake.
+// EarlyDataHandler defines what the application payload is for either the second
+// (if initiator) or third (if responder) handshake message, and defines the
```
MarcoPolo, line 26:
```suggestion
// (if responder) or third (if initiator) handshake message, and defines the
```

Right?

MarcoPolo (commented):
One fix for the comment, but otherwise looks good. Feel free to ping when the change is in and I’ll approve

p2p/security/noise/session_transport.go
MarcoPolo: I’m going to commit this and approve.

MarcoPolo (approved):

PR #1756, opened 2022-09-16
swarm: skip dialing WebTransport addresses when we have QUIC addresses

marten-seemann:
Question: Should we have something similar for WebSocket and TCP?

### DIFF
--- a/p2p/net/swarm/swarm_dial.go
+++ b/p2p/net/swarm/swarm_dial.go
@@ -439,7 +439,7 @@ func (s *Swarm) filterKnownUndialables(p peer.ID, addrs []ma.Multiaddr) []ma.Mul
 		}
 	}
 
-	return ma.FilterAddrs(addrs,
+	return maybeRemoveWebTransportAddrs(ma.FilterAddrs(addrs,
 		func(addr ma.Multiaddr) bool { return !ma.Contains(ourAddrs, addr) },
 		s.canDial,
 		// TODO: Consider allowing link-local addresses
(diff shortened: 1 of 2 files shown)


mxinden (commented):
Do I understand correctly that this would drop all webtransport addresses in case there is a single quic address?

In case a node advertises the following two address, the local node should still dial both, right?

```
"/ip4/1.2.3.4/udp/443/quic
"/ip4/5.6.7.8/udp/443/quic/webtransport
```

In other words, only if the addresses are equal on everything before `quic`, only then should the QUIC address replace the webtransport address.

Does that make sense?

marten-seemann:
> In other words, only if the addresses are equal on everything before quic, only then should the QUIC address replace the webtransport address.

Yes, that would be better, but also more code. Not sure if it's worth the complexity.

marten-seemann:
Actually, this probably won't be too hard. I'll update the PR.

p2p/net/swarm/swarm_dial.go
```diff
@@ -530,3 +530,41 @@ func isRelayAddr(addr ma.Multiaddr) bool {
...
+// Happy-Eyeballs-style dialing.
+func maybeRemoveWebTransportAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
+	var hasQuic, hasWebTransport bool
+	for _, addr := range addrs {
+		if isQUIC(addr) {
+			hasQuic = true
+		}
+		if isWebTransport(addr) {
+			hasWebTransport = true
+		}
```
MarcoPolo, line 556:
```suggestion
		hasQuic = isQuic(addr)
		hasWebTransport = isWebTransport(addr)
```

MarcoPolo (approved):

p2p/net/swarm/swarm_dial.go
marten-seemann: Oh wait, this is not correct. The ifs are in a loop. Will revert.

marten-seemann:
Merging this simple version now. We can still have better logic later (or implement the proper Happy Eyeballs fix later).

PR #1757, opened 2022-09-16
webtransport: move certhash verification to the client

marten-seemann:
As suggested in https://github.com/libp2p/specs/pull/455.

This PR builds on top #1750. **Please only review the last commit**. I'll rebase this PR once #1750 is merged.

### DIFF
--- a/p2p/transport/webtransport/transport.go
+++ b/p2p/transport/webtransport/transport.go
@@ -1,9 +1,11 @@
 package libp2pwebtransport
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"io"
 	"sync"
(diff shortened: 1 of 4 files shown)


p2p/transport/webtransport/transport.go
```diff
@@ -196,32 +198,62 @@ func (t *transport) upgrade(ctx context.Context, sess *webtransport.Session, p p
 
 	// Now run a Noise handshake (using early data) and send all the certificate hashes that we would have accepted.
 	// The server will verify that it advertised all of these certificate hashes.
-	msg := pb.WebTransport{CertHashes: make([][]byte, 0, len(certHashes))}
-	for _, certHash := range certHashes {
-		h, err := multihash.Encode(certHash.Digest, certHash.Code)
+	var verified bool
+	n, err := t.noise.WithSessionOptions(noise.EarlyData(newEarlyDataReceiver(func(b []byte) error {
+		ch, err := decodeCertHashesFromProtobuf(b)
```
MarcoPolo, line 203: Nit: maybe something more descriptive than `ch`?

p2p/transport/webtransport/transport.go
```diff
@@ -196,32 +198,62 @@ func (t *transport) upgrade(ctx context.Context, sess *webtransport.Session, p p
...
+	}
 	return &connSecurityMultiaddrs{
 		ConnSecurity:   c,
 		ConnMultiaddrs: &connMultiaddrs{local: local, remote: remote},
 	}, nil
 }
 
+func decodeCertHashesFromProtobuf(b []byte) ([]multihash.DecodedMultihash, error) {
+	var msg pb.WebTransport
+	if err := msg.Unmarshal(b); err != nil {
```
MarcoPolo, line 242: should we error if len(b) == 0? (The unmarshal will not fail this I think)

MarcoPolo: might be okay. A server sending an empty early data can be interpreted to mean that there is no certhash. Just a bit surprising

MarcoPolo (approved):

p2p/transport/webtransport/transport.go
marten-seemann: `decodedCertHashes`.

PR #1761, opened 2022-09-19
webtransport: add custom resolver to add SNI

### LINKED ISSUE #1760: Add resolver to WebTransport transport
Similar to the changes for websockets.

_Originally posted by @marten-seemann in https://github.com/libp2p/go-libp2p/pull/1719#pullrequestreview-1100300612_

MarcoPolo:
The continuation of https://github.com/libp2p/go-libp2p/pull/1719 for webtransport. Fixes #1760

### DIFF
--- a/p2p/transport/webtransport/transport.go
+++ b/p2p/transport/webtransport/transport.go
@@ -311,7 +311,7 @@ func (t *transport) Close() error {
 // extractSNI returns what the SNI should be for the given maddr. If there is an
 // SNI component in the multiaddr, then it will be returned and
 // foundSniComponent will be true. If there's no SNI component, but there is a
-// DNS-like compoent, then that will be returned for the sni and
+// DNS-like component, then that will be returned for the sni and
 // foundSniComponent will be false (since we didn't find an actual sni component).
 func extractSNI(maddr ma.Multiaddr) (sni string, foundSniComponent bool) {
 	ma.ForEach(maddr, func(c ma.Component) bool {
@@ -334,24 +334,18 @@ func extractSNI(maddr ma.Multiaddr) (sni string, foundSniComponent bool) {
 func (t *transport) Resolve(ctx context.Context, maddr ma.Multiaddr) ([]ma.Multiaddr, error) {
 	sni, foundSniComponent := extractSNI(maddr)
 
-	if foundSniComponent {
-		// The multiaddr already had an sni field, we can keep using it.
+	if foundSniComponent || sni == "" {
+		// The multiaddr already had an sni field, we can keep using it. Or we don't have any sni like thing
 		return []ma.Multiaddr{maddr}, nil
 	}
 
-	if sni != "" {
-		// newMaddr := make(ma.Component)
-		// maddr.
-		beforeQuicMA, afterIncludingQuicMA := ma.SplitFunc(maddr, func(c ma.Component) bool {
-			return c.Protocol().Code == ma.P_QUIC
-		})
-		quicComponent, afterQuicMA := ma.SplitFirst(afterIncludingQuicMA)
-		sniComponent, err := ma.NewComponent(ma.ProtocolWithCode(ma.P_SNI).Name, sni)
-		if err != nil {
-			return nil, err
-		}
-		return []ma.Multiaddr{beforeQuicMA.Encapsulate(quicComponent).Encapsulate(sniComponent).Encapsulate(afterQuicMA)}, nil
+	beforeQuicMA, afterIncludingQuicMA := ma.SplitFunc(maddr, func(c ma.Component) bool {
+		return c.Protocol().Code == ma.P_QUIC
+	})
+	quicComponent, afterQuicMA := ma.SplitFirst(afterIncludingQuicMA)
+	sniComponent, err := ma.NewComponent(ma.ProtocolWithCode(ma.P_SNI).Name, sni)
+	if err != nil {
+		return nil, err
 	}
-
-	return []ma.Multiaddr{maddr}, nil
+	return []ma.Multiaddr{beforeQuicMA.Encapsulate(quicComponent).Encapsulate(sniComponent).Encapsulate(afterQuicMA)}, nil
 }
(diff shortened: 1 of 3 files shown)


p2p/transport/webtransport/listener.go
```diff
@@ -215,7 +215,6 @@ func (l *listener) handshake(ctx context.Context, sess *webtransport.Session) (*
 func (l *listener) checkEarlyData(b []byte) error {
 	var msg pb.WebTransport
 	if err := msg.Unmarshal(b); err != nil {
-		fmt.Println(1)
```
marten-seemann, line 218: oops

p2p/transport/webtransport/transport.go
```diff
@@ -268,3 +275,51 @@ func (t *transport) Close() error {
 	}
 	return nil
 }
+
+// extractSNI returns what the SNI should be for the given maddr. If there is an
+// SNI component in the multiaddr, then it will be returned and
+// foundSniComponent will be true. If there's no SNI component, but there is a
+// DNS-like compoent, then that will be returned for the sni and
+// foundSniComponent will be false (since we didn't find an actual sni component).
+func extractSNI(maddr ma.Multiaddr) (sni string, foundSniComponent bool) {
```
marten-seemann, line 284: Do we need the `bool` here? Would an empty string be sufficient? As far as I know, there's no way to distinguish between empty string and unset in the TLS ClientHello: https://github.com/golang/go/blob/d74bf73be052851e83fb59a40f47d49f4b890ca3/src/crypto/tls/handshake_messages.go#L124-L135.

I think the more idiomatic variable name for the `bool` would be `ok`.

p2p/transport/webtransport/transport.go
```diff
@@ -268,3 +275,51 @@ func (t *transport) Close() error {
...
+			return true
+		}
+		return true
+	})
+	return sni, foundSniComponent
+}
+
+// Resolve implements transport.Resolver
+func (t *transport) Resolve(ctx context.Context, maddr ma.Multiaddr) ([]ma.Multiaddr, error) {
+	sni, foundSniComponent := extractSNI(maddr)
```
marten-seemann, line 303:
```suggestion
	sni, ok := extractSNI(maddr)
```

p2p/transport/webtransport/transport.go
```diff
@@ -268,3 +275,51 @@ func (t *transport) Close() error {
...
+
+	if foundSniComponent {
+		// The multiaddr already had an sni field, we can keep using it.
+		return []ma.Multiaddr{maddr}, nil
+	}
+
+	if sni != "" {
+		// newMaddr := make(ma.Component)
+		// maddr.
```
marten-seemann, line 312: Stray comment?

p2p/transport/webtransport/transport.go
```diff
@@ -268,3 +275,51 @@ func (t *transport) Close() error {
...
+		beforeQuicMA, afterIncludingQuicMA := ma.SplitFunc(maddr, func(c ma.Component) bool {
+			return c.Protocol().Code == ma.P_QUIC
+		})
+		quicComponent, afterQuicMA := ma.SplitFirst(afterIncludingQuicMA)
+		sniComponent, err := ma.NewComponent("sni", sni)
```
marten-seemann, line 317:
```suggestion
		sniComponent, err := ma.NewComponent(ma.ProtocolWithCode(ma.P_SNI).Name, sni)
```

p2p/transport/webtransport/transport.go
```diff
@@ -268,3 +275,51 @@ func (t *transport) Close() error {
...
+// Resolve implements transport.Resolver
+func (t *transport) Resolve(ctx context.Context, maddr ma.Multiaddr) ([]ma.Multiaddr, error) {
+	sni, foundSniComponent := extractSNI(maddr)
+
+	if foundSniComponent {
+		// The multiaddr already had an sni field, we can keep using it.
+		return []ma.Multiaddr{maddr}, nil
+	}
+
+	if sni != "" {
```
marten-seemann, line 310: Nit: Use an early return here to avoid identation:
```go
if sni == "" {
     return []ma.Multiaddr{maddr}, nil
}
```

You can also combine that with the `foundSniComponent` check above.

marten-seemann (commented):
Mostly nits.

p2p/transport/webtransport/transport.go
```diff
@@ -268,3 +275,51 @@ func (t *transport) Close() error {
...
+			foundSniComponent = true
+			return false
+		case ma.P_DNS, ma.P_DNS4, ma.P_DNS6, ma.P_DNSADDR:
+			sni = c.Value()
+			// Keep going in case we find an `sni` component
+			return true
+		}
+		return true
+	})
+	return sni, foundSniComponent
```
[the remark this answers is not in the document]
MarcoPolo: I know we can do this, but I'm generally not a fan. Do you feel strongly?

p2p/transport/webtransport/transport.go
```diff
@@ -268,3 +275,51 @@ func (t *transport) Close() error {
 	}
 	return nil
 }
+
+// extractSNI returns what the SNI should be for the given maddr. If there is an
+// SNI component in the multiaddr, then it will be returned and
+// foundSniComponent will be true. If there's no SNI component, but there is a
+// DNS-like compoent, then that will be returned for the sni and
```
marten-seemann, line 282:
```suggestion
// DNS-like component, then that will be returned for the sni and
```

p2p/transport/webtransport/transport.go
marten-seemann: Not carefully enough, apparently ;)

p2p/transport/webtransport/transport.go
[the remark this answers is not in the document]
marten-seemann: No strong feelings. I usually prefer the naked `return` when using the variables from the function the function signature, but I'm not sure if that's an actual Go convention.

p2p/transport/webtransport/transport.go
MarcoPolo: Resolving this as won't fix since I'm not calling this `ok` for a reason.

p2p/transport/webtransport/transport.go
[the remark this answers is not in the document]
MarcoPolo: https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_avoid_naked_returns

And yes, I know the section right about discourages named return values. I think it's better to keep them so I can reference them in the comment so it's clear from the signature. Maybe returning a struct would be better?

This probably isn't worth the bike-shedding.

marten-seemann (approved):

Issue #1769, opened 2022-09-20
invalid memory address or nil pointer dereference in go-libp2p/p2p/transport/websocket


JamesHertz:
After a bit of pain, I was finally able to compile ipfs using the most up-to-date version of go-libp2p. I literally cloned this repository and redirected the ipfs github.com/libp2p to it.
I had all this work just to add a new protocol to libp2p and run it on ipfs. But after a few times running ipfs I got this. I don't believe the protocol I added made this bug so I'm reporting it. I hope you fix it as soon as possible :)

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x88 pc=0x1805475]

goroutine 1287614 [running]:
github.com/libp2p/go-libp2p/p2p/transport/websocket.(*WebsocketTransport).maDial(0xc001938540, {0x2a442c8, 0xc012642420}, {0x2a58cc0, 0xc00de0c720})
        github.com/libp2p/go-libp2p@v0.22.0/p2p/transport/websocket/websocket.go:186 +0x155
github.com/libp2p/go-libp2p/p2p/transport/websocket.(*WebsocketTransport).Dial(0xc001938540, {0x2a442c8, 0xc012642420}, {0x2a58cc0, 0xc00de0c720}, {0xc00c4e61e0, 0x26})
        github.com/libp2p/go-libp2p@v0.22.0/p2p/transport/websocket/websocket.go:162 +0xc5
github.com/libp2p/go-libp2p/p2p/net/swarm.(*Swarm).dialAddr(0xc000c31ba0, {0x2a442c8, 0xc012642420}, {0xc00c4e61e0, 0x26}, {0x2a58cc0?, 0xc00de0c720})
        github.com/libp2p/go-libp2p@v0.22.0/p2p/net/swarm/swarm_dial.go:483 +0x1f1
github.com/libp2p/go-libp2p/p2p/net/swarm.(*dialLimiter).executeDial(0xc000c1d590, 0xc012d74740)
        github.com/libp2p/go-libp2p@v0.22.0/p2p/net/swarm/limiter.go:219 +0xf0
created by github.com/libp2p/go-libp2p/p2p/net/swarm.(*dialLimiter).addCheckFdLimit
        github.com/libp2p/go-libp2p@v0.22.0/p2p/net/swarm/limiter.go:169 +0x4aa

MarcoPolo:
[...]

PR #1770, opened 2022-09-21
websocket: fix nil pointer in tlsClientConf

MarcoPolo:
It's possible (and common) to have a WebSocket transport without a tls client config. This leads to a nil pointer deref error when we set the ServerName.

This sets a default empty `tlsClientConf` and adds a test case.

Fixes https://github.com/libp2p/go-libp2p/issues/1769

### DIFF
--- a/p2p/transport/websocket/websocket_test.go
+++ b/p2p/transport/websocket/websocket_test.go
@@ -165,10 +165,7 @@ func testWSSServer(t *testing.T, listenAddr ma.Multiaddr) (ma.Multiaddr, peer.ID
 		t.Fatal(err)
 	}
 
-	// l, err := tpt.Listen(ma.StringCast("/ip4/127.0.0.1/tcp/0/wss"))
-	// l, err := tpt.Listen(ma.StringCast("/ip4/127.0.0.1/tcp/0/tls/ws"))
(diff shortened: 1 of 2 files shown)


marten-seemann (commented):
The fix and the test case LGTM, but I’d like to investigate the `gorelease` warning regarding yamux before doing the patch release. Will do so later today.

p2p/transport/websocket/websocket_test.go
```diff
@@ -165,10 +165,7 @@ func testWSSServer(t *testing.T, listenAddr ma.Multiaddr) (ma.Multiaddr, peer.ID
 		t.Fatal(err)
 	}
 
-	// l, err := tpt.Listen(ma.StringCast("/ip4/127.0.0.1/tcp/0/wss"))
-	// l, err := tpt.Listen(ma.StringCast("/ip4/127.0.0.1/tcp/0/tls/ws"))
 	l, err := tpt.Listen(listenAddr)
-	fmt.Println("here", listenAddr)
```
Jorropo, line 171: :thinking: :slightly_smiling_face:

marten-seemann:
I removed the release commit. Let's first make sure we can build kubo before cutting a patch release.

PR #1775, opened 2022-09-22
webtransport: return error before wrapping opened / accepted streams

marten-seemann:


p2p/transport/webtransport/conn.go
```diff
@@ -46,11 +46,17 @@ func newConn(tr tpt.Transport, sess *webtransport.Session, sconn *connSecurityMu
 
 func (c *conn) OpenStream(ctx context.Context) (network.MuxedStream, error) {
 	str, err := c.session.OpenStreamSync(ctx)
+	if err != nil {
+		return nil, err
+	}
 	return &stream{str}, err
```
Jorropo, line 52:
```suggestion
	return &stream{str}, nil
```

p2p/transport/webtransport/conn.go
```diff
@@ -46,11 +46,17 @@ func newConn(tr tpt.Transport, sess *webtransport.Session, sconn *connSecurityMu
...
 }
 
 func (c *conn) AcceptStream() (network.MuxedStream, error) {
 	str, err := c.session.AcceptStream(context.Background())
+	if err != nil {
+		return nil, err
+	}
 	return &stream{str}, err
```
Jorropo, line 60:
```suggestion
	return &stream{str}, nil
```

Jorropo (changes requested):

Jorropo (approved):
Except what I sent before, SGTM

Jorropo:
What I don't get is that there must be broken code elsewhere.
Seems like someone is ignoring err and or checking if the stream is nil.

marten-seemann:
> What I don't get is that there must be broken code elsewhere.
> Seems like someone is ignoring err and or checking if the stream is nil instead.

That's if the bug is upstream. If the bug is downstream, we might get a `nil, nil` return, and this fix won't work (it's still the right thing to do).

PR #1778, opened 2022-09-23
quic: speed up the stateless reset test case

marten-seemann:
This reduces the execution time of this test from ~4.6s to ~0.6s on my machine.


julian88110 (approved):
LGTM.
One side note, it might not be that the kernel is slow to release udp port, could be that go scheduling runtime delays the process of channel Close.

PR #1655, opened 2022-07-12
webrtc: add WebRTC transport

ckousik:
This PR implements the webrtc transport spec according to https://github.com/libp2p/specs/pull/412 .
The webrtc protocol for multiaddr is implemented in this PR and needs to be implemented in `go-multiaddr` prior to merging this PR. This PR also uses multibase encoded multihash for DTLS fingerprint verification after the NOISE handshake.

PR #1784, opened 2022-09-27
add a roadmap for the next 6 months

marten-seemann:
I took a first stab at our roadmap for the next half a year. This document contains that are specific to the go-libp2p stack. General libp2p project items are listed in the [roadmap in the specs repo](https://github.com/libp2p/specs/blob/master/ROADMAP.md). Sections are in no particular order.

**This document is intended as a conversation starter**. 
I intend to keep this PR open for at least a week to give plenty of time for feedback.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,5 +1,4 @@
 
-
 # go-libp2p roadmap Q4’22/Q1’23
 
 ```
@@ -8,6 +7,28 @@ Status: In Progress
 Notes: This document is still in review will may be heavily modified based on stakeholder feedback. Please add any feedback or questions in:
 https://github.com/libp2p/go-libp2p/issues/1806
 ```
+
+## About the Roadmap
+### Sections
+This document consists of two sections: [Milestones](#Milestones) and the [Roadmap Appendix](#Roadmap-Appendix)
+
+[Milestones](#Milestones) is our best educated guess (not a hard commitment) around when we plan to ship the key features.
+Where possible we've broken down a project into discrete sub-projects e.g. project "A" (⚡Handshakes at the Speed of Light) contains three sub-projects and consists of A.1, A.2, and A.3
+A project is signified as "complete" once all of it's sub-projects are shipped.
+
+The [Roadmap Appendix](#Roadmap-Appendix) section describes each project in detail and lists its sub-projects (if applicable.) Here you can find the motivation for each project and goals.
+
+We've deep-linked each section header to it's corresponding GitHub Epic. Latest information on the progress of each project there can be found in the Epics and their child issues (as we will make regular updates.)
+
+### Done criteria
+Our "Definition of Done" for projects/sub-projects that involve writing new protocols or modifications to existing ones will usually consist of the following:
+- [ ] Spec is merged and classified as "Candidate Recommendation"
+- [ ] (by virtue of the above) At least one major reference implementation exists
+- [ ] A well established testing criteria (defined at the outset of the project) is met (testing via Testground or otherwise)
+- [ ] Public documentation (on docs.libp2p) exists
+
+Others supporting projects (like testing or benchmarking) will have different criteria.
+
 
 ## 🛣️ Milestones
 ### 2022
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,58 @@
+# go-libp2p roadmap Q4’22/Q1’23
+
+## 📊 Comprehensive Metrics
+
+Why: For far too long, go-libp2p has been a black box. This has hurt us many times, by allowing trivial bugs to go undetected for a long time ([example](https://github.com/ipfs/kubo/pull/8750)). Having metrics will allow us to track the impact of performance improvements we make over time.
+
+Goal: go-libp2p ships with Grafana dashboards that allow node operators to monitor their nodes in production.
+
+How: This will look similar to how we already expose resource manager metrics. Metrics can be added incrementally for libp2p’s components. First milestone is having metrics for the swarm.
+
+Tracking issue: [https://github.com/libp2p/go-libp2p/issues/1356](https://github.com/libp2p/go-libp2p/issues/1356)
+
+## 📺 Universal Browser Connectivity
+
+Why: A huge part of “the Web” is happening inside the browser. As a universal p2p networking stack, libp2p needs to be able to offer solutions for browser users.
+
+Goal: go-libp2p ships with up-to-date WebTransport and (libp2p-) WebRTC implementations, enabled by default. This allows connections between browsers and public nodes, browsers and non-public nodes as well as two browsers.
+
+1. WebTransport: as the protocol is still under development by IETF and W3C, our implementation needs to follow. To stay up to date, we will have to move as soon as Chrome ships support for a new draft version
+2. WebRTC: while browser to public node is getting close to finalized, there’ll be a push to make the other combinations work as well
+
+## ⚡️Handshakes at the Speed of Light
+
+Why: Historically, libp2p has been very wasteful when it comes to round trips spent during connection establishment. This is slowing down our users, especially their TTFB (time to first byte) metrics.
+
+Goal: go-libp2p optimizes its handshake latency up to the point where only increasing the speed of light would lead to further speedups. In particular, this means (in chronological order of a handshake):
+
+1. cutting off the 1 RTT wasted on security protocol negotiation by including the security protocol in the multiaddr: [https://github.com/libp2p/specs/pull/353](https://github.com/libp2p/specs/pull/353)
+2. cutting off the 1 RTT wasted on muxer negotiation: [https://github.com/libp2p/specs/issues/426](https://github.com/libp2p/specs/issues/426)
+3. using 0.5-RTT data (for TLS) / a Noise Extension to ship the list of Identify protocols, cutting of 1 RTT that many protocols spend waiting on `IdentifyWait`
+
+## 🧠 Smart Dialing
+
+Why: Having a large list of transports to pick from is great. Having an advanced stack that can dial all of them is even greater. But dialing all of them at the same time wastes our, the network’s and the peer’s resources. 
+
+Goal: When given a list of multiaddrs of a peer, go-libp2p is smart enough to pick the address that results in the most performant connection (for example, preferring QUIC over TCP), while also picking the address such that maximizes the likelihood of a successful handshake.
+
+How:
+
+1. implement some kind of “Happy-Eyeballs” style prioritization among all supported transports
+2. estimation of the expected RTT of a connection based on two nodes’ IP addresses, so that Happy Eyeballs Timeouts can be set dynamically
+3. detection of blackholes, especially relevant to detect UDP (QUIC) blackholing
+
+## 🧪 Future-proof Testing
+
+Why: Having lots of transports is great, and shaving off RTTs is awesome. We need to stay backwards-compatible with legacy go-libp2p implementations, and less advanced libp2p stacks.
+
+Goal: We have cross-version and cross-implementation Testground coverage that makes sure that we are able to establish a libp2p connection between two nodes, in the expected number of RTTs. This makes sure that the optimizations don’t break compatibility, actually have the performance impact we expect them to have, and serves as a regression test in the future.
+
+## 📢 Judicious Address Advertisements
+
+Why: A node that advertises lots of addresses hurts itself. Other nodes will have to try dialing a lot of addresses before they find one that actually works, dramatically increasing handshake latencies.
+
+Goal: Nodes only advertise addresses that they are actually reachable at.
+
+How: Unfortunately, the AutoNAT protocol can’t be used to probe the reachability of any particular address (especially due to a bug in the go-libp2p implementation deployed years ago). Most likely, we need a second version of the AutoNAT protocol.
+
+Related discussion: [https://github.com/libp2p/go-libp2p/issues/1480](https://github.com/libp2p/go-libp2p/issues/1480)
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -4,7 +4,7 @@
 
 Why: For far too long, go-libp2p has been a black box. This has hurt us many times, by allowing trivial bugs to go undetected for a long time ([example](https://github.com/ipfs/kubo/pull/8750)). Having metrics will allow us to track the impact of performance improvements we make over time.
 
-Goal: go-libp2p ships with Grafana dashboards that allow node operators to monitor their nodes in production.
+Goal: Export a wider set of metrics across go-libp2p components and enable node operators to monitor their nodes in production. Optionally provide a sample Grafana dashboard similar to the resource manager dashboard.
 
 How: This will look similar to how we already expose resource manager metrics. Metrics can be added incrementally for libp2p’s components. First milestone is having metrics for the swarm.
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,64 +1,118 @@
+
+
 # go-libp2p roadmap Q4’22/Q1’23
 
-## 📊 Comprehensive Metrics
+```
+Date: 2022-10-07
+Status: In Progress
+Notes: This document is still in review will may be heavily modified based on stakeholder feedback. Please add any feedback or questions in:
+https://github.com/libp2p/go-libp2p/issues/1806
+```
 
-Why: For far too long, go-libp2p has been a black box. This has hurt us many times, by allowing trivial bugs to go undetected for a long time ([example](https://github.com/ipfs/kubo/pull/8750)). Having metrics will allow us to track the impact of performance improvements we make over time.
+## 🛣️ Milestones
+### 2022
 
-Goal: Export a wider set of metrics across go-libp2p components and enable node operators to monitor their nodes in production. Optionally provide a sample Grafana dashboard similar to the resource manager dashboard.
+#### Early Q4 (October)
+-   A.1⚡Handshakes at the Speed of Light - **[Early Muxer Negotiation](https://github.com/libp2p/specs/issues/426)**
+- E.1 🧪 Future-proof testing - **RTT test for Early Muxer Negotiation**
 
-How: This will look similar to how we already expose resource manager metrics. Metrics can be added incrementally for libp2p’s components. First milestone is having metrics for the swarm.
+### Mid Q4 (November)
+-   F.1 ⏱ Future-proof Benchmarking - **[Using nix-builders](https://github.com/testground/testground/pull/1425)**
 
-Tracking issue: [https://github.com/libp2p/go-libp2p/issues/1356](https://github.com/libp2p/go-libp2p/issues/1356)
+#### End of Q4 (December)
+-   B.1 🧠 Smart Dialing - **Happy Eyeballs**
+-   C 📊 Comprehensive Metrics - **Swarm metrics**
+-   D.1 📺 Universal Browser Connectivity - **[WebRTC Browser -> Server](https://github.com/libp2p/go-libp2p/pull/1655)**
 
-## 📺 Universal Browser Connectivity
+### 2023
 
-Why: A huge part of “the Web” is happening inside the browser. As a universal p2p networking stack, libp2p needs to be able to offer solutions for browser users.
+#### Early Q1 (January)
+-   B.2 🧠 Smart Dialing - **QUIC Blackhole detector**
+-   A.2 ⚡ Handshakes at the Speed of Light - **[Adding security protocol](https://github.com/libp2p/specs/pull/353)**
 
-Goal: go-libp2p ships with up-to-date WebTransport and (libp2p-) WebRTC implementations, enabled by default. This allows connections between browsers and public nodes, browsers and non-public nodes as well as two browsers.
+#### Mid Q1 (February)
+-   B.3 🧠 Smart Dialing - **RTT estimation**
+    -   🎉 Estimated Project Completion
 
-1. WebTransport: as the protocol is still under development by IETF and W3C, our implementation needs to follow. To stay up to date, we will have to move as soon as Chrome ships support for a new draft version
-2. WebRTC: while browser to public node is getting close to finalized, there’ll be a push to make the other combinations work as well
+#### End of Q1 (March)
+-   F.2 ⏱️ Future-proof benchmarking - **Using first class support for remote runners in Testground**
+-   A.3 ⚡ Handshakes at the Speed of Light - **0.5 RTT data optimization** (for QUIC)
+    -   🎉 Estimated Project Completion
 
-## ⚡️Handshakes at the Speed of Light
+#### Up Next
+-   D.2 📺 Universal Browser Connectivity - **[WebTransport](https://github.com/libp2p/go-libp2p/issues/1717)** update to new draft versions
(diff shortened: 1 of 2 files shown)


ROADMAP.md
```diff
@@ -0,0 +1,58 @@
...
+
+Goal: go-libp2p optimizes its handshake latency up to the point where only increasing the speed of light would lead to further speedups. In particular, this means (in chronological order of a handshake):
+
+1. cutting off the 1 RTT wasted on security protocol negotiation by including the security protocol in the multiaddr: [https://github.com/libp2p/specs/pull/353](https://github.com/libp2p/specs/pull/353)
+2. cutting off the 1 RTT wasted on muxer negotiation: [https://github.com/libp2p/specs/issues/426](https://github.com/libp2p/specs/issues/426)
+3. using 0.5-RTT data (for TLS) / a Noise Extension to ship the list of Identify protocols, cutting of 1 RTT that many protocols spend waiting on `IdentifyWait`
+
+## 🧠 Smart Dialing
+
+Why: Having a large list of transports to pick from is great. Having an advanced stack that can dial all of them is even greater. But dialing all of them at the same time wastes our, the network’s and the peer’s resources.
```
Wondertan, line 34:
```suggestion
Why: Having a large list of transports to pick from is great. Having an advanced stack that can dial all of them is even greater. But dialing all of them at the same time wastes the network’s and the peer’s resources. 
```

marten-seemann: It also wastes our CPU cycles to start that many handshakes.

p-shahi:
This is awesome, my preliminary questions for now are:
1. How did we arrive at this specific list? What other items were up for consideration that did not make it here.
2. What is the priority of each roadmap item in comparison to another? We should come up with the priority here. I can create a PR to this branch.
3. Can I create tracking issues for the items that don't have them?

ROADMAP.md
```diff
@@ -0,0 +1,58 @@
+# go-libp2p roadmap Q4’22/Q1’23
+
+## 📊 Comprehensive Metrics
+
+Why: For far too long, go-libp2p has been a black box. This has hurt us many times, by allowing trivial bugs to go undetected for a long time ([example](https://github.com/ipfs/kubo/pull/8750)). Having metrics will allow us to track the impact of performance improvements we make over time.
+
+Goal: go-libp2p ships with Grafana dashboards that allow node operators to monitor their nodes in production.
+
+How: This will look similar to how we already expose resource manager metrics. Metrics can be added incrementally for libp2p’s components. First milestone is having metrics for the swarm.
```
p-shahi, line 9: Is the goal here to actually check in a Grafana dashboard json model to this repo and direct people to use that for monitoring? or only to slowly add metrics to go-libp2p components and enable users of go-libp2p to craft their own Grafana dashboards

marten-seemann: The goal is to have a ready-to-use Grafana dashboard _somewhere_. Of course, users are free to ignore our dashboards and build whatever they like from the metrics we expose.

Not sure if checking in the json file here is the most user-friendly way, especially as the number of json files will grow quickly. I don't think we need to figure out the best way to publish those right now, that can be part of the work we're planning to do.

p-shahi: Yeah that makes sense. Other projects usually have community members that create their own dashboards for use by others. I'm thinking of something like this: https://github.com/dguess/mina-node-dashboard

> Not sure if checking in the json file here is the most user-friendly way

I agree, "go-libp2p ships with Grafana dashboards" just threw me off. Thanks for the clarification.

marten-seemann:
> How did we arrive at this specific list? 

We didn't, I did ;) I just wrote down what I think would be the most valuable to work on in the next half a year. So this is just my opinion.

This list is intended as a starting point for the roadmap conversation, and shouldn't be viewed as a conclusion that was reached behind closed doors (it really isn't, I made up this entire list without asking for anybody's opinion on it). I'm happy to add and remove items based on feedback on this PR.

> What other items were up for consideration that did not make it here

Everything below "Peer Routing Records" (including) from our [general libp2p roadmap](https://github.com/libp2p/specs/blob/master/ROADMAP.md#evolve) didn't make it. In addition to that, I've not included the part of Protocol Select that would replace multistream (https://github.com/libp2p/specs/pull/349), since we're getting most of the benefit of the Protocol Select proposal by just doing the address change and the early muxer negotiation.

> Are there any tech debt items that should also be a part of this list?

I was debating putting flaky tests on the list. We'll also engage with any bug reports and community PRs that come in, but I'm not sure if that needs a roadmap item or if it's just understood that we'll spend a certain amount of time on those.
One other item that we'll likely need to tackle is transport performance: making sure all of our transports perform well (as compared to plain HTTP for example). Not sure if that's tech debt or should be its own section.

> What is the priority of each roadmap item in comparison to another? We should come up with the priority here. I can create a PR to this branch if you like. This way we can create milestones for go-libp2p

We'd need to assign priorities to sub-items. For example, the early muxer negotiation from the Handshake section has higher priority than the item right on top (putting the security protocol in the multiaddr). How would we capture that?

> Is feasible it is to achieve all of this in the next 6 months?

That's a good question. We have to keep in mind that we're losing 1 - 1.5 months due to Lab Week, Thanksgiving and Xmas. My hope is that we can at least make significant progress on all of these points, but I might be too optimistic here.

ROADMAP.md
marten-seemann: Any suggestion how to rephrase?

ROADMAP.md
```diff
@@ -0,0 +1,58 @@
+# go-libp2p roadmap Q4’22/Q1’23
```
julian88110, line 1: Great write up! 
A general thought is that as libp2p is getting more widely deployed maybe we also need to think about beef up more security features.  May or may not fit in to the next 6 months time frame.
For example behavior based security. The peer is usually not backed by a root of trust, so the traditional identity and chain of trust based approach does not work in this scenario. We may need to consider new approaches to enforce security.

ROADMAP.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. implement some kind of “Happy-Eyeballs” style prioritization among all supported transports
+2. estimation of the expected RTT of a connection based on two nodes’ IP addresses, so that Happy Eyeballs Timeouts can be set dynamically
+3. detection of blackholes, especially relevant to detect UDP (QUIC) blackholing
+
+## 🧪 Future-proof Testing
+
+Why: Having lots of transports is great, and shaving off RTTs is awesome. We need to stay backwards-compatible with legacy go-libp2p implementations, and less advanced libp2p stacks.
+
+Goal: We have cross-version and cross-implementation Testground coverage that makes sure that we are able to establish a libp2p connection between two nodes, in the expected number of RTTs. This makes sure that the optimizations don’t break compatibility, actually have the performance impact we expect them to have, and serves as a regression test in the future.
```
julian88110, line 48: Also thinking about adding the item to convert time-delay based test cases into event sync based test cases if it fits into the time frame, that will reduce a lot of the flakiness.

ROADMAP.md
```diff
@@ -0,0 +1,58 @@
+# go-libp2p roadmap Q4’22/Q1’23
+
+## 📊 Comprehensive Metrics
+
+Why: For far too long, go-libp2p has been a black box. This has hurt us many times, by allowing trivial bugs to go undetected for a long time ([example](https://github.com/ipfs/kubo/pull/8750)). Having metrics will allow us to track the impact of performance improvements we make over time.
+
+Goal: go-libp2p ships with Grafana dashboards that allow node operators to monitor their nodes in production.
```
p-shahi, line 7:
```suggestion
Goal:  Export a wider set of metrics across go-libp2p components and enable node operators to monitor their nodes in production. Optionally provide a sample Grafana dashboards similar to the [rust-libp2p dashboard](https://kademlia-exporter.max-inden.de/d/Pfr0Fj6Mk/rust-libp2p?orgId=1&refresh=30s)
```

ROADMAP.md
p-shahi: I added my suggestion https://github.com/libp2p/go-libp2p/pull/1784/files#r981794624

ROADMAP.md
marten-seemann: @julian88110 Do you have anything specific in mind?
Would that belong into the go-libp2p roadmap, or would that first and foremost be an item for the general libp2p roadmap (https://github.com/libp2p/specs/blob/master/ROADMAP.md)?

ROADMAP.md
marten-seemann, line 7:
```suggestion
Goal: Export a wider set of metrics across go-libp2p components and enable node operators to monitor their nodes in production. Optionally provide a sample Grafana dashboard similar to the resource manager dashboard.
```

ROADMAP.md
marten-seemann: Definitely worth doing. That would probably be part of the general project maintenance / tech debt point that @p-shahi was asking for in https://github.com/libp2p/go-libp2p/pull/1784#issuecomment-1259955399.

ROADMAP.md
marten-seemann: Resolved by pointing to our rcmgr Grafana dashborad.

ROADMAP.md
marten-seemann: Resolving this thread, since I've reworded the goal to be more specific.

MarcoPolo:
> One other item that we'll likely need to tackle is transport performance: making sure all of our transports perform well (as compared to plain HTTP for example). Not sure if that's tech debt or should be its own section.

I'd like to see benchmarks as a section here. I think it's inline with future-proof testing as it's easy for things to get slow if you're not tracking them.

I'll probably have some other comments, but I need to think about this a bit more.

marten-seemann:
@MarcoPolo I added a benchmarking section in 60c4fce. How does that look to you?

ROADMAP.md
julian88110: Yes this should belong to the general roadmap for libp2p. I have some thoughts, may take some time to write it up after the muxer thing is done.

ROADMAP.md
```diff
@@ -0,0 +1,64 @@
...
+
+## 🧠 Smart Dialing
+
+Why: Having a large list of transports to pick from is great. Having an advanced stack that can dial all of them is even greater. But dialing all of them at the same time wastes our, the network’s and the peer’s resources. 
+
+Goal: When given a list of multiaddrs of a peer, go-libp2p is smart enough to pick the address that results in the most performant connection (for example, preferring QUIC over TCP), while also picking the address such that maximizes the likelihood of a successful handshake.
+
+How:
+
+1. implement some kind of “Happy-Eyeballs” style prioritization among all supported transports
```
BigLep, line 40:
```suggestion
1. implement some kind of “[Happy-Eyeballs](https://en.wikipedia.org/wiki/Happy_Eyeballs)” style prioritization among all supported transports
```

ROADMAP.md
```diff
@@ -0,0 +1,64 @@
...
+2. estimation of the expected RTT of a connection based on two nodes’ IP addresses, so that Happy Eyeballs Timeouts can be set dynamically
+3. detection of blackholes, especially relevant to detect UDP (QUIC) blackholing
+
+## 🧪 Future-proof Testing
+
+Why: Having lots of transports is great, and shaving off RTTs is awesome. We need to stay backwards-compatible with legacy go-libp2p implementations, and less advanced libp2p stacks.
+
+Goal: We have cross-version and cross-implementation Testground coverage that makes sure that we are able to establish a libp2p connection between two nodes, in the expected number of RTTs. This makes sure that the optimizations don’t break compatibility, actually have the performance impact we expect them to have, and serves as a regression test in the future.
```
BigLep, line 48: We talk about ensuring compability between versions and implementations, but I assume there is an extra dimension we need to add around transports (and likely encryption and muxer).

Basically want to show 

<go-libp2p, vX, QUIC> can talk with <go-libp2p, vY, QUIC>
<go-libp2p, vX, TCP> can talk with <go-libp2p, vY, TCP>
<go-libp2p, vX, WebTransport> can talk with <go-libp2p, vY, WebTranport>
etc.

BigLep (commented):
@marten-seemann: thanks for getting the ball rolling here.  This is good stuff!

@marten-seemann @p-shahi ;
I know this a dump of milestone definitions in no particular order and that there is more work happening in https://www.notion.so/pl-strflt/go-libp2p-22-Q4-23-Q1-Product-Roadmap-48bc31fd5b394a29812d6cda624fffa0 for faster iteration which will makes it way out hee.

I'll leave my current thoughts from what I'm seeing here so it's public and tied to the milestone definitions.  This is a drive by because I'm not sure when I'll be able to circle back.  I expect manh of these things are on your radar.  Don't feel the need to respond to each comment.  This is a dump while the thoughts were there for me.

1. I like how "Comprehensive Metrics" has a tracking issue.  I think that's useful for communicating if/where there are dependencies between milestone and it allows someone to follow along with progress and or ask questions.  When we settle here, I think it would be good have these for the other items.
2. It seems like some of these things are themes like "Universal Browser Connectivity" and that some items within them are milestones on their own ("WebTransport", "WebRTC").  I think having a way to index into the specific milestone within a theme is good/useful.  That seems to be what Prithvi is doing [here](https://www.notion.so/pl-strflt/Milestone-Roadmap-a90d888d6bf14b999131ea7a802c9960) with "A1.1", "B1.2". etc.   Great.
3. I know this is coming, but we ultimately need to give expected dates (month targets ideally).  A list broken out by month (sort of like what's in https://www.notion.so/pl-strflt/Milestone-Roadmap-a90d888d6bf14b999131ea7a802c9960 ) would be great.  Not required, but showing which milestones are being developed in a given month is nice as well to set expectation on when progress will begin.
4. On the surface "Judicious Address Advertisements" seems like a stretch for the next 6 months and the lowest priority.  I'll be interested to see where that lands on the calendar estimate.
5. Should we bring Futureproof testing to the forefront?  I'm viewing some it as necessary for "Handshakes at the Speed of Light" to be completed and IPDX has laid a good foundation to build on from what they did in libp2p/test-plans.  New/exciting work is still going to be happening with landing WebRTC (with LBL driving this). Part of me wonders if we'll be happy for "eating our veggies" so we have a solid foundation for 2023.

General roadmap hygiene/logistics:
1. When this gets merged, lets make sure we have a link from "libp2p vision ROADMAP"
2. Please also expose within go-libp2p README itself.
3. Make sure we have a way for users to ask questions or comment.  I worry that "open a PR to ask a question or make a suggestion" will prevent people from engaging.
4. Set expectations on when this was last edited and when it will next be reviewed.  Maybe there is a tracking issue for the next review we can link to give the community visibility.  That would also be a place for people to leave comments.

BigLep:
Followup comments are in the followup PR: https://github.com/libp2p/go-libp2p/pull/1804#pullrequestreview-1133893508

ROADMAP.md
```diff
@@ -0,0 +1,118 @@
...
+#### Up Next
+-   D.2 📺 Universal Browser Connectivity - **[WebTransport](https://github.com/libp2p/go-libp2p/issues/1717)** update to new draft versions
+    - Dependency on Chrome needs to support new draft version of WebTransport protocol (still under development by IETF and W3C)
+-  E.2 🧪 Future-proof testing - **Comprehensive testing for non-transport features**
+-  G 📢 Judicious Address Advertisements
+
+# Roadmap Appendix
+## [A.⚡️Handshakes at the Speed of Light](https://github.com/libp2p/go-libp2p/issues/1807)
+
+**Why**: Historically, libp2p has been very wasteful when it comes to round trips spent during connection establishment. This is slowing down our users, especially their TTFB (time to first byte) metrics.
```
BigLep, line 72: I agree this is true.  I would love to have more insight on the actual impact.  For example, it would be great to have anecdotes from users on how much this is an adoption blocker.

This is more general comment.  I don't think there's necessarily anything to do right now.  It's more just to push to find ways to quantify problems in terms of impact as much as possible.

I do agree this issue rolls into one of our key pillars of "performance".  TTFB performance isn't the existential threat I have heard complaints about though (vs. libp2p is slower than HTTP).  I could be tuned in here though.  Getting user input would be great.

ROADMAP.md
```diff
@@ -0,0 +1,118 @@
...
+-   A.1⚡Handshakes at the Speed of Light - **[Early Muxer Negotiation](https://github.com/libp2p/specs/issues/426)**
+- E.1 🧪 Future-proof testing - **RTT test for Early Muxer Negotiation**
+
+### Mid Q4 (November)
+-   F.1 ⏱ Future-proof Benchmarking - **[Using nix-builders](https://github.com/testground/testground/pull/1425)**
+
+#### End of Q4 (December)
+-   B.1 🧠 Smart Dialing - **Happy Eyeballs**
+-   C 📊 Comprehensive Metrics - **Swarm metrics**
+-   D.1 📺 Universal Browser Connectivity - **[WebRTC Browser -> Server](https://github.com/libp2p/go-libp2p/pull/1655)**
```
BigLep, line 46: I think the ordering should be:

-   D.1 📺 Universal Browser Connectivity - **[WebRTC Browser -> Server](https://github.com/libp2p/go-libp2p/pull/1655)**
-   B.1 🧠 Smart Dialing - **Happy Eyeballs**
-   C 📊 Comprehensive Metrics - **Swarm metrics**

Reason being:
1. We're not going to do anything to slow down the browser connectivity story.  This closes up the story as well on "all modern browsers can connect with a go-libp2p node". 
2. Smart dialing - assuming we actually hitting a breaking point for what transport to use

ROADMAP.md
```diff
@@ -0,0 +1,118 @@
...
+
+## [C. 📊 Comprehensive Metrics](https://github.com/libp2p/go-libp2p/issues/1356)
+
+**Why**: For far too long, go-libp2p has been a black box. This has hurt us many times, by allowing trivial bugs to go undetected for a long time ([example](https://github.com/ipfs/kubo/pull/8750)). Having metrics will allow us to track the impact of performance improvements we make over time.
+
+**Goal**: Export a wider set of metrics across go-libp2p components and enable node operators to monitor their nodes in production. Optionally provide a sample Grafana dashboard similar to the resource manager dashboard.
+
+**How**: This will look similar to how we already expose resource manager metrics. Metrics can be added incrementally for libp2p’s components. First milestone is having metrics for the swarm.
+
+## [D. 📺 Universal Browser Connectivity](https://github.com/libp2p/go-libp2p/issues/1811)
```
BigLep, line 100: I know the ordering of milestones is defined above. Maybe in a future PR we move the themes around to be in priority order. For example, as a theme, I think browser connectivity should be before all of these. I would likely put future proof testing at the top.

ROADMAP.md
```diff
@@ -0,0 +1,118 @@
...
+**Why**: A huge part of “the Web” is happening inside the browser. As a universal p2p networking stack, libp2p needs to be able to offer solutions for browser users.
+
+**Goal**: go-libp2p ships with up-to-date WebTransport and (libp2p-) WebRTC implementations, enabled by default. This allows connections between browsers and public nodes, browsers and non-public nodes as well as two browsers.
+
+1. [WebRTC](https://github.com/libp2p/go-libp2p/pull/1655): while browser to public node is getting close to finalized, there’ll be a push to make the other combinations work as well
+2. [WebTransport](https://github.com/libp2p/go-libp2p/issues/1717): as the protocol is still under development by IETF and W3C, our implementation needs to follow. To stay up to date, we will have to move as soon as Chrome ships support for a new draft version
+
+## [E. 🧪 Future-proof Testing](https://github.com/libp2p/go-libp2p/issues/1809)
+
+**Why**: Having lots of transports is great, and shaving off RTTs is awesome. We need to stay backwards-compatible with legacy go-libp2p implementations, and less advanced libp2p stacks.
```
BigLep, line 111: It's bigger than transports

```suggestion
**Why**: Having lots of transports, encryption mechanism, and stream muxers is great, and shaving off RTTs is awesome. We need to stay backwards-compatible with legacy go-libp2p release, other libp2p implementations, and less advanced libp2p stacks.
```

ROADMAP.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Goal**: We have cross-version and cross-implementation Testground coverage that makes sure that we are able to establish a libp2p connection between two nodes, in the expected number of RTTs. This makes sure that the optimizations don’t break compatibility, actually have the performance impact we expect them to have, and serves as a regression test in the future.
+
+1. RTT test for Early Muxer Negotiation (A.1)
```
BigLep, line 115: I'm less convinced this warrants its own milestone.  This is basically part of the done criteria for early muxer integration negotiation.  All new functionality needs to be testing correctly as part of marking it complete.  I think it should be moved up.

I think we need:

Comprehensive interoperability testing across:
libp2p implementations (already happening)
implementation versions (already happening)
transports (I don't think we're exhaustive)
security/encryption (I don't think we're exhaustive)
muxers (I don't think we're exhaustive)

That is the key thing we want I believe so we can move forward with confidence that we're not breaking anything.

ROADMAP.md
```diff
@@ -0,0 +1,118 @@
...
+Notes: This document is still in review will may be heavily modified based on stakeholder feedback. Please add any feedback or questions in:
+https://github.com/libp2p/go-libp2p/issues/1806
+```
+
+## 🛣️ Milestones
+### 2022
+
+#### Early Q4 (October)
+-   A.1⚡Handshakes at the Speed of Light - **[Early Muxer Negotiation](https://github.com/libp2p/specs/issues/426)**
+- E.1 🧪 Future-proof testing - **RTT test for Early Muxer Negotiation**
```
BigLep, line 38: Per below, I think these should be combined.

ROADMAP.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Mid Q4 (November)
+-   F.1 ⏱ Future-proof Benchmarking - **[Using nix-builders](https://github.com/testground/testground/pull/1425)**
```
BigLep, line 41: I think before we spin up benchmarking, lets get interoperability nailed.  The pieces are in place for that already with libp2p/test-plans. I discussed this more below.

ROADMAP.md
BigLep: Per comment other places I believe, I would bump smart dialing and comprehensive metrics until we're more solid on future proofing interoperability and performance.

BigLep (commented):
@p-shahi and @marten-seemann : thanks for the continued work here.  A couple of additional high level things:

1. This doesn't need to be solved now, but I would love to solve surfacing key tenets of libp2p. 
 I personally view these as security, stability (we don't regress with new versions), and performance (we don't regress between releases and there is insignificant performance difference with options like HTTP).  I know those aren't tenets we've formalized in libp2p itself and more discussion needs to be had, but maybe it's worth us stating them for the go-libp2p project itself and we can discuss hoisting them up to libp2p as a whole.

2. what are the thoughts on how we handle open comments that won't be easily resolvable?  We'll want to get this PR merged, but presumably we'll do so without incorporating all the feedback. I want us to be able to make incremental progress without needing to keep PRs open indefinitely.

p-shahi:
> what are the thoughts on how we handle open comments that won't be easily resolvable? We'll want to get this PR merged, but presumably we'll do so without incorporating all the feedback. I want us to be able to make incremental progress without needing to keep PRs open indefinitely.

Open comments will be captured in https://github.com/libp2p/go-libp2p/issues/1806 so that we can merge this PR and make incremental updates

ROADMAP.md
MarcoPolo: I believe TTFB is an important metric to both Saturn and ipfs cluster folks.

Issue #1782, opened 2022-09-27
autorelay: don't discard static relays on disconnections


juligasa:
When autorelay looses connectivity with an static relay (maybe the relay was down, or maybe the client had network switching/problems) it removes the possibility of connecting to it again. Since [handleStaticRelays](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autorelay/relay_finder.go#L106) is only called at the beginning, the only option is to restart autorelay

To reproduce it just insert this test in autorelay_test.go
```go
func TestReconnectToStaticRelays(t *testing.T) {
	var staticRelays []peer.AddrInfo
	const numStaticRelays = 1 //Since candidate selection is random, more than 1 we are not sure which one selects
	relays := make([]host.Host, 0, numStaticRelays)
	for i := 0; i < numStaticRelays; i++ {
		r := newRelay(t)
		t.Cleanup(func() { r.Close() })
		relays = append(relays, r)
		staticRelays = append(staticRelays, peer.AddrInfo{ID: r.ID(), Addrs: r.Addrs()})
	}

	h := newPrivateNode(t,
		autorelay.WithStaticRelays(staticRelays),
		autorelay.WithNumRelays(1),
	)

	defer h.Close()

	require.Eventually(t, func() bool { return numRelays(h) == 1 }, 3*time.Second, 50*time.Millisecond)

	relaysInUse := usedRelays(h)
	oldRelay := relaysInUse[0]
	for _, r := range relays {
		if r.ID() == oldRelay {
			r.Network().ClosePeer(h.ID())
		}
	}

	require.Eventually(t, func() bool { return numRelays(h) == 1 }, 3*time.Second, 100*time.Millisecond)
}
```
A possible solution would be not discarding a static relay on connection errors (and send it back to candidates) or implement a timer to periodically call [handleStaticRelay](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autorelay/relay_finder.go#L273-L288)
<details>
<summary>Version Information</summary>
<pre>

github.com/libp2p/go-libp2p
cloud.google.com/go v0.65.0
cloud.google.com/go/bigquery v1.8.0
cloud.google.com/go/datastore v1.1.0
cloud.google.com/go/pubsub v1.3.1
cloud.google.com/go/storage v1.10.0
dmitri.shuralyov.com/app/changes v0.0.0-20180602232624-0a106ad413e3
[...]

PR #1790, opened 2022-09-30
autorelay: not discarding static relays on connectivity failures

juligasa:
Static relays are a special kind of relays, since the client has spent some time bootstrappin those relays. Thats why if the client (autorelay) looses connection with such a relays, it should not be treated like regular relays (e.g. discarding them) Two cases come to mind:
1. If a static relay goes down (for maintenance purposes), typically the client has more than one static relay, so autorelay will start making slots reservations on the ones online. However, once the first static relay is up again, it should be part again of the candidate list. Current implementation, [discard a failing static relay right away](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autorelay/relay_finder.go#L152-L158).
2. If the client switchs networks or simply the network gets down for a while (not restarting the client since that would imply static relays [being rescanned](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autorelay/relay_finder.go#L106)), then the client would discard all relays, and when back up again, it won't have any static relays available (unles autorelay gets resseted)

This can be solved in two ways, either not discarding a static relay on connection errors (and send it back to candidates) or implementing a timer to periodically call [handleStaticRelay](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autorelay/relay_finder.go#L273-L288). I have implemented both, since the former responses quicker in case no 1 and the latter periodically rescues forgotten static relays in case no 2

This PR solves #1782

### DIFF
--- a/p2p/host/autorelay/options.go
+++ b/p2p/host/autorelay/options.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math/rand"
 	"time"
 
 	"github.com/libp2p/go-libp2p/core/peer"
@@ -16,10 +15,8 @@ type config struct {
 	clock      clock.Clock
 	peerSource func(ctx context.Context, num int) <-chan peer.AddrInfo
 	// minimum interval used to call the peerSource callback
-	peerMinInterval   time.Duration
-	staticRelaySource func(ctx context.Context, num int) <-chan peer.AddrInfo
-	// minimum interval used to call the staticRelaySource callback
-	staticRelayMinInterval time.Duration
+	minInterval  time.Duration
+	staticRelays []peer.AddrInfo
 	// see WithMinCandidates
 	minCandidates int
 	// see WithMaxCandidates
@@ -47,6 +44,11 @@ var defaultConfig = config{
 	maxCandidateAge: 30 * time.Minute,
 }
 
+var (
+	errStaticRelaysMinCandidates = errors.New("cannot use WithMinCandidates and WithStaticRelays")
+	errStaticRelaysPeerSource    = errors.New("cannot use WithPeerSource and WithStaticRelays")
+)
+
 // DefaultRelays are the known PL-operated v1 relays; will be decommissioned in 2022.
 var DefaultRelays = []string{
 	"/ip4/147.75.80.110/tcp/4001/p2p/QmbFgm5zan8P6eWWmeyfncR5feYEMPbht5b1FW1C37aQ7y",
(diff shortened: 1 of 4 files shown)


p2p/host/autorelay/relay_finder.go
```diff
@@ -145,18 +146,29 @@ func (rf *relayFinder) background(ctx context.Context) {
...
+					delete(rf.relays, evt.Peer)
+					rf.notifyMaybeConnectToRelay()
+					rf.notifyMaybeNeedNewCandidates()
+					push = true
+				}
+				rf.relayMx.Unlock()
+			} else if evt.Connectedness == network.Connected {
+				for _, r := range rf.conf.staticRelays {
+					if r.ID == evt.Peer {
+						rf.notifyMaybeConnectToRelay()
```
MarcoPolo, line 163: I don't follow why we need this logic now. Aren't we already connected?

MarcoPolo (changes requested):
First of, thanks a ton for the contribution and apologies for the delay in the review. Follow ups will be much faster here on out.

I don't follow why we need both strategies of trying to connect to all static relays but also keeping them in our candidates. Wouldn't keeping them in our candidates suffice for both scenarios?

Either way, I think we can do something much simpler here. What if we added a new channel similar to `peerChan` in `func findNodes` and call it `staticRelayChan`. We do the same logic (with backoffs and checking if we're at our max) and have a separate similar timer for this. This way we really only need changes in `findNode`. After the new timer fires we can fill staticRelayChan with the static relay nodes that we aren't currently connected to. Backoffs will work similar to how normal candidates are handled, and we don't risk polluting our candidates list with nodes that are unresponsive (if we hit maxCandidates of unresponsive state relay nodes we're stuck). It also keeps all the other logic the same. Candidates will keep getting removed and we don't keep unresponsive relays in the candidate list.

This is a bit like giving static relays priority when sourcing new relays.

What do you think?

p2p/host/autorelay/relay_finder.go
juligasa: That was the attempt to re-reserve slots when we are back online (when the client had intermitent network connection). In that case, the relay [would have dropped all our reservations](https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/relay.go#L581-L591) and we need to re-reserve

MarcoPolo:
> Yes, that could work. I could implement something along those lines.

Great! Very much appreciated if you do, but no expectation here. If you don't get around to it, I'll probably tackle this Thursday or Friday of this week.

juligasa:
Didn't have time at all to have it done. But I was in the process of doing the required changes. I have pushed them so you can work on top of them if you want.

p2p/host/autorelay/options.go
```diff
@@ -74,18 +72,27 @@ func init() {
...
+		if c.staticRelaySource != nil {
 			return errors.New("can't set static relays, static relays already configured")
 		}
 		c.minCandidates = len(static)
-		c.staticRelays = static
+		c.staticRelaySource = func(ctx context.Context, num int) <-chan peer.AddrInfo {
+			c.staticRelayMinInterval = time.Hour
+			rand.Seed(time.Now().UnixNano())
+			var static_candidates = static
+			rand.Shuffle(len(static_candidates), func(i, j int) {
```
MarcoPolo, line 87: Why the shuffle? Users may have a preferred order, no?

p2p/host/autorelay/options.go
```diff
@@ -74,18 +72,27 @@ func init() {
...
+				static_candidates[i], static_candidates[j] = static_candidates[j], static_candidates[i]
+			})
+			for i := 0; i < num%c.minCandidates; i++ {
```
MarcoPolo, line 90: hmm so if num == c.minCandidates we do nothing?

MarcoPolo:
I got a chance to look at this. Possibly dumb thought, but why not simple use `WithPeerSource` to supply the static relays and get rid of the separate code path to handle static relays? Essentially, static relays are just a special type of peer source: https://github.com/libp2p/go-libp2p/pull/1875

p2p/host/autorelay/options.go
juligasa: The rationale here was: if asked only for 1 or 2 candidates (`<len(static_candidates)`), and those candidates happened to be always down, then one wouldn't be able to connect to any. Randomizing the process guarantees than you will pull different candidates as longs as `len(static_candidates)>num`. But yes, if order is important, the user can just increase `num` and get more candidates if previous ones haven't succeeded.

MarcoPolo:
I merged 1875. It includes a helper for creating a peersource from static relays. It should retry static candidates automatically when we loose a candidate.

PR #1793, opened 2022-09-30
webtransport: fix comment in webtransport client auth handshake

MarcoPolo:
I think we forgot to update this comment after changing the handshake protocol.


marten-seemann (approved):
Thanks for catching this!

PR #1794, opened 2022-10-01
p2p/protocol: add ping timeout option

dbadoy:
Would be nice to be able to set a timeout to PingService. There are no changes to the existing interface.

This PR includes a small fix in holepuncher. 
- Coverage the case that doesn't call 'timer.Stop'.
https://github.com/libp2p/go-libp2p/blob/7828f3e0797e0a7b7033fa5e8be9b94f57a4c173/p2p/protocol/holepunch/holepuncher.go#L158-L160


marten-seemann (commented):
Why not just cancel the context?

dbadoy:
Controll by context in `Ping` is feels limited because couldn't go beyond the stream's timeout(1 minute). 
I think this PR is unnecessary If there is definitely no situation where you have to wait more than 1 minute.

p-shahi:
@dbadoy Do you want to close this PR then?

dbadoy:
Yup. I'll close this PR.

PR #1795, opened 2022-10-01
examples: use circuitv2 in relay example

DanielVernall:
Hi, I've never contributed to go-libp2p before and I'm new to the project having come back to it a few times. So, I thought I would start by updating examples/docs that I find need updating as I'm learning libp2p.

I couldn't find much documentation on the circuit-relay v2 implementation, but I have pieced together a sample from the source and tests I have found. The only thing I'm not certain of is the use of WithUseTransient(), as I can't find much info on this. Am I correctly using this configuration in the circuit relay stream creation, or should I be looking to upgrade the connection in some way prior to connecting via the relay?


DanielVernall:
I renamed the h1 through h3 variables to unreachableN and relay1, just to make it really clear in the example for a newcomer which would be acting as which. But I understand you may prefer h1-3 by convention as h is almost always the host name, and because in reality a relay host isn't just a relay host, it's simply another node in the network.

Also, I'm not certain of go-libp2p's naming conventions for Multiaddr and AddrInfo, as I've seen quite a few variables named in all lowercase, eg. "relayaddr" instead of mixed caps, "relayAddr", so I stuck with all lower for some variables in here, but they were quite long (not very Go like, I'm afraid).

examples/relay/main.go
```diff
@@ -20,68 +20,102 @@ func main() {
 }
 
 func run() {
-	// Create three libp2p hosts, enable relay client capabilities on all
-	// of them.
-
-	// Tell the host use relays
-	h1, err := libp2p.New(libp2p.EnableRelay())
+	// Create 2 "unreachable" libp2p hosts that want to communicate.
```
marten-seemann, line 23:
```suggestion
	// Create two "unreachable" libp2p hosts that want to communicate.
```

examples/relay/main.go
```diff
@@ -20,68 +20,102 @@ func main() {
...
+
+	// Attempt to connect the unreachable hosts directly
+	unreachable2info := peer.AddrInfo{
+		ID:    unreachable2.ID(),
+		Addrs: unreachable2.Addrs(),
+	}
+
+	err = unreachable1.Connect(context.Background(), unreachable2info)
+	if err == nil {
+		log.Printf("This actually should have failed. Did you make changes to the example...?")
```
marten-seemann, line 57:
```suggestion
		log.Printf("This actually should have failed.")
```

examples/relay/main.go
```diff
@@ -91,21 +125,30 @@ func run() {
...
-		log.Printf("Failed to connect h1 and h3: %v", err)
+	if err := unreachable1.Connect(context.Background(), unreachable2relayinfo); err != nil {
+		log.Printf("Unexpected error here. Failed to connect unreachable1 and unreachable2: %v", err)
 		return
 	}
 
+	log.Println("Yep, that worked!")
+
+	// **** MIGHT BE WORTH ADDING A COMMENT ADDRESSING WHAT WithUseTransient() is here,
+	// and why it is needed ******************
```
marten-seemann, line 145: Can you do that? :)

marten-seemann:
Looks like this PR needs to be rebased.

examples/relay/main.go
DanielVernall: I'm not really confident with transient connections, as I'm new to libp2p 😅. From what I can piece together from reading some source comments is that transient stream are those that have opened a connection but the stream is not yet fully established, as a protocol hasn't been negotiated yet. 

So, my guess would be that, in the context of a circuit relay, we have established an initial connection to the relay peer (which is marked as transient) but we don't have a connection to our destination peer yet, so when we open our stream to the destination peer we indicate that we are using a transient connection, but it is isn't a problem as the relay will connect to the destination peer and our connection will be upgraded once we've negotiated our stream with the destination peer. But it is possible I'm way off there 😁

marten-seemann: > From what I can piece together from reading some source comments is that transient stream are those that have opened a connection but the stream is not yet fully established, as a protocol hasn't been negotiated yet.

Yeah, confusing terminology here. Sorry for that. That would be transient in the context of the resource manager. In this context, it means something completely different.

Transient here means that it's a connection established on a relayed (as opposed to a direct) connection. Since the relay limits the amount of data that can be exchanged over the relayed connection, application controls need to explicitly opt-in into using a relayed connection. In general, they should only do that if they have low bandwidth requirements, and they're ok with the connection being killed when the relayed connection is replaced with a direct (holepunched) connection.

I hope this explanation makes sense.

DanielVernall: Ah that's interesting, thanks for the details. Is there anything in the docs or source I could read to explore this a bit more? For example, how an application/relay node would go about lifting the relay data limits, or upgrading the connection for devices where hole punching is not possible.

examples/go.work
```diff
@@ -0,0 +1,3 @@
+go 1.18
```
marten-seemann, line 1: I don't think we need this.

marten-seemann (changes requested):

marten-seemann (approved):
Thank you!

PR #1802, opened 2022-10-06
identify: Surface error in IdentifyConn and flaky TestIdentifyPushOnAddrChange

MarcoPolo:
Changes `IdentifyWait` to return a `chan error` instead of a `chan struct{}`. This lets us surface errors to the caller.

This is technically a breaking api change, but I think it is safe because before users would ignore the return value, so they can still ignore the return value and get the same behavior. Callers who are interested if this identify failed can now use the returned error.

The first caller that cares is our flaky `TestIdentifyPushOnAddrChange`. This should tell us why this is failing and give us next steps to fix the flakiness.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -66,8 +66,8 @@ type IDService interface {
 	IdentifyConn(network.Conn)
 	// IdentifyWait triggers an identify (if the connection has not already been
 	// identified) and returns a channel that is closed when the identify protocol
-	// completes.
-	IdentifyWait(network.Conn) <-chan struct{}
+	// completes. Returns any error encountered in the channel.
+	IdentifyWait(network.Conn) <-chan error
 	// OwnObservedAddrs returns the addresses peers have reported we've dialed from
 	OwnObservedAddrs() []ma.Multiaddr
 	// ObservedAddrsFor returns the addresses peers have reported we've dialed from,
@@ -98,7 +98,7 @@ type idService struct {
 
 	// Identified connections (finished and in progress).
 	connsMu sync.RWMutex
-	conns   map[network.Conn]chan struct{}
+	conns   map[network.Conn]chan error
 
 	addrMu sync.Mutex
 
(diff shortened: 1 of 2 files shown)


p2p/protocol/identify/id.go
```diff
@@ -333,6 +334,7 @@ func (ids *idService) IdentifyWait(c network.Conn) <-chan struct{} {
 			if err := ids.identifyConn(c); err != nil {
```
marten-seemann, line 334: Closing the channel seems a bit dangerous to misuse, as it invites users to receive from the channel multiple times. Only in the first case they'll get the error, after that they'll get a `nil`.

Would it be safe to send a `nil` on the channel when no error occurs (without closing)?

MarcoPolo: I'm not sure I follow the reasoning. What's wrong with closing the channel? We can send a nil in case of no error and then close (which is about the same as closing without sending a nil first). 

If a user is receiving multiple times without a closed channel they'll have a waiting recv that will never get completed, right? That seems like a caller bug.

marten-seemann: > What's wrong with closing the channel? We can send a nil in case of no error and then close (which is about the same as closing without sending a nil first).

That's correct. The problem is the error case, where the error is only received once, and after that it _looks_ as if no error occurred (as a receiving from a closed channel returns the zero value).

Not closing would be safer: The API contract would tell users that they can only expect to receive from this channel once.

MarcoPolo: A user could check if the chan is closed with `x, ok := <-c`.

Maybe what we're really saying here is that we wish there was a way to encode that "this channel will return at most one error and then close". And because we can't we're forced to pick two options:

1. Close the channel, and maybe cause a caller to recv multiple times and get nil.
2. Don't close the channel, and maybe cause a caller to block forever on a second recv that will never come.

I think neither maybe case is worth worrying about, and I think you'd prefer the 2nd option if we had to worry about it.

Is that correct? I don't have a strong preference here, so happy to do 2. I just wanted to understand your reasoning.

MarcoPolo: Because of the above later callers may not get the error. I need to rethink this

Issue #1789, opened 2022-09-30
Tracking Issue: early muxer selection


marten-seemann:
eta: 2022Q4

Discussed with @julian88110 yesterday: We'll target the completion of this feature for the v0.24 release. The TLS part is almost done, and it is expected that the Noise part can be completed in the next 1-2 weeks.

Done criteria:
* [x] early muxer selection when using TLS (#1772)
* [x] early muxer selection when using Noise (https://github.com/libp2p/go-libp2p/pull/1813)
* [x] https://github.com/libp2p/go-libp2p/issues/1854
* [x] https://github.com/libp2p/go-libp2p/issues/1836
* [x] double check that the [Testground cross-version test](https://github.com/libp2p/test-plans/blob/master/ping/_compositions/go-cross-versions.toml) hits these new code paths
* [x] spec merged: https://github.com/libp2p/specs/pull/446
* [x] docs: https://github.com/libp2p/docs/issues/213
* [ ] #1889

marten-seemann:
Closing this issue, as this is done. The remaining work is captured by #1889.

PR #1813, opened 2022-10-07
noise: use Noise Extension to negotiate the muxer during the handshake

julian88110:
Muxer negotiation using early data of Noise handshake. 
This is the Noise implementation of issue #1789

### DIFF
--- a/p2p/security/noise/session.go
+++ b/p2p/security/noise/session.go
@@ -40,6 +40,9 @@ type secureSession struct {
 	prologue []byte
 
 	initiatorEarlyDataHandler, responderEarlyDataHandler EarlyDataHandler
+
+	// ConnectionState holds state information releated to the secureSession entity.
+	connectionState network.ConnectionState
 }
 
 // newSecureSession creates a Noise session over the given insecureConn Conn, using
@@ -110,7 +113,7 @@ func (s *secureSession) RemotePublicKey() crypto.PubKey {
 }
 
 func (s *secureSession) ConnState() network.ConnectionState {
-	return network.ConnectionState{}
+	return s.connectionState
 }
 
 func (s *secureSession) SetDeadline(t time.Time) error {
@@ -127,4 +130,11 @@ func (s *secureSession) SetWriteDeadline(t time.Time) error {
 
 func (s *secureSession) Close() error {
 	return s.insecureConn.Close()
+}
+
+func SessionWithConnState(s *secureSession, muxer string) *secureSession {
+	if s != nil {
+		s.connectionState.NextProto = muxer
+	}
+	return s
 }
--- a/p2p/security/noise/transport.go
+++ b/p2p/security/noise/transport.go
@@ -16,9 +16,8 @@ import (
 
 // ID is the protocol ID for noise
 const (
-	ID               = "/noise"
-	MaxExtensionSize = 2048
-	MaxProtoNum      = 100
+	ID          = "/noise"
+	maxProtoNum = 100
 )
 
 var _ sec.SecureTransport = &Transport{}
@@ -54,7 +53,7 @@ func New(privkey crypto.PrivKey, muxers []protocol.ID) (*Transport, error) {
 // SecureInbound runs the Noise handshake as the responder.
 // If p is empty, connections from any peer are accepted.
 func (t *Transport) SecureInbound(ctx context.Context, insecure net.Conn, p peer.ID) (sec.SecureConn, error) {
-	responderEDH := NewTransportEDH(t)
+	responderEDH := newTransportEDH(t)
 	c, err := newSecureSession(t, ctx, insecure, p, nil, nil, responderEDH, false)
 	if err != nil {
 		addr, maErr := manet.FromNetAddr(insecure.RemoteAddr())
@@ -67,7 +66,7 @@ func (t *Transport) SecureInbound(ctx context.Context, insecure net.Conn, p peer
 
 // SecureOutbound runs the Noise handshake as the initiator.
 func (t *Transport) SecureOutbound(ctx context.Context, insecure net.Conn, p peer.ID) (sec.SecureConn, error) {
-	initiatorEDH := NewTransportEDH(t)
+	initiatorEDH := newTransportEDH(t)
 	c, err := newSecureSession(t, ctx, insecure, p, nil, initiatorEDH, nil, true)
 	if err != nil {
 		return c, err
@@ -106,7 +105,9 @@ type transportEarlyDataHandler struct {
 	receivedMuxers []string
 }
 
-func NewTransportEDH(t *Transport) *transportEarlyDataHandler {
+var _ EarlyDataHandler = &transportEarlyDataHandler{}
+
+func newTransportEDH(t *Transport) *transportEarlyDataHandler {
 	return &transportEarlyDataHandler{transport: t}
 }
 
@@ -119,7 +120,7 @@ func (i *transportEarlyDataHandler) Send(context.Context, net.Conn, peer.ID) *pb
 
 func (i *transportEarlyDataHandler) Received(_ context.Context, _ net.Conn, extension *pb.NoiseExtensions) error {
 	// Discard messages with size or the number of protocols exceeding extension limit for security.
-	if extension != nil && extension.XXX_Size() <= MaxExtensionSize && len(extension.StreamMuxers) <= MaxProtoNum {
+	if extension != nil && len(extension.StreamMuxers) <= maxProtoNum {
 		i.receivedMuxers = extension.GetStreamMuxers()
 	}
 	return nil
--- a/p2p/security/noise/transport.go
+++ b/p2p/security/noise/transport.go
@@ -85,19 +85,14 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
 }
 
 func matchMuxers(initiatorMuxers, responderMuxers []string) string {
-	selectedMuxer := ""
 	for _, muxer := range responderMuxers {
 		for _, initMuxer := range initiatorMuxers {
 			if initMuxer == muxer {
-				selectedMuxer = muxer
-				break
+				return muxer
 			}
 		}
-		if selectedMuxer != "" {
-			break
-		}
 	}
-	return selectedMuxer
+	return ""
 }
 
 type transportEarlyDataHandler struct {
@@ -113,8 +108,7 @@ func newTransportEDH(t *Transport) *transportEarlyDataHandler {
 
 func (i *transportEarlyDataHandler) Send(context.Context, net.Conn, peer.ID) *pb.NoiseExtensions {
 	return &pb.NoiseExtensions{
-		WebtransportCerthashes: [][]byte{},
-		StreamMuxers:           i.transport.muxers,
+		StreamMuxers: i.transport.muxers,
 	}
 }
 
--- a/p2p/security/noise/transport.go
+++ b/p2p/security/noise/transport.go
@@ -7,13 +7,19 @@ import (
 	"github.com/libp2p/go-libp2p/core/canonicallog"
 	"github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/protocol"
 	"github.com/libp2p/go-libp2p/core/sec"
+	"github.com/libp2p/go-libp2p/p2p/security/noise/pb"
 
 	manet "github.com/multiformats/go-multiaddr/net"
 )
 
 // ID is the protocol ID for noise
-const ID = "/noise"
+const (
+	ID               = "/noise"
+	MaxExtensionSize = 2048
+	MaxProtoNum      = 100
+)
 
 var _ sec.SecureTransport = &Transport{}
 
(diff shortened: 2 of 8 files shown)


marten-seemann:
@julian88110 Could you rebase (`git rebase`) this PR on top of the current `master`? Looks like there are still commits from your earlier TLS change in the commit history here.

p2p/security/noise/session.go
```diff
@@ -55,6 +58,7 @@ func newSecureSession(tpt *Transport, ctx context.Context, insecure net.Conn, re
 		prologue:                  prologue,
 		initiatorEarlyDataHandler: initiatorEDH,
 		responderEarlyDataHandler: responderEDH,
+		connectionState:           network.ConnectionState{NextProto: ""},
```
marten-seemann, line 61: No need to initialize here, this is the default value.

p2p/security/noise/session.go
```diff
@@ -128,3 +132,10 @@ func (s *secureSession) SetWriteDeadline(t time.Time) error {
 func (s *secureSession) Close() error {
 	return s.insecureConn.Close()
 }
+
+func SessionWithConState(s *secureSession, muxer string) *secureSession {
```
marten-seemann, line 136:
```suggestion
func SessionWithConnState(s *secureSession, muxer string) *secureSession {
```

p2p/security/noise/transport.go
```diff
@@ -7,13 +7,19 @@ import (
...
 
 	manet "github.com/multiformats/go-multiaddr/net"
 )
 
 // ID is the protocol ID for noise
-const ID = "/noise"
+const (
+	ID                 = "/noise"
+	MAX_EXTENSION_SIZE = 2048
+	MAX_PROTO_NUM      = 100
```
marten-seemann, line 21: Please stick with Go variable naming conventions.

p2p/security/noise/transport.go
```diff
@@ -65,3 +82,50 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
 	}
 	return st, nil
 }
+
+func MatchMuxers(initiatorMuxers, responderMuxers []string) string {
```
marten-seemann, line 86: Does this need to be exported?

p2p/security/noise/transport.go
```diff
@@ -65,3 +82,50 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
...
+	respMuxerMap := make(map[string]struct{}, len(responderMuxers))
```
marten-seemann, line 87: This has the best scaling behavior for n -> inf, but our n is limited by the protobuf size to a few hundred values (and in practice, will never be more than a single-digit number).

I think we can live with the O(n^2) scaling of just iterating over both lists. This makes this code easier to read, and it avoids the allocation associated with creating a map.

p2p/security/noise/transport.go
```diff
@@ -65,3 +82,50 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
...
+	for _, muxer := range initiatorMuxers {
+		if _, ok := respMuxerMap[muxer]; ok {
+			selectedMuxer = muxer
+			break
+		}
+	}
+	return selectedMuxer
+}
+
+type TransportEarlyDataHandler struct {
```
marten-seemann, line 102: Does this need to be exported?

p2p/security/noise/transport.go
```diff
@@ -65,3 +82,50 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
...
+func (i *TransportEarlyDataHandler) Send(context.Context, net.Conn, peer.ID) *pb.NoiseExtensions {
+	return &pb.NoiseExtensions{
+		WebtransportCerthashes: [][]byte{},
+		StreamMuxers:           i.transport.muxers,
+	}
+}
+
+func (i *TransportEarlyDataHandler) Received(_ context.Context, _ net.Conn, extension *pb.NoiseExtensions) error {
+	// Discard messages with size or the number of protocols exceeding extension limit for security.
+	if extension != nil && extension.XXX_Size() <= MAX_EXTENSION_SIZE && len(extension.StreamMuxers) <= MAX_PROTO_NUM {
```
marten-seemann, line 120: We need to check the extension size _before_ we parse it.

p2p/security/noise/transport.go
```diff
@@ -65,3 +82,50 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
...
+	}
+	return selectedMuxer
+}
+
+type TransportEarlyDataHandler struct {
+	transport      *Transport
+	receivedMuxers []string
+}
+
+func NewTransportEDH(t *Transport) *TransportEarlyDataHandler {
```
marten-seemann, line 107: Have you considered reusing the `SessionTransport`?

p2p/security/noise/session.go
julian88110: done

p2p/security/noise/session.go
julian88110: OK, thanks for the tip. going with default.

p2p/security/noise/transport.go
julian88110: This is a private utility function that is used by the transport EDH only, so I don't see a need to export. Any other considerations?

p2p/security/noise/transport.go
julian88110: My understanding is tht go constant name convention is all capital separated by underscores, do we have some different convention we adopt in this project?

p2p/security/noise/transport.go
julian88110: we don't have to.

p2p/security/noise/transport.go
julian88110: yes actually my first version of the implementation tried that approach of "withOptions". Then it becomes obvious that there are needs to do condition checks and customizations, my take is that the current approach is cleaner and will introduce less interference with webtransport code.

p2p/security/noise/transport.go
marten-seemann: I don't think that's the convention. Look at https://pkg.go.dev/net/http#pkg-constants for example.

p2p/security/noise/transport.go
marten-seemann: At the time this function is called, the protobuf is already parsed (otherwise we wouldn't have a `pb.NoiseExtensions`).

p2p/security/noise/transport.go
julian88110: I see, changed the names. thanks!

p2p/security/noise/transport.go
julian88110: I see, you mean check the size before the whole handshake message is parsed, right? 
That is more general a case that relates to the whole handshake process.  We can either piggy-back the handshaek message size check in this PR or add it as a separate PR for easier tracking. I am OK with either option.

p2p/security/noise/transport.go
```diff
@@ -65,3 +84,50 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
...
+func (i *transportEarlyDataHandler) Send(context.Context, net.Conn, peer.ID) *pb.NoiseExtensions {
+	return &pb.NoiseExtensions{
+		WebtransportCerthashes: [][]byte{},
+		StreamMuxers:           i.transport.muxers,
+	}
+}
+
+func (i *transportEarlyDataHandler) Received(_ context.Context, _ net.Conn, extension *pb.NoiseExtensions) error {
+	// Discard messages with size or the number of protocols exceeding extension limit for security.
+	if extension != nil && extension.XXX_Size() <= MaxExtensionSize && len(extension.StreamMuxers) <= MaxProtoNum {
```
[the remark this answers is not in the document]
julian88110: Good, I think it somehow works as a defense. At least it is size bounded.

p2p/security/noise/transport.go
```diff
@@ -65,3 +84,50 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
...
+	}
+	return selectedMuxer
+}
+
+type transportEarlyDataHandler struct {
+	transport      *Transport
+	receivedMuxers []string
+}
+
+func NewTransportEDH(t *Transport) *transportEarlyDataHandler {
```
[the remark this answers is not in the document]
julian88110: done.

p2p/security/noise/transport.go
```diff
@@ -65,3 +84,50 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
...
+			}
+		}
+		if selectedMuxer != "" {
+			break
+		}
+	}
+	return selectedMuxer
+}
+
+type transportEarlyDataHandler struct {
```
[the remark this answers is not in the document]
julian88110: Yes, I saw that in libp2p code base a lot, just forgot to apply here, thanks for catching it!

p2p/security/noise/transport.go
```diff
@@ -22,38 +28,51 @@ var _ sec.SecureTransport = &Transport{}
 type Transport struct {
 	localID    peer.ID
 	privateKey crypto.PrivKey
+	muxers     []string
 }
 
 // New creates a new Noise transport using the given private key as its
 // libp2p identity key.
-func New(privkey crypto.PrivKey) (*Transport, error) {
+func New(privkey crypto.PrivKey, muxers []protocol.ID) (*Transport, error) {
```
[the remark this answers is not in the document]
julian88110: I tried to reuse SessionTransport in the first round, but found out it introduces coupling between webtransport and tcp , and needs some awkward condition checks and customization, which makes the solution less attractive than the current one in here, but let me take another look to see if we can reuse it.

p2p/security/noise/transport.go
```diff
@@ -7,13 +7,19 @@ import (
...
 
 	manet "github.com/multiformats/go-multiaddr/net"
 )
 
 // ID is the protocol ID for noise
-const ID = "/noise"
+const (
+	ID               = "/noise"
+	MaxExtensionSize = 2048
+	MaxProtoNum      = 100
```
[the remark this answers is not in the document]
julian88110: done

p2p/security/noise/transport.go
julian88110: as per discussion on size check with uint16, this is resolved.

p2p/security/noise/transport.go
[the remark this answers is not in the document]
julian88110: I reviewed the approach in the context of the overall solution for Muxer negotiation. The current solution still seem to be the cleaner one. 
  1. It needs to customize the SecureConnection obtained by the SecureInbound or SecureOutbound functions. If we customize the above functions for SessionTransport, it introduces an awkward coupling and introduces more cases in webtransport workflow.
  2.  On the other hand customizing the Transport object does not interfere with other cases, and introduces less changes. 
  
What do you think?

p2p/security/noise/transport.go
```diff
@@ -65,3 +83,52 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
 	}
 	return st, nil
 }
+
+func matchMuxers(initiatorMuxers, responderMuxers []string) string {
+	selectedMuxer := ""
+	for _, muxer := range responderMuxers {
+		for _, initMuxer := range initiatorMuxers {
+			if initMuxer == muxer {
+				selectedMuxer = muxer
```
marten-seemann, line 92: Why not return here?

p2p/security/noise/transport.go
```diff
@@ -65,3 +83,52 @@ func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTranspo
...
+
+var _ EarlyDataHandler = &transportEarlyDataHandler{}
+
+func newTransportEDH(t *Transport) *transportEarlyDataHandler {
+	return &transportEarlyDataHandler{transport: t}
+}
+
+func (i *transportEarlyDataHandler) Send(context.Context, net.Conn, peer.ID) *pb.NoiseExtensions {
+	return &pb.NoiseExtensions{
+		WebtransportCerthashes: [][]byte{},
```
marten-seemann, line 116: No need to initialize to zero value.

marten-seemann (commented):
This is looking really good. Just a few nits!

p2p/security/noise/transport.go
julian88110: yes, it can return, it will be simpler to return, thanks!

p2p/security/noise/transport.go
julian88110: done

marten-seemann (commented):
Code LGTM.

Something seems to be wrong with the Protobuf though. This PR adds a single entry to the Noise extension. This shouldn't lead to a 537 LOC change in the generated `.pb.go` file.

julian88110:
> 
This might be related to proto buf compiler version difference, I used the latest protbuf 21.8 and protoc-gen-gogo/1.3.2_2/ to compile the protobuf file, I do notice it changed a lot of the generated file content, even the imported modules are different, indicating a significant code generation change.

marten-seemann:
> This might be related to proto buf compiler version difference, I used the latest protbuf 21.8 and protoc-gen-gogo/1.3.2_2/ to compile the protobuf file, I do notice it changed a lot of the generated file content, even the imported modules are different, indicating a significant code generation change.

I'm not opposed to switching to a different protobuf compiler, but we should be intentional about it, and consistent across the code base. We should also use (Unified) CI to check that all of our `.pb.go` files are up to date.

I've recompiled the Protobuf here, so this PR can make progress.

marten-seemann (approved):

julian88110:
> > This might be related to proto buf compiler version difference, I used the latest protbuf 21.8 and protoc-gen-gogo/1.3.2_2/ to compile the protobuf file, I do notice it changed a lot of the generated file content, even the imported modules are different, indicating a significant code generation change.
> 
> I'm not opposed to switching to a different protobuf compiler, but we should be intentional about it, and consistent across the code base. We should also use (Unified) CI to check that all of our `.pb.go` files are up to date.
> 
> I've recompiled the Protobuf here, so this PR can make progress.

Thanks Marten.
Agreed, that was why I asked for the protobuf compiler versions some time ago. Instead of tracking it off record, We should have something like a tool chain config file or  dev-env setup config to specify our tool chains and their versions.

PR #1817, opened 2022-10-09
basichost: don't wait for Identify before returning host.Connect

marten-seemann:


marten-seemann:
This is messy. The `libp2p.Host` doesn't expose `IDService`, as the `BasicHost` does, so it's not clear how users would wait for identify to finish. Furthermore, `Connect` doesn't return a connection, which would be required so what they could pass it into `IdentifyWait`.

PR #1823, opened 2022-10-12
noise: add an option to allow unknown peer ID in SecureOutbound 

ckousik:
Allows the `SecureOutbound` call to be called with an unknown peer ID (empty string). This is required for https://github.com/libp2p/specs/pull/412#discussion_r977698490

### DIFF
--- a/p2p/security/noise/handshake.go
+++ b/p2p/security/noise/handshake.go
@@ -272,12 +272,14 @@ func (s *secureSession) handleRemoteHandshakePayload(payload []byte, remoteStati
 		return nil, err
 	}
 
-	// check the peer ID for:
-	// * all connections, if we know the peer ID to which we want to connect.
-	if s.remoteID != "" && s.remoteID != id {
-		// use Pretty() as it produces the full b58-encoded string, rather than abbreviated forms.
+	// check the peer ID if enabled
+	if s.checkPeerID && s.remoteID != id {
 		return nil, fmt.Errorf("peer id mismatch: expected %s, but remote key matches %s", s.remoteID.Pretty(), id.Pretty())
 	}
+	// if (s.initiator && s.remoteID != id) || (!s.initiator && s.remoteID != "" && s.remoteID != id) {
+	// 	// use Pretty() as it produces the full b58-encoded string, rather than abbreviated forms.
+	// 	return nil, fmt.Errorf("peer id mismatch: expected %s, but remote key matches %s", s.remoteID.Pretty(), id.Pretty())
+	// }
 
 	// verify payload is signed by asserted remote libp2p key.
 	sig := nhp.GetIdentitySig()
--- a/p2p/security/noise/handshake.go
+++ b/p2p/security/noise/handshake.go
@@ -276,10 +276,6 @@ func (s *secureSession) handleRemoteHandshakePayload(payload []byte, remoteStati
 	if s.checkPeerID && s.remoteID != id {
 		return nil, fmt.Errorf("peer id mismatch: expected %s, but remote key matches %s", s.remoteID.Pretty(), id.Pretty())
 	}
-	// if (s.initiator && s.remoteID != id) || (!s.initiator && s.remoteID != "" && s.remoteID != id) {
-	// 	// use Pretty() as it produces the full b58-encoded string, rather than abbreviated forms.
-	// 	return nil, fmt.Errorf("peer id mismatch: expected %s, but remote key matches %s", s.remoteID.Pretty(), id.Pretty())
-	// }
 
 	// verify payload is signed by asserted remote libp2p key.
 	sig := nhp.GetIdentitySig()
--- a/p2p/security/noise/handshake.go
+++ b/p2p/security/noise/handshake.go
@@ -273,9 +273,8 @@ func (s *secureSession) handleRemoteHandshakePayload(payload []byte, remoteStati
 	}
 
 	// check the peer ID for:
-	// * all outbound connection
-	// * inbound connections, if we know which peer we want to connect to (SecureInbound called with a peer ID)
-	if (s.initiator && s.remoteID != id) || (!s.initiator && s.remoteID != "" && s.remoteID != id) {
+	// * all connections, if we know the peer ID to which we want to connect.
+	if s.remoteID != "" && s.remoteID != id {
 		// use Pretty() as it produces the full b58-encoded string, rather than abbreviated forms.
 		return nil, fmt.Errorf("peer id mismatch: expected %s, but remote key matches %s", s.remoteID.Pretty(), id.Pretty())
 	}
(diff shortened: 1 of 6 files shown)


marten-seemann:
Can we make this a separate function, that has a very clear warning of the risks. Dialing nodes without a peer ID makes you susceptible to MITM attacks, see discussion in https://github.com/libp2p/specs/issues/458.

ckousik:
@marten-seemann Could you take another look at this?

p2p/security/noise/handshake.go
```diff
@@ -272,12 +272,14 @@ func (s *secureSession) handleRemoteHandshakePayload(payload []byte, remoteStati
...
-	if s.remoteID != "" && s.remoteID != id {
-		// use Pretty() as it produces the full b58-encoded string, rather than abbreviated forms.
+	// check the peer ID if enabled
+	if s.checkPeerID && s.remoteID != id {
 		return nil, fmt.Errorf("peer id mismatch: expected %s, but remote key matches %s", s.remoteID.Pretty(), id.Pretty())
 	}
+	// if (s.initiator && s.remoteID != id) || (!s.initiator && s.remoteID != "" && s.remoteID != id) {
+	// 	// use Pretty() as it produces the full b58-encoded string, rather than abbreviated forms.
+	// 	return nil, fmt.Errorf("peer id mismatch: expected %s, but remote key matches %s", s.remoteID.Pretty(), id.Pretty())
+	// }
```
marten-seemann, line 282: No commented code please.

marten-seemann (changes requested):
This seems quite complicated. Have you considered my suggestion of making this a separate method, as I suggested in my last review?

ckousik:
> This seems quite complicated. Have you considered my suggestion of making this a separate method, as I suggested in my last review?

@marten-seemann I considered it but wanted to adhere strictly to the `SecureTransport` interface. Adding that will not be difficult since `noise.Transport` is already exposed.

p2p/security/noise/transport.go
```diff
@@ -39,9 +39,9 @@ func New(privkey crypto.PrivKey) (*Transport, error) {
 }
 
 // SecureInbound runs the Noise handshake as the responder.
-// If p is empty, connections from any peer are accepted.
+// if p is empty accept any peer ID.
```
marten-seemann, line 42: This comment was correct, no need to change it.

p2p/security/noise/session_transport.go
```diff
@@ -77,5 +77,12 @@ func (i *SessionTransport) SecureInbound(ctx context.Context, insecure net.Conn,
 
 // SecureOutbound runs the Noise handshake as the initiator.
 func (i *SessionTransport) SecureOutbound(ctx context.Context, insecure net.Conn, p peer.ID) (sec.SecureConn, error) {
-	return newSecureSession(i.t, ctx, insecure, p, i.prologue, i.initiatorEarlyDataHandler, i.responderEarlyDataHandler, true)
+	return newSecureSession(i.t, ctx, insecure, p, i.prologue, i.initiatorEarlyDataHandler, i.responderEarlyDataHandler, true, true)
+}
+
+// SecureOutboundForAnyPeerID runs the Noise handshake as the initiator but does not check
+// the remote's peer ID. This is the outbound equivalent of calling `SecureInbound` with an empty
+// peer ID.
```
marten-seemann, line 85: Can you add that this allows MITM attacks?

marten-seemann (commented):
I'm not sure I understand why `SecureOutboundForAnyPeerID` is needed on the `Transport` and on the `SessionTransport`.

This PR would also need to be rebased, after recent changes to our Noise transport.

ckousik:
@marten-seemann I want to keep the interface consistent between `SessionTransport` and `Transport`. We also use `SessionTransport` in webrtc for the noise prologue.

marten-seemann:
Sorry for the long response time here, and for changing my mind. I think we should indeed go with an option, implementing this method on both the `Transport` and the `SessionTransport` doesn't seem very nice.
To follow Go conventions (the default should be secure), let's not name it `CheckPeerID(enable bool)` though. What about `DisablePeerIDCheck()`?

p2p/security/noise/transport.go
```diff
@@ -67,15 +67,15 @@ func (t *Transport) SecureInbound(ctx context.Context, insecure net.Conn, p peer
...
 	if err != nil {
 		return c, err
 	}
 	return SessionWithConnState(c, initiatorEDH.MatchMuxers(true)), err
 }
 
-func (t *Transport) WithSessionOptions(opts ...SessionOption) (sec.SecureTransport, error) {
-	st := &SessionTransport{t: t}
+func (t *Transport) WithSessionOptions(opts ...SessionOption) (*SessionTransport, error) {
+	st := &SessionTransport{t: t, disablePeerIdCheck: false}
```
marten-seemann, line 78:
```suggestion
	st := &SessionTransport{t: t}
```

p2p/security/noise/session_transport.go
```diff
@@ -50,22 +50,34 @@ func EarlyData(initiator, responder EarlyDataHandler) SessionOption {
...
+	disablePeerIdCheck bool
 
 	initiatorEarlyDataHandler, responderEarlyDataHandler EarlyDataHandler
 }
 
 // SecureInbound runs the Noise handshake as the responder.
 // If p is empty, connections from any peer are accepted.
 func (i *SessionTransport) SecureInbound(ctx context.Context, insecure net.Conn, p peer.ID) (sec.SecureConn, error) {
-	c, err := newSecureSession(i.t, ctx, insecure, p, i.prologue, i.initiatorEarlyDataHandler, i.responderEarlyDataHandler, false)
+	c, err := newSecureSession(i.t, ctx, insecure, p, i.prologue, i.initiatorEarlyDataHandler, i.responderEarlyDataHandler, false, p != "" || i.disablePeerIdCheck)
```
marten-seemann, line 80: @ckousik Why are you setting `checkPeerID = true` when `i.disalbePeerIdCheck`?

p2p/security/noise/session_transport.go
```diff
@@ -50,22 +50,34 @@ func EarlyData(initiator, responder EarlyDataHandler) SessionOption {
...
 var _ sec.SecureTransport = &SessionTransport{}
 
 // SessionTransport can be used
 // to provide per-connection options
 type SessionTransport struct {
 	t *Transport
 	// options
-	prologue []byte
+	prologue           []byte
+	disablePeerIdCheck bool
```
marten-seemann, line 72:
```suggestion
	disablePeerIDCheck bool
```

marten-seemann (changes requested):

p2p/security/noise/session_transport.go
ckousik: This is to keep behaviour consistent with setting an empty string as peer id.

marten-seemann: Why not `p != "" || !i.disablePeerIDCheck`?

ckousik: I added a quick test and the correct conditional here was `p != "" && !i.disablePeerIDCheck`, since on the inbound call we will only check the peer ID if the ID provided in the argument is not empty, and the ID check is not disabled.

PR #1832, opened 2022-10-17
webtransport: use the rcmgr to control flow control window increases

### LINKED ISSUE #1826: webtransport: use the resource manager to control flow control window increases
Moving from https://github.com/marten-seemann/go-libp2p-webtransport/issues/8.

---

We should use the resource manager to control flow control window increases (similar to https://github.com/libp2p/go-libp2p/blob/d8c4f163a398106a7efd5fca1d3adaa2324573f5/p2p/transport/quic/conn.go#L48-L50).

marten-seemann:
Fixes #1826.

Depends on https://github.com/marten-seemann/webtransport-go/pull/35.

### DIFF
--- a/p2p/transport/webtransport/conn.go
+++ b/p2p/transport/webtransport/conn.go
@@ -27,7 +27,7 @@ func (c *connMultiaddrs) RemoteMultiaddr() ma.Multiaddr { return c.remote }
 type conn struct {
 	*connSecurityMultiaddrs
 
-	transport tpt.Transport
+	transport *transport
 	session   *webtransport.Session
 
 	scope network.ConnScope
@@ -35,7 +35,7 @@ type conn struct {
 
 var _ tpt.CapableConn = &conn{}
 
-func newConn(tr tpt.Transport, sess *webtransport.Session, sconn *connSecurityMultiaddrs, scope network.ConnScope) *conn {
+func newConn(tr *transport, sess *webtransport.Session, sconn *connSecurityMultiaddrs, scope network.ConnScope) *conn {
 	return &conn{
 		connSecurityMultiaddrs: sconn,
 		transport:              tr,
@@ -60,7 +60,18 @@ func (c *conn) AcceptStream() (network.MuxedStream, error) {
 	return &stream{str}, nil
 }
 
-func (c *conn) Close() error             { return c.session.Close() }
+func (c *conn) allowWindowIncrease(size uint64) bool {
+	return c.scope.ReserveMemory(int(size), network.ReservationPriorityMedium) == nil
+}
+
+// Close closes the connection.
+// It must be called even if the peer closed the connection in order for
+// garbage collection to properly work in this package.
+func (c *conn) Close() error {
+	c.transport.removeConn(c.session)
+	return c.session.Close()
+}
+
 func (c *conn) IsClosed() bool           { return c.session.Context().Err() != nil }
 func (c *conn) Scope() network.ConnScope { return c.scope }
 func (c *conn) Transport() tpt.Transport { return c.transport }
--- a/p2p/transport/webtransport/transport_test.go
+++ b/p2p/transport/webtransport/transport_test.go
@@ -1,6 +1,7 @@
 package libp2pwebtransport_test
 
 import (
+	"bytes"
 	"context"
 	"crypto/ecdsa"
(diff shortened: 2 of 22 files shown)


p2p/transport/webtransport/conn.go
```diff
@@ -60,7 +60,18 @@ func (c *conn) AcceptStream() (network.MuxedStream, error) {
 	return &stream{str}, nil
 }
 
-func (c *conn) Close() error             { return c.session.Close() }
+func (c *conn) allowWindowIncrease(size uint64) bool {
+	return c.scope.ReserveMemory(int(size), network.ReservationPriorityMedium) == nil
```
MarcoPolo, line 64: What's the argument for Medium vs High? Medium bounds us to 60% of available memory. So a 4GiB node with defaults would have 0.6(512+128MiB)=384MiB here. I think that sounds fine, just want to check though

p2p/transport/webtransport/transport_test.go
```diff
@@ -582,5 +587,130 @@ func TestSNIIsSent(t *testing.T) {
...
+		str, err := conn.AcceptStream()
+		require.NoError(t, err)
+		_, err = io.CopyBuffer(str, str, make([]byte, 2<<10))
+		require.NoError(t, err)
+		str.CloseWrite()
+	}()
+
+	proxy, err := quicproxy.NewQuicProxy("localhost:0", &quicproxy.Opts{
+		RemoteAddr:  ln.Addr().String(),
+		DelayPacket: func(quicproxy.Direction, []byte) time.Duration { return rtt / 2 },
```
MarcoPolo, line 646: Cool! TIL

p2p/transport/webtransport/transport_test.go
```diff
@@ -582,5 +587,130 @@ func TestSNIIsSent(t *testing.T) {
...
+	conn, err := tr2.Dial(context.Background(), addr, serverID)
+	require.NoError(t, err)
+	str, err := conn.OpenStream(context.Background())
+	require.NoError(t, err)
+	var increasesDone uint32 // to be used atomically
+	go func() {
+		for {
+			_, err := str.Write(bytes.Repeat([]byte{0x42}, 1<<10))
+			require.NoError(t, err)
+			if atomic.LoadUint32(&increasesDone) > 0 {
```
MarcoPolo, line 680: nit: maybe easier to use a select over a done semaphore?
```suggestion
			select {
			    case <-increasesDone:
			        return
			    default:
			}
```

MarcoPolo (approved):

p2p/transport/webtransport/conn.go
marten-seemann: When we're running low on memory, we want to block allocations that are not strictly necessary. Performance will be slightly degraded if you don't get the full receive window, but things will still work. Whereas if you fail to allocate the buffer needed for parsing a protobuf (for example), things will break in a lot worse ways.

p2p/transport/webtransport/transport_test.go
marten-seemann: This will be nicer once we can use the `atomic.Bool` that was introduced in Go 1.19: https://pkg.go.dev/sync/atomic#Bool.

p2p/transport/quic/tracer_metrics.go
```diff
@@ -267,8 +267,8 @@ func (m *metricsConnTracer) SentPacket(hdr *logging.ExtendedHeader, size logging
 	sentPackets.WithLabelValues(m.getEncLevel(logging.PacketTypeFromHeader(&hdr.Header))).Inc()
 }
 
-func (m *metricsConnTracer) ReceivedVersionNegotiationPacket(hdr *logging.Header, v []logging.VersionNumber) {
-	bytesTransferred.WithLabelValues("rcvd").Add(float64(hdr.ParsedLen() + logging.ByteCount(4*len(v))))
+func (m *metricsConnTracer) ReceivedVersionNegotiationPacket(dst, src logging.ArbitraryLenConnectionID, v []logging.VersionNumber) {
+	bytesTransferred.WithLabelValues("rcvd").Add(1 + 4 + 2 + float64(dst.Len()+src.Len()) + float64(4*len(v)))
```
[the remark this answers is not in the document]
marten-seemann: Resolved in #1838.

Issue #1824, opened 2022-10-12
webtransport: deterministically generate certificates, to survive host reboots


marten-seemann:
Copied from https://github.com/marten-seemann/go-libp2p-webtransport/issues/2. I can't move that issue since it's not in the libp2p org.

---

We can generate a (cryptographically secure) (pseudo-)random value from the host's private key as an input to our certificate generation mechanism.
Furthermore, we can set the validity period of the certificate such that it is independent from the boot time of the node.

This would allow us to generate certificates with deterministic (but still unpredictable) hashes, which means that our multiaddr (which includes the certificate hash) would survive reboots of the node, even if all state (except for the private key) is lost.

marten-seemann:
> We can generate a (cryptographically secure) (pseudo-)random value from the host's private key as an input to our certificate generation mechanism.

Adding a bit more context here: We already derive a key from the host key for quic-go's stateless reset key in https://github.com/libp2p/go-libp2p/blob/63b8803005ecca56db249a636453e19158499957/p2p/transport/quic/transport.go#L198-L200

We should probably use the same construction here (with a different `salt` / `info`)

MarcoPolo:
https://pkg.go.dev/crypto/ecdsa

> Signatures generated by this package are not deterministic, but entropy is mixed with the private key and the message, achieving the same level of security in case of randomness source failure.

I wish I saw that sooner.

We're using ECDSA secp256 for the cert key and signature since it's the interopable default.

> The exact list of allowed public key algorithms used in the Subject Public Key Info field (and, as a consequence, in the TLS CertificateVerify message) is [implementation-defined](https://infra.spec.whatwg.org/#implementation-defined); however, it MUST include ECDSA with the secp256r1 (NIST P-256) named group ([[RFC3279]](https://www.w3.org/TR/webtransport/#biblio-rfc3279), Section 2.3.5; [[RFC8422]](https://www.w3.org/TR/webtransport/#biblio-rfc8422)) to provide an interoperable default. It MUST NOT contain RSA keys ([[RFC3279]](https://www.w3.org/TR/webtransport/#biblio-rfc3279), Section 2.3.1).
 
(from https://www.w3.org/TR/webtransport/)

marten-seemann:
Good catch! That's annoying, and I wish there was a way to turn this off.

[...]

PR #1833, opened 2022-10-17
webtransport: use deterministic TLS certificates

MarcoPolo:
Closes https://github.com/libp2p/go-libp2p/issues/1824

### DIFF
--- a/p2p/transport/webtransport/cert_manager.go
+++ b/p2p/transport/webtransport/cert_manager.go
@@ -95,11 +95,11 @@ func (m *certManager) init(hostKey ic.PrivKey) error {
 	// We want to add a random offset to each start time so that not all certs
 	// rotate at the same time across the network. The offset represents moving
 	// the bucket start time some `offset` earlier.
-	offset := (time.Duration(binary.LittleEndian.Uint16(pubkeyBytes)) * time.Minute) % certValidity
+	offset := (time.Duration(binary.LittleEndian.Uint16(pubkeyBytes)) * time.Minute) % (certValidity - clockSkewAllowance)
 
 	// We want our offset to be at least clockSkewAllowance so that the
 	// certificate has at least been valid for one hour.
-	offset = (offset + clockSkewAllowance) % certValidity
+	offset = offset + clockSkewAllowance
 
 	startTime := getCurrentBucketStartTime(start, offset)
 	m.nextConfig, err = newCertConfig(hostKey, startTime, startTime.Add(certValidity))
--- a/p2p/transport/webtransport/cert_manager.go
+++ b/p2p/transport/webtransport/cert_manager.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/benbjohnson/clock"
+	ic "github.com/libp2p/go-libp2p/core/crypto"
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/multiformats/go-multihash"
 )
@@ -26,8 +27,8 @@ type certConfig struct {
 func (c *certConfig) Start() time.Time { return c.tlsConf.Certificates[0].Leaf.NotBefore }
 func (c *certConfig) End() time.Time   { return c.tlsConf.Certificates[0].Leaf.NotAfter }
 
-func newCertConfig(start, end time.Time) (*certConfig, error) {
-	conf, err := getTLSConf(start, end)
+func newCertConfig(key ic.PrivKey, start, end time.Time) (*certConfig, error) {
+	conf, err := getTLSConf(key, start, end)
 	if err != nil {
 		return nil, err
 	}
@@ -57,32 +58,50 @@ type certManager struct {
 	serializedCertHashes [][]byte
 }
 
-func newCertManager(clock clock.Clock) (*certManager, error) {
+func newCertManager(hostKey ic.PrivKey, clock clock.Clock) (*certManager, error) {
 	m := &certManager{clock: clock}
 	m.ctx, m.ctxCancel = context.WithCancel(context.Background())
-	if err := m.init(); err != nil {
+	if err := m.init(hostKey); err != nil {
 		return nil, err
 	}
 
-	m.background()
+	m.background(hostKey)
 	return m, nil
 }
 
-func (m *certManager) init() error {
+type timeBuckets struct {
+	current time.Time
+	next    time.Time
+}
+
+// getTimeBuckets returns a canonical set of times that are bucketed in ranges
+// of certValidity since unix epoch.  This lets you get the same time ranges
+// across reboots without having to persist state.
+func getTimeBuckets(now time.Time) timeBuckets {
+	currentBucket := now.UnixMilli() / certValidity.Milliseconds()
+	return timeBuckets{
+		current: time.UnixMilli((currentBucket) * certValidity.Milliseconds()),
+		next:    time.UnixMilli((currentBucket + 1) * certValidity.Milliseconds()),
+	}
+
+}
+
+func (m *certManager) init(hostKey ic.PrivKey) error {
 	start := m.clock.Now().Add(-clockSkewAllowance)
+	timeBuckets := getTimeBuckets(start)
 	var err error
-	m.nextConfig, err = newCertConfig(start, start.Add(certValidity))
+	m.nextConfig, err = newCertConfig(hostKey, timeBuckets.current, timeBuckets.next)
 	if err != nil {
 		return err
 	}
-	return m.rollConfig()
+	return m.rollConfig(hostKey)
 }
 
-func (m *certManager) rollConfig() error {
+func (m *certManager) rollConfig(hostKey ic.PrivKey) error {
 	// We stop using the current certificate clockSkewAllowance before its expiry time.
 	// At this point, the next certificate needs to be valid for one clockSkewAllowance.
 	nextStart := m.nextConfig.End().Add(-2 * clockSkewAllowance)
-	c, err := newCertConfig(nextStart, nextStart.Add(certValidity))
+	c, err := newCertConfig(hostKey, nextStart, nextStart.Add(certValidity))
 	if err != nil {
 		return err
 	}
@@ -95,7 +114,7 @@ func (m *certManager) rollConfig() error {
 	return m.cacheAddrComponent()
 }
 
-func (m *certManager) background() {
+func (m *certManager) background(hostKey ic.PrivKey) {
 	d := m.currentConfig.End().Add(-clockSkewAllowance).Sub(m.clock.Now())
 	log.Debugw("setting timer", "duration", d.String())
 	t := m.clock.Timer(d)
@@ -111,7 +130,7 @@ func (m *certManager) background() {
 				return
 			case now := <-t.C:
 				m.mx.Lock()
-				if err := m.rollConfig(); err != nil {
+				if err := m.rollConfig(hostKey); err != nil {
 					log.Errorw("rolling config failed", "error", err)
 				}
 				d := m.currentConfig.End().Add(-clockSkewAllowance).Sub(now)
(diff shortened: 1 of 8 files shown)


p2p/transport/webtransport/crypto_test.go
```diff
@@ -130,3 +133,68 @@ func TestCertificateVerification(t *testing.T) {
 		})
 	}
 }
+
+func TestDeterministicCertHashes(t *testing.T) {
+	// Run this test 1000 times since we want to make sure the signatures are deterministic
+	runs := 1000
+	for i := 0; i < runs; i++ {
+		t.Run(fmt.Sprintf("Run=%d", i), func(t *testing.T) {
+			zeroSeed := [32]byte{}
```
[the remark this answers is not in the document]
MarcoPolo: shouldn't matter. The test is to check if the signing process itself is non-deterministic. This seed is fed into the hkdf so the signing process is oblivious to this.

Imagine the signing process changes in the future and it randomly reads a byte from `/dev/urandom` to mix in 10% of the time. Then a single run would succeed 81% of the time. Doing more runs decreases the chance of success iteratively (0.81*0.81*...). The 1000 runs is to detect if there's any chance on non-determinism.

p2p/transport/webtransport/cert_manager.go
```diff
@@ -57,32 +58,50 @@ type certManager struct {
...
 		return nil, err
 	}
 
-	m.background()
+	m.background(hostKey)
 	return m, nil
 }
 
-func (m *certManager) init() error {
+type timeBuckets struct {
```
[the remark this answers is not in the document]
MarcoPolo: The new struct is to provide a useful type signature. It's also nice to have this function here so we can test it.

p2p/transport/webtransport/cert_manager_test.go
```diff
@@ -77,7 +86,7 @@ func TestCertRenewal(t *testing.T) {
 		}
 		return false
 	}, 100*time.Millisecond, 10*time.Millisecond)
-	cl.Add(2 * time.Second)
+	cl.Add(clockSkewAllowance + 2*time.Second)
```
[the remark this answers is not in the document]
MarcoPolo: It's very subtle! It's here: https://github.com/libp2p/go-libp2p/pull/1833/files#diff-6396cfe3758a5a20ba975f8cc2876388c171949cb5bac4419f31ebfa26a2c3a0R93

p2p/transport/webtransport/cert_manager_test.go
```diff
@@ -100,3 +109,47 @@ func TestCertRenewal(t *testing.T) {
 	// check that the 2nd certificate from the beginning was rolled over to be the 1st certificate
 	require.Equal(t, second[1].Value(), third[0].Value())
 }
+
+func TestDeterministicCertsAcrossReboots(t *testing.T) {
```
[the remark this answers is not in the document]
MarcoPolo: Do you mean we should move this test somewhere else or another test that’s an integration test?

p2p/transport/webtransport/cert_manager_test.go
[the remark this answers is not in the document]
MarcoPolo: Changed!

p2p/transport/webtransport/cert_manager_test.go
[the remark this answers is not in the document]
MarcoPolo: Added

p2p/transport/webtransport/cert_manager.go
```diff
@@ -57,32 +60,60 @@ type certManager struct {
...
 
-func (m *certManager) init() error {
-	start := m.clock.Now().Add(-clockSkewAllowance)
-	var err error
-	m.nextConfig, err = newCertConfig(start, start.Add(certValidity))
+// getCurrentTimeBucket returns the canonical start time of the given time are bucketed by
+// ranges of certValidity since unix epoch. This lets you get the same time
+// ranges across reboots without having to persist state.  timeBuckets represent
+// our current time bbucket and our next time bucket.
+// These overlap by 2*clock skew.
```
marten-seemann, line 78: This comment needs to be updated.

p2p/transport/webtransport/cert_manager.go
```diff
@@ -57,32 +60,60 @@ type certManager struct {
...
+	start := m.clock.Now()
+	pubkeyBytes, err := hostKey.GetPublic().Raw()
+	if err != nil {
+		return err
+	}
+
+	// We want to add a random offset to each start time so that not all certs
+	// rotate at the same time across the network. The offset represents moving
+	// the bucket start time some `offset` earlier.
+	offset := (time.Duration(binary.LittleEndian.Uint16(pubkeyBytes)) * time.Minute) % (certValidity - clockSkewAllowance)
```
marten-seemann, line 98: Why is this module certValidity minus a single clockSkewAllowance?

p2p/transport/webtransport/cert_manager_test.go
```diff
@@ -39,14 +43,16 @@ func certHashFromComponent(t *testing.T, comp ma.Component) []byte {
...
+	require.NoError(t, err)
+	m, err := newCertManager(priv, cl)
 	require.NoError(t, err)
 	defer m.Close()
 
 	conf := m.GetConfig()
 	require.Len(t, conf.Certificates, 1)
 	cert := conf.Certificates[0]
-	require.Equal(t, cl.Now().Add(-clockSkewAllowance).UTC(), cert.Leaf.NotBefore)
+	require.GreaterOrEqual(t, cl.Now(), cert.Leaf.NotBefore)
```
marten-seemann, line 55: Why not test for skew here?

p2p/transport/webtransport/cert_manager_test.go
```diff
@@ -59,7 +65,10 @@ func TestInitialCert(t *testing.T) {
 
 func TestCertRenewal(t *testing.T) {
 	cl := clock.NewMock()
-	m, err := newCertManager(cl)
+	cl.Set(time.UnixMilli(0))
```
marten-seemann, line 68: We probably want to avoid the first epoch here.

p2p/transport/webtransport/cert_manager_test.go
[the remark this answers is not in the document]
marten-seemann: Do we still need the skew here?

p2p/transport/webtransport/cert_manager_test.go
```diff
@@ -100,3 +110,66 @@ func TestCertRenewal(t *testing.T) {
 	// check that the 2nd certificate from the beginning was rolled over to be the 1st certificate
 	require.Equal(t, second[1].Value(), third[0].Value())
 }
+
+func TestDeterministicCertsAcrossReboots(t *testing.T) {
```
marten-seemann, line 114: Do we need to skip on OSX?

p2p/transport/webtransport/cert_manager_test.go
```diff
@@ -100,3 +110,66 @@ func TestCertRenewal(t *testing.T) {
...
+	// Run this test 100 times to make sure it's deterministic
+	runs := 100
+	for i := 0; i < runs; i++ {
+		t.Run(fmt.Sprintf("Run=%d", i), func(t *testing.T) {
+			cl := clock.NewMock()
+			priv, _, err := test.RandTestKeyPair(crypto.Ed25519, 256)
```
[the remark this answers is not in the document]
marten-seemann: Not sure I understand. Isn’t there the chance we hit the offset for any key?

p2p/transport/webtransport/transport_test.go
```diff
@@ -584,3 +592,192 @@ func TestSNIIsSent(t *testing.T) {
...
+		if err != nil {
+			return false
+		}
+		l, err := tr.Listen(ma.StringCast("/ip4/127.0.0.1/udp/0/quic/webtransport"))
+		if err != nil {
+			return false
+		}
+		defer l.Close()
+
+		getLogWriter, err := utils.GetQLOGWriter()
```
marten-seemann, line 629: Leftover debug code?

p2p/transport/webtransport/transport_test.go
```diff
@@ -584,3 +592,192 @@ func TestSNIIsSent(t *testing.T) {
 	}
 
 }
+
+func TestServerSendsBackValidCert(t *testing.T) {
+	var maxTimeoutErrors = 10
+
+	require.NoError(t, quick.Check(func(timeSinceUnixEpoch time.Duration, keySeed int64, randomClientSkew time.Duration) bool {
```
marten-seemann, line 599: Do we need to repeat this test 100 time or so?

p2p/transport/webtransport/cert_manager_test.go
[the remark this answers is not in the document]
MarcoPolo: Kind of. In this test we have this setup
```
 v-- start here-ish
|---------|
        |---------|
```

with `cl.Set(m.currentConfig.End().Add(-(2*clockSkewAllowance + time.Second)))`, we advance the clock:

```
        v-- clock is here now (minus one second)
|---------|
        |---------|
```


Remember the transition point is one clock skew before the end.
```
        v-- clock is here now (minus one second)
|---------|
         | <- transition point
        |---------|
```

So we need to move one clock skew forward and a couple seconds
```
         v-- clock is here now (plus one second)
|---------|
         | <- transition point
        |---------|
```

--- 

Maybe this makes more sense if I change it to only advance the clock to one clock skew before the end?
with `cl.Set(m.currentConfig.End().Add(-(clockSkewAllowance + time.Second)))`, we advance the clock:

p2p/transport/webtransport/cert_manager_test.go
MarcoPolo: The MacOS issue was around listening, this should be okay (after using a deterministic seed, since otherwise we have random offsets and adding an hour might get us over the border).

p2p/transport/webtransport/transport_test.go
MarcoPolo: That's the default: https://pkg.go.dev/testing/quick?utm_source=gopls#Config

p2p/transport/webtransport/cert_manager_test.go
MarcoPolo: added

p2p/transport/webtransport/cert_manager_test.go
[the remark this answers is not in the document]
MarcoPolo: This test makes sure that if we reboot one hour in the future then we get the same key assuming we aren't near the boundary. The problem with the random key is that it means we have a random offset (offset is defined by the key). So by setting the key to a predetermined one we can uphold the assumption. The zero seed is just one such key, but any key that doesn't put us near the bounds is fine.

marten-seemann:
Looks like one of these tests is failing:
```
 === RUN   TestServerSendsBackValidCert
  2022/10/22 14:53:07 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
  Error: CRYPTO_ERROR (0x12a): Times are not valid: server_now=2062-11-24 13:43:26.165 +0000 UTC client_now=2062-11-24 13:25:30.7370715 +0000 UTC certstart=2062-11-24 13:30:00 +0000 UTC certend=2062-12-08 13:30:00 +0000 UTC
      transport_test.go:598: 
          	Error Trace:	/home/runner/work/go-libp2p/go-libp2p/p2p/transport/webtransport/transport_test.go:598
          	            	#54: failed on input -6053665406165266670, -6699227244707765831, -642676675427928500
          	Test:       	TestServerSendsBackValidCert
  --- FAIL: TestServerSendsBackValidCert (0.76s)

p2p/transport/webtransport/cert_manager.go
MarcoPolo: I was including the clockSkew allowance into the offset, but that was wrong (and also tricky). I've done the simpler thing of subtracting the `start` I pass to `getCurrentBucketStartTime` by `clockSkewAllowance`. This works because `getCurrentBucketStartTime` will never give a time later than `start + certValidity - 2*clockSkew`.

marten-seemann (approved):

PR #1835, opened 2022-10-21
routed host: search for new multi addresses upon connect failure

dennis-tra:
For context: https://github.com/ipfs/kubo/issues/9264

---

If we cannot connect to a remote host, check if we can find more up-to-date Multiaddresses in the DHT. If we have found at least one additional Multiaddress, attempt to connect again.

### DIFF
--- a/p2p/host/routed/routed.go
+++ b/p2p/host/routed/routed.go
@@ -113,9 +113,9 @@ func (rh *RoutedHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
 		// up-to-date addresses for the given peer. If there
 		// are addresses we didn't know about previously, we
 		// try to connect again.
-		newAddrs, addrErr := rh.findPeerAddrs(ctx, pi.ID)
-		if addrErr != nil {
-			return err // return outer error
+		newAddrs, err := rh.findPeerAddrs(ctx, pi.ID)
+		if err != nil {
+			return fmt.Errorf("failed to find peers: %w", err)
 		}
 
 		// Build lookup map
--- a/p2p/host/routed/routed.go
+++ b/p2p/host/routed/routed.go
@@ -107,7 +107,39 @@ func (rh *RoutedHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
 
 	// if we're here, we got some addrs. let's use our wrapped host to connect.
 	pi.Addrs = addrs
-	return rh.host.Connect(ctx, pi)
+	err := rh.host.Connect(ctx, pi)
+	if err != nil {
+		// We couldn't connect. Let's check if we have the most
+		// up-to-date addresses for the given peer. If there
+		// are addresses we didn't know about previously, we
+		// try to connect again.
+		newAddrs, addrErr := rh.findPeerAddrs(ctx, pi.ID)
+		if addrErr != nil {
+			return err // return outer error
+		}
+
+		// Build lookup map
+		lookup := make(map[ma.Multiaddr]struct{}, len(addrs))
+		for _, addr := range addrs {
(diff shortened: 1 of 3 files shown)


p2p/host/routed/routed.go
```diff
@@ -107,7 +107,39 @@ func (rh *RoutedHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
...
-	return rh.host.Connect(ctx, pi)
+	err := rh.host.Connect(ctx, pi)
+	if err != nil {
+		// We couldn't connect. Let's check if we have the most
+		// up-to-date addresses for the given peer. If there
+		// are addresses we didn't know about previously, we
+		// try to connect again.
+		newAddrs, addrErr := rh.findPeerAddrs(ctx, pi.ID)
+		if addrErr != nil {
+			return err // return outer error
```
dennis-tra, line 118: @marten-seemann which error would you return?

marten-seemann: I think we should return `addrErr` here. Ideally wrapped with `fmt.Errorf`.

p2p/host/routed/routed_test.go
```diff
@@ -0,0 +1,73 @@
...
+	callCount  int
+	findPeerFn func(ctx context.Context, id peer.ID) (peer.AddrInfo, error)
+}
+
+func (m *mockRouting) FindPeer(ctx context.Context, pid peer.ID) (peer.AddrInfo, error) {
+	m.callCount += 1
+	return m.findPeerFn(ctx, pid)
+}
+
+func TestRoutedHost_Connect_obsoleteAddresses(t *testing.T) {
```
marten-seemann, line 27:
```suggestion
func TestRoutedHostConnectToObsoleteAddresses(t *testing.T) {
```

p2p/host/routed/routed_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+var _ Routing = (*mockRouting)(nil)
+
+type mockRouting struct {
+	callCount  int
+	findPeerFn func(ctx context.Context, id peer.ID) (peer.AddrInfo, error)
+}
+
+func (m *mockRouting) FindPeer(ctx context.Context, pid peer.ID) (peer.AddrInfo, error) {
+	m.callCount += 1
```
marten-seemann, line 23:
```suggestion
	m.callCount++
```

p2p/host/routed/routed.go
dennis-tra: You're thinking of something like?

```go
return fmt.Errorf("%s: %s", err, addrErr)
```

p2p/host/routed/routed.go
```diff
@@ -107,7 +107,39 @@ func (rh *RoutedHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
...
-	return rh.host.Connect(ctx, pi)
+	err := rh.host.Connect(ctx, pi)
+	if err != nil {
+		// We couldn't connect. Let's check if we have the most
+		// up-to-date addresses for the given peer. If there
+		// are addresses we didn't know about previously, we
+		// try to connect again.
+		newAddrs, addrErr := rh.findPeerAddrs(ctx, pi.ID)
+		if addrErr != nil {
+			return fmt.Errorf("%s: %s", err, addrErr)
```
marten-seemann, line 118: This error message doesn't make a lot of sense. What about this:
```go
return fmt.Errorf("failed to find peers: %w", addrErr)
```

You can then also rename the `addrErr` to `err`.

dennis-tra: done

marten-seemann (approved):

Issue #1836, opened 2022-10-21
Create integration test for libp2p early muxer negotiation


julian88110:
The code changes that are specified in [#446] changes muxer negotiation behavior, an integration test plan should be put in  place to test the correct muxer negotiation behavior.

[#446]:  https://github.com/libp2p/specs/pull/446/files

marten-seemann:
Specifically, we should test that if the client has muxer preference `a > b` and the server has muxer preference `b < a`, TLS, Noise and multistream negotiate the same muxer.

PR #1840, opened 2022-10-27
Muxer early negotiation integration test

julian88110:
Integration test for early muxer negotiation in security handshake.

### DIFF
--- a/p2p/test/muxer-negotiation/muxer_negotiation_test.go
+++ b/p2p/test/muxer-negotiation/muxer_negotiation_test.go
@@ -77,7 +77,7 @@ func TestMuxerNegotiatin(t *testing.T) {
 		}()
 
 		<-ready
-		runClientAndCheckMuxer(t, cctx, clih, getHostAddress(svrh), cSec, sSec, expected)
+		runClientAndCheckMuxer(t, cctx, clih, getHostAddress(t, svrh), cSec, sSec, expected)
 		clih.Close()
 		svrh.Close()
 		cCancel()
@@ -116,15 +116,12 @@ func makeHost(t *testing.T, transportType string, muxers []MuxerEntity, port int
 	return secTrans, h, err
 }
 
-func getHostAddress(ha host.Host) *peer.AddrInfo {
+func getHostAddress(t *testing.T, ha host.Host) *peer.AddrInfo {
 	// Build host multiaddress
 	hostAddr := ma.StringCast(fmt.Sprintf("/p2p/%s", ha.ID().Pretty()))
 	addr := ha.Addrs()[0]
 	addrInfo, err := peer.AddrInfoFromString(addr.Encapsulate(hostAddr).String())
-	if err != nil {
-		log.Fatal("Failed to get address info ", err)
-		return nil
-	}
+	require.NoError(t, err)
 	return addrInfo
 }
 
--- a/p2p/test/muxer-negotiation/muxer_negotiation_test.go
+++ b/p2p/test/muxer-negotiation/muxer_negotiation_test.go
@@ -118,7 +118,7 @@ func makeHost(t *testing.T, transportType string, muxers []MuxerEntity, port int
 
 func getHostAddress(ha host.Host) *peer.AddrInfo {
 	// Build host multiaddress
-	hostAddr, _ := ma.NewMultiaddr(fmt.Sprintf("/p2p/%s", ha.ID().Pretty()))
+	hostAddr := ma.StringCast(fmt.Sprintf("/p2p/%s", ha.ID().Pretty()))
 	addr := ha.Addrs()[0]
 	addrInfo, err := peer.AddrInfoFromString(addr.Encapsulate(hostAddr).String())
 	if err != nil {
--- a/p2p/test/muxer-negotiation/muxer_negotiation_test.go
+++ b/p2p/test/muxer-negotiation/muxer_negotiation_test.go
@@ -90,6 +90,8 @@ func TestMuxerNegotiatin(t *testing.T) {
 
 		<-ready
 		runClientAndCheckMuxer(cctx, clih, getHostAddress(svrh), cSec, sSec, expected, t)
+		clih.Close()
+		svrh.Close()
 		cCancel()
 		sCancel()
 	}
--- a/p2p/test/muxer-negotiation/muxer_negotiation_test.go
+++ b/p2p/test/muxer-negotiation/muxer_negotiation_test.go
@@ -19,107 +19,89 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/muxer/yamux"
 	"github.com/stretchr/testify/require"
 
-	/*
-		"github.com/libp2p/go-libp2p/p2p/muxer/yamux"
-	*/
-
-	golog "github.com/ipfs/go-log/v2"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
 const testData = "Muxer Test Data\n"
 
 type muxerTest struct {
-	svrMuxers     []string
-	svrTrans      []network.Multiplexer
-	cliMuxers     []string
-	cliTrans      []network.Multiplexer
+	svrMuxers     []MuxerEntity
+	cliMuxers     []MuxerEntity
 	expectedMuxer string
 }
 
 func TestMuxerNegotiatin(t *testing.T) {
 	secTypes := []string{Tls, Noise}
 	testCases := []muxerTest{
-		{svrMuxers: []string{"/mplex/6.7.0"},
-			svrTrans:      []network.Multiplexer{mplex.DefaultTransport},
-			cliMuxers:     []string{"/yamux/1.0.0"},
-			cliTrans:      []network.Multiplexer{yamux.DefaultTransport},
+		{svrMuxers: []MuxerEntity{{"/mplex/6.7.0", mplex.DefaultTransport}},
+			cliMuxers:     []MuxerEntity{{"/yamux/1.0.0", yamux.DefaultTransport}},
 			expectedMuxer: ""},
-		{svrMuxers: []string{"/mplex/6.7.0"},
-			svrTrans:      []network.Multiplexer{mplex.DefaultTransport},
-			cliMuxers:     []string{"/mplex/6.7.0"},
-			cliTrans:      []network.Multiplexer{mplex.DefaultTransport},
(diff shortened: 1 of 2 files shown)


p2p/test/muxer-negotiation/muxer_negotiation_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peerstore"
+	"github.com/libp2p/go-libp2p/p2p/muxer/mplex"
+	"github.com/libp2p/go-libp2p/p2p/muxer/yamux"
+	"github.com/stretchr/testify/require"
+
+	/*
+		"github.com/libp2p/go-libp2p/p2p/muxer/yamux"
+	*/
```
marten-seemann, line 24: No comments please.

p2p/test/muxer-negotiation/muxer_negotiation_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+		<-ready
+		runClientAndCheckMuxer(cctx, clih, getHostAddress(svrh), cSec, sSec, expected, t)
+		clih.Close()
+		svrh.Close()
+		cCancel()
+		sCancel()
+	}
+
+	golog.SetAllLoggers(golog.LevelInfo)
```
marten-seemann, line 99: Why is this needed?

p2p/test/muxer-negotiation/muxer_negotiation_test.go
```diff
@@ -0,0 +1,199 @@
...
+		for i, testCase := range testCases {
+			testName := "Test muxer negotiation for " + secType + ", case " + fmt.Sprint(i)
+			t.Run(testName, func(t *testing.T) {
+				doMuxerNegotiation(t, secType, testCase.svrMuxers, testCase.cliMuxers, testCase.svrTrans, testCase.cliTrans, testCase.expectedMuxer)
+			})
+		}
+	}
+}
+
+func makeHost(transportType string, muxers []string, muxTrans []network.Multiplexer, port int) (*TransportWithMuxer, host.Host, error) {
```
marten-seemann, line 111: It's probably easier to pass a `testing.T` to these helper functions.

p2p/test/muxer-negotiation/muxer_negotiation_test.go
```diff
@@ -0,0 +1,199 @@
...
+	for i := 0; i < len(muxers); i++ {
+		opts = append(opts, libp2p.Muxer(muxers[i], muxTrans[i]))
+	}
+	h, err := libp2p.New(opts...)
+	return secTrans, h, err
+}
+
+func getHostAddress(ha host.Host) string {
+	// Build host multiaddress
+	hostAddr, _ := ma.NewMultiaddr(fmt.Sprintf("/p2p/%s", ha.ID().Pretty()))
```
marten-seemann, line 139:
```suggestion
	hostAddr := ma.StringCast(fmt.Sprintf("/p2p/%s", ha.ID().Pretty()))
```

p2p/test/muxer-negotiation/muxer_negotiation_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+	addr := ha.Addrs()[0]
+	return addr.Encapsulate(hostAddr).String()
+}
+
+func runServer(ctx context.Context, ha host.Host) {
+	ha.SetStreamHandler("/echo/1.0.0", streamHandler)
+}
+
+func runClientAndCheckMuxer(ctx context.Context, ha host.Host, targetPeer string, cSec, sSec *TransportWithMuxer, expected string, t *testing.T) {
```
marten-seemann, line 149: 1. You don't need `getHostAddress` if you pass a `peer.AddrInfo` to this function.
2. By convention, `*testing.T` should be the first parameter.

p2p/test/muxer-negotiation/muxer_negotiation_test.go
```diff
@@ -0,0 +1,199 @@
...
+	cliTrans      []network.Multiplexer
+	expectedMuxer string
+}
+
+func TestMuxerNegotiatin(t *testing.T) {
+	secTypes := []string{Tls, Noise}
+	testCases := []muxerTest{
+		{svrMuxers: []string{"/mplex/6.7.0"},
+			svrTrans:      []network.Multiplexer{mplex.DefaultTransport},
+			cliMuxers:     []string{"/yamux/1.0.0"},
```
marten-seemann, line 45: It's probably nicer to introduce a new struct to group the name and muxer together.

p2p/test/muxer-negotiation/muxer_negotiation_test.go
julian88110: 1. Not sure I am exactly following, but yes we don't have to convert it to string and back to addinfo, if that is what you mean. Updated the code.
2. yes, changed code.

p2p/test/muxer-negotiation/muxer_negotiation_test.go
julian88110: Not needed anymore, used for debugging. thanks for catching it

p2p/test/muxer-negotiation/muxer_negotiation_test.go
julian88110: 1. Not sure I am exactly following, but yes we don't have to convert it to string and back to addinfo, if that is what you mean. Updated the code.
2. yes, changed code.

marten-seemann (changes requested):
I must have missed this in my first review, sorry for that.
Why is the `TransportWithMuxer` needed? It seems like defining a new `SecureTransport` here defeats the purpose of an integration test: We should just be testing a host constructed with standard components using `libp2p.New`.

julian88110:
> I must have missed this in my first review, sorry for that. Why is the `TransportWithMuxer` needed? It seems like defining a new `SecureTransport` here defeats the purpose of an integration test: We should just be testing a host constructed with standard components using `libp2p.New`.

From host point of view, we get exactly the same host by either supplying the security transport constructor or by supplying a security transport directly.  The reason for TransportWithMuxer is to provide a pass-through wrapper to the security transport to enable some visibility of security transport state such as NextProto selected. Otherwise we don't have direct access to the internal state of selected muxer from a host, and that makes the test less direct. If we don't have the internal state of the security transport, it is not possible to know if the early muxer negotiation succeeded or not. Yes, we provided a  pass-through muxer wrapper, but the system behavior is not changed, so it does not compromise any test fidelity.

marten-seemann:
The `TransportWithMuxer` has some significant complexity, and doesn't reflect very well what a host constructed with `libp2p.New` would do. For example, this is not how the security protocol is selected. As such, this is not a proper integration test. What we should do is have a test that constructs two hosts via:
```go
h, err := libp2p.New(
    libp2p.Transport(tcp.New),
    libp2p.Security(...), // TLS, Noise, or Plaintext
    libp2p.Muxer(...), // mplex and yamux, ordered by priority, in different combinations for client and server
)
```
We then connect the two hosts and observe the result of the negotiation.

In order to satisfy the requirements listed in https://github.com/libp2p/go-libp2p/issues/1836#issuecomment-1288097608, we need to make sure that the correct muxer is selected, specifically, that _the same_ muxer is selected no matter if we run on top of TLS, Noise or Plaintext (which would exercise the non-optimized code path and do the negotiation via multistream).
We can find out which muxer was selected by exposing it somewhere, of by using reflections on the connection that was returned.

BigLep (commented):
> The TransportWithMuxer has some significant complexity, and doesn't reflect very well what a host constructed with libp2p.New would do. For example, this is not how the security protocol is selected.

I'm doing a drive-by and I obviously don't have experience on the codebase (or with go), but I'm surprised by this.  In my reading of the PR as it stands today:
1. The test setup is using public libp2p API (``libp2p.new`` with a passed in ``libp2p.Muxer`` to "inject" a ``TransportWithMuxer`` so we can capture what is happening when the two nodes do their security handshake.).
2. ``TransportWithMuxer`` doesn't seem to have complex logic to me.  It seems to be a simple wrapper around a ``SecureTransport`` for capturing the selected Muxer.  (maybe some top-level comments would help with clarity?)

I'm only commenting here because it seems like this integration test is snowballing and dragging on.  If we're asserting the wrong thing in the test, then that is one thing and needs to be fixed.  But if we're 1) not doing gross hacks like reflection in our test setup or as part of asserts and 2) we're asserting the right muxer is being selected, then I'm hoping we can get this merged without a lot of extra refactoring.

marten-seemann:
> 2\. `TransportWithMuxer` doesn't seem to have complex logic to me.  It seems to be a simple wrapper around a `SecureTransport` for capturing the selected Muxer.  (maybe some top-level comments would help with clarity?)

It's the constructor of `TransportWithMuxer` (the `New` method) that's problematic, as it constructs the security implementation (TLS / Noise)l: https://github.com/libp2p/go-libp2p/pull/1840/files#diff-6d7556f17e0bef65cc1474e0c2bd6bd5030146d5c89c7a119928393557c13eccR40-R44. This is problematic since the security implementations needs to be constructed with the slice of the muxers, and the order of the items in the slice expresses the preference. For example here: https://github.com/libp2p/go-libp2p/blob/c334288f8fe4d659f290043f788509e14f28cdde/p2p/security/tls/transport.go#L35
This test bypasses the normal code path used for the construction of the security implementation.

---

An integration test for muxer negotiation at the bare minimum needs to cover (using the exact same code path that `libp2p.New` uses):
1. initialization of the security implementation (TLS, Noise) and
2. exercise the negotiation

As it stands, this test doesn't do (1).

> If we're asserting the wrong thing in the test, then that is one thing and needs to be fixed. But if we're 1) not doing gross hacks like reflection in our test setup or as part of asserts and 2) we're asserting the right muxer is being selected, then I'm hoping we can get this merged without a lot of extra refactoring.

I'm open to different ways of asserting that the negotiation did the right thing. I don't have any particularly strong dislike of reflections in test code. If that's the easiest way to do this, so be it.
On the other hand, code that doesn't use reflections is always preferred to code that does :) And we'll need to expose the muxer anyway as part of our [swarm metrics effort](https://github.com/libp2p/go-libp2p/blob/master/ROADMAP.md#end-of-q4-december), which is on our roadmap for this quarter.

PR #1842, opened 2022-10-30
chore: preallocate slices

estensen:
When we know the final size of a slice preallocate for it to save memory.
Assigning directly is slightly faster than using append.

### DIFF
--- a/core/peer/addrinfo.go
+++ b/core/peer/addrinfo.go
@@ -86,7 +86,7 @@ func AddrInfoFromP2pAddr(m ma.Multiaddr) (*AddrInfo, error) {
 
 // AddrInfoToP2pAddrs converts an AddrInfo to a list of Multiaddrs.
 func AddrInfoToP2pAddrs(pi *AddrInfo) ([]ma.Multiaddr, error) {
-	var addrs []ma.Multiaddr
+	addrs := make([]ma.Multiaddr, len(pi.Addrs))
 	p2ppart, err := ma.NewComponent("p2p", Encode(pi.ID))
 	if err != nil {
 		return nil, err
@@ -94,8 +94,8 @@ func AddrInfoToP2pAddrs(pi *AddrInfo) ([]ma.Multiaddr, error) {
 	if len(pi.Addrs) == 0 {
 		return []ma.Multiaddr{p2ppart}, nil
 	}
-	for _, addr := range pi.Addrs {
-		addrs = append(addrs, addr.Encapsulate(p2ppart))
+	for i, addr := range pi.Addrs {
+		addrs[i] = addr.Encapsulate(p2ppart)
 	}
 	return addrs, nil
 }
--- a/core/peer/addrinfo.go
+++ b/core/peer/addrinfo.go
@@ -93,9 +93,9 @@ func AddrInfoToP2pAddrs(pi *AddrInfo) ([]ma.Multiaddr, error) {
 	if len(pi.Addrs) == 0 {
 		return []ma.Multiaddr{p2ppart}, nil
 	}
-	addrs := make([]ma.Multiaddr, len(pi.Addrs))
-	for i, addr := range pi.Addrs {
-		addrs[i] = addr.Encapsulate(p2ppart)
+	addrs := make([]ma.Multiaddr, 0, len(pi.Addrs))
+	for _, addr := range pi.Addrs {
+		addrs = append(addrs, addr.Encapsulate(p2ppart))
 	}
 	return addrs, nil
 }
--- a/core/peer/addrinfo.go
+++ b/core/peer/addrinfo.go
@@ -86,7 +86,6 @@ func AddrInfoFromP2pAddr(m ma.Multiaddr) (*AddrInfo, error) {
 
 // AddrInfoToP2pAddrs converts an AddrInfo to a list of Multiaddrs.
 func AddrInfoToP2pAddrs(pi *AddrInfo) ([]ma.Multiaddr, error) {
-	addrs := make([]ma.Multiaddr, len(pi.Addrs))
 	p2ppart, err := ma.NewComponent("p2p", Encode(pi.ID))
 	if err != nil {
 		return nil, err
@@ -94,6 +93,7 @@ func AddrInfoToP2pAddrs(pi *AddrInfo) ([]ma.Multiaddr, error) {
 	if len(pi.Addrs) == 0 {
 		return []ma.Multiaddr{p2ppart}, nil
 	}
+	addrs := make([]ma.Multiaddr, len(pi.Addrs))
 	for i, addr := range pi.Addrs {
 		addrs[i] = addr.Encapsulate(p2ppart)
 	}
--- a/p2p/net/connmgr/connmgr_test.go
+++ b/p2p/net/connmgr/connmgr_test.go
@@ -520,10 +520,10 @@ func TestPeerProtectionSingleTag(t *testing.T) {
 	}
 
 	// protect the first 5 peers.
-	var protected []network.Conn
-	for _, c := range conns[0:5] {
+	protected := make([]network.Conn, len(conns[0:5]))
+	for i, c := range conns[0:5] {
 		cm.Protect(c.RemotePeer(), "global")
-		protected = append(protected, c)
+		protected[i] = c
 		// tag them negatively to make them preferred for pruning.
 		cm.TagPeer(c.RemotePeer(), "test", -100)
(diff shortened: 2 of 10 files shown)


MarcoPolo:
Thanks! Did you use a tool to find these spots or something else?

estensen:
yeah, repro by running:
`golangci-lint run ./... -E prealloc --disable-all`

p2p/protocol/circuitv2/util/pbconv.go
```diff
@@ -36,13 +36,10 @@ func PeerToPeerInfoV1(p *pbv1.CircuitRelay_Peer) (peer.AddrInfo, error) {
 }
 
 func PeerInfoToPeerV1(pi peer.AddrInfo) *pbv1.CircuitRelay_Peer {
-	var addrs [][]byte
-	if len(pi.Addrs) > 0 {
-		addrs = make([][]byte, 0, len(pi.Addrs))
-	}
+	addrs := make([][]byte, len(pi.Addrs))
```
MarcoPolo, line 39: Let’s keep the old behaviour of having this be nil. In this case I don’t think this matters, but generally we can’t be sure if callers check this.

core/peer/addrinfo.go
```diff
@@ -86,16 +86,16 @@ func AddrInfoFromP2pAddr(m ma.Multiaddr) (*AddrInfo, error) {
 
 // AddrInfoToP2pAddrs converts an AddrInfo to a list of Multiaddrs.
 func AddrInfoToP2pAddrs(pi *AddrInfo) ([]ma.Multiaddr, error) {
-	var addrs []ma.Multiaddr
+	addrs := make([]ma.Multiaddr, len(pi.Addrs))
```
marten-seemann, line 89: Can you move this down to where it's actually needed?

marten-seemann (changes requested):
I don't understand why you're replacing `append`s with assignments to a specific index. `append` is the nicer way to do this, please don't change that.

MarcoPolo:
> I don't understand why you're replacing `append`s with assignments to a specific index. `append` is the nicer way to do this, please don't change that.

I don’t have a preference, but assignments are technically faster. That said, I don’t think this matters here.

marten-seemann:
> > I don't understand why you're replacing `append`s with assignments to a specific index. `append` is the nicer way to do this, please don't change that.
> 
> I don’t have a preference, but assignments are technically faster. That said, I don’t think this matters here.

I'd expect it to compile to exactly the same code. I ran a benchmark, and the two perform exactly the same. If at all, appending is slightly faster, but probably not statistically significant.
```
goos: darwin
goarch: arm64
pkg: github.com/libp2p/go-libp2p/p2p/protocol/holepunch
BenchmarkAppend-10      24834030                48.23 ns/op
BenchmarkAssign-10      24644346                48.32 ns/op
```

```go
const l = 128

func BenchmarkAppend(b *testing.B) {
	a := make([]int, 0, l)
	for i := 0; i < b.N; i++ {
		a = a[:0]
		for j := 0; j < l; j++ {
			a = append(a, j)
		}
	}
}

func BenchmarkAssign(b *testing.B) {
	a := make([]int, l)
	for i := 0; i < b.N; i++ {
		for j := 0; j < l; j++ {
			a[j] = j
		}
	}
}
```

The reason people tend to prefer `append` is that it's safer: If you get your length calculation wrong, all that happens is that you do an additional alloc. If you get your length calculation wrong and do an assignment, your process will panic.

estensen:
I've always gotten better perf from assignment than appending. Same code on my machine:
```
goos: darwin
goarch: amd64
pkg: github.com/estensen/benchmarks/tmp
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkAppend-12    	15084253	        68.56 ns/op
BenchmarkAssign-12    	17883320	        66.12 ns/op
PASS
ok  	github.com/estensen/benchmarks/tmp	2.536s
```

But I do agree that it is safer to use `append`. I'll revert it.

p2p/net/connmgr/connmgr_test.go
```diff
@@ -520,7 +520,7 @@ func TestPeerProtectionSingleTag(t *testing.T) {
 	}
 
 	// protect the first 5 peers.
-	var protected []network.Conn
+	protected := make([]network.Conn, 0, len(conns[0:5]))
```
marten-seemann, line 523: Let's not do this, this looks error-prone. This is a test anyway, so I wouldn't care about allocs.

p2p/net/connmgr/connmgr_test.go
```diff
@@ -599,7 +599,7 @@ func TestPeerProtectionMultipleTags(t *testing.T) {
 	}
 
 	// protect the first 5 peers under two tags.
-	var protected []network.Conn
+	protected := make([]network.Conn, 0, len(conns[0:5]))
```
marten-seemann, line 602: Same here.

p2p/net/swarm/swarm_listen.go
```diff
@@ -39,7 +39,7 @@ func (s *Swarm) Listen(addrs ...ma.Multiaddr) error {
 
 // ListenClose stop and delete listeners for all of the given addresses.
 func (s *Swarm) ListenClose(addrs ...ma.Multiaddr) {
-	var listenersToClose []transport.Listener
+	listenersToClose := make([]transport.Listener, 0, len(s.listeners.m))
```
marten-seemann, line 42: This will massively over-allocate.

p2p/protocol/circuitv2/relay/relay.go
```diff
@@ -492,7 +492,7 @@ func (r *Relay) writeResponse(s network.Stream, status pbv2.Status, rsvp *pbv2.R
 func (r *Relay) makeReservationMsg(p peer.ID, expire time.Time) *pbv2.Reservation {
 	expireUnix := uint64(expire.Unix())
 
-	var addrBytes [][]byte
+	addrBytes := make([][]byte, 0, len(r.host.Addrs()))
```
marten-seemann, line 495: This will also overallocate if there are many non-public addresses.

p2p/protocol/internal/circuitv1-deprecated/util.go
```diff
@@ -38,9 +38,9 @@ func peerToPeerInfo(p *pb.CircuitRelay_Peer) (peer.AddrInfo, error) {
 }
 
 func peerInfoToPeer(pi peer.AddrInfo) *pb.CircuitRelay_Peer {
-	addrs := make([][]byte, len(pi.Addrs))
-	for i, addr := range pi.Addrs {
-		addrs[i] = addr.Bytes()
+	addrs := make([][]byte, 0, len(pi.Addrs))
+	for _, addr := range pi.Addrs {
+		addrs = append(addrs, addr.Bytes())
```
marten-seemann, line 43: This is deprecated code. No need to make any changes here.

marten-seemann (changes requested):

p2p/net/swarm/swarm_listen.go
estensen: I see. I don't find any good ways to provide a size hint here, so removing this

marten-seemann (approved):

marten-seemann:
Thank you!

MarcoPolo:
Let me preface this by saying this doesn't matter at all :)

> I'd expect it to compile to exactly the same code.

I went into a rabbit hole here because this doesn't make sense to me. Append _has_ to do more work than an assign (which should be a single instruction). So it _should_ be slower. Sure enough if we load up the code in [godbolt](https://godbolt.org/z/dzhr8fYzq) we see that the append has one more compare and mov than assign _should have_. I say  _should have_ because this benchmark actually is flawed since the compiler optimizes the assignment and makes it the same as the noop (you don't see a `MOVQ` in the asm output). To fix this you actually should do something like [this](https://gist.github.com/MarcoPolo/18b39366c6f57270059d0ef03c29351b#file-main_test-go-L55). This optimization actually makes it a bit more surprising that append (an extra `cmp` and 2 `mov`s) is as fast as a `noop`, but then again CPUs are pretty good at branch prediction and pipelining and parallelizing especially in this very simple case.

Also I'm impressed with how well the compiler optimized the append loop. It really keeps the hot path very small. If we run without compiler optimizations (`gcflags=-N`) we also see the expected speed diffs.

Results from my tests here: https://gist.github.com/MarcoPolo/18b39366c6f57270059d0ef03c29351b#file-results-md

PR #1848, opened 2022-10-31
core/peerstore: removed unused provider addr ttl constant

marten-seemann:
Extracting this commit out of #1835, to better preserve history when squashing.


reddragon09:
@marten-seemann 
You're forgot remove ProviderAddrTTL in https://github.com/libp2p/go-libp2p-core/blob/master/peerstore/peerstore.go#L27

marten-seemann:
I did not. That repo is archived.

PR #1855, opened 2022-11-04
swarm: return filtered addrs in addrsForDial

dennis-tra:
Commit [`f654b4b`](https://github.com/libp2p/go-libp2p/commit/f654b4bd6970d41737b2098fa0998ca686916118) introduced multiaddress resultion by transport and changed the return value from `goodAddrs` to `resolved`. I guess this should be `goodAddrs` again?

cc @MarcoPolo

### DIFF
--- a/p2p/net/swarm/swarm_dial.go
+++ b/p2p/net/swarm/swarm_dial.go
@@ -334,7 +334,7 @@ func (s *Swarm) addrsForDial(ctx context.Context, p peer.ID) ([]ma.Multiaddr, er
 
 	s.peers.AddAddrs(p, goodAddrs, peerstore.TempAddrTTL)
 
-	return resolved, nil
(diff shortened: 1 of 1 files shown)


marten-seemann (approved):
LGTM, but I'll let @MarcoPolo do another review, as this was introduced in #1719.

marten-seemann:
We should probably also have a test, stuff like this _should_ have been caught be a unit test.

p-shahi:
I think this could be a candidate to include in `0.23.4` along with connection manager map fix
Wdyt @MarcoPolo @julian88110 @marten-seemann?

MarcoPolo (approved):
I can't think of any reason why I would prefer resolved. Looks like my bug. Thanks for the catch! :bug: 🥾

Issue #1847, opened 2022-10-31
connmgr: Concurrent Map Misuse


p-shahi:
## Description

Reported in: https://github.com/ipfs/kubo/issues/9374#issuecomment-1296416337

When running `Kubo 0.16.0` users report a crash.
There is a provided stacktrace: 

```
WebUI: http://127.0.0.1:5001/webui
Gateway (readonly) server listening on /ip4/127.0.0.1/tcp/8080
Daemon is ready
fatal error: concurrent map iteration and map write

goroutine 222 [running]:
github.com/libp2p/go-libp2p/p2p/net/connmgr.peerInfos.SortByValueAndStreams.func1.1(0xc0179d00f0?)
        github.com/libp2p/go-libp2p@v0.23.2/p2p/net/connmgr/connmgr.go:267 +0x6a
github.com/libp2p/go-libp2p/p2p/net/connmgr.peerInfos.SortByValueAndStreams.func1(0x16b?, 0x21b?)
        github.com/libp2p/go-libp2p@v0.23.2/p2p/net/connmgr/connmgr.go:277 +0x127
sort.partition_func({0xc0119d9370?, 0xc003e967e0?}, 0x0, 0x2d4, 0x32?)
        sort/zsortfunc.go:157 +0x191
```

https://github.com/libp2p/go-libp2p/blob/master/p2p/net/connmgr/connmgr.go#L255

## Versions affected

`Kubo 0.16.0` uses `go-libp2p v0.23.2`
But this issue could have existed for some time.

## Root Cause Analysis
```
<Once determined, fill RCA here>
```

### Open Questions:
If it's valid that concurrent goroutines need to read `peerInfos` as it's being written to, shouldn't it be locked inside the sync read write mutex?
https://github.com/libp2p/go-libp2p/blob/7465a509d01195b38d00125814457d2b37d29fd0/p2p/net/connmgr/connmgr.go#L374-L377
https://github.com/libp2p/go-libp2p/blob/7465a509d01195b38d00125814457d2b37d29fd0/p2p/net/connmgr/connmgr.go#L405-L407
https://github.com/libp2p/go-libp2p/blob/7465a509d01195b38d00125814457d2b37d29fd0/p2p/net/connmgr/connmgr.go#L456-L468

p-shahi:
Some more logs shared here: https://github.com/ipfs/kubo/issues/9374#issuecomment-1306547871

@julian88110 @MarcoPolo would either of you like to take a look?
[...]

Issue #1707, opened 2022-07-04
Warn if limits conflict with connmgr


MarcoPolo:
We should emit some warning if the current limits conflict with the current settings for the connection manager. Not sure if this should live here or somewhere else. This issue is mostly to track it.

piersy:
Hi @MarcoPolo, id be interested in picking this up, but I would need a bit of clarification from you about how you think this should work.

It looks to me like a good place to insert the check would be just after this block, since everything should have been configured by that point:

https://github.com/libp2p/go-libp2p/blob/4caa4e5bd45e5c834c2d417d06d276508b2f9621/config/config.go#L297-L302

However  the resource manager in `rcmgr` is not exported, were you planning on exporting it in order to get at the conn limit? The alternative I see would be to add a method to `network.ResourceManager` to expose the conn limit. Similarly were you planning on adding a method to `connmgr.ConnManager` to check the limit extracted from the resource manger?

Thanks!

MarcoPolo:
Thanks @piersy!

Here's what I'm thinking:

1. We can do this at the start of the NewNode function since everything we need is in the Config.
2. The ConnManager can have a new method called `CheckLimits(limits GetterConnLimits)`.
3. GetConnLimits is an interface with a single method `GetConnLimits`.
4. The private resourceManager type implements this interface and returns the total system limits.

Conn manager defines a new interface for the thing it checks its limits against. And the resourceManager implements this interface. The glue code in NewNode checks to see if the configured resource manager implements the interface and calls the `CheckLimits` method on the conn manager.

There's probably some subtleties here, but hopefully this is enough to get you started :)

piersy:
Thanks @MarcoPolo, that sounds good to me!

PR #1857, opened 2022-11-08
connmgr: fix concurrent map misuse.

cachalots:
fix https://github.com/libp2p/go-libp2p/issues/1847 https://github.com/ipfs/kubo/issues/9374
https://github.com/libp2p/go-libp2p/blob/21dc42bd72fc6064c27a098db680e94ced5a76d6/p2p/net/connmgr/connmgr.go#L450
Loop conns is not locked, causing concurrent map misuse.

This repair method needs to modify more code, so I didn't choose to use this method.
```
type peerInfos  []peerInfo

// replace to
type peerInfos struct {
	peerInfo []peerInfo
	segments segments
}
```

### DIFF
--- a/p2p/net/connmgr/connmgr.go
+++ b/p2p/net/connmgr/connmgr.go
@@ -252,7 +252,7 @@ type peerInfos []peerInfo
 // will sort peers with no streams before those with streams (all else being
 // equal). If `sortByMoreStreams` is true it will sort peers with more streams
 // before those with fewer streams. This is useful to prioritize freeing memory.
-func (p peerInfos) SortByValueAndStreams(sortByMoreStreams bool) {
+func (p peerInfos) SortByValueAndStreams(sortByMoreStreams bool, segments segments) {
 	sort.Slice(p, func(i, j int) bool {
 		left, right := p[i], p[j]
 		// temporary peers are preferred for pruning.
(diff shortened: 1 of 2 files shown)


marten-seemann (changes requested):
That doesn't seem like a very intuitive API. `peerInfo` is just a `[]peerInfo`, and has no (obvious) connection to `segments` (which just holds a `map` of `peerInfo`s).

cachalots:
> That doesn't seem like a very intuitive API. `peerInfo` is just a `[]peerInfo`, and has no (obvious) connection to `segments` (which just holds a `map` of `peerInfo`s).

I just put forward this idea. The code just add parameter `segments` to the  method `SortByValueAndStreams` to get lock.

p-shahi:
Thanks for the initiative and providing logs @cachalots! I will close this in favor of https://github.com/libp2p/go-libp2p/pull/1860

MarcoPolo:
Thanks for looking into this @cachalots. I put a fix here: https://github.com/libp2p/go-libp2p/pull/1860. It's worth pointing out that this PR has two possible deadlocks. 

1. both left and right are in the same segment.
2. goroutine 1 locks A then B, goroutine 2 locks B then A.

s hard Concurrency i.

I'm closing this pr and proceeding with #1860 (feel free to give that a review!), but I appreciate the effort. In the mean time, if you're looking for other places to contribute you could take a look at: https://github.com/libp2p/go-libp2p/issues/1707 especially since you've already familiarized yourself with the connmgr.

Issue #1854, opened 2022-11-03
Remove multistream.transport as an entity of type network.multiplexer


julian88110:
In the process of building the integration test for early muxer negotiation https://github.com/libp2p/go-libp2p/issues/1836, this enhancement came into being as suggested by @marten-seemann  while we discussed the options of making the implementation of #1851 cleaner. 

The existence of multistream.transport as a network.multiplexer type becomes an obvious obstacle of sharing muxer state info internally, especially in the concurrent context which may require locking to avoid data race. 

This issue ticket proposes to get rid of the multistream.transport entity and merge its function into the upgrader.  This solution may require fair amount of refactoring of the config and initialization code too, a lot of test code will also be affected, but in long term it should be worth the effort for simplified upgrade logic and cleaner code structure. There is no reason the multistream.transport should be a network.Transport type and be restricted to that interface, which becomes awkward in facilitating new features such as early muxer negotiation.

BigLep:
@julian88110 : lets reference https://github.com/libp2p/go-libp2p/pull/1851#issuecomment-1302450166 in the description.

julian88110:
> @julian88110 : lets reference [#1851 (comment)](https://github.com/libp2p/go-libp2p/pull/1851#issuecomment-1302450166) in the description.
Good point. updated the description.

PR #1861, opened 2022-11-08
Remove muxer-multistream.Transport and refactor upgrader

julian88110:
Details can be found in Issue #1854.

 - Remove muxer-multistream.Transport as an entity of type network.Multiplexer.
 -  Define a new type MsTransport to encapsulate the multistream selection function.
 -  Revise upgrader code to use the new MsTransport for muxer negotiation. 
 -  Update the relevant tests.


marten-seemann (changes requested):
I don't think we should have the `MsTransport` interface, or its implementation. The point of #1854 was to get rid of this complexity, not to move it to a different package.

p2p/net/upgrader/upgrader_test.go
```diff
@@ -51,23 +53,41 @@ func (m *negotiatingMuxer) NewConn(c net.Conn, isServer bool, scope network.Peer
...
 		return nil, err
 	}
-	return yamux.DefaultTransport.NewConn(c, isServer, scope)
+
+	return &upgrader.StmMuxer{
+		ID:          "/yamux/1.0.0",
+		StreamMuxer: yamux.DefaultTransport,
+	}, nil
+
+	//return yamux.DefaultTransport.NewConn(c, isServer, scope)
```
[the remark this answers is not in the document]
julian88110: missed this in clean up. 
Cleaned it up.

p2p/net/upgrader/multistream.go
```diff
@@ -14,6 +12,17 @@ import (
 
 var DefaultNegotiateTimeout = time.Second * 60
 
+type StmMuxer struct {
+	ID          string
+	StreamMuxer network.Multiplexer
+}
+
+type MsTransport interface {
```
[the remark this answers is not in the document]
julian88110: > 

Yes the multistream.transport can be merged into upgrader, having the upgrader lump all the functions together. But I prefer a modularized solution by encapsulating the muxer negotiation function in MsTransport. Single responsibility for single entitiy. The root cause of the issue is that the original multistream.Transport was forced to comply with network.Multiplexer interface and assume awkward position in the function chain. I believe by giving MsTranport its own module clarifies the function and keeps a clean modularized structure. Glad to hear more of your thoughts on this if you have any concerns.

p2p/net/upgrader/multistream.go
```diff
@@ -14,6 +12,17 @@ import (
 
 var DefaultNegotiateTimeout = time.Second * 60
 
+type Multiplexer struct {
+	ID          string
+	StreamMuxer network.Multiplexer
+}
+
+type MsTransport interface {
```
MarcoPolo, line 20: Generally callers define the interface: https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_let_callers_define_the_interface_they_require.

There's a tendency to follow Java patterns and say "here's an interface that clients can use". But that's backwards, it's really here's what I provide as my concrete type, and callers can program against the interface they need. If your type doesn't conform to what they need they'll get a compiler error. More context here too: https://github.com/libp2p/go-libp2p/pull/1758#issuecomment-1267623184

Also see https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#interfaces

p2p/transport/websocket/websocket_test.go
```diff
@@ -36,10 +36,16 @@ import (
...
+	muxer := tptu.NewMsTransport()
+	muxer.AddMuxer("/yamux/1.0.0", yamux.DefaultTransport)
+	return muxer
+}
+
 func newUpgrader(t *testing.T) (peer.ID, transport.Upgrader) {
 	t.Helper()
 	id, m := newInsecureMuxer(t)
-	u, err := tptu.New(m, yamux.DefaultTransport)
+	u, err := tptu.New(m, makeMsTransport())
```
MarcoPolo, line 48: This is different than the old behavior. Why this change?

p2p/net/upgrader/upgrader_test.go
```diff
@@ -77,12 +95,18 @@ func (m *blockingMuxer) Unblock() {
 // errorMuxer is a muxer that errors while setting up
 type errorMuxer struct{}
 
-var _ network.Multiplexer = &errorMuxer{}
+var _ upgrader.MsTransport = &errorMuxer{}
 
-func (m *errorMuxer) NewConn(c net.Conn, isServer bool, scope network.PeerScope) (network.MuxedConn, error) {
+func (m *errorMuxer) NegotiateMuxer(c net.Conn, isServer bool) (*upgrader.Multiplexer, error) {
```
MarcoPolo, line 100: This is a different test now. The old test errored when creating a muxed connection. This now errors when negotiation. I don't think we should change this test, but it would maybe we should add another test if it makes sense.

p2p/net/upgrader/upgrader.go
```diff
@@ -212,10 +210,14 @@ func (u *upgrader) setupMuxer(ctx context.Context, conn sec.SecureConn, server b
 
 	var smconn network.MuxedConn
 	var err error
-	// TODO: The muxer should take a context.
+	var streamMuxer *Multiplexer
+
 	go func() {
 		defer close(done)
-		smconn, err = u.muxer.NewConn(conn, server, scope)
+		streamMuxer, err = u.mstream.NegotiateMuxer(conn, server)
```
MarcoPolo, line 217: I think `NegotiateMuxer` should take a ctx and then we don't have to do this select dance below and we don't have to spawn a go routine.

p2p/net/upgrader/multistream.go
```diff
@@ -24,21 +33,21 @@ type Transport struct {
...
 
-func (t *Transport) AddTransport(path string, tpt network.Multiplexer) {
+func (t *Transport) AddMuxer(path string, tpt network.Multiplexer) {
 	t.mux.AddHandler(path, nil)
 	t.tpts[path] = tpt
 	t.OrderPreference = append(t.OrderPreference, path)
 }
 
-func (t *Transport) NewConn(nc net.Conn, isServer bool, scope network.PeerScope) (network.MuxedConn, error) {
+func (t *Transport) NegotiateMuxer(nc net.Conn, isServer bool) (*Multiplexer, error) {
```
MarcoPolo, line 50: This is the only function you need – and it's just a thin wrapper around `go-multistream`. I would say:

1. we move this to the upgrader (No one else really cares about this package (I think)).
2. Use a ctx to set the negotiate timeout
3. Make it private
4. Delete this whole package. less code!

MarcoPolo (changes requested):
I agree with @marten-seemann , let's remove the `muxer_multistream` package. I left some notes that I think give a rough idea of the steps required. But really you can probably start by copying the `NegotiateMuxer` func, deleting the `muxer_multistream` package, and then fixing errors/tests.

You got yourself a great refactor here! and I would expect net negative code and fewer public things :)

p2p/net/upgrader/multistream.go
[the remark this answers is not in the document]
marten-seemann: Yes, it definitely should be moved into the upgrader. I'm all for modularizing when it makes, but I don't think in this case it does. The `GetTransportByKey` shows that this interface creates more friction than it removes.

As @MarcoPolo pointed out in https://github.com/libp2p/go-libp2p/pull/1861/files#r1018559886, this is really just a few lines of code.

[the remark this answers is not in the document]
julian88110: merged the function of MsTransport into upgrader.

p2p/net/upgrader/multistream.go
julian88110: MsTransport removed, thanks for the info link!

p2p/transport/websocket/websocket_test.go
julian88110: that's because of the change of the way upgrader is constructed, the final product is the upgrader, so that the end result is the same. The existing code makes the multistream.Transport conform to network.Multiplexer interface so you can pass a yamux to upgrader, but that's not the case anymore after the multistream.Transport is removed.

p2p/net/upgrader/upgrader_test.go
julian88110: As the multistream transport itself is removed, this test is no longer relevant so I think it should be removed now.

p2p/net/upgrader/upgrader.go
```diff
@@ -74,15 +75,21 @@ type upgrader struct {
...
+
+	msmuxer    *mss.MultistreamMuxer //
+	muxers     map[string]network.Multiplexer
+	preference []string
 }
 
 var _ transport.Upgrader = &upgrader{}
 
-func New(secureMuxer sec.SecureMuxer, muxer network.Multiplexer, opts ...Option) (transport.Upgrader, error) {
+func New(secureMuxer sec.SecureMuxer, muxers map[string]network.Multiplexer, preference []string, opts ...Option) (transport.Upgrader, error) {
```
marten-seemann, line 86: These constructor options are very unclean.

What you want to use here is list of `<id>, <muxer>`, ordered by preference. Introduce a type that reflects that.

config/muxer.go
```diff
@@ -43,21 +42,23 @@ func MuxerConstructor(m interface{}) (MuxC, error) {
 	}, nil
 }
 
-func makeMuxer(h host.Host, tpts []MsMuxC) (network.Multiplexer, error) {
-	muxMuxer := msmux.NewBlankTransport()
+func makeMuxers(h host.Host, tpts []MsMuxC) (map[string]network.Multiplexer, []string, error) {
 	transportSet := make(map[string]struct{}, len(tpts))
+	muxers := make(map[string]network.Multiplexer)
+	preference := make([]string, 0)
```
marten-seemann, line 48: Preallocate slices with the correct capacity, if you know the capacity in advance.

marten-seemann (commented):
This looks a lot better than the last iteration!

Issue #1852, opened 2022-11-02
"panic: runtime error: invalid memory address or nil pointer dereference" in p2p/host/autorelay/relay_finder.go:209 if EnableAutoRelay()


alx696:
After update to v0.23.2, an error will occur:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc77f31]

goroutine 8824 [running]:
github.com/libp2p/go-libp2p/p2p/host/autorelay.(*relayFinder).findNodes(0xc00010ed00, {0x146f198?, 0xc002d3f940})
        /home/m/go/pkg/mod/github.com/libp2p/go-libp2p@v0.23.2/p2p/host/autorelay/relay_finder.go:209 +0x51
github.com/libp2p/go-libp2p/p2p/host/autorelay.(*relayFinder).background.func2()
        /home/m/go/pkg/mod/github.com/libp2p/go-libp2p@v0.23.2/p2p/host/autorelay/relay_finder.go:112 +0x66
created by github.com/libp2p/go-libp2p/p2p/host/autorelay.(*relayFinder).background
        /home/m/go/pkg/mod/github.com/libp2p/go-libp2p@v0.23.2/p2p/host/autorelay/relay_finder.go:110 +0x192

    at ChildProcess.exithandler (node:child_process:407:12)
    at ChildProcess.emit (node:events:527:28)
    at maybeClose (node:internal/child_process:1092:16)
    at Socket.<anonymous> (node:internal/child_process:451:11)
    at Socket.emit (node:events:527:28)
    at Pipe.<anonymous> (node:net:709:12)
```

If comment out `libp2p.EnableAutoRelay()` , no errors.

```
libp2p.New(
		// Use the keypair we generated
		libp2p.Identity(*myKey),
		// Multiple listen addresses
		libp2p.ListenAddrStrings(
			fmt.Sprint("/ip4/0.0.0.0/tcp/", port),          // regular tcp connections
			fmt.Sprint("/ip4/0.0.0.0/udp/", port, "/quic"), // a UDP endpoint for the QUIC transport
			fmt.Sprint("/ip6/::/udp/", port, "/quic"),      // a UDP endpoint for the QUIC transport
		),
		// support TLS connections
		libp2p.Security(libp2p_tls.ID, libp2p_tls.New),
		// support noise connections
		libp2p.Security(noise.ID, noise.New),
[...]

PR #1863, opened 2022-11-08
fix: autorelay: add some guard rails and docs

MarcoPolo:
Fixes https://github.com/libp2p/go-libp2p/issues/1852 by adding guard rails and some more documentation.

### DIFF
--- a/options.go
+++ b/options.go
@@ -245,7 +245,9 @@ func EnableRelayService(opts ...relayv2.Option) Option {
 //
 // Dependencies:
 //   - Relay (enabled by default)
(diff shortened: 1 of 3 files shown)


marten-seemann (approved):
LGTM.

In the long term, we can think about a nicer API, as suggested in https://github.com/libp2p/go-libp2p/issues/1852#issuecomment-1307995595. That won't make it into a patch release though.

MarcoPolo:
Agreed. Opened an issue here: https://github.com/libp2p/go-libp2p/issues/1866


Let's merge this for the patch release.

PR #1869, opened 2022-11-09
core/protocol: remove deprecated Negotiator.NegotiateLazy

marten-seemann:
This function was deprecated more than half a year ago.


MarcoPolo (approved):
We still need to remove it from the multistream package.

marten-seemann:
> We still need to remove it from the multistream package.

@MarcoPolo https://github.com/multiformats/go-multistream/pull/92

PR #1874, opened 2022-11-11
libp2p + HTTP

MarcoPolo:
A draft PR that adds first class support for HTTP as a protocol in libp2p. This came out of various discussions around IPFS camp and 1:1 discussions.

This means:
* You can write a protocol using HTTP
* You can use HTTP over libp2p streams.
* You can use your HTTP protocol over a standard TCP + TLS HTTP connection (even curl works to communicate with a libp2p node, i.e. this works: `curl --insecure --cert ./client.cert --key ./client.key https://127.0.0.1:9561/echo -d "Hello World"`).
* This enables you to communicate with a peer over standard TCP + TLS HTTP, _even if that peer is a dumb static server_.

More details in the draft spec: <todo link>

---

Note: this is **not** in the roadmap. This came from seeing the desire when talking to various users around IPFS camp. This work for now will be prioritized below other items in the roadmap unless we agree otherwise.

---  

Do not merge this PR until:

- [ ] We discuss this in the libp2p spec
- [ ] Agree on a protocol ID
- [ ] Agree on the API shown off here
- [ ] Clean up the code. There are various TODOs left since the first pass here is a PoC.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -108,6 +108,8 @@ type Config struct {
 
 	EnableHolePunching  bool
 	HolePunchingOptions []holepunch.Option
+
+	HTTPConfig bhost.HTTPConfig
 }
 
 func (cfg *Config) makeSwarm() (*swarm.Swarm, error) {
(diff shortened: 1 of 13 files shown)


marten-seemann:
Closing since libp2p now has HTTP support.

PR #1875, opened 2022-11-11
fix: autorelay: treat static relays as just another peer source

MarcoPolo:
fixes https://github.com/libp2p/go-libp2p/issues/1782 by making static relays a specific type of peer source. Then re-using all the peersource logic. @marten-seemann  was there a reason you didn't do this originally?

### DIFF
--- a/p2p/host/autorelay/options.go
+++ b/p2p/host/autorelay/options.go
@@ -15,8 +15,7 @@ type config struct {
 	clock      clock.Clock
 	peerSource func(ctx context.Context, num int) <-chan peer.AddrInfo
 	// minimum interval used to call the peerSource callback
-	minInterval  time.Duration
-	staticRelays []peer.AddrInfo
+	minInterval time.Duration
 	// see WithMinCandidates
 	minCandidates int
 	// see WithMaxCandidates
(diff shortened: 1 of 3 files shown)


p2p/host/autorelay/options.go
```diff
@@ -75,17 +73,25 @@ type Option func(*config) error
...
-			return errors.New("can't set static relays, static relays already configured")
+			return errAlreadyHavePeerSource
 		}
-		c.minCandidates = len(static)
-		c.staticRelays = static
+
+		WithPeerSource(func(ctx context.Context, numPeers int) <-chan peer.AddrInfo {
+			if len(static) < numPeers {
+				numPeers = len(static)
+			}
```
marten-seemann, line 83: The documentation for `WithPeerSource` says:
> Implementations must send *at most* numPeers, and close the channel when they don't intend to provide any more peers.

marten-seemann: We can use `WithMaxCandidaters` to make sure that the `relayFinder` is asking us for enough peers. We should probably also add some documentation that there's little point in providing a huge number of static relays.

marten-seemann (commented):
We probably could be a lot smarter about static relays (for example, doing exponential backoff for the ones we can't connect to), but we can leave that for a future PR.

This logic looks fine, and I'm not sure why I didn't implement it like that back then.

p2p/host/autorelay/options.go
MarcoPolo: This check _is_ to make sure we send at most numPeers. If we don’t have numPeers we’ll only send what we have. Essentially the ‘min(len(static),numPeers))’. I debated using with ‘WithMaxCandidaters’, but your right I’ll add it.

marten-seemann: For some reason I thought the `if` was the other way round (`>` instead of `<`). Sorry for that!

marten-seemann (approved):

PR #1878, opened 2022-11-14
rcmgr: fix missing service peer default in LimitConfig.Apply

neelvirdy:
ServicePeerDefault wasn't being applied in the Apply method, which led to these limits being 0 instead of mirroring the limits from the Apply call in some usage patterns.


neelvirdy:
tagging @marten-seemann for review since i can't add reviewers via GH, thanks!

marten-seemann (approved):
Thank you!

PR #1881, opened 2022-11-15
feat: quic,webtransport: enable both quic-draft29 and quic-v1 addrs on quic. only quic-v1 on webtransport

### LINKED ISSUE #1841: quic: advertise both quic-draft29 and quic-v1 addresses
This issue assumes that we introduce a new code point for QUIC v1 as suggested in https://github.com/multiformats/multiaddr/issues/145.

For a transition period (6 months?) we should advertise both QUIC draft-29 and QUIC v1. This makes sure that legacy nodes will be able to use QUIC to connect to us for a while. After that, they can still use TCP. Really, they _should_ upgrade their node.

To make this happen, we need to:
* [x] https://github.com/multiformats/multicodec/pull/298
* [x] make the corresponding change in [go-multiaddr](https://github.com/multiformats/go-multiaddr)
* [ ] change `transport.Listener.Multiaddr` method to return `[]ma.Multiaddr` instead of a single `ma.Multiaddr`
* [ ] use that updated interface to return QUIC draft-29 and QUIC v1 multiaddrs from the QUIC transport
* [ ] disable QUIC draft-29 for Webtransport (both for dialing and listening)
* [ ] pick the correct version in the QUIC transport for dialing (according to the multiaddr we're dialing)

MarcoPolo:
Fixes #1841 see that issue for more context.

This lets transports advertise multiple multiaddrs. Allows the quic transport to advertise both QUIC versions draft29 and 1. Changes webtransport to only support QUIC version 1.

Before merge: 
- [ ] Fix go.mod to use updated quic-go after https://github.com/lucas-clemente/quic-go/pull/3620 is merged and released.

### DIFF
--- a/p2p/net/swarm/swarm_addr.go
+++ b/p2p/net/swarm/swarm_addr.go
@@ -18,7 +18,7 @@ func (s *Swarm) ListenAddresses() []ma.Multiaddr {
 func (s *Swarm) listenAddressesNoLock() []ma.Multiaddr {
 	addrs := make([]ma.Multiaddr, 0, len(s.listeners.m))
 	for l := range s.listeners.m {
-		addrs = append(addrs, l.Multiaddr())
+		addrs = append(addrs, l.Multiaddrs()...)
 	}
 	return addrs
 }
--- a/p2p/net/swarm/swarm_addr.go
+++ b/p2p/net/swarm/swarm_addr.go
@@ -16,7 +16,7 @@ func (s *Swarm) ListenAddresses() []ma.Multiaddr {
 }
 
 func (s *Swarm) listenAddressesNoLock() []ma.Multiaddr {
-	addrs := make([]ma.Multiaddr, 0, len(s.listeners.m))
+	addrs := make([]ma.Multiaddr, 0, len(s.listeners.m)+10) // A bit extra so we may avoid an extra allocation in the for loop below.
 	for l := range s.listeners.m {
 		addrs = append(addrs, l.Multiaddrs()...)
 	}
--- a/p2p/net/swarm/swarm_dial.go
+++ b/p2p/net/swarm/swarm_dial.go
@@ -12,6 +12,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/core/transport"
+	"github.com/lucas-clemente/quic-go"
 
 	ma "github.com/multiformats/go-multiaddr"
 	madns "github.com/multiformats/go-multiaddr-dns"
@@ -439,15 +440,17 @@ func (s *Swarm) filterKnownUndialables(p peer.ID, addrs []ma.Multiaddr) []ma.Mul
 		}
 	}
 
-	return maybeRemoveWebTransportAddrs(ma.FilterAddrs(addrs,
-		func(addr ma.Multiaddr) bool { return !ma.Contains(ourAddrs, addr) },
-		s.canDial,
-		// TODO: Consider allowing link-local addresses
-		func(addr ma.Multiaddr) bool { return !manet.IsIP6LinkLocal(addr) },
-		func(addr ma.Multiaddr) bool {
-			return s.gater == nil || s.gater.InterceptAddrDial(p, addr)
-		},
-	))
+	return maybeRemoveWebTransportAddrs(
+		maybeRemoveQUICDraft29(
+			ma.FilterAddrs(addrs,
+				func(addr ma.Multiaddr) bool { return !ma.Contains(ourAddrs, addr) },
+				s.canDial,
+				// TODO: Consider allowing link-local addresses
+				func(addr ma.Multiaddr) bool { return !manet.IsIP6LinkLocal(addr) },
+				func(addr ma.Multiaddr) bool {
+					return s.gater == nil || s.gater.InterceptAddrDial(p, addr)
+				},
+			)))
 }
 
 // limitedDial will start a dial to the given peer when
@@ -536,9 +539,32 @@ func isWebTransport(addr ma.Multiaddr) bool {
 	return err == nil
 }
 
-func isQUIC(addr ma.Multiaddr) bool {
-	_, err := addr.ValueForProtocol(ma.P_QUIC)
-	return err == nil && !isWebTransport(addr)
+func quicVersion(addr ma.Multiaddr) (quic.VersionNumber, bool) {
+	found := false
+	foundWebTransport := false
+	var version quic.VersionNumber
+	ma.ForEach(addr, func(c ma.Component) bool {
+		switch c.Protocol().Code {
+		case ma.P_QUIC:
+			version = quic.VersionDraft29
+			found = true
+			return true
+		case ma.P_QUIC_V1:
+			version = quic.Version1
+			found = true
+			return true
+		case ma.P_WEBTRANSPORT:
+			version = quic.Version1
+			foundWebTransport = true
+			return false
+		default:
+			return true
+		}
+	})
+	if foundWebTransport {
+		return 0, false
+	}
+	return version, found
 }
 
 // If we have QUIC addresses, we don't want to dial WebTransport addresses.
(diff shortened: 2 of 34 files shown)


p2p/net/swarm/swarm_addr.go
```diff
@@ -18,7 +18,7 @@ func (s *Swarm) ListenAddresses() []ma.Multiaddr {
 func (s *Swarm) listenAddressesNoLock() []ma.Multiaddr {
 	addrs := make([]ma.Multiaddr, 0, len(s.listeners.m))
```
marten-seemann, line 19: Should we preallocate a few more here? Maybe 5 or 10?

p2p/net/swarm/swarm_listen.go
```diff
@@ -43,7 +43,7 @@ func (s *Swarm) ListenClose(addrs ...ma.Multiaddr) {
 
 	s.listeners.Lock()
 	for l := range s.listeners.m {
-		if !containsMultiaddr(addrs, l.Multiaddr()) {
+		if !containsSomeMultiaddr(addrs, l.Multiaddrs()) {
```
marten-seemann, line 46: We should extend the documentation for this function that we close a listener listening on multiple addresses if _any_ address matches.

p2p/transport/quic/transport.go
```diff
@@ -261,7 +272,15 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
 		return nil, err
 	}
-	qconn, err := quicDialContext(ctx, pconn, addr, host, tlsConf, t.clientConfig)
+
+	clientConfig := t.clientConfig
+	if v == quic.Version1 {
+		// The endpoint has explicit support for version 1, so we'll only use that version.
+		clientConfig = t.clientConfig.Clone()
+		clientConfig.Versions = []quic.VersionNumber{quic.Version1}
+	}
```
marten-seemann, line 281: Should we have the equivalent logic for `v == quic.VersionDraft29`?

marten-seemann (commented):
This turned out to be a much larger change than I expected. Good work here.

Contrary to what the PR title suggests, we're not advertising draft-29 for WebTransport, right? This is the reasonable thing to do. We don't need draft-29 to support WebTransport (it's unclear if it's even RFC-compliant to run WebTransport over draft versions).

p2p/transport/quic/listener.go
```diff
@@ -19,18 +20,18 @@ var quicListen = quic.Listen // so we can mock it in tests
...
-	privKey        ic.PrivKey
-	localPeer      peer.ID
-	localMultiaddr ma.Multiaddr
+	quicListener    quic.Listener
+	conn            pConn
+	transport       *transport
+	rcmgr           network.ResourceManager
+	privKey         ic.PrivKey
+	localPeer       peer.ID
+	localMultiaddrs map[quic.VersionNumber]ma.Multiaddr
```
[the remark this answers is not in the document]
MarcoPolo: It's a bit cleaner when we access them: https://github.com/libp2p/go-libp2p/blob/marco/quic-v1/p2p/transport/quic/listener.go#L143

Otherwise we'd have to branch and update a variable. Not really a problem, just a bit less tidy imo. I don't have a strong opinion here so if you feel strongly we can use two vars.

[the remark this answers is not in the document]
marten-seemann: No strong opinion either, but using two vars would make the return order of `LocalMultiAddr` deterministic, which would be nice (see https://github.com/libp2p/go-libp2p/pull/1881#discussion_r1023319759).

p2p/transport/quic/quic_multiaddr.go
```diff
@@ -1,30 +1,62 @@
...
+	case quic.Version1:
+		return udpMA.Encapsulate(quicV1MA), nil
+	default:
+		return nil, errors.New("unknown quic version")
+	}
 }
 
-func fromQuicMultiaddr(addr ma.Multiaddr) (net.Addr, error) {
-	return manet.ToNetAddr(addr.Decapsulate(quicMA))
+func fromQuicMultiaddr(addr ma.Multiaddr) (net.Addr, quic.VersionNumber, error) {
```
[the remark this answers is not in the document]
MarcoPolo: Slight preference for keeping this such that we iterate the multiaddr fewer times. (I've even made a change to join the multiaddrs once at the end instead of each step).

p2p/host/resource-manager/limit_defaults.go
```diff
@@ -492,8 +492,8 @@ var DefaultLimits = ScalingLimitConfig{
 	},
 
 	PeerBaseLimit: BaseLimit{
-		ConnsInbound:    4,
-		ConnsOutbound:   8,
+		ConnsInbound:    6,
+		ConnsOutbound:   6,
```
marten-seemann, line 496: Why are you modifying the limits here?

p2p/net/swarm/swarm_dial.go
```diff
@@ -568,3 +594,38 @@ func maybeRemoveWebTransportAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
...
 	return addrs[:c]
 }
+
+// If we have QUIC V1 addresses, we don't want to dial QUIC draft29 addresses.
+// This is a similar hack to the above. If we add one more hack like this, let's
+// define a `Filterer` interface like the `Resolver` interface that transports
+// can optionally implement if they want to filter the multiaddrs.
+//
+// This mutates the input
+func maybeRemoveQUICDraft29(addrs []ma.Multiaddr) []ma.Multiaddr {
```
marten-seemann, line 604: Good catch!

p2p/transport/quic/transport.go
```diff
@@ -446,13 +478,20 @@ func (t *transport) Proxy() bool {
...
 }
 
 func (t *transport) String() string {
 	return "QUIC"
 }
 
 func (t *transport) Close() error {
-	return t.connManager.Close()
+	// We may register multiple protocols, but they are all the same transport, so only support getting closed once.
+	t.closeOnce.Do(func() {
```
marten-seemann, line 493: Why is this needed? Can't `connManager.Close` be called multiple times?

p2p/host/resource-manager/limit_defaults.go
MarcoPolo: A bit more info in the commit: https://github.com/libp2p/go-libp2p/commit/a969c6d41aa21a18901a5d45954bc91560c8a53c

We see flakiness in CI because a client will dial with concurrency 8 but the limit on the server is 4 so we get rcmgr failures. We notice this when we have a lot of addrs to dial

p2p/transport/quic/transport.go
MarcoPolo: no it can't be. Should it be?

p2p/transport/quic/listener.go
```diff
@@ -164,6 +182,10 @@ func (l *listener) Addr() net.Addr {
 }
 
 // Multiaddr returns the multiaddress of this listener.
-func (l *listener) Multiaddr() ma.Multiaddr {
-	return l.localMultiaddr
+func (l *listener) Multiaddrs() []ma.Multiaddr {
+	mas := make([]ma.Multiaddr, 0, len(l.localMultiaddrs))
+	for _, a := range l.localMultiaddrs {
+		mas = append(mas, a)
```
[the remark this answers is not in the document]
MarcoPolo: Slight preference for keeping this non-deterministic so we don't encode an assumption here.

marten-seemann:
@MarcoPolo I just released quic-go v0.31.0. Let's get https://github.com/libp2p/go-libp2p/pull/1882 merged and then rebase this PR?

p2p/transport/quic/quic_multiaddr.go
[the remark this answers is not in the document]
marten-seemann: In general, it feels "wrong" to do that much multiaddr magic in libp2p packages. Maybe the API we expose in `multiaddr` is insufficient? Or are we over-optimizing things here?

Anyway, no strong feelings, I'm ok with keeping this as is, just something to think about.

p2p/transport/quic/listener.go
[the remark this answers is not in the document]
marten-seemann: Fair point.

p2p/transport/quic/transport.go
MarcoPolo: Let's try to dedupe this at the higher level

p2p/transport/quic/quic_multiaddr.go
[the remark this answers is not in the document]
MarcoPolo: > Maybe the API we expose in multiaddr is insufficient

I think so. I think there are some things we can add that would replace 80% of our uses of `ForEach`.

marten-seemann (approved):

PR #1883, opened 2022-11-16
webtransport: use more randomness in tests of deterministic cert hashes

marten-seemann:


p2p/transport/webtransport/cert_manager_test.go
```diff
@@ -113,11 +114,11 @@ func TestCertRenewal(t *testing.T) {
 
 func TestDeterministicCertsAcrossReboots(t *testing.T) {
 	// Run this test 100 times to make sure it's deterministic
-	runs := 100
+	const runs = 100
 	for i := 0; i < runs; i++ {
 		t.Run(fmt.Sprintf("Run=%d", i), func(t *testing.T) {
 			cl := clock.NewMock()
-			priv, _, err := test.SeededTestKeyPair(crypto.Ed25519, 256, 0)
+			priv, _, err := crypto.GenerateEd25519Key(rand.Reader)
```
MarcoPolo, line 121: We don't want to do this because this will cause the test to be flakey.

Imagine the cert is generated that will transition to the next cert in 30 minutes, this test would fail.

We already have other quic check tests that use random keys, we don't want to do this here. This test should simply make sure the most basic test with a specific key works no matter how many times we run it. This test fails if any bit of our cert generation is non-deterministic, as was the case before our single byte reader hack.

MarcoPolo (changes requested):
I don't think we should do this for the point listed above.

p2p/transport/webtransport/cert_manager_test.go
MarcoPolo: Maybe we should add this as a comment so we don't try to do this in the future. That's my fault.

PR #1884, opened 2022-11-16
webtransport: tame log output of tests

marten-seemann:
Every `t.Run` emits one log line, making the test output very lengthy.

### DIFF
--- a/p2p/transport/webtransport/cert_manager_test.go
+++ b/p2p/transport/webtransport/cert_manager_test.go
@@ -3,7 +3,6 @@ package libp2pwebtransport
 import (
 	"crypto/sha256"
 	"crypto/tls"
-	"fmt"
 	"testing"
 	"testing/quick"
 	"time"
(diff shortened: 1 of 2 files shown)


MarcoPolo:
Is this really a problem? You don't see these lines when the test pass. And when they don't you get information for how flaky the test is.

p2p/transport/webtransport/cert_manager_test.go
```diff
@@ -115,30 +114,28 @@ func TestDeterministicCertsAcrossReboots(t *testing.T) {
...
-			newCerts := m.serializedCertHashes
-
-			require.Equal(t, oldCerts, newCerts)
-		})
+		cl := clock.NewMock()
+		priv, _, err := test.SeededTestKeyPair(crypto.Ed25519, 256, 0)
+		require.NoError(t, err)
+		m, err := newCertManager(priv, cl)
+		require.NoError(t, err)
+		defer m.Close()
```
MarcoPolo, line 122: This shouldn't be here, since we are closing this below (this is my bug)

p2p/transport/webtransport/cert_manager_test.go
```diff
@@ -115,30 +114,28 @@ func TestDeterministicCertsAcrossReboots(t *testing.T) {
...
+		require.Len(t, conf.Certificates, 1)
+		oldCerts := m.serializedCertHashes
+
+		m.Close()
+
+		cl.Add(time.Hour)
+		// reboot
+		m, err = newCertManager(priv, cl)
+		require.NoError(t, err)
+		defer m.Close()
```
MarcoPolo, line 134: This defer doesn't work as before anymore

MarcoPolo (approved):
I don't feel strongly about having these in the a `t.Run` or not. We should fix the `defer` though. Besides that this looks good.

PR #1885, opened 2022-11-16
upgrader: absorb the muxer_multistream.Transport into the upgrader

### LINKED ISSUE #1854: Remove multistream.transport as an entity of type network.multiplexer

### LINKED ISSUE #1873: config: use a more concrete type to pass muxers to upgrader
With #1854, we'll need to pass the muxer along with its ID to the upgrader. This will resolve the potential for confusion in the future here. I suggest merging this PR as is, given that the fix for that issue is imminent.

_Originally posted by @marten-seemann in https://github.com/libp2p/go-libp2p/pull/1858#discussion_r1019430022_

marten-seemann:
Fixes #1873. Fixes #1854. Closes #1861.


config/config.go
```diff
@@ -168,31 +167,21 @@ func (cfg *Config) addTransports(h host.Host) error {
 		return fmt.Errorf("swarm does not support transports")
 	}
 
-	muxers := make([]protocol.ID, 0, len(cfg.Muxers))
```
MarcoPolo, line 171: Good riddance!

MarcoPolo (approved):

PR #1887, opened 2022-11-16
add an integration test for muxer selection

### LINKED ISSUE #1836: Create integration test for libp2p early muxer negotiation

marten-seemann:
Fixes #1836. Depends on #1885. Closes #1851. Closes #1840.

The test currently fails since apparently the early muxer negotiation respects the server's preferences, not the client's. We should fix that. Will do in another PR.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -182,6 +182,7 @@ func (cfg *Config) addTransports(h host.Host) error {
 			fx.ParamTags(`group:"security"`),
 		)),
 		fx.Supply(cfg.Muxers),
+		fx.Supply(h.ID()),
 		fx.Provide(func() host.Host { return h }),
 		fx.Provide(func() crypto.PrivKey { return h.Peerstore().PrivKey(h.ID()) }),
 		fx.Provide(func() connmgr.ConnectionGater { return cfg.ConnectionGater }),
--- a/p2p/net/upgrader/upgrader.go
+++ b/p2p/net/upgrader/upgrader.go
@@ -171,7 +171,7 @@ func (u *upgrader) upgrade(ctx context.Context, t transport.Transport, maconn ma
 		}
 	}
 
-	smconn, err := u.setupMuxer(ctx, sconn, server, connScope.PeerScope())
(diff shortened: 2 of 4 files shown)


MarcoPolo:
fyi I moved the base to `remove-muxer-multistream`

p2p/net/upgrader/upgrader.go
```diff
@@ -257,17 +263,18 @@ func (u *upgrader) setupMuxer(ctx context.Context, conn sec.SecureConn, server b
 		if err != nil {
 			return
 		}
+		muxerID = m.ID
 		smconn, err = m.Muxer.NewConn(conn, server, scope)
 	}()
 
 	select {
 	case <-done:
```
MarcoPolo, line 271: Minor nit, I think it would be better to return the value here (a la message passing) rather than mutating shared state.

config/config.go
```diff
@@ -182,6 +182,7 @@ func (cfg *Config) addTransports(h host.Host) error {
 			fx.ParamTags(`group:"security"`),
 		)),
 		fx.Supply(cfg.Muxers),
+		fx.Supply(h.ID()),
```
MarcoPolo, line 185: Why this change? Did this not work before?

MarcoPolo (approved):
lg, one nit.

config/config.go
marten-seemann: The insecure transport consume the `peer.ID`: https://github.com/libp2p/go-libp2p/blob/d8d2efafdd368bc4911c734ee284ff285fcdfbca/core/sec/insecure/insecure.go#L43

The other transports consume the private key.

marten-seemann:
Rebased and changed merge target back to `master`.

p2p/net/upgrader/upgrader.go
marten-seemann: I think I agree. Changed.

PR #1888, opened 2022-11-16
tls / noise: prefer the client's muxer preferences

marten-seemann:
As the integration test (#1887) has revealed, TLS and Noise both selected the muxer according to the server's preferences, whereas the multistream negotiation selects according to the client's preferences. This PR changes TLS' and Noise's behavior to match the multistream behavior.

Since TLS' ALPN selection algorithm also selects according to the server's preferences, we have to play a little trick to overwrite that behavior.

### DIFF
--- a/p2p/security/tls/transport.go
+++ b/p2p/security/tls/transport.go
@@ -68,7 +68,26 @@ func (t *Transport) SecureInbound(ctx context.Context, insecure net.Conn, p peer
 	for _, muxer := range t.muxers {
 		muxers = append(muxers, string(muxer))
 	}
-	// Prepend the prefered muxers list to TLS config.
+	// TLS' ALPN selection lets the server select the protocol, preferring the server's preferences.
+	// We want to prefer the client's preference though.
+	getConfigForClient := config.GetConfigForClient
+	config.GetConfigForClient = func(info *tls.ClientHelloInfo) (*tls.Config, error) {
+	alpnLoop:
+		for _, proto := range info.SupportedProtos {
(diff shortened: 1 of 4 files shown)


p2p/security/tls/transport.go
```diff
@@ -63,7 +63,24 @@ func (t *Transport) SecureInbound(ctx context.Context, insecure net.Conn, p peer
...
-	// Prepend the prefered muxers list to TLS config.
+	// TLS' ALPN selection lets the server select the protocol, preferring the server's preferences.
+	// We want to prefer the client's preference though.
+	getConfigForClient := config.GetConfigForClient
+	config.GetConfigForClient = func(info *tls.ClientHelloInfo) (*tls.Config, error) {
+	alpnLoop:
+		for _, proto := range info.SupportedProtos {
+			for _, m := range muxers {
+				if m == proto {
+					config.NextProtos = []string{proto}
```
MarcoPolo, line 74: Don't we have to append the "libp2p" ALPN here? (It seems like were doing that below? with `	config.NextProtos = append(muxers, config.NextProtos...)`). I'm pretty sure this doesn't matter, since we know the client supports the ALPN we set here. So maybe just a comment would be fine.

MarcoPolo (approved):
one nit about the var names, but otherwise lgtm

p2p/security/tls/transport.go
marten-seemann: Technically we have to, but we can short-circuit it, since we know that we'll select the muxer. I'll add a comment.

PR #1895, opened 2022-11-17
webtransport: update webtransport-go to v0.3.0

marten-seemann:
Not really sure what to do with the error codes here. At some point, we probably should standardize libp2p error codes, and at that point, all the codes we've been using so far will become deprecated, so that we can distinguish between new and legacy nodes.

### DIFF
--- a/p2p/transport/webtransport/transport.go
+++ b/p2p/transport/webtransport/transport.go
@@ -34,6 +34,8 @@ var log = logging.Logger("webtransport")
 
 const webtransportHTTPEndpoint = "/.well-known/libp2p-webtransport"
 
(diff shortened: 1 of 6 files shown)


p2p/transport/webtransport/transport.go
```diff
@@ -34,6 +34,8 @@ var log = logging.Logger("webtransport")
 
 const webtransportHTTPEndpoint = "/.well-known/libp2p-webtransport"
 
+const errorCodeConnectionGating = 0x47415445 // GATE in ASCII
```
MarcoPolo, line 37: nice

MarcoPolo (approved):

Issue #1816, opened 2022-10-09
circuitv2: flaky TestReservationFailures


marten-seemann:
https://github.com/libp2p/go-libp2p/actions/runs/3208002096/jobs/5243462098

```
  === RUN   TestReservationFailures/invalid_time
      reservation_test.go:100: 
          	Error Trace:	/home/runner/work/go-libp2p/go-libp2p/p2p/protocol/circuitv2/client/reservation_test.go:100
          	Error:      	"error writing reservation message: Application error 0x0: peer:12D3KooWQMrNzVJdPSsgvtwFvZ4jPLYNjXZ1Xc4BAPWtA4MGAtMy: cannot reserve inbound connection: resource limit exceeded" does not contain "received reservation with expiration date in the past"
          	Test:       	TestReservationFailures/invalid_time
  === RUN   TestReservationFailures/invalid_voucher
  --- FAIL: TestReservationFailures (0.89s)
      --- PASS: TestReservationFailures/unsupported_protocol (0.20s)
      --- PASS: TestReservationFailures/wrong_message_type (0.14s)
      --- PASS: TestReservationFailures/unknown_status (0.17s)
      --- FAIL: TestReservationFailures/invalid_time (0.18s)
      --- PASS: TestReservationFailures/invalid_voucher (0.20s)

Issue #1765, opened 2022-09-20
circuitv2: flaky TestReservationFailures


marten-seemann:
https://github.com/libp2p/go-libp2p/actions/runs/3092616835/jobs/5004089751

```
  === RUN   TestReservationFailures/unsupported_protocol
      reservation_test.go:100: 
          	Error Trace:	D:\a\go-libp2p\go-libp2p\p2p\protocol\circuitv2\client\reservation_test.go:100
          	Error:      	"Application error 0x0: peer:12D3KooWKfsfaScKKyfhMAsdiRPKqDiK7TYWf5Lq8EmLXKELS67X: cannot reserve inbound connection: resource limit exceeded" does not contain "protocol not supported"
          	Test:       	TestReservationFailures/unsupported_protocol

Issue #1730, opened 2022-09-02
circuitv2: flaky TestReservationFailures/unsupported_protocol


marten-seemann:
```
  === RUN   TestReservationFailures/unsupported_protocol
      reservation_test.go:100: 
          	Error Trace:	D:\a\go-libp2p\go-libp2p\p2p\protocol\circuitv2\client\reservation_test.go:100
          	Error:      	"Application error 0x0: peer:12D3KooWDuwgsbf9CcVe1euAXgjn8CxtHKjQdhuLGiJ7Na4NWinE: cannot reserve inbound connection: resource limit exceeded" does not contain "protocol not supported"
          	Test:       	TestReservationFailures/unsupported_protocol
  === RUN   TestReservationFailures/wrong_message_type
  === RUN   TestReservationFailures/unknown_status
  === RUN   TestReservationFailures/invalid_time
  === RUN   TestReservationFailures/invalid_voucher
  --- FAIL: TestReservationFailures (0.64s)
      --- FAIL: TestReservationFailures/unsupported_protocol (0.20s)
      --- PASS: TestReservationFailures/wrong_message_type (0.09s)
      --- PASS: TestReservationFailures/unknown_status (0.08s)
      --- PASS: TestReservationFailures/invalid_time (0.17s)
      --- PASS: TestReservationFailures/invalid_voucher (0.10s)

Issue #1668, opened 2022-08-09
circuitv2: flaky TestReservationFailures


marten-seemann:
https://github.com/libp2p/go-libp2p/runs/7748150365?check_suite_focus=true

```
  === RUN   TestReservationFailures
  === RUN   TestReservationFailures/unsupported_protocol
  2022/08/09 14:31:47 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
  === RUN   TestReservationFailures/wrong_message_type
  === RUN   TestReservationFailures/unknown_status
  === RUN   TestReservationFailures/invalid_time
      reservation_test.go:101: 
          	Error Trace:	/home/runner/work/go-libp2p/go-libp2p/p2p/protocol/circuitv2/client/reservation_test.go:101
          	Error:      	"error writing reservation message: Application error 0x0: peer:QmQkiUAs4viJ4L6VYx1f4QTcc8oWcCsk8qxDxU9efYK3Fs: cannot reserve inbound connection: resource limit exceeded" does not contain "received reservation with expiration date in the past"
          	Test:       	TestReservationFailures/invalid_time
  === RUN   TestReservationFailures/invalid_voucher
  --- FAIL: TestReservationFailures (8.07s)
      --- PASS: TestReservationFailures/unsupported_protocol (2.04s)
      --- PASS: TestReservationFailures/wrong_message_type (1.53s)
      --- PASS: TestReservationFailures/unknown_status (1.77s)
      --- FAIL: TestReservationFailures/invalid_time (1.72s)
      --- PASS: TestReservationFailures/invalid_voucher (1.01s)

PR #1899, opened 2022-11-17
circuitv2: disable the resource manager in tests

marten-seemann:
Fixes https://github.com/libp2p/go-libp2p/issues/1816
Fixes https://github.com/libp2p/go-libp2p/issues/1765
Fixes https://github.com/libp2p/go-libp2p/issues/1730
Fixes https://github.com/libp2p/go-libp2p/issues/1668


MarcoPolo:
Let's prefer: https://github.com/libp2p/go-libp2p/pull/1898

marten-seemann:
We should do both. There’s no need to run the resource manager in these tests, as that’s not what we’re testing here.

PR #1901, opened 2022-11-17
identify: simplify TTL handling of existing addresses

marten-seemann:
While debugging flaky identify tests I came across the address TTL logic.

Current logic:
* take all addrs with `RecentlyConnectedAddrTTL` and set to `TempAddrTTL`
* take all addrs with `ConnectedAddrTTL` and set to `TempAddrTTL`
* add new addrs with `RecentlyConnectedAddrTTL` / ` ConnectedAddrTTL`
* remove all addrs with `TempAddrTTL`

This should be equivalent to:
* remove all addrs with `RecentlyConnectedAddrTTL`, `ConnectedAddrTTL` and `TempAddrTTL`
* add new addrs with `RecentlyConnectedAddrTTL` / ` ConnectedAddrTTL`

**Question**: Why not just just clear all addresses at the beginning (using `Peerstore.ClearAddrs`)? I assume this is because we might have learned other addresses this peer might have from somewhere other than Identify. However, those addresses would only be protected if they had a TTL other than `ConnectedAddrTTL`, `RecentlyConnectedAddrTTL` or `TempAddrTTL`. It seems like identifying how protection-worthy an addresses is by its TTL is kind of crude. @Stebalien, do you have more context here?

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -611,12 +611,13 @@ func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn) {
 		ttl = peerstore.ConnectedAddrTTL
 	}
 
-	// Downgrade connected and recently connected addrs to a temporary TTL.
+	// Expire all temporary addrs, connected and recently connected addrs.
 	for _, ttl := range []time.Duration{
 		peerstore.RecentlyConnectedAddrTTL,
 		peerstore.ConnectedAddrTTL,
+		peerstore.TempAddrTTL,
 	} {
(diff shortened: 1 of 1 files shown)


Stebalien:
I think I did it this way so we'd always have addresses. Otherwise, we might be connected to some peer but not have addresses for that peer, which was, IIRC, a problem? But this was ages ago and it may not be a real issue anymore.

I totally agree, this mechanism is terrible.

MarcoPolo (approved):

MarcoPolo:
This has changed, this PR is no longer applicable.

Issue #1906, opened 2022-11-20
security protocol ordering preference broken


marten-seemann:
Since fx groups don't provide ordering (see https://pkg.go.dev/go.uber.org/fx#hdr-Value_Groups), we end up not adhering to the ordering of the configured security protocols.

We need to:
- [x] fix this bug
- [x] add an integration test

This is a blocker for the v0.24.0 release.

vyzo:
fx.... lol.

marten-seemann:
@vyzo Enjoy your told-you-so moment.

The fact that we didn't have any tests for security protocol selection is the real culprit here, as I see it.

vyzo:
it was a small triumph :)

Issue #1356, opened 2022-03-11
Exposing Metrics


marten-seemann:
libp2p currently collects a few metrics (for example, for [TCP](https://github.com/libp2p/go-tcp-transport/blob/master/metrics.go) and [QUIC](https://github.com/libp2p/go-libp2p-quic-transport/blob/master/tracer_metrics.go)). We also collect metrics for the resource manager, however, that is currently left as a responsibility to the application. This is how we end up with slightly different implementations (and metrics names) in [go-ipfs](https://github.com/ipfs/go-ipfs/pull/8785) and [lotus](https://github.com/filecoin-project/lotus/pull/8089).
In the (very near!) future, we want to expose more metrics from the swarm regarding transports, security protocols and muxers. This would have allowed us to detect the muxer prioritisation bug (https://github.com/ipfs/go-ipfs/pull/8750) a lot earlier.

**We therefore should have a coherent libp2p metrics story.**

---

Open questions:
1. In the past, we debated if the metrics code should live within the respective repository, or if that code should live separately and hook in via tracers. Note that this issue mostly goes away once we make progress with our repo consolidation (and that's what we should base our designs on).
1. Some people have expressed concerns introducing an OpenCensus / Prometheus dependency into the default build introduces too much bloat. Is that a valid concern? Does it really add too much overhead? If so, would it make sense to hide all metrics collection across our code base behind a build flag?
1. OpenCensus vs. Prometheus: In the past, we've used both tools, very inconsistently. If I understand correctly, OpenCensus is essentially another abstraction layer between the tracer and Prometheus, and would _in theory_ allow the usage of other tools than Prometheus, although it might be debatable how much we care about that.

Thoughts, @vyzo @aschmahmann @mxinden @Stebalien @lanzafame?

---

Tracking the various components we want to instrument:
* [x] Resource Manager
* [x] Swarm: https://github.com/libp2p/go-libp2p/issues/1910
    * [x] https://github.com/libp2p/go-libp2p/issues/2021
    * [x] https://github.com/libp2p/go-libp2p/issues/2106
* [x] AutoNAT: https://github.com/libp2p/go-libp2p/issues/2017
* [x] Identify: https://github.com/libp2p/go-libp2p/issues/2019
* [x] Relay service: https://github.com/libp2p/go-libp2p/issues/2018
* [x] Eventbus: https://github.com/libp2p/go-libp2p/issues/2020
[...]

PR #1907, opened 2022-11-20
expose the security protocol on the ConnectionState

marten-seemann:
In preparation for the integration test needed for #1906. Will also be needed to report swarm metrics (#1356).


core/network/conn.go
```diff
@@ -34,12 +34,12 @@ type Conn interface {
...
+// ConnectionState holds information about the connection.
 type ConnectionState struct {
-	// The next protocol used for stream muxer selection. This is derived from
-	// security protocol handshake, for example, Noise handshake payload or
-	// TLS/ALPN negotiation.
-	NextProto string
+	// The stream multiplexer used on this connection (if any).
+	StreamMultiplexer string
+	// The security protocol used on this connection (if any).
+	Security string
```
MarcoPolo, line 42: 💯 Thanks for this rename. I think this makes a lot more sense.

MarcoPolo (approved):

PR #1908, opened 2022-11-20
respect the user's security protocol preference order

### LINKED ISSUE #1906: security protocol ordering preference broken

marten-seemann:
Fixes #1906. Depends on #1907.


marten-seemann:
GitHub wtf: I deleted the `expose-security` branch, because #1907 was merged. Then GitHub automatically closed this PR, which was still targeting `expose-security`. Now I can't retarget it to `master`, because for some reason you can't retarget closed PRs. I also can't reopen it because `expose-security` is gone. Well thanks for that.

PR #1912, opened 2022-11-21
respect the user's security protocol preference order

### LINKED ISSUE #1906: security protocol ordering preference broken

marten-seemann:
Recreating #1908, which was killed by GitHub (see https://github.com/libp2p/go-libp2p/pull/1908#issuecomment-1322694258).

Fixes #1906.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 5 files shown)


MarcoPolo (approved):
This is a bit gross, but I don't really have a better idea.

Issue #1432, opened 2019-08-13
Private network support


Stebalien:

Kubuxu:
Hey, I missed this. I will review in next few days.

marten-seemann:
@Kubuxu Nothing new here. I just moved the issue to go-libp2p because we've moved go-libp2p-quic-transport into go-libp2p. No idea why GitHub is sending out notifications for that :(

Issue #1249, opened 2021-11-30
PrivateNetwork option broken by making QUIC a default


hsanjuan:
Making QUIC a default transport automatically breaks any node initialized with `PrivateNetwork`.

This should have probably been handled somehow. There is no constructor option to disable QUIC, so we are left to manually choosing the transports.

marten-seemann:
We should:
1. if the default transports are set: disable QUIC if `PrivateNetwork` is set, then (re?) run the check if any transports are configured
2. if QUIC is configured explicitly: throw an error

PR #1921, opened 2022-11-28
use a different set of default transports when PSK is enabled

RobotSail:
### Problem

The current behavior in `go-libp2p` is that the transports are provided through `fx` by default when no other options are specified. 
QUIC is listed in the defaults but does not support PSK usage, so creating a new app will always result in an error.
The way that libp2p is currently bootstrapped does not allow for dynamically enabling/disabling transports, so we cannot disable QUIC once it's been provided as a value group.

### Solution

This PR adds a new default transport option that gets activated when `cfg.Transports == nil && cfg.PSK != nil`. The new option is mutually exclusive with the current one on the condition of `cfg.PSK == nil`.  

Signed-off-by: Oleg <97077423+RobotSail@users.noreply.github.com>

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -45,6 +45,15 @@ var DefaultTransports = ChainOptions(
 	Transport(ws.New),
 )
 
+// DefaultPrivateTransports are the default libp2p transports when a PSK is supplied.
+//
+// Use this option when you want to *extend* the set of transports used by
+// libp2p instead of replacing them.
+var DefaultPrivateTransports = ChainOptions(
+	Transport(tcp.NewTCPTransport),
+	Transport(ws.New),
+)
+
 // DefaultPeerstore configures libp2p to use the default peerstore.
 var DefaultPeerstore Option = func(cfg *Config) error {
 	ps, err := pstoremem.NewPeerstore()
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
There's a bigger problem here, namely that most transports don't support PSK at all. We just made it explicit for QUIC, but not for the others (see https://github.com/libp2p/go-libp2p/issues/1432 for context).

The correct fix would be to throw an error when constructing transports that don't support PSK, when a PSK is configured. Unfortunately, that's not a trivial change.

The even more correct fix would be to get rid of PSK, as it's clearly not useful if it can't be used with most of our transports.

RobotSail:
@marten-seemann Is there any transport that supports PSK which we could use? The PSK is important to our use-case in particular.

marten-seemann:
TCP and WebSocket support PSK. Would you mind describing your use case?

RobotSail:
We're building an IPFS Operator to run in Kubernetes. One of our key features is orchestrating large-scale private networks, but we're using the libp2p circuit relay in order to allow individual IPFS nodes to be dialable by each other due to the amount of network abstractions that take place in a cloud environment. 
The reason for this PR is so that we can initialize `go-libp2p-relay-circuit` with a PSK and be able to properly handle peer-to-peer circuits while maintaining privacy.

marten-seemann:
I'm not sure I understand how this helps maintain privacy. There's no such thing as unencrypted libp2p connections, so all connections are private anyway.
What PSK allows you is to conveniently separate your nodes from other libp2p / IPFS nodes, but at the heavy cost of double-encryption and limiting your transports to TCP and WebSocket.

RobotSail:
@marten-seemann We're doing this so that enterprise users can create their own private IPFS networks that would run at scale but without advertising their CIDs to the global DHT. I think it would be fine if the transports are limited to TCP & WebSocket, but we haven't been able to try running it at scale yet to actually see what the drawbacks would be.

marten-seemann:
Thank you for the explanation. I think this should be part of a larger discussion around PSK, and I’ve opened https://github.com/libp2p/specs/issues/489 for that.

marten-seemann:
There’s an issue for this PR from a while ago: https://github.com/libp2p/go-libp2p/issues/1249.
For some reason, GitHub doesn’t allow me to link it to this PR.

marten-seemann (approved):
Actually, this PR is fine, and an improvement independent from the larger discussion about the future of pnet.

PR #1923, opened 2022-11-29
transport.Listener,quic: Support multiple QUIC versions with the same Listener. Only return a single multiaddr per listener.

### LINKED ISSUE #1904: quic: QUIC listener doesn't respect the QUIC version of the listen address
```go
func TestListenOnlyOnOneVersion(t *testing.T) {
	t.Run("only one draft-29", func(t *testing.T) {
		h1, err := libp2p.New(
			libp2p.Transport(libp2pquic.NewTransport),
			libp2p.ListenAddrStrings(
				"/ip4/127.0.0.1/udp/12345/quic", // QUIC draft-29
			),
		)
		require.NoError(t, err)
		defer h1.Close()

		addrs := h1.Addrs()
		require.Len(t, addrs, 1)
		require.Equal(t, ma.P_QUIC, getQUICMultiaddrCode(addrs[0]))
	})
}
```

This test fails, as the listener is listening on both versions.

MarcoPolo:
Fixes #1904

Undoes part of the work in https://github.com/libp2p/go-libp2p/pull/1881. The QUIC transport now returns a virtual listener that only handles one QUIC version. If a user then listens on another QUIC version with the same multiaddr, we return another virtual listener that uses the same underlying listener but only advertises the other QUIC version.

This is ready for review, but I see some flakiness in the tests (maybe also from #1905 ) so I'll continue debugging that.

### DIFF
--- a/p2p/transport/quic/virtuallistener.go
+++ b/p2p/transport/quic/virtuallistener.go
@@ -29,9 +29,10 @@ func (l *virtualListener) Multiaddr() ma.Multiaddr {
 }
 
 func (l *virtualListener) Close() error {
+	l.acceptRunnner.rmAcceptForVersion(l.version)
+
 	l.t.listenersMu.Lock()
 	defer l.t.listenersMu.Unlock()
-	l.acceptRunnner.rmAcceptForVersion(l.version)
 
 	var err error
 	listeners := l.t.listeners[l.udpAddr]
@@ -55,7 +56,16 @@ func (l *virtualListener) Close() error {
 }
 
 func (l *virtualListener) Accept() (tpt.CapableConn, error) {
-	v, ok := <-l.acceptChan
+	var v acceptVal
+	var ok bool
+	select {
+	// Check if we have a pending connection first
+	case v, ok = <-l.acceptChan:
+	default:
+		// No? Let's call Accept and wait for a connection
+		go l.acceptRunnner.Accept(l.listener, l.version)
+		v, ok = <-l.acceptChan
+	}
 	if !ok {
 		return nil, errors.New("listener closed")
 	}
@@ -69,6 +79,8 @@ type acceptVal struct {
 }
 
 type acceptLoopRunner struct {
+	listenerMu sync.Mutex
+
 	muxerMu sync.Mutex
 	muxer   map[quic.VersionNumber]chan acceptVal
 }
@@ -93,7 +105,7 @@ func (r *acceptLoopRunner) rmAcceptForVersion(v quic.VersionNumber) {
 
 	ch, ok := r.muxer[v]
 	if !ok {
-		panic("unexpected chan already found in accept muxer")
+		panic("expected chan in accept muxer")
 	}
 	ch <- acceptVal{err: errors.New("listener Accept closed")}
 	delete(r.muxer, v)
@@ -112,9 +124,12 @@ func (r *acceptLoopRunner) sendErrAndClose(err error) {
 	}
 }
 
-func (r *acceptLoopRunner) run(l *listener) error {
+func (r *acceptLoopRunner) Accept(l *listener, expectedVersion quic.VersionNumber) error {
 	for {
+		r.listenerMu.Lock()
 		conn, err := l.Accept()
+		r.listenerMu.Unlock()
+
 		if err != nil {
 			r.sendErrAndClose(err)
 			return err
@@ -143,6 +158,11 @@ func (r *acceptLoopRunner) run(l *listener) error {
 			// We dropped the connection, close it
 			conn.Close()
 			continue
+		}
+
+		if version == expectedVersion {
+			// We got the version we were expecting, we can exit.
+			return nil
 		}
 	}
 }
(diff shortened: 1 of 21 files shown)


p2p/transport/quic/virtuallistener.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+type acceptLoopRunner struct {
+	listenerMu sync.Mutex
+
+	muxerMu sync.Mutex
+	muxer   map[quic.VersionNumber]chan acceptVal
+}
+
+func (r *acceptLoopRunner) acceptForVersion(v quic.VersionNumber) chan acceptVal {
```
marten-seemann, line 88: Nit: `AcceptForVersion`. Makes it easier to spot that this is called from outside this struct.

p2p/transport/quic/virtuallistener.go
```diff
@@ -0,0 +1,168 @@
...
+
+	if _, ok := r.muxer[v]; ok {
+		panic("unexpected chan already found in accept muxer")
+	}
+
+	r.muxer[v] = ch
+	return ch
+}
+
+func (r *acceptLoopRunner) rmAcceptForVersion(v quic.VersionNumber) {
```
marten-seemann, line 102: Same here.

p2p/transport/quic/virtuallistener.go
```diff
@@ -0,0 +1,168 @@
...
+	return v.conn, v.err
+}
+
+type acceptVal struct {
+	conn tpt.CapableConn
+	err  error
+}
+
+type acceptLoopRunner struct {
+	listenerMu sync.Mutex
```
marten-seemann, line 82: Why is this mutex needed? I assume it's to protect against concurrent calls to `Accept` from multiple listeners? If so, I'd prefer to call this a semaphore (and make it a `sem chan struct` with cap 1).

p2p/transport/quic/virtuallistener.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (l *virtualListener) Accept() (tpt.CapableConn, error) {
+	var v acceptVal
+	var ok bool
+	select {
+	// Check if we have a pending connection first
+	case v, ok = <-l.acceptChan:
+	default:
+		// No? Let's call Accept and wait for a connection
+		go l.acceptRunnner.Accept(l.listener, l.version)
```
marten-seemann, line 66: Not sure I understand this logic here, especially the pending first connection.

I was wondering if `Accept` could this be driven by the Go routines that called `Accept`? It would be really nice if we didn't have to start another Go routine here. 
You could then retrieve connections from the underlying listener, and return them directly (if they're intended for this virtual listener), or enqueue them (if they're intended for a virtual listener of a different version).

MarcoPolo: Refactored to get rid of the goroutine. We still check first if we have previously enqueued a connection.

p2p/transport/quic/virtuallistener.go
```diff
@@ -0,0 +1,179 @@
...
+		err = l.listener.Close()
+		delete(l.t.listeners, l.udpAddr)
+		return err
+	}
+
+	for i := 0; i < len(listeners); i++ {
+		// Swap remove
+		if l == listeners[i] {
+			listeners[i] = listeners[len(listeners)-1]
+			listeners = listeners[0 : len(listeners)-1]
```
marten-seemann, line 50:
```suggestion
			listeners = listeners[:len(listeners)-1]
```

p2p/transport/quic/virtuallistener.go
```diff
@@ -0,0 +1,179 @@
...
+		// Nothing to handle this connection version. Close it
+		conn.Close()
+		return nil, nil
+	}
+
+	// Non blocking
+	select {
+	case ch <- acceptVal{conn: conn}:
+	default:
+		// We dropped the connection, close it
```
marten-seemann, line 160:
```suggestion
		// accept queue filled up, drop the connection
```

marten-seemann: Can you also add a log line here?

p2p/transport/quic/virtuallistener.go
```diff
@@ -0,0 +1,179 @@
...
+var _ tpt.Listener = &virtualListener{}
+
+func (l *virtualListener) Multiaddr() ma.Multiaddr {
+	return l.listener.localMultiaddrs[l.version]
+}
+
+func (l *virtualListener) Close() error {
+	l.acceptRunnner.RmAcceptForVersion(l.version)
+
+	l.t.listenersMu.Lock()
```
marten-seemann, line 34: It would be nice if the `virutalListener` didn't have to interact with that much internal state of the transport. Can we move all this logic into a method on the transport?

marten-seemann (approved):
LGTM modulo the review comments here.

Can you add some tests similar to the one suggested in https://github.com/libp2p/go-libp2p/issues/1904#issue-1456706966 in `p2p/test`:
* create a draft-29 listener and check that a draft-29 connection is accepted, but a v1 connection is closed
* the other way round, for a v1 listener
* create a draft-29 + v1 listener and check that both versions are accepted

PR #1927, opened 2022-12-01
webtransport: disable support for signed/static TLS certificates

MarcoPolo:
This needs more thought before we enable support.

### DIFF
--- a/p2p/transport/webtransport/transport.go
+++ b/p2p/transport/webtransport/transport.go
@@ -48,16 +48,6 @@ func WithClock(cl clock.Clock) Option {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


marten-seemann (approved):
Should we `t.Skip` the tests instead of deleting them? We do want this feature at some point in the future.

It would probably also makes sense to create an issue in this repo, and include a link in the error message and in the skip message.

marten-seemann:
Yes.

MarcoPolo:
If we use t.Skip we have to make changes since we removed the `WithTLSConfig` option. So I removed them. I think it's pretty easy to revive dead code from the git history, so I would prefer to keep it deleted and revive it when needed.

PR #1929, opened 2022-12-02
webtransport: add PSK to constructor, and fail if it is used

marten-seemann:
That way, it won't be possible to construct a host with a PSK when WebTransport is enabled. This is desireable since WebTransport doesn't support private network (same as QUIC).


MarcoPolo (approved):
Thanks `fx`!

PR #1938, opened 2022-12-08
webtransport: fix flaky accept queue test

### LINKED ISSUE #1870: webtransport: flaky TestAcceptQueueFilledUp
```
=== RUN   TestAcceptQueueFilledUp
    transport_test.go:539: 
        	Error Trace:	/home/runner/work/go-libp2p/go-libp2p/p2p/transport/webtransport/transport_test.go:539
        	            	error reading handshake message: stream reset, but failed to convert stream error 386759528: error code outside of expected range
        	Test:       	TestAcceptQueueFilledUp
--- FAIL: TestAcceptQueueFilledUp (10.04s)

### PR DESCRIPTION
**Pull request by marten-seemann**
Fixes #1870. 

The problem here is that it's not valid assumption that the 16 connection we dial first are all accepted (by the QUIC listener) in that order.
What we should assert that out of 17 connections dialed, 16 are accepted and 1 is rejected.

### DIFF
--- a/p2p/transport/webtransport/transport_test.go
+++ b/p2p/transport/webtransport/transport_test.go
@@ -473,17 +473,50 @@ func TestAcceptQueueFilledUp(t *testing.T) {
 		return cl.Dial(context.Background(), ln.Multiaddr(), serverID)
 	}
 
-	for i := 0; i < 16; i++ {
-		conn, err := newConn()
-		require.NoError(t, err)
-		defer conn.Close()
+	const num = 16 + 1 // one more than the accept queue capacity
+	// Dial one more connection than the accept queue can hold.
+	errChan := make(chan error, num)
+	for i := 0; i < num; i++ {
+		go func() {
+			conn, err := newConn()
+			if err != nil {
+				errChan <- err
+				return
+			}
+			_, err = conn.AcceptStream()
+			errChan <- err
+		}()
 	}
 
-	conn, err := newConn()
-	if err == nil {
-		_, err = conn.AcceptStream()
+	// Since the handshakes complete asynchronously, we won't know _which_ one is rejected,
+	// so the only thing we can test for is that exactly one connection attempt is rejected.
+	select {
+	case <-errChan:
+	case <-time.After(time.Second):
+		t.Fatal("expected one connection to be rejected")
 	}
-	require.Error(t, err)
+	select {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: p2p/transport/webtransport/transport_test.go [hunk 1]
```diff
@@ -473,17 +473,29 @@ func TestAcceptQueueFilledUp(t *testing.T) {
...
 	}
 
-	conn, err := newConn()
-	if err == nil {
-		_, err = conn.AcceptStream()
-	}
-	require.Error(t, err)
+	// Since the handshakes complete asynchronously, we won't know _which_ one is rejected,
+	// so the only thing we can test for is that exactly one connection attempt is rejected.
+	require.Eventually(t, func() bool { return len(errChan) > 0 }, time.Second, 10*time.Millisecond, "expected an error to occur")
```
MarcoPolo, line 493: Can we recv from this channel with a timeout instead of polling?

p2p/transport/webtransport/transport_test.go
```diff
@@ -473,17 +473,29 @@ func TestAcceptQueueFilledUp(t *testing.T) {
...
+	require.Never(t, func() bool { return len(errChan) > 1 }, 100*time.Millisecond, 10*time.Millisecond, "only expected one error to occur")
```
MarcoPolo, line 494: Can we recv with timeout here as well

p2p/transport/webtransport/transport_test.go
```diff
@@ -473,17 +473,29 @@ func TestAcceptQueueFilledUp(t *testing.T) {
...
+
+	// test shutdown
+	require.NoError(t, ln.Close())
+	require.Eventually(t, func() bool { return len(errChan) == num }, time.Second, 10*time.Millisecond)
```
MarcoPolo, line 498: Same here

MarcoPolo (approved):
Looks good, but I think would be nice to use receives insteads

p2p/transport/webtransport/transport_test.go
marten-seemann: Makes sense.

p2p/transport/webtransport/transport_test.go
marten-seemann: Yes.

PR #1946, opened 2022-12-12
routed host: return Connect error if FindPeer doesn't yield new addresses

marten-seemann:


p2p/host/routed/routed.go
```diff
@@ -135,10 +134,10 @@ func (rh *RoutedHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
 			pi.Addrs = newAddrs
 			return rh.host.Connect(ctx, pi)
 		}
-
-		return err
+		// No appropriate new address found.
+		// Return the original dial error.
+		return cerr
```
Jorropo, line 139: I'm just thinking but maybe it's worth returning an `errors.Join` of findPeerAddrs's error and cerr.
For now let's keep it as-is.

p2p/host/routed/routed_test.go
```diff
@@ -61,13 +55,36 @@ func TestRoutedHostConnectToObsoleteAddresses(t *testing.T) {
...
+	// Now, that function will return the correct multi addresses for host 2
+	// (we have fetched the most up-to-date data from the DHT)
+	mr := &mockRouting{findPeerFn: func(context.Context, peer.ID) (peer.AddrInfo, error) { return pi, nil }}
 
-	// The mocked FindPeer function should have been called
-	assert.Equal(t, 1, mr.callCount)
+	// Build routed host
+	rh := Wrap(h1, mr)
+	// Connection establishment should have worked without an error
+	require.Error(t, rh.Connect(context.Background(), pi))
```
Jorropo, line 88: This comment and check doesn't make sense,
> // Connection establishment should have worked **without an error**

https://pkg.go.dev/github.com/stretchr/testify/require#Error
> Error asserts that a function returned **an error** (i.e. not `nil`). 

AFAIT you are testing that two back to back useless addresses do result in an error returned, so the comment is wrong and the code is correct.

Jorropo (changes requested):

p2p/host/routed/routed_test.go
marten-seemann: Yes, copy-pasted from the test above.

Jorropo (approved):
LGTM

Issue #857, opened 2020-03-26
Submit libp2p to oss-fuzz


willscott:
go-libp2p probably gets enough usage that it would be valuable to have a background fuzz testing framework running against it.

if we have a docker image building a `go-fuzz` entrypoint to libp2p, then we can submit to google's oss-fuzz (https://github.com/google/oss-fuzz/tree/master/projects) and the google cloud will do much of the automation.

docs: https://google.github.io/oss-fuzz//getting-started/new-project-guide/go-lang/

Issue #1694, opened 2021-02-13
can't decode an encoded peer.ID, for certain peer ID values


marten-seemann:
Minimal example:

```go
package main

import (
	"log"

	"github.com/libp2p/go-libp2p-core/peer"
)

func main() {
	data := []byte{
		0x00, 0x01, 0x72, 0x94, 0xe4, 0xc5, 0x30, 0x1b, 0x30, 0x30, 0x30, 0x0e, 0x30, 0xbd, 0xef,
		0x30, 0x30, 0x30, 0x22, 0x77, 0xff, 0x22, 0x30, 0xef, 0x30, 0x30, 0x30, 0xbd, 0x30, 0x30, 0x30,
		0x30, 0xbd, 0x30, 0x30,
	}
	var id peer.ID
	if err := id.UnmarshalText(data); err != nil {
		log.Fatal("unmarshal failed")
	}
	encoded := peer.Encode(id)
	id2, err := peer.Decode(encoded)
	if err != nil {
		log.Fatal(err)
	}
	if id != id2 {
		log.Fatal("expected ids to match")
	}
}
```

The `peer.Decode(encoded)` step here fails: `failed to parse peer ID: expected 1 as the cid version number, got: 3261359138256`.

As far as I can tell, the reason for the failure is that `Encode` always uses base58 encoding: https://github.com/libp2p/go-libp2p-core/blob/83ac1d370dfe3f04aee38a904f73e49070be4863/peer/peer.go#L144-L146
Encoding uses the CID code to decode, if the encoded value doesn't start with "1" or "Qm": https://github.com/libp2p/go-libp2p-core/blob/83ac1d370dfe3f04aee38a904f73e49070be4863/peer/peer.go#L172-L187

I'm not sure what to do here. Can someone with more IPLD experience help me out here? @rvagg, @warpfork?

bshastry:
A smaller input that triggers the same bug (found via https://github.com/google/oss-fuzz/pull/9202)

```
package main

import (
	"log"
	"github.com/libp2p/go-libp2p/core/peer"
)

func main() {
	var data string = "\x00\x01r0\x010"
	var id peer.ID
	if err := id.UnmarshalText([]byte(data)); err != nil {
		log.Print("Unmarshal failed")
		return
	}
	encoded := peer.Encode(id)
	id2, err := peer.Decode(encoded)
	if err != nil {
		log.Fatal(err)
	}
	if id != id2 {
		log.Fatal("expected ids to match")
[...]

PR #1949, opened 2022-12-12
Add fuzz function.

bshastry:
Fixes https://github.com/libp2p/go-libp2p/issues/857 by integrating an existing fuzzer (see https://github.com/libp2p/go-libp2p/issues/1694) that encodes a random byte stream and decodes it checking if (1) decode failed for a successful encode; and (2) encoded and decoded IDs match.

See: https://github.com/google/oss-fuzz/pull/9185

### DIFF
--- a/fuzz/encode_decode_fuzz.go
+++ b/fuzz/encode_decode_fuzz.go
@@ -1,0 +1,20 @@
+package fuzz
+
+import (
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+func FuzzEncodeDecodeID(data []byte) int {
+	var id peer.ID
+	if err := id.UnmarshalText(data); err == nil {
+		encoded := peer.Encode(id)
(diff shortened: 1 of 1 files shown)


marten-seemann:
I'd love to spend more time on fuzzing, but I don't think this is the right approach. This PR is adding a fuzz test for _a single_ method, which happens to be deprecated. It most definitely doesn't fix #857.

bshastry:
Thanks for reviewing this. It would have been nice to discuss what could make this work more interesting and potentially mergeable before closing the PR. Honestly, I'm a bit sad that it was closed right away.

For what it's worth, my arguments for keeping this open would be

- best effort work i.e., something is better than nothing
- projects like prysm make use of go-libp2p so bugs have impact
- finding bugs early is good

I'm still open to expanding this PR with the hope of adding more security critical functions to the test harness with your guidance. Have a nice day!

PR #1950, opened 2022-12-13
core/crypto (openssl): fix public key returned when generating keys

marten-seemann:
This looks scary, but as far as I can tell, it doesn't have any real-world consequences in our stack.


Jorropo (approved):
LGTM

Jorropo (changes requested):
Actually we should have a test that call generate key, then verify that the public key match `priv.GetPublic` key.

marten-seemann:
This is incorrect.

PR #1953, opened 2022-12-14
core/crypto: drop all OpenSSL code paths

### LINKED ISSUE #1539: tls: support OpenSSL
We should be able to support using OpenSSL to perform the actual encryption/decryption. See https://godoc.org/github.com/libp2p/go-openssl#Client and https://godoc.org/github.com/libp2p/go-openssl#Server.

Motivation: OpenSSL is faster and has better side-channel protection.

Requirement: This feature should only be enabled when the "openssl" build flag is enabled.

### LINKED ISSUE #1892: flaky TestRSASignZero (using OpenSSL)
```
=== RUN   TestRSASignZero
    rsa_test.go:74: failed to generate RSA key
--- FAIL: TestRSASignZero (1.38s)

### LINKED ISSUE #1951: core/crypto (openssl): RSA public key generation broken
This is probably not the right way to export a public key:
https://github.com/libp2p/go-libp2p/blob/47f9ea1306dfb4704280e9fe309269ae33f7934e/core/crypto/rsa_openssl.go#L33
https://github.com/libp2p/go-libp2p/blob/47f9ea1306dfb4704280e9fe309269ae33f7934e/core/crypto/rsa_openssl.go#L38

As far as I can see, the fix would need to happen in go-openssl, and involve a bunch of super-ugly Go-to-C and vice versa code.

### LINKED ISSUE #1952: OpenSSL vs. BoringSSL vs. Go standard library crypto
## Context

We're currently maintaining a [fork of go-openssl](https://github.com/libp2p/go-openssl) which has created a lot of maintenance burden in the past, and I'm not optimistic that the situation will improve in the future. Also, I don't _want_ to be maintain this library.

RSA used to be the standard key type in the early days of IPFS / libp2p, and we still have a minority of nodes in the network that use RSA keys. I just instrumented a kubo node, and 15 - 20% of nodes it connects to use RSA keys (the rest is almost exclusively Ed25519).

PL is running some nodes as a service to the network. Of those, bootstrap nodes handle a lot of short-lived connections, i.e. perform a lot of handshakes, during which we need to perform a signature verification.

## Benchmarks

@Jorropo ran some benchmarks of signature verification using different key types a while back in https://github.com/libp2p/go-libp2p/pull/1686#issuecomment-1219412636. However, there's currently a bug in our key generation code, rendering those results invalid.  Benchmarks need to be run with https://github.com/libp2p/go-libp2p/tree/fix-crypto-benchmark due to #1951.

It looks like RSA signature verification is about 4x faster than the respective Go standard library code path.

  | Go Standard Library | OpenSSL | BoringSSL
-- | -- | -- | --
BenchmarkVerifyRSA1B-2 | 85.905 | 26.792 | 20.336
BenchmarkVerifyRSA10B-2 | 106.754 | 27.453 | 19.681
BenchmarkVerifyRSA100B-2 | 94.862 | 26.88 | 20.728
BenchmarkVerifyRSA1000B-2 | 101.62 | 28.35 | 19.747
BenchmarkVerifyRSA10000B-2 | 110.342 | 32.806 | 25.194
BenchmarkVerifyRSA100000B-2 | 170.742 | 100.674 | 80.228
BenchmarkVerifyEd255191B-2 | 76.216 | 82.31 | 80.681
BenchmarkVerifyEd2551910B-2 | 79.42 | 78.811 | 82.704
BenchmarkVerifyEd25519100B-2 | 78.767 | 81.161 | 89.527
BenchmarkVerifyEd255191000B-2 | 79.369 | 84.34 | 81.65
BenchmarkVerifyEd2551910000B-2 | 94.968 | 93.985 | 101.424
BenchmarkVerifyEd25519100000B-2 | 256.963 | 247.958 | 282.375

(all values in microseconds)

## Conclusion

Go standard library signature verification is slightly (10%?) slower for RSA than for Ed25519. A significant speedup of the RSA code path can be achieved by using OpenSSL or BoringSSL, but it's questionable if that will have a big impact on the node's CPU load, given that RSA peers are just a small (and decreasing) fraction of the network. Furthermore, verifying host key signatures is not the only cost during the handshake (there's a regular TLS or Noise handshake as well). It also doesn't seem appropriate to introduce significant complexity to optimize the RSA code path, even though it's not significantly slower than the Ed25519 code path.

Am I missing something here? @vyzo @Stebalien @guseggert @lidel @p-shahi @Jorropo

marten-seemann:
Fixes #1952. Closes #1951. Closes #1892. Closes #1539. Closes #1894.

See #1952 for motivation and discussion.


MarcoPolo (commented):
More context here as well: https://github.com/ipfs/kubo/issues/9507#issuecomment-1370165446

MarcoPolo (approved):

Issue #1910, opened 2022-11-21
swarm: minimal set of metrics


marten-seemann:
The objective is to define a _minimal_ set of metrics that the swarm should expose. There are a lot of interesting things we _could_ measure (and we should in the long term, see #1356), but we need to start somewhere.

I propose to define the minimal set as those metrics that allow us to measure the effect of our smarter dialing logic (#1785).

Suggested metrics:
* conns opened (grouped by: direction, transport / security / muxer)
* conns closed (grouped by: direction, transport / security / muxer)
* conn duration histogram: duration between handshake completion and closing (grouped by: transport / security / muxer)
* dial duration histogram
* dial errors (group by: error type (timeout, cancellation, other), transport / security / muxer)

When implementing smarter dialing, we expect:
* the dial errors to decrease dramatically (for cancelations)
* the dial duration to just increase slightly (otherwise this dialing logic wouldn't be very smart)
* the number of very short-lived incoming connection (as other nodes upgrade)

marten-seemann:
Adding to the list of suggested metrics:
* when closing: amount of stream data exchanged, total number of streams opened / accepted
* canceled handshakes (on the listening side)

The last one is a tricky one, for multiple reasons:
1. It’s not something we can learn about in the `swarm` package, as the transport implementation perform the entire libp2p handshake before handing a connection to the swarm. We therefore need to instrument _every transport_ with a `handshake_started` event. The number of canceled handshakes is then `handshake_completed - handshake_started`.
2. A connection might finish the handshake and then be canceled (there’s a race condition on the sender side here). As long as we don’t have proper close reasons (https://github.com/libp2p/specs/issues/479), we have no way of definitely telling, but we can look at connections that only existed for a very short time, or even better, didn’t accept any streams

GlenDC:
Hi I can pick this issue up with the aim to ship it within the next 10 days.
I do have some questions that I would like to align on before putting potentially wasted dev time into this:

1. Reporter is exposed as an interface (https://github.com/libp2p/go-libp2p/blob/master/core/metrics/reporter.go#L20-L21) but normally with Go you expose from your package concrete types, and it are the users that can (do not have to) use it as an interface.
  - that and given the fact that we're the only users and do not have multiple reporter production use cases, we might as well keep it at the single reporter use case?
[...]

PR #1959, opened 2022-12-20
roadmap: remove optimizations of the TCP-based handshake

marten-seemann:
As an early result of the swarm metrics effort (#1910), we've discovered that >80% of the connections that a (full) node on the IPFS network establishes / accepts are QUIC connections. We can expect similar numbers for other libp2p networks that use go-libp2p (and that enabled the QUIC transport).

It therefore makes sense to focus our attention on cutting down roundtrips for QUIC users. This makes sense for multiple reasons:
1. At the risk of stating the obvious, improving the handshake latency for >80% of the connections is more impactful than improving the handshake latency for <20% of the connections.
2. Optimizing the TCP-based handshake by getting rid of the security protocol negotiation reduces the latency of a TCP handshake 4 to 3 RTTs (including Identify in the calculation here), that's 25%. Optimizing the QUIC handshake (by using 0.5-RTT data for Identify) allows us to bring down the latency from 2 to 1 RTT, a speedup of 50%.
3. Moving the security protocol is a complicated change, requiring changes to the libp2p specification (https://github.com/libp2p/specs/pull/353), and requires both nodes to upgrade to the latest versions. Using 0.5-RTT data in QUIC on the other hand is a backwards-compatible change, as it doesn't require the client to upgrade. Thus, the performance impact will materialize much faster.

I suggest _removing_ the TCP-based optimizations instead of just _deprioritizing_ them, since given our current staffing situation, and the recent addition of the libp2p+HTTP work, we already have a lot on our plate.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -31,8 +31,7 @@ https://github.com/libp2p/go-libp2p/issues/1806
     - [3. WebTransport: Update to new draft versions](#3-webtransport-update-to-new-draft-versions)
   - [B. ⚡️ Handshakes at the Speed of Light](#b-️-handshakes-at-the-speed-of-light)
     - [1. Early Muxer Negotiation](#1-early-muxer-negotiation)
-    - [2. Adding security protocol](#2-adding-security-protocol)
-    - [3. 0.5 RTT data optimization](#3-05-rtt-data-optimization)
+    - [2. 0.5 RTT data optimization](#2-05-rtt-data-optimization-(for-QUIC))
   - [C. 🧠 Smart Dialing](#c--smart-dialing)
     - [1. Happy Eyeballs](#1-happy-eyeballs)
     - [2. QUIC Blackhole detector](#2-quic-blackhole-detector)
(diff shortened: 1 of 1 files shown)


marten-seemann:
> > As an early result of the swarm metrics effort (#1910), we've discovered that >80% of the connections that a (full) node on the IPFS network establishes / accepts are QUIC connections.
> 
> Have these results been shared anywhere? I don't doubt the veracity of the claim but it links to an issue that is in progress without a PR.

Not yet. The code is still work in progress, and my dev branch is only hooked up to my own Grafana instance at this point. I'm happy to share two screenshots:
image
image

@mxinden has shared some similar numbers in https://filecoinproject.slack.com/archives/C03K82MU486/p1671440249540809?thread_ts=1671373570.578789&cid=C03K82MU486.

BigLep (approved):

mxinden (commented):
For what my input is worth, reasoning here makes sense to me :+1:

PR #1960, opened 2022-12-21
config: initialize the null rcmgr in fx, not in the transports

marten-seemann:
This would fix https://github.com/ipfs/kubo/issues/9516.


MarcoPolo (approved):

p-shahi:
Should this be merged?

MarcoPolo:
A lot has changed here. I don't think this applies any longer.

PR #1971, opened 2023-01-02
quic, tcp: only register Prometheus counters when metrics are enabled

marten-seemann:
Due to the performance impact, we're only collecting metrics when explicitly enabled anyway. We shouldn't even register the Prometheus metrics if not.

### DIFF
--- a/p2p/transport/quicreuse/tracer_metrics.go
+++ b/p2p/transport/quicreuse/tracer_metrics.go
@@ -85,7 +85,9 @@ func (c *aggregatingCollector) RemoveConn(id string) {
 
 var collector *aggregatingCollector
 
-func init() {
+var initMetricsOnce sync.Once
+
+func initMetrics() {
 	const (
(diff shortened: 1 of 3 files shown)


p2p/transport/quicreuse/tracer_metrics.go
```diff
@@ -85,7 +85,9 @@ func (c *aggregatingCollector) RemoveConn(id string) {
 
 var collector *aggregatingCollector
 
-func init() {
+var initMetricsOnce sync.Once
+
+func initMetrics() {
 	const (
 		direction = "direction"
 		encLevel  = "encryption_level"
```
MarcoPolo, line 93: On L103 we're using the default prometheus Registerer. I think in general we should accept this as a parameter rather than hardcoded since we are a library.

No change needed now. If you agree, let's open an issue.

MarcoPolo (approved):

PR #1972, opened 2023-01-02
circuitv2: correctly set the transport in the ConnectionState

marten-seemann:


p2p/protocol/circuitv2/client/conn.go
```diff
@@ -143,3 +144,20 @@ func (c *Conn) untagHop() {
...
+	network.ConnStat
+}
+
+type capableConn struct {
+	capableConnWithStat
+}
+
+var transportName = ma.ProtocolWithCode(ma.P_CIRCUIT).Name
+
+func (c *capableConn) ConnState() network.ConnectionState {
```
MarcoPolo, line 159: Should we make this a value receiver instead? That lets us avoid pointer chasing.

MarcoPolo (approved):

p2p/protocol/circuitv2/client/conn.go
marten-seemann: Done.

p2p/protocol/circuitv2/client/transport.go
```diff
@@ -63,7 +63,11 @@ func (c *Client) Dial(ctx context.Context, a ma.Multiaddr, p peer.ID) (transport
 		return nil, err
 	}
 	conn.tagHop()
-	return c.upgrader.Upgrade(ctx, c, conn, network.DirOutbound, p, connScope)
+	cc, err := c.upgrader.Upgrade(ctx, c, conn, network.DirOutbound, p, connScope)
+	if err != nil {
+		return nil, err
+	}
+	return &capableConn{cc.(capableConnWithStat)}, nil
```
MarcoPolo, line 70:
```suggestion
	return capableConn{cc.(capableConnWithStat)}, nil
```

MarcoPolo (changes requested):
I don't think you need to create a pointer there

MarcoPolo (approved):
sorry I meant to approve this with the suggestion.

p2p/protocol/circuitv2/client/transport.go
marten-seemann: Done.

Issue #1955, opened 2022-12-14
rcmgr: Use prometheus for metrics (OpenCensus generates garbage)


guseggert:
I've been profiling a large [hydra-booster](https://github.com/libp2p/hydra-booster) deployment, and recently added the [built-in Resource Manager metrics](https://github.com/libp2p/go-libp2p/tree/master/p2p/host/resource-manager/obs), which caused CPU usage to significantly increase, caused by increased GC frequency, which seems to be caused by a large amount of garbage generated by adding tags to OpenCensus metrics inside of Resource Manager:

![allocs](https://user-images.githubusercontent.com/877588/207710324-cef6f6d2-ca92-4de1-8b63-7bb9b6e2cfa5.png)

(Note that this is using go-libp2p v0.21, which was before go-libp2p-resource-manager was conslidated into the go-libp2p repo, but the RM metrics code has not changed significantly in recent versions, so likely still a problem. I am still working to upgrade the hydras to v0.24, after which I'll try again.)

marten-seemann:
Interesting. This is a huge amount of garbage we're creating. Thanks for reporting, @guseggert!

We should aim for making recording of metrics allocation-free, ideally. This seems to be easy with `tag.Upsert`, but there might be other code paths where this is not that easy.

This is highly relevant for our broader metrics effort (#1356, #1910).

guseggert:
Seems related: https://github.com/census-instrumentation/opencensus-go/issues/1265

marten-seemann:
There doesn’t seem to be a lot of appetite for these optimizations in OpenCensus.

This would be an argument against using OpenCensus and for using Prometheus directly.

guseggert:
Agreed, it doesn't matter what kind of flexibility OpenCensus provides, if it's unusable in hot code paths.

marten-seemann:
Putting this on the list for the v0.25 release. Resource manager metrics are an integral part of our resource observability story, they need to “just work” under all work loads.

guseggert:
[...]

PR #1973, opened 2023-01-02
swarm: add a basic metrics tracer

### LINKED ISSUE #1910: swarm: minimal set of metrics

marten-seemann:
Fixes #1910.

Due to our experience with OpenCensus (#1955), this now uses Prometheus. There's a benchmark test that confirms that (by using a `sync.Pool` for the slices containing the labels) we're not allocating at all.

### DIFF
--- a/p2p/net/swarm/swarm_dial.go
+++ b/p2p/net/swarm/swarm_dial.go
@@ -490,11 +490,15 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (tra
 		return nil, ErrNoTransport
 	}
 
+	start := time.Now()
 	connC, err := tpt.Dial(ctx, addr, p)
 	if err != nil {
+		recordDialFailed(addr, err)
 		return nil, err
 	}
 	canonicallog.LogPeerStatus(100, connC.RemotePeer(), connC.RemoteMultiaddr(), "connection_status", "established", "dir", "outbound")
+	recordConnectionOpened(network.DirOutbound, connC.RemotePublicKey(), connC.ConnState())
+	recordHandshakeLatency(time.Since(start), connC.ConnState())
 
 	// Trust the transport? Yeah... right.
 	if connC.RemotePeer() != p {
--- a/p2p/net/swarm/swarm_dial.go
+++ b/p2p/net/swarm/swarm_dial.go
@@ -493,12 +493,17 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (tra
 	start := time.Now()
 	connC, err := tpt.Dial(ctx, addr, p)
 	if err != nil {
-		recordDialFailed(addr, err)
+		if s.metricsTracer != nil {
+			s.metricsTracer.FailedDialing(addr, err)
+		}
 		return nil, err
 	}
 	canonicallog.LogPeerStatus(100, connC.RemotePeer(), connC.RemoteMultiaddr(), "connection_status", "established", "dir", "outbound")
-	recordConnectionOpened(network.DirOutbound, connC.RemotePublicKey(), connC.ConnState())
-	recordHandshakeLatency(time.Since(start), connC.ConnState())
+	if s.metricsTracer != nil {
+		connState := connC.ConnState()
+		s.metricsTracer.OpenedConnection(network.DirOutbound, connC.RemotePublicKey(), connState)
+		s.metricsTracer.CompletedHandshake(time.Since(start), connState)
+	}
 
 	// Trust the transport? Yeah... right.
 	if connC.RemotePeer() != p {
--- a/p2p/net/swarm/swarm_metrics.go
+++ b/p2p/net/swarm/swarm_metrics.go
@@ -16,50 +16,56 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-const metricNamespace = "libp2p_swarm_"
+const metricNamespace = "libp2p_swarm"
 
 var (
 	connsOpened = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
-			Name: metricNamespace + "connections_opened_total",
-			Help: "Connections Opened",
+			Namespace: metricNamespace,
+			Name:      "connections_opened_total",
+			Help:      "Connections Opened",
 		},
 		[]string{"dir", "transport", "security", "muxer"},
 	)
 	keyTypes = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
-			Name: metricNamespace + "key_types_total",
-			Help: "key type",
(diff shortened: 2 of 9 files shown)


marten-seemann:
Here's some preliminary results, obtaining from running a Kubo node with this branch:
image

marten-seemann:
I added the Grafana dashboard. There's probably a lot that can be improved on that front, happy about suggestions (and PRs...).

What's really annoying here is that Grafana is too stupid to apply consistent colors on different panels. For example, it makes it soooo much easier to consume if QUIC always has the same color. You can set colors manually via so-called overrides, but those are per panel. So what I did is I defined the overrides for all our transports in one panel, and then copy-pasted a long block of JSON over to the other panels. Apparently that's the only way to do it. 🤢

MarcoPolo:
I know it's asking a bit, but is there an open grafana dashboard I can see?

p2p/net/swarm/swarm_metrics_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"testing"
+
+	"github.com/libp2p/go-libp2p/core/crypto"
+	"github.com/libp2p/go-libp2p/core/network"
+
+	"github.com/stretchr/testify/require"
+)
+
+func BenchmarkMetricsConnOpen(b *testing.B) {
+	b.ReportAllocs()
```
MarcoPolo, line 14: Can we check that this is indeed some low number of allocs?

p2p/net/swarm/swarm_dial.go
```diff
@@ -490,11 +490,20 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (tra
...
+		if s.metricsTracer != nil {
+			s.metricsTracer.FailedDialing(addr, err)
+		}
 		return nil, err
 	}
 	canonicallog.LogPeerStatus(100, connC.RemotePeer(), connC.RemoteMultiaddr(), "connection_status", "established", "dir", "outbound")
+	if s.metricsTracer != nil {
+		connState := connC.ConnState()
+		s.metricsTracer.OpenedConnection(network.DirOutbound, connC.RemotePublicKey(), connState)
+		s.metricsTracer.CompletedHandshake(time.Since(start), connState)
```
MarcoPolo, line 505: Do we want to do this on the listen side as well?

p2p/net/swarm/swarm_metrics.go
```diff
@@ -0,0 +1,185 @@
...
+	keyTypes = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: metricNamespace + "key_types_total",
+			Help: "key type",
+		},
+		[]string{"dir", "key_type"},
+	)
+	connsClosed = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: metricNamespace + "connections_closed_total",
```
MarcoPolo, line 38:
```suggestion
            Namespace: metricNamespace,
			Name: "connections_closed_total",
```

Likewise with the others.

p2p/net/swarm/swarm_metrics.go
```diff
@@ -0,0 +1,185 @@
...
+	*tags = append(*tags, getDirection(dir))
+	*tags = appendConnectionState(*tags, cs)
+	connDuration.WithLabelValues(*tags...).Observe(duration.Seconds())
+}
+
+func (m *metricsTracer) CompletedHandshake(t time.Duration, cs network.ConnectionState) {
+	tags := getStringSlice()
+	defer putStringSlice(tags)
+	*tags = appendConnectionState(*tags, cs)
+	connHandshakeLatency.WithLabelValues(*tags...).Observe(t.Seconds())
```
MarcoPolo, line 159: It would be great if we could know if this was an early muxer negotiation.

MarcoPolo (commented):
This is so cool! I can't believe we haven't had this yet.

Just some nits. I'll approve as soon as they're fixed.

p2p/net/swarm/swarm_metrics_test.go
marten-seemann: In a unit test? Benchmarks are currently not run on CI at all.

p2p/net/swarm/swarm_dial.go
marten-seemann: That would require us to put it in the transport. The swarm doesn't know when the handshake of an incoming connection started.

p2p/host/peerstore/pstoremem/addr_book.go
```diff
@@ -46,7 +46,10 @@ type addrSegment struct {
 }
 
 func (segments *addrSegments) get(p peer.ID) *addrSegment {
-	return segments[byte(p[len(p)-1])]
+	if len(p) == 0 { // it's not terribly useful to use an empty peer ID, but at least we should not panic
```
MarcoPolo, line 49: `p.Validate()` instead?

marten-seemann: How did that change sneak into this PR? That was https://github.com/libp2p/go-libp2p/pull/2006. Will rebase.

p2p/net/swarm/swarm_metrics.go
marten-seemann: Done.

p2p/net/swarm/swarm_metrics_test.go
MarcoPolo: Yup

p2p/net/swarm/swarm_dial.go
MarcoPolo: Could we add this to the ConnectionState struct?

p2p/net/swarm/swarm_metrics.go
MarcoPolo: Could we add this to the ConnectionState struct?

MarcoPolo (approved):

MarcoPolo:
Merging because I can't wait for this feature!

PR #1975, opened 2023-01-03
identify: remove support for Identify Delta

### LINKED ISSUE #977: Identify Push does not emit eventbus events
Unlike its siblings `identify` and `identify/delta`, `identify/push` does not seem to emit events on the eventbus such as `EvtPeerIdentificationCompleted`, or `EvtPeerProtocolsUpdated`. It should emit both of these events.

Question: Is there any assumption in `EvtPeerIdentificationCompleted` that would prevent it from being used in identify push?

@Stebalien @aarshkshah1992

### LINKED ISSUE #1974: identify: remove support Delta
Identify Delta creates a lot of complexity. Instrumenting a Kubo node yielded the following result for Identify messages *received* within a 6h time frame:
image
Identify Delta traffic is 1/100 of the traffic caused by normal Identify messages (which are exchanged right at the beginning of the connection).

It's questionable if there's a scenario where Delta actually helps: Since Delta transfers a list of protocols that were added / removed. Usually, nodes don't add / remove protocols _on a regular basis_, but only occasionally. For example, when a node discovers that it's public, it might add the relay and the DHT protocol. However, this happens rarely, and it seems ok to send O(1000) (number of peers) 1 kB (size of an Identify message) messages in that case.

We can just remove support for Delta both on the sender and the receiver side, since Delta has always been an optional feature: if it's not available, nodes will just use Push instead.

cc @aarshkshah1992

marten-seemann:
Fixes #1974. Fixes #977.

-700 LOC, this is starting to pay off. We'll probably be able to shave off even more later.


p2p/protocol/identify/id.go
```diff
@@ -565,11 +560,49 @@ func (ids *idService) getSignedRecord(snapshot *identifySnapshot) []byte {
...
+
+func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn, isPush bool) {
 	p := c.RemotePeer()
 
-	// mes.Protocols
+	supported, _ := ids.Host.Peerstore().GetProtocols(p)
+	added, removed := diff(supported, mes.Protocols)
 	ids.Host.Peerstore().SetProtocols(p, mes.Protocols...)
+	if isPush {
+		ids.emitters.evtPeerProtocolsUpdated.Emit(event.EvtPeerProtocolsUpdated{
```
Stebalien, line 600: Hm. We should have been doing this all along...

marten-seemann: Yes. A lot of libp2p users rely on this event, and Push is _much_ more common than Delta.
My Kubo node receives 10x as many Push messages (see Grafana screenshot in #1974). I wonder how many subtle failures were caused by this...

marten-seemann: Turns out we even had an issue for this: #977.

MarcoPolo (approved):
Nice work 🔥 🔥 🔥

p2p/protocol/identify/id.go
```diff
@@ -268,20 +267,18 @@ func (ids *idService) loop() {
 					select {
 					case phs[pid].pushCh <- struct{}{}:
 					default:
-						log.Debugf("dropping addr updated message for %s as buffer full", pid.Pretty())
+						log.Debugf("dropping addr updated message for %s as buffer full", pid.String())
```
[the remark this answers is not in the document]
marten-seemann: Fixed.

PR #1980, opened 2023-01-05
chore: use generic LRU cache

muXxer:
This PR upgrades golang-lru to the generics branch v2.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/golang/mock v1.6.0
 	github.com/google/gopacket v1.1.19
 	github.com/gorilla/websocket v1.5.0
-	github.com/hashicorp/golang-lru v0.5.4
+	github.com/hashicorp/golang-lru/v2 v2.0.1
 	github.com/ipfs/go-cid v0.3.2
 	github.com/ipfs/go-datastore v0.6.0
 	github.com/ipfs/go-ds-badger v0.3.0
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -51,11 +51,11 @@ require (
 	github.com/raulk/go-watchdog v1.3.0
 	github.com/stretchr/testify v1.8.1
 	go.opencensus.io v0.24.0
-	go.uber.org/fx v1.18.2
-	go.uber.org/goleak v1.1.12
-	golang.org/x/crypto v0.3.0
+	go.uber.org/fx v1.19.0
```
marten-seemann, line 54: There’s something wrong with the dependencies here. There’s no reason to update fx and qtls in this PR.

marten-seemann (changes requested):
This looks like a reasonable improvement, but there something wrong with the dependencies.

go.mod
muXxer: Should I revert the second commit, which was basically a `go get -u`?
Maybe you can update the other dependencies some other day :+1:

marten-seemann: Yes, please do. It’s good to keep PRs limited to one change at a time.

muXxer: I knew it wasn't clean when I did it, but I like to live dangerously :trollface: 
Should be fine now, I reverted the second commit :+1:

marten-seemann (approved):
Nice! Always happy to see generic code (being used)!

PR #1982, opened 2023-01-05
websocket: Replace gorilla websocket transport with nhooyr websocket transport

### LINKED ISSUE #1970: Migrate off github.com/gorilla/websocket
The github.com/gorilla/websocket module, [used for WebSocket support by go-libp2p](https://github.com/libp2p/go-libp2p/blob/82315917f76ca1e9e1c3aca3ad957080f0ac870c/go.mod#L13), was [recently archived](https://github.com/gorilla#gorilla-toolkit) and is no longer maintained. 

The only other serious option I can find is https://github.com/nhooyr/websocket, but it does not support HTTP/2. There is also an attempt to fork the Gorilla repos here: https://github.com/gorillaincubator. Or go-libp2p could maintain its own fork.

MarcoPolo:
fixes #1970 

- [x] Run this on the kubo sharness tests and see if anything comes up. 
- [x] Test this with the test-plans interop tester.
- [ ] If that looks okay, ping @thibmeu since I know they use websockets a lot.


p2p/transport/websocket/addrs.go
```diff
@@ -10,52 +10,28 @@ import (
...
 
-// Addr is an implementation of net.Addr for WebSocket.
-type Addr struct {
+// addrWrapper is an implementation of net.Addr for WebSocket.
+type addrWrapper struct {
 	*url.URL
 }
 
-var _ net.Addr = (*Addr)(nil)
+var _ net.Addr = (addrWrapper)(addrWrapper{})
```
marten-seemann, line 18:
```suggestion
var _ net.Addr = &addrWrapper{}
```

p2p/transport/websocket/websocket.go
```diff
@@ -179,7 +173,26 @@ func (t *WebsocketTransport) maDial(ctx context.Context, raddr ma.Multiaddr) (ma
 		return nil, err
 	}
 	isWss := wsurl.Scheme == "wss"
-	dialer := ws.Dialer{HandshakeTimeout: 30 * time.Second}
+	wsurlCopy := *wsurl
+	remoteAddr := addrWrapper{URL: &wsurlCopy}
+	localAddrChan := make(chan addrWrapper, 1)
+
+	transport := &http.Transport{
+		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
```
marten-seemann, line 181: Why are we setting the `DialContext` here and below (in the `isWss` branch)?

p2p/transport/websocket/conn.go
```diff
@@ -25,7 +28,8 @@ func (c conn) RemoteAddr() net.Addr {
 
 func (c conn) Read(b []byte) (int, error) {
 	n, err := c.Conn.Read(b)
-	if err == nil && n == 0 {
+	if err == nil && n == 0 && c.readAttempts < maxReadAttempts {
```
marten-seemann, line 31: Why is this necessary? An `io.Reader` should not return `n == 0 && err == nil`. Is this a bug upstream?

marten-seemann (commented):
Should the PR title say `websocket:` instead of `webtransport:`?

p2p/transport/websocket/websocket.go
MarcoPolo: To fill the `localAddr` into localAddrChan in case `!isWss`. This is the hackiest part of this. It doesn't seem like the websocket library lets us see the localAddr for the connection (which we need to fulfill the `net.Conn` interface). We can get it as part of the underlying dial, which is what I'm doing here.

p2p/transport/websocket/conn.go
MarcoPolo: Gorilla did the same thing and we had a for loop then.

From the Go doc on `io.Reader`:

> Implementations of Read are discouraged from returning a zero byte count with a nil error, except when len(p) == 0. Callers should treat a return of 0 and nil as indicating that nothing happened; in particular it does not indicate EOF.

It's discouraged, but allowed to return `0, nil`.

marten-seemann: Why do we have to deal with it here? Can't we just have the user deal with it?

p2p/transport/websocket/websocket.go
marten-seemann: Can we a `!isWss` code path and set it there? That would make the code slightly easier to understand.

MarcoPolo: It's harder to understand imo: https://gist.github.com/MarcoPolo/ddb3c159d1358d6e261a78282425839c

p2p/transport/websocket/conn.go
MarcoPolo: The user is the upgrader in this case, and it doesn't expect this. It fails without this.

marten-seemann: I'm not sure I understand how this fix helps. If the reader return `0, nil`. I assume that happens when there's no data to read. So retrying immediately will yield the same result, unless new data was received in the few nanoseconds it took to retry the `Read` call.

Really, this sounds like a bug upstream. `Read` should block until new data is available. Otherwise every consumer of this `io.Reader` will into this problem. It's a common pattern to call `Read` in a loop until it errors, and every such loop will be a busy-loop if it returns `0, nil`.

MarcoPolo: >  I assume that happens when there's no data to read. So retrying immediately will yield the same result, unless new data was received in the few nanoseconds it took to retry the Read call.


I think this has to do with the end of a data frame. Before we start a new reader for the next frame: https://github.com/nhooyr/websocket/blob/master/netconn.go#L118. We do the same thing today with Gorilla: https://github.com/libp2p/go-libp2p/blob/master/p2p/transport/websocket/conn.go#L65. The difference here is that we bound it so we aren't looping forever.

MarcoPolo:
Sharness is fine: https://app.circleci.com/pipelines/github/ipfs/kubo/8686/workflows/de7238e4-fcbf-4a02-a497-0ccbef57979b/jobs/94088

MarcoPolo:
I'm going to test this with the interop tester as well :)

p2p/transport/websocket/conn.go
marten-seemann: That makes sense, thank you for the explanation. I'm ok with keeping this workaround for now. What do you think of:
1. Opening an issue upstream. Maybe even submit a PR, this seems pretty straightforward to fix with a few lines of code.
2. Add a comment with your explanation from above (and a link to the issue) to the code.

Jorropo:
Sorry for the dumb question but why ?

I know that gorilla websocket is no longer maintained, it's not like the websocket protocol is evolving and it probably wont in the near to medium future. Did we found issues in gorilla websocket that is forcing us to change ?

https://pkg.go.dev/nhooyr.io/websocket?tab=importedby has far less imports than https://pkg.go.dev/github.com/gorilla/websocket?tab=importedby (~500 vs ~18k), just by the shear number of users gorilla websocket and the "background fuzzing" this yielded I would most of it's bug to have been fixed already.

Also, I don't know what is the status on nhooyr websocket support, but 7 commits in 2 years doesn't seems very active to me, maybe nhooyr websocket is feature complete (again websocket isn't a protocol that has seen any recent activity) justifying such low activity idk.

I just would like to see more rational, the very remote view I see is that we are replacing a well tested explicitly abandoned project that had recent activity, with a less tested sparsely recently active one.

---

I think we should wait until we find an issue in gorilla websocket to try to pick an other one , if it's working why are we trying to change it ?
This also leaves more time for someone else to pick up the project, this is an extremely popular project, I would be very surprised if in a year we don't see maintained forks of it.

marten-seemann:
> I know that gorilla websocket is no longer maintained, it's not like the websocket protocol is evolving and it probably wont in the near to medium future. Did we found issues in gorilla websocket that is forcing us to change ?

It's not a good idea to depend on unmaintained code. We're not having any issues with Gorilla right now, but these are the kind of things that come back and bite you 6 months in the future, if you don't take care of them.

> https://pkg.go.dev/nhooyr.io/websocket?tab=importedby has far less imports than https://pkg.go.dev/github.com/gorilla/websocket?tab=importedby (~500 vs ~18k), just by the shear number of users gorilla websocket and the "background fuzzing" this yielded I think most of it's bug to have been fixed already.

I can see this as an argument when comparing a library with 5 vs 180 imports, but 500 seems plenty.

> This also leaves more time for someone else to pick up the project, this is an extremely popular project, I would be very surprised if in a year we don't see maintained forks of it.

Gorilla had been searching for a new maintainer since 2018 (!!), and nobody stepped up to take the job. I'm not very confident that somebody will do so now that it's abandoned. There'll certainly be people forking it, and making minor changes, but that's a very different thing than actually maintaining a project.

p2p/transport/websocket/addrs.go
```diff
@@ -10,52 +10,28 @@ import (
...
 
-// Addr is an implementation of net.Addr for WebSocket.
-type Addr struct {
+// addrWrapper is an implementation of net.Addr for WebSocket.
+type addrWrapper struct {
 	*url.URL
 }
 
-var _ net.Addr = (*Addr)(nil)
+var _ net.Addr = addrWrapper{}
```
aviscode, line 18: Why is this one needed

p2p/transport/websocket/addrs.go
```diff
@@ -10,52 +10,28 @@ import (
...
 
 // Network returns the network type for a WebSocket, "websocket".
-func (addr *Addr) Network() string {
+func (a addrWrapper) Network() string {
```
aviscode, line 21: why not  a *addrWrapper ?

marten-seemann: Why would it? This is faster without a pointer dereference.

p2p/transport/websocket/addrs.go
marten-seemann: This is a pretty standard compile time interface assertion.

p2p/transport/websocket/addrs.go
Jorropo: I know I am being pedentic, but `a` / `addr` is not used, this LITERALLY does nothing:
```x86asm
        TEXT    "".addrWrapper.Value(SB), NOSPLIT|ABIInternal, $0-8
        LEAQ    go.string."websocket"(SB), AX
        MOVL    $9, BX
        RET

        TEXT    "".(*addrWrapper).Pointer(SB), NOSPLIT|ABIInternal, $0-8
        LEAQ    go.string."websocket"(SB), AX
        MOVL    $9, BX
        RET
```
Both functions literally have the same assembly.
Here value is correct but this just because you are not modifying the `addrWrapper`. Even if you did, a pointer like this should cost no more than 3ns. Caring about performance here does not make sense.

Rules I personally follow:
- First start by using value receiver, this is to be clear that your methods are read only.
- However if you need shared state then use pointers on all methods (to get around promotion issues).

aviscode: The only thing is it was a pointer shouldn't we leave it a pointer ?

Jorropo: > The only thing is it was a pointer shouldn't we leave it a pointer ?

Why ? It's readonly.

p2p/transport/websocket/conn.go
marten-seemann: @MarcoPolo Thoughts?

marten-seemann (commented):
This is now getting interop-tested using our handshake test, right?

p2p/transport/websocket/conn.go
MarcoPolo: done

marten-seemann (approved):

PR #1991, opened 2023-01-11
CI: Fast multidimensional Interop tests

MarcoPolo:
Runs the current branch against other supported versions. Tests all transport, muxer, and security combinations. Tests dialing and listening. Runs in [<10 min](https://github.com/libp2p/go-libp2p/actions/runs/3894748269) compared to the existing interop test of [~30 min](https://github.com/libp2p/go-libp2p/actions/runs/3868886818) (but effectively ~15min since there are two tests that run in parallel).

In total this currently runs 95 integration tests compared to the 2 testground tests. To be fair the testground test run a test where all different nodes are on the same network and try to ping each other. That would be equivalent to roughly 5-15 tests here.

Draft until the corresponding test-plans PR is merged: https://github.com/libp2p/test-plans/pull/99


test-plans/cmd/ping/main.go
```diff
@@ -0,0 +1,152 @@
...
+		Password:    "",
+		DB:          0,
+	})
+	defer rClient.Close()
+
+	// Make sure redis is ready
+	_, err := rClient.Ping(ctx).Result()
+	if err != nil {
+		fmt.Println(err)
+		panic("Failed to connect to redis")
```
marten-seemann, line 54:
```suggestion
		log.Fatalf("Failed to connect to redis: %s", err)
```

test-plans/cmd/ping/main.go
```diff
@@ -0,0 +1,152 @@
...
+		options = append(options, libp2p.Transport(libp2pquic.NewTransport))
+		listenAddr = fmt.Sprintf("/ip4/%s/udp/0/quic", ip)
+	case "quic-v1":
+		options = append(options, libp2p.Transport(libp2pquic.NewTransport))
+		listenAddr = fmt.Sprintf("/ip4/%s/udp/0/quic-v1", ip)
+	case "webtransport":
+		options = append(options, libp2p.Transport(libp2pwebtransport.New))
+		listenAddr = fmt.Sprintf("/ip4/%s/udp/0/quic-v1/webtransport", ip)
+	default:
+		panic("Unsupported transport")
```
marten-seemann, line 79:
```suggestion
		log.Fatalf("Unsupported transport: %s", transport)
```

Same for the `panic`s below.

test-plans/cmd/ping/main.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+func main() {
+	var (
+		transport     = os.Getenv("transport")
+		secureChannel = os.Getenv("security")
+		muxer         = os.Getenv("muxer")
+		isDialerStr   = os.Getenv("is_dialer")
+		ip            = os.Getenv("ip")
+		redis_addr    = os.Getenv("REDIS_ADDR")
```
marten-seemann, line 31: Nit: No underscores in Go variable names

test-plans/go.mod
```diff
@@ -0,0 +1,99 @@
...
+	golang.org/x/net v0.3.0 // indirect
+	golang.org/x/sync v0.1.0 // indirect
+	golang.org/x/sys v0.3.0 // indirect
+	golang.org/x/text v0.5.0 // indirect
+	golang.org/x/tools v0.3.0 // indirect
+	google.golang.org/protobuf v1.28.1 // indirect
+	lukechampine.com/blake3 v1.1.7 // indirect
+)
+
+replace github.com/libp2p/go-libp2p => ../
```
marten-seemann, line 99: Does this need to be a separate module? These `replace` statements have proven to be incredibly painful in the past (when you update a dependency, you also have to run `go mod tidy` here, otherwise CI fill fail your build).

test-plans/ping-versions.template.json
```diff
@@ -0,0 +1,21 @@
+[
+    {
+        "id": "go-libp2p-head",
+        "containerImageID": "FILL THIS IN",
```
marten-seemann, line 4: This is supposed to fill in the hash of the image, right? Do we need to do this? 
It would be nicer if we could just use a tag.

test-plans/PingDockerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM golang:1.19-alpine AS builder
+
+WORKDIR /app/
+
+COPY ./ .
+WORKDIR /app/test-plans
+RUN go mod download
+RUN go build -o /testplan ./cmd/ping 
+
+FROM golang:1.19-alpine
```
marten-seemann, line 13: Don't need a Go image here, can use a blank image.

marten-seemann (commented):
Nice work! Mostly nits here.

I'm wondering if we can use Docker tags instead of capturing the container ID, that might simplify the build process a tiny bit.

.github/workflows/interop-test.yml
```diff
@@ -1,23 +1,40 @@
...
-  # NOTE: during a pull request run, github creates a merge commit referenced in `github.sha`
-  # that merge commit is not a regular commit. You won't find it with a regular `git checkout SHA` and
-  # tools like `go get repo@SHA` won't find it.
-  #
-  # As a workaround, we generate a path to the actual pull request's commit, it looks like:
-  # `github.com/external-org/go-libp2p@latest-commit-on-their-branch`
-  run-ping-interop-cross-version:
-    uses: "libp2p/test-plans/.github/workflows/run-composition.yml@master"
+  build-ping-container:
+    name: Build Ping interop container
```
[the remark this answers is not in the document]
MarcoPolo: It's just the one?

.github/workflows/interop-test.yml
```diff
@@ -1,23 +1,40 @@
...
+          name: ping-versions
+          path: ./test-plans/ping-versions.json
+      - name: Upload image tar
+        uses: actions/upload-artifact@v3
+        with:
+          name: ping-image
+          path: ./test-plans/ping-image.tar
+  run-multidim-interop:
+    needs: build-ping-container
+    uses: "libp2p/test-plans/.github/workflows/run-testplans.yml@marco/wo-testground-opts"
```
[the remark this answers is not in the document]
MarcoPolo: This needs the branch that's in review. It will be master once that's merged.

test-plans/go.mod
MarcoPolo: Doesn't have to be a separate module. Although if this isn't a separate module, then the test dependencies bubble up to the main module. So Redis would show up as a dependency to go-libp2p, even though it's only used for this test.

Unless, I'm wrong about that. In which case I'll happily not have this be a separate module.

.github/workflows/interop-test.yml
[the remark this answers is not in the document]
marten-seemann: Ok, then I'm misunderstanding something here. Where are all the other containers built?

test-plans/ping-versions.template.json
MarcoPolo: Maybe there's a misunderstanding?

The overall idea is this is how you run the current branch against all released versions. This enables interop tests per go-libp2p PR. This extends the defined tests in test-plans by adding this version to the versions already defined.

Once a version is released, the source of truth is the test-plans repo. That defines supported released versions. When we release a version we also push a container image to some container registry (TBD). Then update the test-plans repo to define version A of go-libp2p support X,Y,Z and here's the link to the container image.

This template defines what the current go-libp2p head supports. The `"FILL THIS IN"` is filled in by the Makefile which will put the container image it build for this branch. I should rephrase this to `"This will be filled in automatically"`.

test-plans/PingDockerfile
MarcoPolo: When using `FROM scratch` I get:
```
$ docker run --rm -it $( jq -r '.imageID' < ping-image.json)
exec /testplan: no such file or directory
```

MarcoPolo: but from alpine works fine :)

.github/workflows/interop-test.yml
```diff
@@ -1,23 +1,40 @@
...
-  # As a workaround, we generate a path to the actual pull request's commit, it looks like:
-  # `github.com/external-org/go-libp2p@latest-commit-on-their-branch`
-  run-ping-interop-cross-version:
-    uses: "libp2p/test-plans/.github/workflows/run-composition.yml@master"
+  build-ping-container:
+    name: Build Ping interop container
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
```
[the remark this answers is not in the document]
MarcoPolo: You're right on non-windows platforms bash is already default.

test-plans/go.mod
marten-seemann: > So Redis would show up as a dependency to go-libp2p, even though it's only used for this test.

Yeah, let's avoid that.

marten-seemann: We can (later) add a pre-commit git hook that checks that the `go.mod` here is up to date.

test-plans/Makefile
```diff
@@ -0,0 +1,10 @@
+all: ping-image.tar
+
+ping-image.tar: PingDockerfile go.mod go.sum cmd/ping/
+	cd .. && docker build -t go-libp2p-head -f test-plans/PingDockerfile .
```
marten-seemann, line 4: Please double check this suggestion.
```suggestion
	cd .. && docker build -t go-libp2p-head -f PingDockerfile .
```

test-plans/cmd/ping/main.go
```diff
@@ -0,0 +1,152 @@
...
+	})
+	defer rClient.Close()
+
+	// Make sure redis is ready
+	_, err := rClient.Ping(ctx).Result()
+	if err != nil {
+		log.Fatalf("Failed to connect to redis: %s", err)
+	}
+
+	is_dialer := isDialerStr == "true"
```
marten-seemann, line 57: Here as well 😇

marten-seemann (approved):
LGTM. Please merge the test-plans PR first, and update the ref to master here.

test-plans/Makefile
MarcoPolo: This is run from the parent directory. This change would break it.

PR #1996, opened 2023-01-12
ci: use GitHub token to install protoc

marten-seemann:


marten-seemann:
Thank you @galargh! It works :)

PR #1997, opened 2023-01-12
chore: Update release.md

p-shahi:
Remove updating Lotus in the Release checklist. Afaik, it's not something we do anymore


marten-seemann:
We weren't very diligent with that recently, but I actually want to work more closely with the Lotus stewards. The current update is particularly painful, as we're still in the late stages of finishing up the transition to the monorepo (something similar to https://github.com/ipfs/kubo/pull/9423 will be needed).

Due to the nature of the network, the Filecoin network updates much faster than the IPFS network. That's an opportunity for us to validate the expected impact of our protocol and implementation improvements faster than we currently do.

PR #2000, opened 2023-01-19
feat!: rcmgr: Change LimitConfig to use LimitVal type

MarcoPolo:
This solves the problem around the encoding having different semantics than the config that is passed into the resource manager [context](https://github.com/libp2p/go-libp2p/pull/1998#issuecomment-1386348540). This makes special values explicit (Unlimited, default, block all). And I think it's a more ergonomic API (create a struct with sane defaults, rather than having to copy some pre-initialized struct, then modify relevant values).

The high level idea is you have some custom configuration called `PartialLimitConfig` that gets constructed with a `ConcreteLimitConfig` (a configuration that has no special or missing values, zero means zero, max means max). The `.Build` method fills in any missing information from the `PartialLimitConfig` and returns a complete `ConcreteLimitConfig`. Another way to think about it is that `PartialLimitConfig` is a partial overrides config, and you add it to a complete `ConcreteLimitConfig` to get another `ConcreteLimitConfig`.  

Also take a look at the README for some updates on the description there.

Changes:
* Introduces a new type `LimitVal` which can explicitly specify "use default", "unlimited", "block all", as well as any positive number. The zero value of `LimitVal` (the value when you create the object in Go) is "Use default".
  * The JSON marshalling of this is straightforward.
* Introduces a new `ResourceLimits` type which uses `LimitVal` instead of ints so it can encode the above for the resources.
* Changes `LimitConfig` to `PartialLimitConfig` and uses `ResourceLimits`. This along with the marshalling changes means you can now marshal the fact that some resource limit is set to block all.
  * Because the default is to use the defaults, this avoids the footgun of initializing the resource manager with 0 limits (that would block everything).

Before:
```
	cfg := rcmgr.DefaultLimits.AutoScale()
	cfg.System.ConnsInbound = 3
	cfg.System.ConnsOutbound = 1024
	cfg.System.Conns = 1024
	cfg.PeerDefault.Conns = 1
	cfg.PeerDefault.ConnsInbound = 1
	cfg.PeerDefault.ConnsOutbound = 1
```

After:
```
	cfg := rcmgr.PartialLimitConfig{
		System: &rcmgr.ResourceLimits{
			ConnsInbound:  3,
			ConnsOutbound: 1024,
			Conns:         1024,
		},
		PeerDefault: &rcmgr.ResourceLimits{
			ConnsInbound:  1,
			ConnsOutbound: 1,
			Conns:         1,
		},
	}.Build(rcmgr.DefaultLimits.AutoScale())
```

And the JSON encoding:

```
	bl := ResourceLimits{
		Streams:         DefaultLimit,
		StreamsInbound:  10,
		StreamsOutbound: BlockAllLimit,
		Conns:           10,
		// ConnsInbound:    DefaultLimit,
		ConnsOutbound: Unlimited,
		Memory:        Unlimited64,
	}

	jsonEncoded, err := json.Marshal(bl)
	require.NoError(t, err)
	require.Equal(t, string(jsonEncoded), `{"StreamsInbound":10,"StreamsOutbound":"blockAll","Conns":10,"ConnsOutbound":"unlimited","Memory":"unlimited"}`)
```

This doesn't affect the ScalingLimitConfig. That has a lot of knobs and represents the developer's knowledge of how the resource usage should scale. LimitConfig on the other hand is more like the manual tweaks an end-user would make. It's used in conjunction with the defaults created from a ScalingLimitConfig.

### DIFF
--- a/p2p/host/resource-manager/README.md
+++ b/p2p/host/resource-manager/README.md
@@ -28,9 +28,21 @@ scalingLimits := rcmgr.DefaultLimits
 // Add limits around included libp2p protocols
 libp2p.SetDefaultServiceLimits(&scalingLimits)
 
-// Turn the scaling limits into a static set of limits using `.AutoScale`. This
+// Turn the scaling limits into a reified set of limits using `.AutoScale`. This
 // scales the limits proportional to your system memory.
-limits := scalingLimits.AutoScale()
+scaledDefaultLimits := scalingLimits.AutoScale()
+
+// Tweak certain settings
+cfg := rcmgr.LimitConfig{
+  System: &rcmgr.ResourceLimits{
+    // Allow unlimited outbound streams
+    StreamsOutbound: rcmgr.Unlimited,
+  },
+  // Everything else is default. The exact values will come from `scaledDefaultLimits` above.
+}
+
+// Create our limits by using our cfg and replacing the default values with values from `scaledDefaultLimits`
+limits := cfg.Reify(scaledDefaultLimits)
 
 // The resource manager expects a limiter, se we create one from our limits.
 limiter := rcmgr.NewFixedLimiter(limits)
@@ -50,6 +62,54 @@ if err != nil {
 // Create a libp2p host
 host, err := libp2p.New(libp2p.ResourceManager(rm))
 ```
+
+### Saving the limits config
+The easiest way to save the defined limits is to serialize the `LimitConfig`
+type as JSON.
+
+```go
+noisyNeighbor, _ := peer.Decode("QmVvtzcZgCkMnSFf2dnrBPXrWuNFWNM9J3MpZQCvWPuVZf")
+cfg := rcmgr.LimitConfig{
+  System: &rcmgr.ResourceLimits{
+    // Allow unlimited outbound streams
+    StreamsOutbound: rcmgr.Unlimited,
+  },
+  Peer: map[peer.ID]rcmgr.ResourceLimits{
+    noisyNeighbor: {
+      // No inbound connections from this peer
+      ConnsInbound: rcmgr.BlockAllLimit,
+      // But let me open connections to them
+      Conns:         rcmgr.DefaultLimit,
+      ConnsOutbound: rcmgr.DefaultLimit,
+      // No inbound streams from this peer
+      StreamsInbound: rcmgr.BlockAllLimit,
+      // And let me open unlimited (by me) outbound streams (the peer may have their own limits on me)
+      StreamsOutbound: rcmgr.Unlimited,
+    },
+  },
+}
+jsonBytes, _ := json.Marshal(&cfg)
+
+// string(jsonBytes)
+// {
+//   "System": {
+//     "StreamsOutbound": "unlimited"
+//   },
+//   "Peer": {
+//     "QmVvtzcZgCkMnSFf2dnrBPXrWuNFWNM9J3MpZQCvWPuVZf": {
+//       "StreamsInbound": "blockAll",
+//       "StreamsOutbound": "unlimited",
+//       "ConnsInbound": "blockAll"
+//     }
+//   }
+// }
+```
+
+This will omit defaults from the JSON output. It will also serialize the
+blockAll, and unlimited values explicitly.
+
+The `Memory` field is serialized as a string to workaround the JSON limitation
+of 32 bit integers (`Memory` is an int64).
 
 ## Basic Resources
 
--- a/p2p/host/resource-manager/README.md
+++ b/p2p/host/resource-manager/README.md
@@ -33,7 +33,7 @@ libp2p.SetDefaultServiceLimits(&scalingLimits)
 scaledDefaultLimits := scalingLimits.AutoScale()
 
 // Tweak certain settings
-cfg := rcmgr.LimitConfig{
+cfg := rcmgr.PartialLimitConfig{
   System: &rcmgr.ResourceLimits{
     // Allow unlimited outbound streams
     StreamsOutbound: rcmgr.Unlimited,
@@ -42,7 +42,7 @@ cfg := rcmgr.LimitConfig{
 }
 
 // Create our limits by using our cfg and replacing the default values with values from `scaledDefaultLimits`
-limits := cfg.Reify(scaledDefaultLimits)
+limits := cfg.Build(scaledDefaultLimits)
 
 // The resource manager expects a limiter, se we create one from our limits.
 limiter := rcmgr.NewFixedLimiter(limits)
@@ -64,12 +64,12 @@ host, err := libp2p.New(libp2p.ResourceManager(rm))
 ```
 
 ### Saving the limits config
-The easiest way to save the defined limits is to serialize the `LimitConfig`
+The easiest way to save the defined limits is to serialize the `PartialLimitConfig`
 type as JSON.
 
 ```go
 noisyNeighbor, _ := peer.Decode("QmVvtzcZgCkMnSFf2dnrBPXrWuNFWNM9J3MpZQCvWPuVZf")
-cfg := rcmgr.LimitConfig{
+cfg := rcmgr.PartialLimitConfig{
   System: &rcmgr.ResourceLimits{
     // Allow unlimited outbound streams
     StreamsOutbound: rcmgr.Unlimited,
@@ -338,7 +338,7 @@ This is done using the `ScalingLimitConfig`. For every scope, this configuration
 struct defines the absolutely bare minimum limits, and an (optional) increase of
 these limits, which will be applied on nodes that have sufficient memory.
 
-A `ScalingLimitConfig` can be converted into a `LimitConfig` (which can then be
+A `ScalingLimitConfig` can be converted into a `ConcreteLimitConfig` (which can then be
 used to initialize a fixed limiter with `NewFixedLimiter`) by calling the `Scale` method.
 The `Scale` method takes two parameters: the amount of memory and the number of file
 descriptors that an application is willing to dedicate to libp2p.
@@ -406,7 +406,7 @@ go-libp2p process. For the default definitions see [`DefaultLimits` and
 
 If the defaults seem mostly okay, but you want to adjust one facet you can
 simply copy the default struct object and update the field you want to change. You can
-apply changes to a `BaseLimit`, `BaseLimitIncrease`, and `LimitConfig` with
+apply changes to a `BaseLimit`, `BaseLimitIncrease`, and `ConcreteLimitConfig` with
 `.Apply`.
 
 Example
--- a/p2p/host/resource-manager/README.md
+++ b/p2p/host/resource-manager/README.md
@@ -28,7 +28,7 @@ scalingLimits := rcmgr.DefaultLimits
 // Add limits around included libp2p protocols
 libp2p.SetDefaultServiceLimits(&scalingLimits)
 
-// Turn the scaling limits into a reified set of limits using `.AutoScale`. This
+// Turn the scaling limits into a concrete set of limits using `.AutoScale`. This
 // scales the limits proportional to your system memory.
 scaledDefaultLimits := scalingLimits.AutoScale()
 
--- a/p2p/host/resource-manager/limit.go
+++ b/p2p/host/resource-manager/limit.go
@@ -110,8 +110,8 @@ func valueOrBlockAll64(n int64) LimitVal64 {
 }
 
 // ToResourceLimits converts the BaseLimit to a ResourceLimits
-func (l BaseLimit) ToResourceLimits() ResourceLimits {
-	return ResourceLimits{
+func (l BaseLimit) ToResourceLimits() *ResourceLimits {
+	return &ResourceLimits{
 		Streams:         valueOrBlockAll(l.Streams),
 		StreamsInbound:  valueOrBlockAll(l.StreamsInbound),
 		StreamsOutbound: valueOrBlockAll(l.StreamsOutbound),
--- a/p2p/host/resource-manager/limit.go
+++ b/p2p/host/resource-manager/limit.go
@@ -100,12 +100,16 @@ type BaseLimit struct {
 func valueOrBlockAll(n int) LimitVal {
 	if n == 0 {
 		return BlockAllLimit
+	} else if n == math.MaxInt {
+		return Unlimited
 	}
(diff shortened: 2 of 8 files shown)


p2p/host/resource-manager/README.md
```diff
@@ -28,9 +28,21 @@ scalingLimits := rcmgr.DefaultLimits
...
+cfg := rcmgr.LimitConfig{
+  System: &rcmgr.ResourceLimits{
+    // Allow unlimited outbound streams
+    StreamsOutbound: rcmgr.Unlimited,
+  },
+  // Everything else is default. The exact values will come from `scaledDefaultLimits` above.
+}
+
+// Create our limits by using our cfg and replacing the default values with values from `scaledDefaultLimits`
+limits := cfg.Reify(scaledDefaultLimits)
```
BigLep, line 45: A couple of thoughts:
1. Is the word "reify" intuitive enough?  This may certainly just be me.  I didn't know what the word meant, and haven't seen it used in these kind of contexts.  (But if this is a convention in golang or libraries then great.  I just wanted to make sure we're not being too clever here.)
2. The comment sentence reads weird to me.  Is that really what is happening?  I would expect that we have some scaled/computed defaults as a result of the AutoScale call and then we are overriding some of those values by the user-supplied overrides.

BigLep: I guess you're hitting on this in the overall PR description.  I think I need to come at this with a clear head.  Feel free to disregard this for now.

MarcoPolo: Happy to use different terminology. `PartialConfig` vs `ConcreteConfig`? Reification is familiar to me, but if it feels weird for other folks we should change.

ajnavarro: @MarcoPolo that terminology sounds better.

p2p/host/resource-manager/limit.go
```diff
@@ -97,6 +97,81 @@ type BaseLimit struct {
...
+	if out.Streams == DefaultLimit &&
+		out.StreamsInbound == DefaultLimit &&
+		out.StreamsOutbound == DefaultLimit &&
+		out.Conns == DefaultLimit &&
+		out.ConnsInbound == DefaultLimit &&
+		out.ConnsOutbound == DefaultLimit &&
+		out.FD == DefaultLimit &&
+		out.Memory == DefaultLimit64 {
+		return nil
+	}
```
ajnavarro, line 170: Maybe this can be a `IsEmpty()` function on `ResourceLimits`.

p2p/host/resource-manager/README.md
MarcoPolo: What about `Build`? `Materialize` has a specific meaning in databases that I don’t want to invoke.

MarcoPolo: Changed:
`LimitConfig` => `PartialLimitConfig`
`ReifiedLimitConfig` => `ConcreteLimitConfig`
`.Reify(` => `.Build(`

ajnavarro:
LGTM. I'll get this PR and do some integration tests on Kubo to see how it looks.

MarcoPolo:
after a sync conversation with @marten-seemann. 0 in the JSON should mean "block all". This is because a user who manually edits a limit from 1 to 0 would be surprised to see 0 actually mean default instead of block all.

p2p/host/resource-manager/limit.go
```diff
@@ -202,65 +270,65 @@ func (l *BaseLimit) GetMemoryLimit() int64 {
 }
 
 func (l *fixedLimiter) GetSystemLimits() Limit {
-	return &l.System
+	return &l.system
```
ajnavarro, line 273: Is there any reason to make these values private? When trying to integrate this on Kubo, we have some cases where we want to tweak final values after scaling, and we can't do that anymore if these values are private.

MarcoPolo: Yes. We only want one way of changing things. I think fiddling with the `ConcreteLimitConfig` struct is tricky, since you have no defaults. You have to make sure everything is properly initialized.

If you want to tweak these values, the best way is to use a `PartialConfig` for the things you want to tweak and then Build it with a `ConcreteLimitConfig`. The readme has an example

ajnavarro:
@MarcoPolo I have a question about how to integrate this when using `rcmgr.ResourceScopeLimiter` to set new limits. This is how I did it: 

I had to copy `infiniteBaseLimits` because it is private:

```go
var infiniteBaseLimit = rcmgr.BaseLimit{
	Streams:         math.MaxInt,
	StreamsInbound:  math.MaxInt,
	StreamsOutbound: math.MaxInt,
	Conns:           math.MaxInt,
	ConnsInbound:    math.MaxInt,
	ConnsOutbound:   math.MaxInt,
	FD:              math.MaxInt,
	Memory:          math.MaxInt64,
}
```

baseLimit is a `rcmgr.ResourceLimits` object:

```go
		limiter, ok := s.(rcmgr.ResourceScopeLimiter)
		if !ok { 
			return ...
		}

		baseLimit := limit.Build(infiniteBaseLimit)
		limiter.SetLimit(&rcmgr.BaseLimit{
			Streams:         baseLimit.Streams,
			StreamsInbound:  baseLimit.StreamsInbound,
			StreamsOutbound: baseLimit.StreamsOutbound,

			Conns:         baseLimit.Conns,
			ConnsInbound:  baseLimit.ConnsInbound,
			ConnsOutbound: baseLimit.ConnsOutbound,

			FD:     baseLimit.FD,
			Memory: baseLimit.Memory,
		})
```

Shall we set as public `infiniteBaseLimit` and have a way to transform from a `ResourceLimit` object to a `BaseLimit` one?

p2p/host/resource-manager/limit_defaults.go
```diff
@@ -108,53 +110,295 @@ func (cfg *ScalingLimitConfig) AddProtocolPeerLimit(proto protocol.ID, base Base
...
+		if _, ok := (*this)[k]; !ok {
+			(*this)[k] = l
+		}
+	}
+}
+
+func (cfg *PartialLimitConfig) Apply(c PartialLimitConfig) {
 	cfg.System.Apply(c.System)
 	cfg.Transient.Apply(c.Transient)
 	cfg.AllowlistedSystem.Apply(c.AllowlistedSystem)
```
ajnavarro, line 404: These lines can cause nil pointer exceptions now.

ajnavarro (changes requested):

p2p/host/resource-manager/limit.go
```diff
@@ -97,6 +97,68 @@ type BaseLimit struct {
 	Memory          int64 `json:",omitempty"`
 }
 
+func valueOrBlockAll(n int) LimitVal {
+	if n == 0 {
+		return BlockAllLimit
+	}
```
ajnavarro, line 103:
```suggestion
	if i == math.MaxInt {
		return Unlimited
	} else if i == 0 {
		return BlockAllLimit
	}
```

p2p/host/resource-manager/limit.go
```diff
@@ -97,6 +97,68 @@ type BaseLimit struct {
...
+	return LimitVal(n)
+}
+func valueOrBlockAll64(n int64) LimitVal64 {
+	if n == 0 {
+		return BlockAllLimit64
+	}
```
ajnavarro, line 109:
```suggestion
	if i == math.MaxInt64 {
		return Unlimited
	} else if i == 0 {
		return BlockAllLimit
	}
```

ajnavarro (commented):
added small things I saw wen integrating with Kubo here: https://github.com/ipfs/kubo/pull/9612

MarcoPolo:
`func (l *ResourceLimits) Build(defaults BaseLimit) BaseLimit {` Should instead take a `Limit` interface instead. This will make it nicer to use with the `ResourceScopeLimiter` interface.

p2p/host/resource-manager/limit_defaults.go
MarcoPolo: zero values for `ResourceLimits` is `DefaultLimit`.

MarcoPolo: I don't understand your comment. Am I missing something?

ajnavarro: No, I was missing something. I thought that zero value will be zero, but is -2 (BlockAllLimit)

p2p/host/resource-manager/limit_defaults.go
```diff
@@ -167,145 +473,212 @@ func (cfg *LimitConfig) Apply(c LimitConfig) {
...
+
+	conn   BaseLimit
+	stream BaseLimit
+}
+
+// ToLimitConfigWithDefaults converts a ConcreteLimitConfig to a PartialLimitConfig.
+// Uses the defaults config to know what was specifically set and what was left
+// as default. Returns a minimal PartialLimitConfig. Build the returned PartialLimitConfig
+// with the defaults to get back to the original ConcreteLimitConfig.
+func (cfg ConcreteLimitConfig) ToLimitConfigWithDefaults(defaults ConcreteLimitConfig) PartialLimitConfig {
```
MarcoPolo, line 564: We can probably get rid of this function.

p2p/host/resource-manager/limit_defaults.go
```diff
@@ -167,145 +473,212 @@ func (cfg *LimitConfig) Apply(c LimitConfig) {
...
-		if _, ok := cfg.Peer[s]; !ok {
-			cfg.Peer[s] = l
-		}
+
+	return out
+}
+
+// ToLimitConfig converts a ConcreteLimitConfig to a PartialLimitConfig. The returned
+// PartialLimitConfig will have no default values.
+func (cfg ConcreteLimitConfig) ToLimitConfig() PartialLimitConfig {
```
MarcoPolo, line 609: This should be called `ToConcreteLimitConfig`

marten-seemann (approved):
I'm still not fully convinced that `ConcreteLimitConfig.ToPartialLimitConfig()` is a good idea, but I don't want to block this PR on that.

@MarcoPolo Could you add a paragraph about the rcmgr changes to the v0.25 release issue please?

p2p/host/resource-manager/limit_defaults.go
```diff
@@ -108,53 +111,356 @@ func (cfg *ScalingLimitConfig) AddProtocolPeerLimit(proto protocol.ID, base Base
...
-type LimitConfig struct {
-	System    BaseLimit `json:",omitempty"`
-	Transient BaseLimit `json:",omitempty"`
+type LimitVal int
+
+const (
+	// DefaultLimit is the default value for resources. The exact value depends on the context, but will get values from `DefaultLimits`.
+	DefaultLimit LimitVal = 0
+	// Unlimited is the value for unlimited resources. An arbitrarily high number will also work.
+	Unlimited LimitVal = -1
```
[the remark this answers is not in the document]
MarcoPolo: Purely preference. A -1 looks different than some large number `9223371036854775807` in case this is printed out. Is that large number a Max Int? (it isn't)

p2p/host/resource-manager/README.md
```diff
@@ -51,6 +63,54 @@ if err != nil {
...
+    // Allow unlimited outbound streams
+    StreamsOutbound: rcmgr.Unlimited,
+  },
+  Peer: map[peer.ID]rcmgr.ResourceLimits{
+    noisyNeighbor: {
+      // No inbound connections from this peer
+      ConnsInbound: rcmgr.BlockAllLimit,
+      // But let me open connections to them
+      Conns:         rcmgr.DefaultLimit,
+      ConnsOutbound: rcmgr.DefaultLimit,
```
BigLep, line 83: Do you even have to specify rcmgr.DefaultLimit or are those just added for clarity?

p2p/host/resource-manager/README.md
```diff
@@ -51,6 +63,54 @@ if err != nil {
...
+// string(jsonBytes)
+// {
+//   "System": {
+//     "StreamsOutbound": "unlimited"
+//   },
+//   "Peer": {
+//     "QmVvtzcZgCkMnSFf2dnrBPXrWuNFWNM9J3MpZQCvWPuVZf": {
+//       "StreamsInbound": "blockAll",
+//       "StreamsOutbound": "unlimited",
+//       "ConnsInbound": "blockAll"
```
BigLep, line 102: Is the order that they serialize as?  If we are in control, I think it would be good to go from outwards to in

Conns
ConnsInbound
CoonnsOutbound
Streams
StreamsInbound
StreamsOutbound

p2p/host/resource-manager/README.md
```diff
@@ -51,6 +63,54 @@ if err != nil {
...
+//     }
+//   }
+// }
+```
+
+This will omit defaults from the JSON output. It will also serialize the
+blockAll, and unlimited values explicitly.
+
+The `Memory` field is serialized as a string to workaround the JSON limitation
```
BigLep, line 111: To drive this home, maybe set a Memory limit in your example above so it shows up in the serialization?

p2p/host/resource-manager/README.md
```diff
@@ -278,7 +338,7 @@ This is done using the `ScalingLimitConfig`. For every scope, this configuration
 struct defines the absolutely bare minimum limits, and an (optional) increase of
 these limits, which will be applied on nodes that have sufficient memory.
 
-A `ScalingLimitConfig` can be converted into a `LimitConfig` (which can then be
+A `ScalingLimitConfig` can be converted into a `ConcreteLimitConfig` (which can then be
 used to initialize a fixed limiter with `NewFixedLimiter`) by calling the `Scale` method.
 The `Scale` method takes two parameters: the amount of memory and the number of file
 descriptors that an application is willing to dedicate to libp2p.
```
BigLep, line 344: (minor) Maybe also mention the AutoScale method for added convenience?

BigLep (commented):
Is there any API breaking changes with this work?

Regardless I didn't see any mention of the changes in https://github.com/libp2p/go-libp2p/releases/tag/v0.25.0 

(it makes me think that having changelog files would be useful so that this text gets captured as part of the PR itself.  Kubo follows this approach).

p2p/host/resource-manager/README.md
MarcoPolo: Just added for clarity, and to show that default doesn't get serialized into json. You'll also notice Memory is implicitly default.

Issue #1886, opened 2022-11-16
clean up usage of string vs. protocol.ID


marten-seemann:
It's all over the place, and we're converting between the two all the time.

sukunrt:
can i pick this up?

sukunrt:
I'm looking into this. Discussed the issue with @marten-seemann 
Here a simple generic type constraint of ```~string``` won't suffice since ```multistream.MultistreamMuxer ``` is a struct and method parameters cannot be generic. To make a method like 
```func (msm *MultistreamMuxer) AddHandlerWithFunc(protocol string, match func(string) bool, handler HandlerFunc) ``` 
generic we'd need MultistreamMuxer to be generic with a signature like  
```func (msm *MultistreamMuxer[T ~string]) AddHandlerWithFunc(protocol T, match func(T) bool, handler HandlerFunc) ``` 
that seems tricky to do without breaking the api since this constructor ```NewMultistreamMuxer()``` expects to create a muxer without any type being specified. I'll come back on the possible options here.

**Comment by marten-seemann** [comment]
> generic we'd need MultistreamMuxer to be generic with a signature like
> `func (msm *MultistreamMuxer[T ~string]) AddHandlerWithFunc(protocol T, match func(T) bool, handler HandlerFunc) `

Yes, that would be great!

> that seems tricky to do without breaking the api since this constructor `NewMultistreamMuxer()` expects to create a muxer without any type being specified. I'll come back on the possible options here.

I wouldn’t worry about that. Breaking the API is ok here, we can release this as a new major version.

If you really, really want to avoid breaking API changes you could do this
```go
func NewGenericMultistreamMuxer[T ~string]() MultistreamMuxer[T] {
    …
}
```
and then have `NewMultistreamMuxer` return the concrete string type:
```go
func NewMultistreamMuxer() MultistreamMuxer[string] {
    return NewGenericMultistreamMuxer[string]()
}
```

Not sure if that’s worth it here though, we don’t use it at so many places that avoiding the API change would actually pay off.

sukunrt:
For go-libp2p the changes are here. I have changed PeerStore. There are a couple of other places where we can remove the converstions before integrating the changes go-multistream repo. I'll update those changes soon. 

https://github.com/libp2p/go-libp2p/pull/2004

sukunrt:
[...]

PR #2004, opened 2023-01-20
consistently use protocol.ID instead of strings

sukunrt:
This reduces the string to protocol.ID translations happening at various places in the code

Fixes https://github.com/libp2p/go-libp2p/issues/1886

### DIFF
--- a/core/network/conn.go
+++ b/core/network/conn.go
@@ -42,7 +42,7 @@ type ConnectionState struct {
 	// The security protocol used on this connection (if any). For example: /tls/1.0.0
 	Security protocol.ID
 	// the transport used on this connection. For example: tcp
-	Transport protocol.ID
+	Transport string
 }
 
 // ConnSecurity is the interface that one can mix into a connection interface to
--- a/core/network/conn.go
+++ b/core/network/conn.go
@@ -6,6 +6,7 @@ import (
 
 	ic "github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/protocol"
 
 	ma "github.com/multiformats/go-multiaddr"
 )
@@ -37,11 +38,11 @@ type Conn interface {
 // ConnectionState holds information about the connection.
 type ConnectionState struct {
 	// The stream multiplexer used on this connection (if any). For example: /yamux/1.0.0
-	StreamMultiplexer string
+	StreamMultiplexer protocol.ID
 	// The security protocol used on this connection (if any). For example: /tls/1.0.0
-	Security string
+	Security protocol.ID
 	// the transport used on this connection. For example: tcp
-	Transport string
+	Transport protocol.ID
 }
 
 // ConnSecurity is the interface that one can mix into a connection interface to
--- a/core/protocol/switch.go
+++ b/core/protocol/switch.go
@@ -3,7 +3,13 @@ package protocol
 
 import (
 	"io"
+
+	"github.com/multiformats/go-multistream"
 )
+
+type StringLike interface {
+	~string
+}
 
 // HandlerFunc is a user-provided function used by the Router to
 // handle a protocol/stream.
@@ -11,7 +17,7 @@ import (
 // Will be invoked with the protocol ID string as the first argument,
 // which may differ from the ID used for registration if the handler
 // was registered using a match function.
-type HandlerFunc = func(protocol string, rwc io.ReadWriteCloser) error
+type HandlerFunc = multistream.HandlerFunc[ID]
 
 // Router is an interface that allows users to add and remove protocol handlers,
 // which will be invoked when incoming stream requests for registered protocols
@@ -25,7 +31,7 @@ type Router interface {
 
 	// AddHandler registers the given handler to be invoked for
 	// an exact literal match of the given protocol ID string.
-	AddHandler(protocol string, handler HandlerFunc)
+	AddHandler(protocol ID, handler HandlerFunc)
 
 	// AddHandlerWithFunc registers the given handler to be invoked
 	// when the provided match function returns true.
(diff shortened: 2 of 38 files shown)


p2p/protocol/identify/id.go
```diff
@@ -592,9 +592,10 @@ func diff(a, b []string) (added, removed []string) {
 func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn, isPush bool) {
 	p := c.RemotePeer()
 
+	// TODO (sukunrt): Fix this
 	supported, _ := ids.Host.Peerstore().GetProtocols(p)
-	added, removed := diff(supported, mes.Protocols)
-	ids.Host.Peerstore().SetProtocols(p, mes.Protocols...)
+	added, removed := diff(protocol.ConvertToStrings(supported), mes.Protocols)
+	ids.Host.Peerstore().SetProtocols(p, protocol.ConvertFromStrings(mes.Protocols)...)
```
marten-seemann, line 598: Probably makes the most sense to make `diff` take `protocol.ID`s instead of `string`s. Since we're getting the list of strings from the Protobuf, we don't really have a choice but to convert from strings once.

p2p/host/basic/basic_host.go
```diff
@@ -654,6 +652,8 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
 		}, nil
 	}
 
+	pidStrings := protocol.ConvertToStrings(pids)
```
marten-seemann, line 655: This won't be needed any more once you pull in https://github.com/multiformats/go-multistream/pull/95, right?

marten-seemann (commented):
Can you pull in https://github.com/multiformats/go-multistream/pull/95, so we can see how this will look like in the end?

p2p/host/basic/basic_host.go
sukunrt: yes. This would be resolved.

p2p/protocol/identify/id.go
sukunrt: Changed the diff function. It looks much better now.

sukunrt:
Should I change network.ConnectionState?

sukunrt:
> Can you pull in [multiformats/go-multistream#95](https://github.com/multiformats/go-multistream/pull/95), so we can see how this will look like in the end?

Do you mean to point the multistream module to the new changes and check? If so, sukunrt/fix-pid-string-conv-ms-merged
This is the branch with the merged changes.

marten-seemann:
> > Can you pull in [multiformats/go-multistream#95](https://github.com/multiformats/go-multistream/pull/95), so we can see how this will look like in the end?
> 
> Do you mean to point the multistream module to the new changes and check? If so, sukunrt/fix-pid-string-conv-ms-merged (https://github.com/sukunrt/go-libp2p/tree/sukunrt/fix-pid-string-conv-ms-merged) This is the branch with the merged changes.

Nice!!!

> Should I change network.ConnectionState?

I think it would make sense. What do you think? Does it make the code easier?

sukunrt:
I think changing network.ConnectionState is correct. That way internally it is always protocol.ID and we only convert to string when dealing with external lib like tls or translating to protobufs.

examples/ipfs-camp-2019/05-Discovery/protocol.go
```diff
@@ -51,7 +51,7 @@ func chatInputLoop(ctx context.Context, h host.Host, donec chan struct{}) {
 	for scanner.Scan() {
 		msg := scanner.Text()
 		for _, peer := range h.Network().Peers() {
-			if _, err := h.Peerstore().SupportsProtocols(peer, string(chatProtocol)); err == nil {
+			if _, err := h.Peerstore().SupportsProtocols(peer, chatProtocol); err == nil {
```
marten-seemann, line 54: Don't update the examples. They have a separate `go.mod` file. We only update them once we release a new libp2p version.

sukunrt: fixed this.

marten-seemann: Am I right to assume you forgot to push?

sukunrt: sorry about that. fixed. I had updated the other branch sukunrt/fix-pid-string-conv-ms-merged

marten-seemann:
@sukunrt I just cut a new release of go-multistream with all your changes: https://github.com/multiformats/go-multistream/tree/v0.4.0.

Can you update this PR to use the release?

p2p/host/basic/basic_host.go
sukunrt: this is fixed.

sukunrt:
Updated the branch with all changes merged.

core/network/conn.go
```diff
@@ -37,11 +38,11 @@ type Conn interface {
...
 type ConnectionState struct {
 	// The stream multiplexer used on this connection (if any). For example: /yamux/1.0.0
-	StreamMultiplexer string
+	StreamMultiplexer protocol.ID
 	// The security protocol used on this connection (if any). For example: /tls/1.0.0
-	Security string
+	Security protocol.ID
 	// the transport used on this connection. For example: tcp
-	Transport string
+	Transport protocol.ID
```
marten-seemann, line 45: The transport is not really a `protocol.ID`. Can we leave this as a `string`?

core/protocol/switch.go
```diff
@@ -11,7 +13,7 @@ import (
 // Will be invoked with the protocol ID string as the first argument,
 // which may differ from the ID used for registration if the handler
 // was registered using a match function.
-type HandlerFunc = func(protocol string, rwc io.ReadWriteCloser) error
+type HandlerFunc = multistream.HandlerFunc[ID]
```
marten-seemann, line 16: Nice one!

marten-seemann (commented):
I did a full review, and this looks really good to me. One minor thing, then this should be good to merge. 🎉

core/network/conn.go
sukunrt: fixed. For my understanding, protocol.ID is used for any protocol we use *over* the transport, right?

marten-seemann (approved):

Issue #2001, opened 2023-01-19
Multiaddrs with peer ID are not considered dialable


masih:
A multiaddr can optionally contain the peer ID, i.e. `../p2p/<peer-ID>`. When such multiaddrs are added to the libp2p host peerstore, they are not considered to be dialable, and filtered on `host.Connect` resulting in `no good addresses` error.

For example, the following test fails with error `no good addresses`:

```go 
func TestHost_ConnectWithMultiaddrAndID(t *testing.T) {
	host, err := libp2p.New()
	require.NoError(t, err)
	pid, err := peer.Decode("12D3KooWNSRG5wTShNu6EXCPTkoH7dWsphKAPrbvQchHa5arfsDC")
	require.NoError(t, err)

	addrWithPid, err := multiaddr.NewMultiaddr("/ip4/209.94.92.6/tcp/24001/p2p/" + pid.String())
	require.NoError(t, err)

	host.Peerstore().AddAddr(pid, addrWithPid, peerstore.PermanentAddrTTL)

	addrs := host.Peerstore().Addrs(pid)
	t.Log(addrs)

	err = host.Connect(context.Background(), peer.AddrInfo{ID: pid})
	require.NoError(t, err)
}

```

While this one succeeds:

```go
func TestHost_ConnectWithMultiaddr(t *testing.T) {
	host, err := libp2p.New()
	require.NoError(t, err)
	pid, err := peer.Decode("12D3KooWNSRG5wTShNu6EXCPTkoH7dWsphKAPrbvQchHa5arfsDC")
	require.NoError(t, err)

	addrWithPid, err := multiaddr.NewMultiaddr("/ip4/209.94.92.6/tcp/24001")
	require.NoError(t, err)

	host.Peerstore().AddAddr(pid, addrWithPid, peerstore.PermanentAddrTTL)

	addrs := host.Peerstore().Addrs(pid)
	t.Log(addrs)

	err = host.Connect(context.Background(), peer.AddrInfo{ID: pid})
	require.NoError(t, err)
}
```

The only difference being that the multiaddr added to peerstore in the first test contained peer ID.

In the specific case above, looks like the address is being filtered by `CanDial` of `TcpTransport` which uses [this](https://github.com/libp2p/go-libp2p/blob/master/p2p/transport/tcp/tcp.go#L154) matcher.

Is this behaviour expected and or documented? Otherwise, it seems surprising that a valid multiaddr being successfully added to the peerstore and yet no good address are found on connection.

[...]

PR #2007, opened 2023-01-23
Clean addresses with peer id before adding to addrbook

sukunrt:
Fixes: https://github.com/libp2p/go-libp2p/issues/2001

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -709,6 +709,11 @@ func (h *BasicHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
 		}
 	}
 
+	addrs := h.Peerstore().Addrs(pi.ID)
+	if len(addrs) == 0 {
+		return fmt.Errorf("no valid address found for peer %s", pi.ID)
+	}
+
 	return h.dialPeer(ctx, pi.ID)
 }
 
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -709,11 +709,6 @@ func (h *BasicHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
 		}
 	}
 
-	addrs := h.Peerstore().Addrs(pi.ID)
-	if len(addrs) == 0 {
-		return fmt.Errorf("no valid address found for peer %s", pi.ID)
-	}
-
 	return h.dialPeer(ctx, pi.ID)
 }
 
--- a/p2p/host/peerstore/pstoreds/addr_book.go
+++ b/p2p/host/peerstore/pstoreds/addr_book.go
@@ -276,7 +276,7 @@ func (ab *dsAddrBook) AddAddrs(p peer.ID, addrs []ma.Multiaddr, ttl time.Duratio
 	if ttl <= 0 {
 		return
 	}
-	addrs = cleanAddrs(addrs)
+	addrs = cleanAddrs(addrs, p)
 	ab.setAddrs(p, addrs, ttl, ttlExtend, false)
 }
 
@@ -302,7 +302,7 @@ func (ab *dsAddrBook) ConsumePeerRecord(recordEnvelope *record.Envelope, ttl tim
 		return false, nil
 	}
 
-	addrs := cleanAddrs(rec.Addrs)
+	addrs := cleanAddrs(rec.Addrs, rec.PeerID)
 	err = ab.setAddrs(rec.PeerID, addrs, ttl, ttlExtend, true)
 	if err != nil {
 		return false, err
@@ -385,7 +385,7 @@ func (ab *dsAddrBook) SetAddr(p peer.ID, addr ma.Multiaddr, ttl time.Duration) {
 
 // SetAddrs will add or update the TTLs of addresses in the AddrBook.
 func (ab *dsAddrBook) SetAddrs(p peer.ID, addrs []ma.Multiaddr, ttl time.Duration) {
-	addrs = cleanAddrs(addrs)
+	addrs = cleanAddrs(addrs, p)
 	if ttl <= 0 {
 		ab.deleteAddrs(p, addrs)
 		return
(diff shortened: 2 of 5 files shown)


sukunrt:
I'm cleaning the address before adding it to addressbook. The other option is to change the check in all transports. This seemed cleaner to me.

p2p/host/peerstore/test/addr_book_suite.go
```diff
@@ -141,6 +141,14 @@ func testAddAddress(ab pstore.AddrBook, clk *mockClock.Mock) func(*testing.T) {
...
 			AssertAddressesEqual(t, addrs, ab.Addrs(""))
 		})
+
+		t.Run("add a /p2p address", func(t *testing.T) {
+			peerId := GeneratePeerIDs(1)[0]
+			addr := GenerateAddrs(1)
+			p2pAddr := addr[0].Encapsulate(Multiaddr("/p2p/" + peerId.String()))
+			ab.AddAddr(peerId, p2pAddr, time.Hour)
+			AssertAddressesEqual(t, addr, ab.Addrs(peerId))
+		})
```
MarcoPolo, line 151: We should have a case that fails if the peer id in the multiaddr is != to the peer id passed in.

MarcoPolo: The annoying thing is that `AddAddr` doesn't return an error. So the best we can do here is to fail silently and log an error in the implementations.

p2p/host/peerstore/pstoreds/addr_book.go
```diff
@@ -601,10 +601,12 @@ func (ab *dsAddrBook) deleteAddrs(p peer.ID, addrs []ma.Multiaddr) (err error) {
 func cleanAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
```
MarcoPolo, line 601: This should now take a peer id as well, and filter out multiaddrs that have an id component but don't match (and log an error when this happens)

p2p/host/peerstore/pstoremem/addr_book.go
```diff
@@ -238,19 +238,21 @@ func (mab *memoryAddrBook) addAddrsUnlocked(s *addrSegment, p peer.ID, addrs []m
 
 	exp := mab.clock.Now().Add(ttl)
 	for _, addr := range addrs {
-		if addr == nil {
+		// Remove suffix of /p2p/peer-id from address
+		taddr, _ := peer.SplitAddr(addr)
```
MarcoPolo, line 242: same comment as above about the peer id check. Also I think it's okay to shadow the `addr variable` with the new multiaddr.

p2p/host/peerstore/pstoremem/addr_book.go
```diff
@@ -283,17 +285,18 @@ func (mab *memoryAddrBook) SetAddrs(p peer.ID, addrs []ma.Multiaddr, ttl time.Du
 
 	exp := mab.clock.Now().Add(ttl)
 	for _, addr := range addrs {
-		if addr == nil {
+		taddr, _ := peer.SplitAddr(addr)
```
MarcoPolo, line 288: same comment as above.

MarcoPolo (changes requested):
Thanks for picking this up! This is really helpful 🎉. There's a couple of things we should do before merging this:

* The changes mentioned in the review.
* We should update the implementations of [`Host.Connect`](https://pkg.go.dev/github.com/libp2p/go-libp2p/core/host#Host) so that they check that the multiaddrs with a `p2p` component in the `Addrs` field match the `ID` field. Filter out any mismatching multiaddrs (maybe log a warning). And return an error if no addresses are dialable because of mismatched peer ids. As well as removing the `p2p` component of the multiaddr before passing it down the call stack.

sukunrt:
> * We should update the implementations of [`Host.Connect`](https://pkg.go.dev/github.com/libp2p/go-libp2p/core/host#Host) so that they check that the multiaddrs with a `p2p` component in the `Addrs` field match the `ID` field. Filter out any mismatching multiaddrs (maybe log a warning). And return an error if no addresses are dialable because of mismatched peer ids. As well as removing the `p2p` component of the multiaddr before passing it down the call stack.

All Host.Connect implementations use PeerStore.AddAddrs which will filter out for invalid ID, should I do it in the connect implementation as well? 
I was thinking of doing this:
```
func (h *BasicHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
	// absorb addresses into peerstore
	h.Peerstore().AddAddrs(pi.ID, pi.Addrs, peerstore.TempAddrTTL)
    addrs := h.Peerstore().Addrs(pi.ID)
	if len(addrs) == 0 {
		return fmt.Errorf("no valid address found for peer %s", pi.ID)
	}

```

p2p/host/peerstore/test/addr_book_suite.go
sukunrt: Added another test case with filtering on invalid peerId

p2p/host/peerstore/pstoreds/addr_book.go
sukunrt: fixed now

p2p/host/peerstore/pstoremem/addr_book.go
sukunrt: This should be fixed

p2p/host/peerstore/pstoremem/addr_book.go
sukunrt: fixed now.

sukunrt:
@MarcoPolo Addressed review comments. Do the changes in Host.Connect look okay?

p2p/host/peerstore/pstoreds/addr_book.go
sukunrt: I'm logging a warning which was previously logged for nil addresses, should I make it an error?

MarcoPolo:
> All Host.Connect implementations use PeerStore.AddAddrs which will filter out for invalid ID and log the warning, 

Good point. If there are no available addresses we'll still get that error from `Connect`. I think that's okay.

p2p/host/basic/basic_host.go
```diff
@@ -709,6 +709,11 @@ func (h *BasicHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
 		}
 	}
 
+	addrs := h.Peerstore().Addrs(pi.ID)
+	if len(addrs) == 0 {
```
MarcoPolo: On second thought we already return a similar error here https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/swarm_dial.go#L340. So this change is not required. Sorry I added confusion here.

MarcoPolo (approved):
Some small nits. I pushed a change. Thanks a ton for fixing this :)

MarcoPolo:
Will merge after CI passes.

p2p/host/basic/basic_host.go
sukunrt: that's fair. Thanks for fixing this!

masih:
Thanking you folks for getting this issue resolved so quickly 🙇

PR #2012, opened 2023-01-24
Expose muxer ids

aschmahmann:
The protocol IDs of the stream multiplexers are not exposed as constants, which makes using them a pretty miserable experience since you have to hunt through the code to find them. This is particularly true for mplex which, while still being the only multiplexer used in parts of the libp2p ecosystem, does not have its protocol ID referenced in go-libp2p outside of tests anymore (also the constructor documentation said `mplux` instead of `mplex`).

This exposes the protocol IDs with an exported constant `ID` as is done for Noise and TLS.

Also, fixes a documentation bug indicating that mplex was used by default.

Note: the various ipfs-camp-2019 examples can be updated as well post-release so that the constants are only defined once within go-libp2p.

### DIFF
--- a/libp2p.go
+++ b/libp2p.go
@@ -37,7 +37,7 @@ func ChainOptions(opts ...Option) Option {
 // transport protocols;
 //
 // - If no multiplexer configuration is provided, the node is configured by
-// default to use the "yamux/1.0.0" stream connection multiplexer;
+// default to use yamux;
 //
 // - If no security transport is provided, the host uses the go-libp2p's noise
 // and/or tls encrypted transport to encrypt all traffic;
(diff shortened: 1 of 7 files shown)


libp2p.go
```diff
@@ -37,8 +37,7 @@ func ChainOptions(opts ...Option) Option {
 // transport protocols;
 //
 // - If no multiplexer configuration is provided, the node is configured by
-// default to use the "yamux/1.0.0" and "mplux/6.7.0" stream connection
-// multiplexers;
+// default to use the "yamux/1.0.0" stream connection multiplexer;
```
aschmahmann, line 40: I left this as it was before but should this be either just the word `yamux` or `"/yamux/1.0.0"`? If so any preference? Maintainers can feel free to merge without this change, or make any change they see fit here.

marten-seemann, line 40:
```suggestion
// default to use yamux;
```

marten-seemann (approved):

PR #2015, opened 2023-01-27
quic: disable sending of Version Negotiation packets

### LINKED ISSUE #1433: quic: disable sending of Version Negotiation packets
QUIC is susceptible to Request Forgery Attacks: https://www.rfc-editor.org/rfc/rfc9000.html#name-request-forgery-with-versio

In short, the client controls both the source and the destination connection ID (both up to 255 bytes in length), which the server will echo in a Version Negotiation Packet. The client therefore is able to control byte 2 - 511 of a Version Negotiation Packet (the first bit of byte 1 is given by the wire format of the VN packet, the remaining 7 bits are chosen by random).

In a p2p setting, there's no reason to do version negotiation at all: as addresses are advertised and discovered, nothing prevents us from including the QUIC version in the advertisement. We didn't do this for draft version, as we didn't want to burn too many multiaddr code points (and the request forgery attack described above wasn't discovered yet at that point).

With QUIC becoming stable, we'll be at a point to clean this up soon: Once draft-29 is phased out of the network, we can change the [definition](https://github.com/multiformats/multiaddr/blob/master/protocols.csv#L22) of the `quic` multiaddr codepoint to mean "QUIC v1 (RFC 9000)", and disable sending of Version Negotiation packets.

Once a new version of QUIC is standardized, we'd need to define a new code point for that version.

cc @Stebalien and @aschmahmann, as we discussed this a while ago

marten-seemann:
Fixes #1433.

### DIFF
--- a/p2p/test/quic/quic_test.go
+++ b/p2p/test/quic/quic_test.go
@@ -3,6 +3,7 @@ package quic_test
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/libp2p/go-libp2p"
 	"github.com/libp2p/go-libp2p/core/network"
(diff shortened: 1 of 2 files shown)


marten-seemann:
Haven't had the time to debug this yet, but this looks like a bug. We should not have been VNPs anywhere...

marten-seemann:
> Haven't had the time to debug this yet, but this looks like a bug. We should not have been VNPs anywhere...

No bug here, just a test that was relying on an error that would be generated when receiving a VNP. See comment in the code for a more detailed explanation.

marten-seemann:
I confirmed that this won't break the IPFS network by running a Kubo node that logs sending and receiving of Version Negotiation packets. It turns out that there are still some ancient go-ipfs nodes around that use draft-27. They're using go-ipfs versions v0.5.1 (and earlier). Later Kubo versions were already using draft-29.

These nodes aren't able to establish QUIC connections these days, since they _only_ speak draft-27. For them, disabling version negotiation will mean that they'll have to wait for the handshake timeout instead of being able to abort the connection attempt right after receiving the VNP. I don't care about that.

MarcoPolo (approved):

Issue #1866, opened 2022-11-09
autorelay: refactor `libp2p.EnableAutoRelay` option


MarcoPolo:
We should split the function into two options, one that _requires_ a list of static relays and one that _requires_ a peer source. Having options, some of which are required, is not a nice API pattern.

_Originally posted by @marten-seemann in https://github.com/libp2p/go-libp2p/issues/1852#issuecomment-1307995595_

marten-seemann:
Suggestion: `libp2p.EnableAutoRelayWithStaticRelays([]peer.AddrInfo)` and `libp2p.EnableAutoRelayWithAutoDiscovery(func(ctx context.Context, num int) <-chan peer.AddrInfo)`?

I'd be grateful for better naming suggestions :)

p-shahi:
@sukunrt Flagging this for you in case you are interested in taking on more issues. Thanks for all the great contributions up till now, no pressure to take this up however :pray:

sukunrt:
sure! I'd love to!

sukunrt:
As I understand. 

We want two new functions to configure autorelay
```
libp2p.EnableAutoRelayWithStaticRelays([]peer.AddrInfo, opts ...autorelay.Option)
libp2p.EnableAutoRelayWithPeerSource(func(ctx context.Context, numPeers int) <-chan peer.AddrInfo, opts ...autorelay.Option)
```

We would need to add the second options parameter because otherwise there would be no way of providing options like WithCircuitV1Support()

In EnableAutoRelayWithPeerSource there is a configurable mininterval for callbacks which is set to 30 seconds for static relays. Should I allow for setting this minInterval too and make the function:

```
libp2p.EnableAutoRelayWithPeerSource(func(ctx context.Context, numPeers int) <-chan peer.AddrInfo, minInterval time.Duration, opts ...autorelay.Option)
```

What should I do with the old function libp2p.EnableAutoRelay? Deprecate it or remove it?

marten-seemann:
> We want two new functions to configure autorelay
> 
> ```
> libp2p.EnableAutoRelayWithStaticRelays([]peer.AddrInfo, opts ...autorelay.Option)
> libp2p.EnableAutoRelayWithPeerSource(func(ctx context.Context, numPeers int) <-chan peer.AddrInfo, opts ...autorelay.Option)
> ```

That's correct.



[...]

PR #2022, opened 2023-01-27
autorelay: Split libp2p.EnableAutoRelay into 2 functions

sukunrt:
Provide two specific ways to enable the autorelay subsystem libp2p.EnableAutoRelayWithStaticRelays
libp2p.EnableAutoRelayWithPeerSource

fixes: https://github.com/libp2p/go-libp2p/issues/1866

### DIFF
--- a/options.go
+++ b/options.go
@@ -320,7 +320,7 @@ func EnableAutoRelay(opts ...autorelay.Option) Option {
 }
 
 // EnableAutoRelayWithStaticRelays configures libp2p to enable the AutoRelay subsystem using
-// the provided slice of relays as relay candidates
+// the provided relays as relay candidates.
 // This subsystem performs automatic address rewriting to advertise relay addresses when it
 // detects that the node is publicly unreachable (e.g. behind a NAT).
 func EnableAutoRelayWithStaticRelays(static []peer.AddrInfo, opts ...autorelay.Option) Option {
@@ -332,14 +332,14 @@ func EnableAutoRelayWithStaticRelays(static []peer.AddrInfo, opts ...autorelay.O
 }
 
 // EnableAutoRelayWithPeerSource configures libp2p to enable the AutoRelay subsystem using
-// the provided peerSource callback to get more relay candidates
+// the provided peerSource callback to get more relay candidates.
 // This subsystem performs automatic address rewriting to advertise relay addresses when it
 // detects that the node is publicly unreachable (e.g. behind a NAT).
 func EnableAutoRelayWithPeerSource(peerSource func(context.Context, int) <-chan peer.AddrInfo,
 	opts ...autorelay.Option) Option {
 	return func(cfg *Config) error {
 		cfg.EnableAutoRelay = true
-		cfg.AutoRelayOpts = append([]autorelay.Option{autorelay.WithPeerSource(peerSource, 30*time.Second)}, opts...)
+		cfg.AutoRelayOpts = append([]autorelay.Option{autorelay.WithPeerSource(peerSource)}, opts...)
 		return nil
 	}
 }
--- a/options.go
+++ b/options.go
@@ -4,7 +4,6 @@ package libp2p
 // those are in defaults.go).
 
 import (
-	"context"
 	"crypto/rand"
 	"encoding/binary"
 	"errors"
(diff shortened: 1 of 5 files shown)


options.go
```diff
@@ -315,6 +319,31 @@ func EnableAutoRelay(opts ...autorelay.Option) Option {
...
+func EnableAutoRelayWithStaticRelays(static []peer.AddrInfo, opts ...autorelay.Option) Option {
+	return func(cfg *Config) error {
+		cfg.EnableAutoRelay = true
+		cfg.AutoRelayOpts = append([]autorelay.Option{autorelay.WithStaticRelays(static)}, opts...)
+		return nil
+	}
+}
+
+// EnableAutoRelayWithPeerSource configures libp2p to enable the AutoRelay subsystem using
+// the provided peerSource callback to get more relay candidates
```
marten-seemann, line 335:
```suggestion
// the provided peerSource callback to get more relay candidates.
```

options.go
```diff
@@ -315,6 +319,31 @@ func EnableAutoRelay(opts ...autorelay.Option) Option {
 	}
 }
 
+// EnableAutoRelayWithStaticRelays configures libp2p to enable the AutoRelay subsystem using
+// the provided slice of relays as relay candidates
```
marten-seemann, line 323:
```suggestion
// the provided relays as relay candidates.
```

p2p/host/autorelay/options.go
```diff
@@ -174,3 +174,12 @@ func WithClock(cl clock.Clock) Option {
 		return nil
 	}
 }
+
+// WithMinInterval sets the minimum interval after which peerSource callback will be called for more
+// candidates even if AutoRelay needs new candidates
+func WithMinInterval(interval time.Duration) Option {
```
marten-seemann, line 180: Why is the minimum interval both a parameter to `WithPeerSource` and a separate option here?

marten-seemann (commented):
One questions about the API. There's also a failing staticcheck check.

p2p/host/autorelay/options.go
sukunrt: I want to fix that, just wanted to confirm once if it's fine to change WithPeerSource method signature. 

https://github.com/libp2p/go-libp2p/issues/1866#issuecomment-1406184775

sukunrt: fixed. Removed from WithPeerSource

sukunrt:
@marten-seemann addressed review comments. 
I've removed the test TestIncorrectInit since that should be fixed by this PR and shouldn't need this test that checks for panic on incorrect initialisation.

marten-seemann (approved):
LGTM. @MarcoPolo, could you have a look at this PR please? Just want to make sure we definitely get the API right this time.

MarcoPolo:
Pushed a change to make the `PeerSource` function a separate type so that it can be documented properly

MarcoPolo (approved):

p2p/host/autorelay/options.go
```diff
@@ -10,9 +10,22 @@ import (
...
+// of the relays. Implementations must send *at most* numPeers, and close the
+// channel when they don't intend to provide any more peers. AutoRelay will not
+// call the callback again until the channel is closed. Implementations should
+// send new peers, but may send peers they sent before. AutoRelay implements a
+// per-peer backoff (see WithBackoff). See WithMinInterval for setting the
+// minimum interval between calls to the callback. The context.Context passed
+// MAY be canceled when AutoRelay feels satisfied, it will be canceled when the
+// node is shutting down. If the channel is canceled you MUST close the output
+// channel at some point.
+type PeerSource func(ctx context.Context, num int) <-chan peer.AddrInfo
```
marten-seemann, line 24: I don't think this is correct. We're passing in a `context.Context` here, not the `context.CancelFunc`, so AutoRelay has no way of canceling this context.

marten-seemann (changes requested):

p2p/host/autorelay/options.go
MarcoPolo: AutoRelay is the one passing in the context. There is a typo in this comment. It should say If the context is canceled instead of if the channel is canceled.

marten-seemann: Right, sorry for the stupid mistake.

p2p/host/autorelay/options.go
```diff
@@ -10,9 +10,22 @@ import (
...
 
+// AutoRelay will call this function when it needs new candidates because it is
+// not connected to the desired number of relays or we get disconnected from one
+// of the relays. Implementations must send *at most* numPeers, and close the
+// channel when they don't intend to provide any more peers. AutoRelay will not
+// call the callback again until the channel is closed. Implementations should
+// send new peers, but may send peers they sent before. AutoRelay implements a
+// per-peer backoff (see WithBackoff). See WithMinInterval for setting the
+// minimum interval between calls to the callback. The context.Context passed
+// MAY be canceled when AutoRelay feels satisfied, it will be canceled when the
```
marten-seemann, line 21: This is not a RFC 2119 "may".
```suggestion
// may be canceled when AutoRelay feels satisfied, it will be canceled when the
```

marten-seemann (approved):

marten-seemann (approved):

PR #2025, opened 2023-01-27
quic: rcmgr: Fix connection accounting

MarcoPolo:
The quic transport didn't end the connection reservation in some error cases. The fix here is to use the "err defer" pattern to handle all error cases by checking for errors in a defer statement and finishing the scope then. This pattern is more resilient to future changes since future code doesn't have to remember to free resources on every err return statement.

Tested by having Kubo run for a while.

Before:
Screenshot 2023-01-27 at 10 09 03 AM

After:
Screenshot 2023-01-27 at 12 30 16 PM

This was debugged with this [commit](https://github.com/libp2p/go-libp2p/commit/016861124a5ee0c6f7207ddeabe9bf1321ff04bf). It would give some memory to the connections in the swarm and then print out any connection scopes that didn't have memory and were not done. This meant that they were created but not passed down to the swarm and not finished.

### DIFF
--- a/p2p/transport/quic/listener.go
+++ b/p2p/transport/quic/listener.go
@@ -56,15 +56,13 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 		}
 		c, err := l.setupConn(qconn)
 		if err != nil {
-			qconn.CloseWithError(1, err.Error())
-			continue
-		}
-		if l.transport.gater != nil && !(l.transport.gater.InterceptAccept(c) && l.transport.gater.InterceptSecured(network.DirInbound, c.remotePeerID, c)) {
-			c.scope.Done()
-			qconn.CloseWithError(errorCodeConnectionGating, "connection gated")
 			continue
 		}
 		l.transport.addConn(qconn, c)
+		if l.transport.gater != nil && !(l.transport.gater.InterceptAccept(c) && l.transport.gater.InterceptSecured(network.DirInbound, c.remotePeerID, c)) {
+			c.closeWithError(errorCodeConnectionGating, "connection gated")
+			continue
+		}
 
 		// return through active hole punching if any
 		key := holePunchKey{addr: qconn.RemoteAddr().String(), peer: c.remotePeerID}
@@ -84,7 +82,7 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 	}
 }
 
-func (l *listener) setupConn(qconn quic.Connection) (*conn, error) {
+func (l *listener) setupConn(qconn quic.Connection) (_c *conn, _err error) {
 	remoteMultiaddr, err := quicreuse.ToQuicMultiaddr(qconn.RemoteAddr(), qconn.ConnectionState().Version)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


p2p/transport/quic/listener.go
```diff
@@ -84,7 +82,7 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 	}
 }
 
-func (l *listener) setupConn(qconn quic.Connection) (*conn, error) {
+func (l *listener) setupConn(qconn quic.Connection) (_c *conn, _err error) {
```
marten-seemann, line 85: Is the `_` needed? I've never seen this pattern.

marten-seemann:
There's a flaky rcmgr test:
```
  === RUN   TestResourceManagerAcceptDenied/reuseport_off
      conn_test.go:59: using a RSA key: QmUJNXf8tyYMn2LJbsBoTBBT9Xf2YXDBKSK4KV9dc3G5Yh
      conn_test.go:59: using a Secp256k1 key: 16Uiu2HAm5X88sRQ9t6x9JVuR3h9Bp9jApD83kRUXF9AawHGbtvA2
      controller.go:269: missing call(s) to *mocknetwork.MockConnManagementScope.Done() /home/runner/work/go-libp2p/go-libp2p/p2p/transport/quic/conn_test.go:238
      controller.go:269: aborting test due to missing call(s)
  --- FAIL: TestResourceManagerAcceptDenied (1.17s)
      --- PASS: TestResourceManagerAcceptDenied/reuseport_on (0.48s)
      --- FAIL: TestResourceManagerAcceptDenied/reuseport_off (0.69s)

p2p/transport/quic/listener.go
MarcoPolo: not needed, I don't want to shadow or deal with shadowed vars. I also don't want this to be used in the body of the function.

I also want to avoid the use of the named return values _except_ for this deferred error handling.

If you have a better naming suggestion I'm open to it.

BigLep:
Thanks @MarcoPolo .   Do you think this would have impacted System.ConsInbound?  If so, I agree it would be nice to include in [Kubo 0.18.1](https://github.com/ipfs/kubo/issues/9579) .  If not, I think it can wait.

MarcoPolo:
>  Do we have the same problem in the WebTransport transport?

WebTransport doesn't use the err defer pattern but looks okay.

MarcoPolo:
> Do you think this would have impacted System.ConsInbound

Looking at the data and the code, conns inbound should be safe from this miscount.

marten-seemann:
> Thanks @MarcoPolo . Do you think this would have impacted System.ConsInbound? If so, I agree it would be nice to include in [Kubo 0.18.1](https://github.com/ipfs/kubo/issues/9579) . If not, I think it can wait.

It looks like this won't affect the code path that Kubo is using for its resource management in v0.18. Cutting and bubbling up a patch release is pretty low overhead though, so we should consider doing it anyway. This has confused Antonio and it has come up in https://github.com/protocol/bifrost-infra/issues/2308. We might be able to avoid some spurious bug reports if we get this fix out asap.

PR #2031, opened 2023-01-31
circuitv2: fix resource accounting when connection upgrading fails

marten-seemann:
Very similar to #2015.


MarcoPolo:
you mean #2025 ?

MarcoPolo (commented):
Where is the missing connScope.Done call?

marten-seemann:
> Where is the missing connScope.Done call?

Here: https://github.com/libp2p/go-libp2p/blob/34dc11547a10e7abe0a3fc615dbf839818554cab/p2p/protocol/circuitv2/client/transport.go#L66-L69

PR #2036, opened 2023-01-31
chore: remove textual roadmap in favor for Starmap

p-shahi:
The textual roadmap is a chore to maintain and update. The PL organization is using Starmap, I propose we rely on Starmap to display the timeline instead of maintaining it in a markdown file.
Furthermore, when this Starmap feature https://github.com/pln-planning-tools/Starmap/issues/120 is shipped, we can leverage that to have a textual view


MarcoPolo (approved):

SgtPooki:
Thanks for referencing the starmap issue.. this helps us to know how important the list feature is. I've bumped up it's priority a bit.. and I think it can be done outside of the d3 refactor (https://github.com/pln-planning-tools/Starmap/issues/237), which should help for getting the list feature done sooner than later.

PR #2037, opened 2023-01-31
rcmgr: *: Always close connscope

### LINKED ISSUE #2033: rcmgr: Finish reservation in one place when connection handshake fails
The correct flow when negotiating a connection is:

1. Get a connection reservation.
2. Handshake/negotiate the connection.
3. On error, finish the reservation.
4. On success return a connection that will finish the reservation on connection `.Close()`.

Currently for step 3, we are finishing the reservation in multiple code paths. This is error prone since it's easy to forget to finish a reservation somewhere and have a connection reservation leak. See #2025 for an example.

The fix here is to refactor the code so that the on error step always happens. The two approaches are to either:
1. Wrap the logic in another function, and finish the reservation by the caller on error.
2. Handle the error in a `defer` func (aka err defer).

1 is probably more idiomatic Go code, so prefer that.

MarcoPolo:
Fixes #2033.

Closes #2025  and closes #2031 .

The general pattern here is replacing 

```
func ...Dial() {
  connScope := ...
  
  err := ...
  if err != nil {
    connScope.Done()
  }
  
    err := ...
  if err != nil {
    connScope.Done()
  }
  
    err := ...
  if err != nil {
    connScope.Done()
  }
}
```

With: 

```
func ...Dial() {
  connScope := ...
  
  c, err := dialWithScope
  
  if err != nil {
    connScope.Done()
    return nil, err
  }
  
  return c, nil
}

func dialWithScope(..., connScope) {
  if err != nil {
     return nil, err
  }
}
```

### DIFF
--- a/p2p/net/swarm/swarm_conn.go
+++ b/p2p/net/swarm/swarm_conn.go
@@ -207,9 +207,18 @@ func (c *Conn) NewStream(ctx context.Context) (network.Stream, error) {
 	if err != nil {
 		return nil, err
 	}
-	ts, err := c.conn.OpenStream(ctx)
+
+	s, err := c.openAndAddStream(ctx, scope)
 	if err != nil {
 		scope.Done()
+		return nil, err
+	}
+	return s, nil
+}
+
+func (c *Conn) openAndAddStream(ctx context.Context, scope network.StreamManagementScope) (network.Stream, error) {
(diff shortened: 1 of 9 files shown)


marten-seemann (approved):
Good idea to fix this class of issues once and for all!

PR #2038, opened 2023-02-01
eventbus: add metrics

sukunrt:
Related to: https://github.com/libp2p/go-libp2p/issues/2020

### DIFF
--- a/p2p/host/eventbus/basic.go
+++ b/p2p/host/eventbus/basic.go
@@ -6,7 +6,6 @@ import (
 	"reflect"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/libp2p/go-libp2p/core/event"
 )
@@ -37,15 +36,12 @@ func (e *emitter) Emit(evt interface{}) error {
 	if atomic.LoadInt32(&e.closed) != 0 {
 		return fmt.Errorf("emitter is closed")
 	}
-	st := time.Now()
 
 	e.n.emit(evt)
 	e.w.emit(evt)
 
 	if e.metricsTracer != nil {
-		d := time.Since(st)
 		e.metricsTracer.EventEmitted(e.typ)
-		e.metricsTracer.NotificationTime(e.typ, d)
 	}
 	return nil
 }
--- a/p2p/host/eventbus/basic.go
+++ b/p2p/host/eventbus/basic.go
@@ -60,21 +60,15 @@ func (e *emitter) Close() error {
 	return nil
 }
 
-func NewBus() event.Bus {
-	return &basicBus{
+func NewBus(opts ...Option) event.Bus {
+	bus := &basicBus{
 		nodes:    map[reflect.Type]*node{},
-		wildcard: new(wildcardNode),
+		wildcard: &wildcardNode{},
 	}
-}
-
-func NewBusWithMetrics() event.Bus {
-	metricsTracer := NewMetricsTracer()
-	wildcardNode := &wildcardNode{metricsTracer: metricsTracer}
-	return &basicBus{
-		nodes:         map[reflect.Type]*node{},
-		wildcard:      wildcardNode,
-		metricsTracer: metricsTracer,
+	for _, opt := range opts {
+		opt(bus)
 	}
+	return bus
 }
 
 func (b *basicBus) withNode(typ reflect.Type, cb func(*node), async func(*node)) {
--- a/p2p/host/eventbus/basic.go
+++ b/p2p/host/eventbus/basic.go
@@ -129,6 +129,10 @@ func (w *wildcardSub) Close() error {
 	return nil
 }
 
+func (w *wildcardSub) Name() string {
+	return w.name
+}
+
 type namedSink struct {
 	name string
 	ch   chan interface{}
@@ -140,6 +144,10 @@ type sub struct {
 	dropper       func(reflect.Type)
 	metricsTracer MetricsTracer
 	name          string
+}
+
+func (s *sub) Name() string {
+	return s.name
 }
 
 func (s *sub) Out() <-chan interface{} {
--- a/p2p/host/eventbus/basic.go
+++ b/p2p/host/eventbus/basic.go
@@ -6,6 +6,7 @@ import (
 	"reflect"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/libp2p/go-libp2p/core/event"
 )
@@ -36,9 +37,14 @@ func (e *emitter) Emit(evt interface{}) error {
 	if atomic.LoadInt32(&e.closed) != 0 {
 		return fmt.Errorf("emitter is closed")
 	}
+	st := time.Now()
+
 	e.n.emit(evt)
 	e.w.emit(evt)
+
+	d := time.Since(st)
 	e.metricsTracer.EventEmitted(e.typ)
+	e.metricsTracer.NotificationTime(e.typ, d)
 	return nil
 }
 
@@ -53,10 +59,12 @@ func (e *emitter) Close() error {
 }
 
 func NewBus() event.Bus {
+	metricsTracer := NewMetricsTracer()
+	wildcard := &wildcardNode{metricsTracer: metricsTracer}
 	return &basicBus{
 		nodes:         map[reflect.Type]*node{},
-		wildcard:      new(wildcardNode),
-		metricsTracer: NewMetricsTracer(),
+		wildcard:      wildcard,
+		metricsTracer: metricsTracer,
 	}
 }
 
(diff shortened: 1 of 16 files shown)


sukunrt:
You'll need to change 
```
"datasource": {
                                "type": "prometheus",
                                "uid": "aMik1H0Vz"
                        },
                        
  ```
datasource.uid to your prometheus datasource uid.

File: p2p/host/eventbus/basic_metrics.go [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+	prometheus.MustRegister(eventsEmitted)
+}
+
+func NewMetricsTracer() MetricsTracer {
+	initMetricsOnce.Do(initMetrics)
+	return &metricsTracer{}
+}
+
+func (m *metricsTracer) EventEmitted(typ reflect.Type) {
+	eventsEmitted.WithLabelValues(typ.String()).Inc()
```
marten-seemann, line 43: Should we filter out the package name here (the `event.`)?

This is how it looks like in Grafana:
image

sukunrt: fixed.

p2p/host/eventbus/basic_metrics.go
```diff
@@ -0,0 +1,147 @@
...
+		[]string{"subscriber_name"},
+	)
+	subscriberEventQueued = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metricNamespace,
+			Name:      "subscriber_event_queued",
+			Help:      "Event Queued for subscriber",
+		},
+		[]string{"subscriber_name"},
+	)
```
marten-seemann, line 64: I'd like to avoid having subscribers pass in a name. Can we just make this per event?

p2p/host/eventbus/basic.go
```diff
@@ -57,12 +67,22 @@ func NewBus() event.Bus {
 	}
 }
 
+func NewBusWithMetrics() event.Bus {
```
marten-seemann, line 70: This works totally fine, but we usually use the option pattern:
```go
type Option func(*basicBus) error

func NewBus(opts ...Option) event.Bus
```

p2p/host/eventbus/basic_metrics.go
sukunrt: But the queues are per subscriber and not per event. 
We can remove the subscriber metrics and just track whether for a given event type one of the subscriber queues was full.

p2p/host/eventbus/basic.go
sukunrt: will fix. This would also be backwards compatible. nice :)

p2p/host/eventbus/basic_metrics.go
```diff
@@ -0,0 +1,147 @@
...
+	} else {
+		observer.Set(0)
+	}
+}
+
+func (m *metricsTracer) SubscriberEventQueued(name string) {
+	subscriberEventQueued.WithLabelValues(name).Inc()
+}
+
+func stripPrefix(s string) string {
```
marten-seemann, line 141: Nit: https://pkg.go.dev/strings#TrimPrefix

p2p/host/eventbus/basic.go
marten-seemann: This is what we did for the swarm: https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/swarm.go#L74-L79

sukunrt: fixed. Not returning error for now since there's only one option that can't error.

p2p/host/eventbus/basic_metrics.go
sukunrt: fixed.

sukunrt:
@marten-seemann Addressed review comments. Removed custom passed in names from all subscribers for now. they are set to the filename. If you want I'll add them back.

p2p/host/eventbus/basic_metrics.go
```diff
@@ -0,0 +1,139 @@
...
+			Namespace: metricNamespace,
+			Name:      "events_emitted_total",
+			Help:      "Events Emitted",
+		},
+		[]string{"event"},
+	)
+	notificationTime = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metricNamespace,
+			Name:      "notification_time_microseconds",
```
marten-seemann, line 26: There's a convention in Prometheus to use SI units: https://prometheus.io/docs/practices/naming/. So this should be `notification_time_seconds`.

p2p/host/eventbus/basic_metrics.go
```diff
@@ -0,0 +1,139 @@
...
+			Help:      "Time taken to notify all subscribers",
+			Buckets:   prometheus.ExponentialBuckets(1.0, 2, 10),
+		},
+		[]string{"event"},
+	)
+	numSubscribers = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: metricNamespace,
+			Name:      "num_subscribers",
```
marten-seemann, line 35:
```suggestion
			Name:      "subscribers_total",
```

marten-seemann: In Grafana, this would look pretty cool as a Gauge: 
image

p2p/host/eventbus/grafana-dashboards/eventbus.json
```diff
@@ -0,0 +1,835 @@
...
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "builder",
+          "expr": "sum by(event) (rate(libp2p_eventbus_events_emitted_total[1m]))",
```
marten-seemann, line 237: It's better to use the `$__rate_interval` variable instead of hardcoding a value. Also, I don't think you need the `sum`. This configuration should work:
image

p2p/host/eventbus/basic_metrics.go
marten-seemann: Thinking about this a bit more, this measurement doesn't seem as useful any more:
* If all channels don't block, this takes basically no time at all, we're just adding an event to every channel.
* If a channel blocks because the subscriber is not reading from the channel, it will block indefinitely, and then this event won't be emitted.

I suggest remote it.

sukunrt: fixed

p2p/host/eventbus/basic_metrics.go
sukunrt: fixed

p2p/host/eventbus/grafana-dashboards/eventbus.json
sukunrt: removed hardcoding

sukunrt:
@marten-seemann  addressed review comments.

sukunrt:
made subscriber queue a state timeline

p2p/host/eventbus/basic_metrics.go
```diff
@@ -0,0 +1,121 @@
...
+	)
+	totalSubscribers = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: metricNamespace,
+			Name:      "subscribers_total",
+			Help:      "Number of subscribers for an event type",
+		},
+		[]string{"event"},
+	)
+	subscriberQueueLength = prometheus.NewHistogramVec(
```
MarcoPolo, line 30: Why is this a histogram? I think this should be a gauge. What do you think was the value in having this be a histogram?

If it's a gauge, the promql query also becomes simpler and correct. I think the current query in the grafana dashboard is wrong. It's showing you the increase in queue length. If you're full, you'll see that number go to 0 since it's not increasing.

MarcoPolo (dismissed):

p2p/host/eventbus/basic_metrics.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+type metricsTracer struct{}
+
+var _ MetricsTracer = &metricsTracer{}
+
+var initMetricsOnce sync.Once
+
+func initMetrics() {
+	prometheus.MustRegister(eventsEmitted, totalSubscribers, subscriberQueueLength, subscriberQueueFull, subscriberEventQueued)
```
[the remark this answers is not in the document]
sukunrt: fixed. I'm using 
```
type MetricsTracerOption = func(*metricsTracerSetting)

func NewMetricsTracer(opts ...MetricsTracerOption) MetricsTracer
```

p2p/host/eventbus/basic_metrics.go
sukunrt: I wanted to look at things like 90%ile queue length. I guess we can ignore that since if the consumer is slow the queue will get full eventually.  Having a gauge alerts us to increasing queue length which should be enough. 

I didn't realise the current query would go to 0, I thought sum / count would stay constant. I just checked, it does go to 0 😅

marten-seemann (approved):
I added two minor changes to this PR:
1. Removing the version number from the subscriber event name. This makes it more pleasant to view in Grafana.
2. A number of minor tweaks to the Grafana dashboard.

I think this is ready to go now. Thanks to @sukunrt for doing all the heavy lifting here!

PR #2039, opened 2023-02-01
Migrate to test-plan composite action

thomaseizinger:
I wasn't sure if you wanted to keep the Makefile. I deleted it for now because we no longer create an archive. Feel free to take over to fix up any additional things.


MarcoPolo (approved):

thomaseizinger:
I love Rust but I am jealous that you can build a `go-libp2p` application without any caching in a new docker container in less than a minute: https://github.com/libp2p/go-libp2p/actions/runs/4069702343/jobs/7009727962#step:3:59

MarcoPolo:
And it's statically linked! Go tooling is quite nice.

thomaseizinger:
Is this an optimized binary or a debug build? We are currently running release builds in the tests but they take even longer so I am thinking of switching to debug builds unless we really need the absolute best performance.

MarcoPolo:
This is a release build. Debug should be fine. Most of the overhead is in the container namespacing and networking.

PR #2050, opened 2023-02-03
notifications: fix TestListenAddressNotif

sukunrt:
The order of the received addresses isn't fixed. It's received from Swarm.listenAddressesNoLock which is iterating over a map.


sukunrt:
closed in favour of https://github.com/libp2p/go-libp2p/pull/2051

Issue #2019, opened 2023-01-27
identify: expose metrics


marten-seemann:
This has already proved valuable in https://github.com/libp2p/go-libp2p/issues/1974.

For both sender and receiver side, we should track:
* number of Identify / Identify Push messages sent
* number of addresses in the Identify message
* number of protocols added / removed

Issue #2055, opened 2023-02-04
metrics: make it easier to update / discover Grafana dashboard


marten-seemann:
Currently all of our Grafana dashboards are available right next to the instrumented code. 
They are:
* [swarm](https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/grafana-dashboards/swarm.json)
* [eventbus](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/eventbus/grafana-dashboards/eventbus.json)
* [resource-manager](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/resource-manager/obs/grafana-dashboards/resource-manager.json)
* [identify](https://github.com/libp2p/go-libp2p/pull/2069/files#diff-6c19b257ba3054d4f415bf581870ee279f98c2ab42cd5a7fc1a07c708eac7920)

We need to make it as easy as possible for our users to import all the dashboards. At the moment, we define 3 dashboards (swarm, rcmgr, eventbus), but with #1356, we'll soon have ~10 dashboards.

One option to simplify things a bit would be to create a single `dashboards` directory in the root directory that holds all dashboards. Users would still have to open every JSON file and copy-paste it into their Grafana instance, but at least they don't have to walk our directory tree.

Another option would be to evaluate the grafana.com import option:
<img width="650" alt="image" src="https://user-images.githubusercontent.com/1478487/216792365-2f92e0af-70af-43f6-880b-c07f34bf1a4d.png">
Apparently you can import a dashboard by URL or by ID. Unfortunately, a URL pointing to a GitHub repository doesn't work, it has to be a grafana.com URL. Maybe there's an easy way to publish our dashboards on grafana.com via a GitHub Action when we cut a release?

MarcoPolo:
We should also have a README in that directory.

MarcoPolo:
This is done with all dashboards in a single folder. We can't have a single one-click to import all the dashboards with what Grafana provides.

PR #2069, opened 2023-02-07
identify: add some basic metrics

marten-seemann:
We're introducing some pretty substantial changes to the Identify protocol. It would be good to have some basic metrics.
We can probably do a lot more here, but this is a beginning so we're not flying completely blind.

Part of #2019.

### DIFF
--- a/p2p/protocol/identify/grafana-dashboards/dashboard.json
+++ b/p2p/protocol/identify/grafana-dashboards/dashboard.json
@@ -1,0 +1,355 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__elements": {},
+  "__requires": [
+    {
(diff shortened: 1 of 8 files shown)


p2p/protocol/identify/grafana-dashboards/dashboard.json
```diff
@@ -0,0 +1,355 @@
+{
```
p-shahi, line 1: Wdyt about moving this dashboard and others to a common top-level folder? So this could be `grafana-dashboards/identify/dashboard.json` and similarly `grafana-dashboards/swarm/swarm.json` and so on

This would make it easy to discover all of them at once and not make them buried inside each module.

marten-seemann: See https://github.com/libp2p/go-libp2p/issues/2055. I suggest doing that in a separate PR, and probably after the v0.25 release.

p-shahi: good stuff, hadn't seen that.

MarcoPolo (approved):

MarcoPolo (approved):

PR #2080, opened 2023-02-09
chore: update examples to v0.25.1

MarcoPolo:
To v0.25

This uses a version of kademlia from https://github.com/libp2p/go-libp2p-kad-dht/pull/815. I think we can merge this and update this version later so that we unblock https://github.com/libp2p/go-libp2p/pull/2073


marten-seemann (changes requested):
I think there are more go.mod files hiding in the examples directory.

marten-seemann (approved):
Fixed the missing examples.

PR #2084, opened 2023-02-10
chore: Introduce a Changelog

MarcoPolo:
In a draft for a discussion.

I like the idea of having a changelog file in the repo that can be updated as features are rolled out. Kubo does this and rust-libp2p does this. The less work we can do at release time the better. Ideally releases should feel almost like a no-op.

For the initial version, I copied the release notes from the tracking issue. In the future, I imagine the tracking issue is used as coordination/planning (as it's done today) and breaking changes/features are added in the changelog as we go. This can then be copy/pasted into the release notes.

I would also like to see the semantic PR titles being used to highlight breaking changes and such. But that's somewhat orthogonal to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,126 @@
+# Table Of Contents <!-- omit in toc -->
+- [v0.25.1](#v0251)
+- [v0.25.0](#v0250)
+  - [🔦 Highlights](#-highlights)
+    - [Metrics](#metrics)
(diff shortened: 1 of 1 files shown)


p-shahi (approved):
👍 this is great to have

p-shahi:
Maintainers agreed to try the changelog out in the 2023-02-13 occurrence of the go-libp2p triage meeting, so I'm merging this

Issue #2056, opened 2023-02-04
Go 1.19: use atomic.Pointer instead of atomic.Value


marten-seemann:
There's a few places in our codebase where we're currently using an `atomic.Value`:
```
p2p/net/connmgr/decay.go
41:     lastTick atomic.Value

p2p/net/mock/mock_stream.go
34:     protocol atomic.Value

p2p/host/autonat/autonat.go
36:     status atomic.Value

p2p/net/swarm/swarm_stream.go
27:     protocol atomic.Value

p2p/net/swarm/swarm.go
149:    streamh atomic.Value
```

Go 1.19 added a `atomic.Pointer`, which makes use of generics, and allows us to do a type-safe atomic store and load operation. We should replace all uses of `atomic.Value` with `atomic.Pointer`.

This can be done as soon as Unified CI is updated and we've dropped support for Go 1.18.

marten-seemann:
#2073 was just merged, so this is ready to go now. @sukunrt Do you want to give it a shot?

sukunrt:
sure! I'll pick this.

PR #2088, opened 2023-02-13
change atomic.Value to atomic.Pointer

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2056

### DIFF
--- a/p2p/net/connmgr/decay.go
+++ b/p2p/net/connmgr/decay.go
@@ -38,7 +38,7 @@ type decayer struct {
 	knownTags map[string]*decayingTag
 
 	// lastTick stores the last time the decayer ticked. Guarded by atomic.
-	lastTick atomic.Value
+	lastTick atomic.Pointer[time.Time]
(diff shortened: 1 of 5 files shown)


p2p/net/connmgr/decay.go
```diff
@@ -163,7 +164,8 @@ func (d *decayer) process() {
 	for {
 		select {
 		case now = <-ticker.C:
-			d.lastTick.Store(now)
+			nn := now
+			d.lastTick.Store(&nn)
```
marten-seemann, line 168: Why not `&now`?

sukunrt: now is declared outside the block and reassigned at line 219. storing a pointer to now would cause it to use this value of 219 sometimes.

marten-seemann: Good catch!

marten-seemann (approved):

Issue #2043, opened 2023-02-02
Go 1.19: use atomic.Bool


marten-seemann:
We're using `atomic.{U}Int{32,64}` to store a 0 or a 1 at a few places across the code base. We should replace these with the `atomic.Bool` introduced in Go 1.19, as this will make the code easier to read.

This can be done as soon as Unified CI is updated and we've dropped support for Go 1.18.

marten-seemann:
#2073 was just merged, so this is ready to go now. @sukunrt Do you want to give it a shot?

sukunrt:
yes! I'll pick this after https://github.com/libp2p/go-libp2p/issues/2056

PR #2089, opened 2023-02-13
use atomic.Bool instead of int32 operations

sukunrt:
Fixes: https://github.com/libp2p/go-libp2p/issues/2043

@marten-seemann 
There are a bunch of uses of int32 and int64 as atomic counters with atomic.AddInt32 or atomic.AddInt64. We can replace those with atomic.Int32 and atomic.Int64. Should I change those too?

### DIFF
--- a/p2p/host/eventbus/basic.go
+++ b/p2p/host/eventbus/basic.go
@@ -27,13 +27,13 @@ type emitter struct {
 	n             *node
 	w             *wildcardNode
 	typ           reflect.Type
-	closed        int32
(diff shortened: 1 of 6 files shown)


marten-seemann (approved):

marten-seemann:
> There are a bunch of uses of int32 and int64 as atomic counters with atomic.AddInt32 or atomic.AddInt64. We can replace those with atomic.Int32 and atomic.Int64. Should I change those too?

That would be great! Thank you!

Issue #2046, opened 2023-02-03
autonat (?): node frequently changes its mind about its reachability status


marten-seemann:
Using the event bus metrics (#2038) on a Kubo node with Accelerated DHT client enabled, it looks like the node somewhat frequently changes its mind about its reachability status. From the event metrics, we won't be able to tell what it thinks the availability is, but on a public node I wouldn't expect _any_ changes on a node that has been running for more than 5 minutes or so.

<img width="688" alt="image" src="https://user-images.githubusercontent.com/1478487/216482074-ea3efc20-6b8c-4dae-b503-21d5ef99139d.png">

This can have interesting consequences on higher layers. For example, when a node goes private, it will leave the DHT by switching to client mode. I'm wondering how much of the observed churn can be attributed to this.

As expected, this is accompanied by a change in supported protocols, presumably the DHT switching back and forth between client and server mode:
<img width="695" alt="image" src="https://user-images.githubusercontent.com/1478487/216482761-841e4467-febc-4ac2-b830-cc99c7804b9a.png">

Unexpectedly, we don't observe any change in local addresses. It's not clear to my why we're not obtaining a relay reservation. Maybe we're switching back and forth too quickly to actually obtain the reservation? Alternatively, there could also be a bug in AutoRelay.
<img width="685" alt="image" src="https://user-images.githubusercontent.com/1478487/216483167-f03576a3-2769-42da-82cd-d35a730d1399.png">

This issue suggests that it would be valuable to pick up the AutoNAT metrics (https://github.com/libp2p/go-libp2p/issues/2017) next. This will hopefully give us a better understanding of what's going on.

cc @Jorropo @dennis-tra @yiannisbot

sukunrt:
I added lots of logs on a kubo node locally and observed that all these changes on my node are happening 
when the current autonat.status is Public has address with protocol "quic" and the autonat server we contact dialsback on address with protocol "quic-v1" or if the current address is "quic-v1" and server dialsback with "quic"

the reason(bug?) for emitting a reachability change event are these lines: 
https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autonat/autonat.go#L306-L309
They emit an event even when the observed address has changed. This seems incorrect since reachability is still public and has not changed. 

LocalAddresses don't change here because both "quic" and "quic-v1" are used to communicate with some peers and they're both available from ObservedAddrsManager. 

On this branch: https://github.com/libp2p/go-libp2p/pull/2086
[...]

PR #2092, opened 2023-02-14
autonat: don't emit reachability changed events on address change

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2046

@marten-seemann I've changed:

1. No events are emitted if only the observed address is changed.
2. All observations with reachability = Public and addr = nil are ignored. 
    - previously this caused an increase in confidence.
3. transition from unknown to private now begins with confidence = 0. 
    - previously this begin with confidence = 1

### DIFF
--- a/p2p/host/autonat/autonat.go
+++ b/p2p/host/autonat/autonat.go
@@ -287,11 +287,6 @@ func (as *AmbientAutoNAT) scheduleProbe() time.Duration {
 func (as *AmbientAutoNAT) recordObservation(observation autoNATResult) {
 	currentStatus := as.status.Load()
 
-	// Ignore public observations with no address
-	if observation.Reachability == network.ReachabilityPublic && observation.address == nil {
-		return
-	}
-
 	if observation.Reachability == network.ReachabilityPublic {
 		log.Debugf("NAT status is public")
 		changed := false
--- a/p2p/host/autonat/autonat.go
+++ b/p2p/host/autonat/autonat.go
@@ -286,10 +286,18 @@ func (as *AmbientAutoNAT) scheduleProbe() time.Duration {
 // Update the current status based on an observed result.
 func (as *AmbientAutoNAT) recordObservation(observation autoNATResult) {
 	currentStatus := as.status.Load()
+
+	// Ignore public observations with no address
+	if observation.Reachability == network.ReachabilityPublic && observation.address == nil {
+		return
+	}
+
 	if observation.Reachability == network.ReachabilityPublic {
 		log.Debugf("NAT status is public")
 		changed := false
 		if currentStatus.Reachability != network.ReachabilityPublic {
+			// Aggressively switch to public from other states ignoring confidence
+
 			// we are flipping our NATStatus, so confidence drops to 0
 			as.confidence = 0
 			if as.service != nil {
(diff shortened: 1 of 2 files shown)


p2p/host/autonat/autonat.go
```diff
@@ -286,10 +286,18 @@ func (as *AmbientAutoNAT) scheduleProbe() time.Duration {
 // Update the current status based on an observed result.
 func (as *AmbientAutoNAT) recordObservation(observation autoNATResult) {
 	currentStatus := as.status.Load()
+
+	// Ignore public observations with no address
+	if observation.Reachability == network.ReachabilityPublic && observation.address == nil {
```
marten-seemann, line 291: If I understand correctly, this can't really happen. A successful observation will always contain a multiaddress, and if the peer sends a multiaddress that we can't decode, we'd record this as a `ReachabilityUnknown` observation. We shouldn't do that either, we should discard that measurement.

marten-seemann (commented):
Is there any reason to keep the `address` field on the `autoNATresult`?

p2p/host/autonat/autonat.go
sukunrt: I checked. It cannot happen. cli.DialBack(ctx, pi.ID) ensures that address is not nil when err is nil. 
I added this because TestAutoNATObservationRecording specifically has a check for this case. 

I'll remove both.

sukunrt: fixed.

p2p/host/autonat/autonat_test.go
```diff
@@ -252,6 +239,21 @@ func TestAutoNATObservationRecording(t *testing.T) {
...
 
+	// don't emit events on observed address change
+	newAddr, _ := ma.NewMultiaddr("/ip4/127.0.0.1/udp/12345")
+	an.recordObservation(autoNATResult{network.ReachabilityPublic, newAddr})
+	if an.Status() != network.ReachabilityPublic {
+		t.Fatalf("reachability should stay public")
+	}
+	select {
+	case e := <-s.Out():
+		_, ok := e.(event.EvtLocalReachabilityChanged)
```
marten-seemann, line 250: Nit: remove the `ok` assertion here. It's ok to panic (especially since it's a test), and would make the test more robust.

marten-seemann (approved):
The `recordObservation` logic looks correct, and is a nice simplication.

p2p/host/autonat/autonat_test.go
sukunrt: So I removed the assertion completely. The subscription is for reachability change, any event received here is an error in this case.

p2p/host/autonat/autonat_test.go
```diff
@@ -252,6 +239,17 @@ func TestAutoNATObservationRecording(t *testing.T) {
...
+	// don't emit events on observed address change
+	newAddr, _ := ma.NewMultiaddr("/ip4/127.0.0.1/udp/12345")
+	an.recordObservation(autoNATResult{network.ReachabilityPublic, newAddr})
+	if an.Status() != network.ReachabilityPublic {
+		t.Fatalf("reachability should stay public")
+	}
+	select {
+	case <-s.Out():
+		t.Fatal("received event without state transition")
+	case <-time.After(1 * time.Second):
```
marten-seemann, line 251:
```suggestion
	case <-time.After(300 * time.Millisecond):
```

marten-seemann (approved):
Thank you @sukunrt!

Issue #2091, opened 2023-02-14
relay circuitv2: relaymanager starts new relays without closing older ones


sukunrt:
While debugging https://github.com/libp2p/go-libp2p/issues/2046 , we found out that autonat would emit events when the observed address has changed even though the node reachability hasn't changed. 

I found out that the LocalProtocolsUpdated event was coming from RelayManager.reachabilityChanged which starts a new relay if the observed event is reachability.Public
https://github.com/libp2p/go-libp2p/blob/master/p2p/host/relaysvc/relay.go#L65-L85

This is incorrect. It should only start a new relay if the previous state was not public or at least close any previous running relays and then start a new one. 

cc: @marten-seemann

PR #2093, opened 2023-02-14
relaymanager: do not start new relay if one already exists

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2091

### DIFF
--- a/p2p/host/relaysvc/relay.go
+++ b/p2p/host/relaysvc/relay.go
@@ -65,14 +65,18 @@ func (m *RelayManager) background(ctx context.Context) {
 func (m *RelayManager) reachabilityChanged(r network.Reachability) error {
 	switch r {
 	case network.ReachabilityPublic:
+		m.mutex.Lock()
+		defer m.mutex.Unlock()
+		// ignore events with no reachability change
+		if m.relay != nil {
+			return nil
+		}
 		relay, err := relayv2.New(m.host, m.opts...)
 		if err != nil {
 			return err
 		}
-		m.mutex.Lock()
-		defer m.mutex.Unlock()
 		m.relay = relay
-	case network.ReachabilityPrivate:
+	default:
 		m.mutex.Lock()
 		defer m.mutex.Unlock()
 		if m.relay != nil {
--- a/p2p/host/relaysvc/relay.go
+++ b/p2p/host/relaysvc/relay.go
@@ -67,7 +67,8 @@ func (m *RelayManager) reachabilityChanged(r network.Reachability) error {
 	case network.ReachabilityPublic:
 		m.mutex.Lock()
 		defer m.mutex.Unlock()
-		// ignore events with no reachability change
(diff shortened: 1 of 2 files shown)


p2p/host/relaysvc/relay_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/libp2p/go-libp2p/core/network"
+	bhost "github.com/libp2p/go-libp2p/p2p/host/blank"
+	swarmt "github.com/libp2p/go-libp2p/p2p/net/swarm/testing"
+	"github.com/stretchr/testify/require"
+)
+
+func TestReachabilityChangeEvent(t *testing.T) {
+	h := bhost.NewBlankHost(swarmt.GenSwarm(t))
+	rmgr := NewRelayManager(h)
+	rmgr.reachabilityChanged(network.ReachabilityPublic)
```
marten-seemann, line 15: I have a slight preference for not calling unexported functions in tests (where possible). In this case, you could get the event bus (via `h.EventBus()`), and then emit reachability events.

p2p/host/relaysvc/relay.go
```diff
@@ -65,14 +65,18 @@ func (m *RelayManager) background(ctx context.Context) {
 func (m *RelayManager) reachabilityChanged(r network.Reachability) error {
 	switch r {
 	case network.ReachabilityPublic:
+		m.mutex.Lock()
+		defer m.mutex.Unlock()
+		// ignore events with no reachability change
```
marten-seemann, line 70:
```suggestion
		// This could happen if two consecutive EvtLocalReachabilityChanged report the same reachability.
		// This shouldn't happen, but it's safer to double-check.
```

p2p/host/relaysvc/relay_test.go
sukunrt: that's neat!

sukunrt: this is fixed now

p2p/host/relaysvc/relay_test.go
```diff
@@ -0,0 +1,67 @@
...
+	emitter, err := rmgr.host.EventBus().Emitter(new(event.EvtLocalReachabilityChanged), eventbus.Stateful)
+	if err != nil {
+		t.Fatal(err)
+	}
+	evt := event.EvtLocalReachabilityChanged{Reachability: network.ReachabilityPublic}
+	emitter.Emit(evt)
+	require.Eventually(
+		t,
+		func() bool { rmgr.mutex.Lock(); defer rmgr.mutex.Unlock(); return rmgr.relay != nil },
+		200*time.Millisecond,
```
marten-seemann, line 28: Better to increase this to a second, CI is super slow (sometimes), which then leads to flaky tests.

marten-seemann (approved):

p2p/host/relaysvc/relay_test.go
sukunrt: fixed.

PR #2094, opened 2023-02-14
feat: make ConcreteLimitConfig's field public

Jorropo:
This allows consumers to set and verify their own hard limits, such as how we do in Kubo.

Also could this be released to v0.25.2 if merged please ? This is blocking https://github.com/ipfs/kubo/pull/9647


MarcoPolo:
Let's hold off on this until I try out how the existing API works in https://github.com/ipfs/kubo/pull/9647. I made these fields not public on purpose so that these struct wouldn't be used directly. This struct is hard to use properly, 0 means block all, and you can accidentally block some resource you actually need.

Instead users should use PartialLimitConfig always and build from that.

Jorropo:
We are not gonna go forward with that, we are gonna use `.ToPartialConfig` and add a `.MustConcrete` in v0.26.0 (with a work around using `.Build(rcmgr.ConcreteLimitConfig{})` as a temporary workaround for `.MustConrete` in v0.25).
See https://filecoinproject.slack.com/archives/C03K82MU486/p1676402407266349?thread_ts=1676364991.780789&cid=C03K82MU486

PR #2105, opened 2023-02-16
identify: deduplicate Identify snapshots

marten-seemann:
As a result of (the now fixed) https://github.com/libp2p/go-libp2p/issues/2046, we were sending a lot of Identify Push messages into the network that were exactly identical to the previous message.

Since sending pushes is relatively expensive (we contact every peer we're connected to) it pays off to be a bit more careful. This PR now adds an explicit check that the new snapshot actually changes protocols, addresses or record.

### DIFF
--- a/p2p/protocol/identify/id_test.go
+++ b/p2p/protocol/identify/id_test.go
@@ -894,6 +894,47 @@ func TestIncomingIDStreamsTimeout(t *testing.T) {
 	}
 }
 
+func TestDeduplicateIdentifySnapshots(t *testing.T) {
+	h1 := blhost.NewBlankHost(swarmt.GenSwarm(t))
+	h2 := blhost.NewBlankHost(swarmt.GenSwarm(t))
+	defer h1.Close()
+	defer h2.Close()
+
+	ids1, err := identify.NewIDService(h1)
+	require.NoError(t, err)
+	defer ids1.Close()
+	ids1.Start()
+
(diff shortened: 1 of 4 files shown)


p2p/protocol/identify/id_test.go
```diff
@@ -894,6 +894,47 @@ func TestIncomingIDStreamsTimeout(t *testing.T) {
...
+	emitter, err := h1.EventBus().Emitter(new(event.EvtLocalProtocolsUpdated))
+	require.NoError(t, err)
+	// Emit the event. Nothing changed about the snapshot, so we don't expect any Identify Push.
+	require.NoError(t, emitter.Emit(event.EvtLocalProtocolsUpdated{}))
+	sub2, err := h2.EventBus().Subscribe(new(event.EvtPeerProtocolsUpdated))
+	require.NoError(t, err)
+	select {
+	case ev := <-sub2.Out():
+		t.Fatalf("received duplicate event: %+v", ev)
+	case <-time.After(250 * time.Millisecond):
```
MarcoPolo, line 934: Do we have another test that checks that a change does indeed propagate?

MarcoPolo (approved):

p-shahi:
Can this be merged?

marten-seemann:
We now have https://github.com/libp2p/go-libp2p/pull/2299, we don't need this any more.

PR #2112, opened 2023-02-17
quic: don't send detailed error messages when closing connections

### LINKED ISSUE #1928: quic: Error from peer should be explicit that it is coming from remote peer rather than a local error
We lost some time debugging this. Basically Kubo folks saw this log line

```
{12D... Application error 0x0: conn-27571160: system: cannot reserve inbound connection: resource limit exceeded}
```

Even with the resource manager disabled. And naturally assumed we were still blocked despite using a null resource manager. This error actually came from the peer when it closed the connection [here](https://github.com/libp2p/go-libp2p/blob/9ade7609f770c31ec6d78d04b8b5efbe0e7ff5a2/p2p/transport/quic/listener.go#L71)

We should be more explicit locally that this is a remote error from the peer, and not an error we generated locally.

On the flip side, it's cool that the local peer can see the remote peer is blocking connections because of rcmgr.

marten-seemann:
Fixes #1928.

I know this PR will make some people unhappy, but we really shouldn't send the peer a detailed error message here, as this might allow an attacker to gain a lot of insight into how a node under attack is coping with the attack. This error message has proved useful in debugging recent incidents on the IPFS network, but ultimately it's the wrong solution to the problem.
For example:
* it only works on QUIC
* and only for errors that occur during connection setup

Instead, we should have error codes: https://github.com/libp2p/specs/issues/479

### DIFF
--- a/p2p/transport/quic/listener.go
+++ b/p2p/transport/quic/listener.go
@@ -96,7 +96,7 @@ func (l *listener) setupConn(qconn quic.Connection) (*conn, error) {
 	c, err := l.setupConnWithScope(qconn, connScope, remoteMultiaddr)
 	if err != nil {
 		connScope.Done()
-		qconn.CloseWithError(1, err.Error())
+		qconn.CloseWithError(1, "")
 		return nil, err
 	}
 


MarcoPolo (approved):
The "from remote" change is in quic-go, correct? Is that still planned?

marten-seemann:
> The "from remote" change is in quic-go, correct? Is that still planned?

I don't know if that will actually be very helpful, once the long and confusing error message is gone.

Issue #2113, opened 2023-02-17
swarm metrics: connection counting seems broken


marten-seemann:
<img width="1648" alt="image" src="https://user-images.githubusercontent.com/1478487/219556309-6ce4ca22-7b02-45e6-a4f8-352951447cd1.png">

While at the same time:
```
❯ ipfs swarm peers | wc
   1487    1487  135713
```

config:
```
  "Swarm": {
    "AddrFilters": null,
    "ConnMgr": {
      "GracePeriod": "20s",
      "HighWater": 1500,
      "LowWater": 900,
      "Type": "basic"
    },
```

sukunrt:
Do you have a long running node? 
On my setup the primary source of divergence is ```MetricsTracer.ClosedConnection``` not being called when two dials succeed and one of them is dropped [here](https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/limiter.go#L222-L225)

We can just call ```MetricsTracer.ClosedConnection``` there or we can move the OpenedConnection call to ```swarm.addConn```

PR #2115, opened 2023-02-18
swarm: fix swarm closed connection count

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2113

### DIFF
--- a/p2p/net/swarm/limiter.go
+++ b/p2p/net/swarm/limiter.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/transport"
 
@@ -20,11 +21,12 @@ type dialResult struct {
 }
 
 type dialJob struct {
-	addr    ma.Multiaddr
-	peer    peer.ID
-	ctx     context.Context
-	resp    chan dialResult
-	timeout time.Duration
+	addr          ma.Multiaddr
+	peer          peer.ID
+	ctx           context.Context
+	resp          chan dialResult
+	timeout       time.Duration
+	metricsTracer MetricsTracer
 }
 
 func (dj *dialJob) cancelled() bool {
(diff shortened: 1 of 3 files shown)


marten-seemann:
Nice debugging! Why did you decide to not move the `OpenedConnection` call to `addConn`?

sukunrt:
I chose this because if we moved OpenedConnection to swarm.addConn we'd not track connections dropped [here](https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/limiter.go#L222-L225)
I thought this tracking of concurrent successful dials and some of them immediately dropped might be useful if we change something in the future. 
A fix I thought of was to drain ```w.resch``` [here](https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/dial_worker.go#L88-L90) but that didn't look very clean.

MarcoPolo:
Nice work! Let me also say that I'm sorry the dial logic is a bit messy.

While I think this PR is reasonable, I'm afraid it still misses some close cases [e.g.](https://github.com/libp2p/go-libp2p/blob/8529a340a1f1c70a0f60a0685576ca9a066c22b9/p2p/net/swarm/swarm_dial.go#L510). I think the strategy should be to make sure we make fewer calls not more, since more is an indicator that this is brittle and requires us to remember to make the right calls when things error or close. My suggestion then is to make sure we record this metric in the `.Close()` call itself. This is good for two reasons: 1. Less duplication – we can just reuse the existing Close calls to record the metric, and 2. If we notice this number being wrong we can be pretty sure that there's a connection leak issue rather than us forgetting to call this metric.

Since we're on a time crunch I'm going to push a PR with these changes. But thanks for this PR and debugging this, it helped me at least :)

sukunrt:
> I'm afraid it still misses some close cases [e.g.](https://github.com/libp2p/go-libp2p/blob/8529a340a1f1c70a0f60a0685576ca9a066c22b9/p2p/net/swarm/swarm_dial.go#L510).

oops 

I agree. My PR has ClosedConnection calls all over the place. your change looks much cleaner. Thanks for fixing this :)

Issue #2047, opened 2023-02-03
Metrics should be able to accept a prometheus registerer parameter


MarcoPolo:
Instead of always using the default registerer.

sukunrt:
Should I add An option to create a metrics tracer with a given registrar like done in eventbus metrics? 
https://github.com/libp2p/go-libp2p/blob/master/p2p/host/eventbus/basic_metrics.go#L96-L100

```
func MustRegisterWith(reg prometheus.Registerer) MetricsTracerOption {
	return func(s *metricsTracerSetting) {
		s.reg = reg
	}
}
```

marten-seemann:
I'd be in favor of renaming the option. `MustRegisterWith` sounds like it's a global function (which it isn't really, it's an option), and it also sounds like it could panic. Better: `WithRegisterer`.

If I understand correctly, when allowing multiple `prometheus.Registerer`s, we need to be careful with the `sync.Once`. We need to register once _with every `Registerer`_.

We probably need to introduce some kind of metrics helper: A (thread-safe) set to keep track of which `Registerer`s we've already registered with.

sukunrt:
Agreed on ```WithRegisterer```. 
Why do we need to register just once? The prometheus Registerer interface has a Register method which would give an error ```AlreadyRegisteredError``` on reregistration. We can ignore this error and panic/propagate other errors when doing registerer.Register. Is this good enough?

marten-seemann:
> Why do we need to register just once? The prometheus Registerer interface has a Register method which would give an error `AlreadyRegisteredError` on reregistration. We can ignore this error and panic/propagate other errors when doing registerer.Register. Is this good enough?

Looks like I didn't look closely enough at the Prometheus API. This is even better! Catching `AlreadyRegisteredError` and panicking for all others sounds reasonable to me.

PR #2116, opened 2023-02-19
metrics: add options to disable metrics and to set Prometheus registerer

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2047

### DIFF
--- a/p2p/metricshelper/registerer.go
+++ b/p2p/metricshelper/registerer.go
@@ -1,6 +1,8 @@
 package metricshelper
 
 import (
+	"errors"
+
 	"github.com/prometheus/client_golang/prometheus"
 )
 
@@ -10,7 +12,7 @@ func RegisterCollectors(reg prometheus.Registerer, collectors ...prometheus.Coll
 	for _, c := range collectors {
 		err := reg.Register(c)
 		if err != nil {
-			if _, ok := err.(prometheus.AlreadyRegisteredError); !ok {
+			if ok := errors.As(err, &prometheus.AlreadyRegisteredError{}); !ok {
 				panic(err)
 			}
 		}
--- a/p2p/metricshelper/registerer.go
+++ b/p2p/metricshelper/registerer.go
@@ -1,0 +1,18 @@
+package metricshelper
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// RegisterCollectors registers the collectors with reg ignoring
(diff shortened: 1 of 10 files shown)


sukunrt:
I made an implementation with generics too. But that seems too complex for 50 lines of code reuse. 
It's in my branch: ```sukunrt:metrics-registerer-generic```

p2p/metricshelper/registerer.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// RegisterCollectors registers the collectors with reg ignoring
+// reregistration error and panics on any other error
+func RegisterCollectors(reg prometheus.Registerer, collectors ...prometheus.Collector) {
+	for _, c := range collectors {
+		err := reg.Register(c)
+		if err != nil {
+			if _, ok := err.(prometheus.AlreadyRegisteredError); !ok {
```
marten-seemann, line 13: Probably slightly safer to do use `errors.Is`

marten-seemann (commented):
Agreed to not make this generic.

There now needs to be some way to set the `Registerer` via a libp2p option.

p2p/metricshelper/registerer.go
sukunrt: the returned error is an instance of the type 
```
type AlreadyRegisteredError struct {
	ExistingCollector, NewCollector Collector
}
```
Checking with errors.Is will not work since we do not know the existing collector, so I changed this to use errors.As

There is a tricky case here:
If an error is returned where we are registering a collector where a different collector with the same fully qualified name was registered before, this method will suppress it and the collector used as argument to the method will not be collected for metrics. In this example:
```
	reg := prometheus.NewRegistry()
	c1 := prometheus.NewCounter(
		prometheus.CounterOpts{
			Name: "counter",
		},
	)
	c2 := prometheus.NewCounter(
		prometheus.CounterOpts{
			Name: "counter",
		},
	)
	RegisterCollectors(reg, c1)
    RegisterCollectors(reg, c2) // c2 is not registered
    c1.Inc()
    c2.Inc() // This will have no effect
```
I'm leaving it as it is because this situation doesn't arise anywhere in the code yet since we are using namespaces for our metrics

MarcoPolo (approved):

marten-seemann (approved):

Issue #1993, opened 2023-01-12
host: use dependency injection (Fx) to construct services, and start and stop them


marten-seemann:
Now that we’ve started using Fx to construct our transports, muxers and security protocols, we should start using it to construct libp2p services like AutoNAT, Identify, the hole punching service etc.

## Constructing, Starting and Stopping Services

These services might depend on each other. For example, the hole puncher needs the identity service. Currently, these interdependencies are hardcoded (and the host keeps references to particular services), but we could just have Fx take care of resolving this dependency tree.

We should also clearly define what a libp2p service is:
```go
type Service interface {
     Start() error
     io.Closer
}
```
In general, service _should not_ start any Go routines in their constructor, but wait for the explicit start signal provided with the `Start` method.

The host would then just have to keep track of a slice of services, so it can cleanly shut them down. Starting and stopping the services should also be done by Fx, so services are started in the right order. Maybe we won’t even need the slice of services, assuming that Fx keeps track the dependency tree for shutdown.

## How to expose existing services to the Application

An application might want to be able to interact with particular services, once they’re constructed. We could add getter functions to the host, but this quickly gets messy, and doesn’t work with injected services at all.

Instead we could have users pass in a service struct that Fx would fill with concrete values, e.g.:
```go
libp2p.FillHost(struct { *identify.Service, *autonat.Service }) (host.Host, error)
```

MarcoPolo:
I was in this code recently and was reminded of this issue. I think it's still a good idea :+1:

PR #2118, opened 2023-02-20
use Fx to start and stop the host, swarm, autorelay and quicreuse

### LINKED ISSUE #2117: quicreuse: is not closed when closing the host
The reason is that it's (if not provided by the user) created by Fx, and since we're not using the Fx application lifecycle yet, we have no way of telling when a host is shut down.

marten-seemann:
Fixes #2117. Part of #1993. Part of #2514.


config/config.go
```diff
@@ -322,45 +308,98 @@ func (cfg *Config) NewNode() (host.Host, error) {
...
+				OnStop: func(context.Context) error { return sw.Close() },
+			})
+			return sw
+		}),
+		fx.Provide(cfg.newBasicHost),
+		fx.Provide(func(h *bhost.BasicHost, lifecycle fx.Lifecycle) host.Host {
+			lifecycle.Append(fx.StartHook(h.Start))
+			return h
+		}),
+		fx.Provide(func(h host.Host) peer.ID { return h.ID() }),
```
MarcoPolo, line 352: Shouldn't dependents of a peer.ID take a host instead? Then you know for sure the peer ID is yours. same with the privkey below.

config/config.go
```diff
@@ -322,45 +308,98 @@ func (cfg *Config) NewNode() (host.Host, error) {
...
-			h.Close()
-			return nil, fmt.Errorf("cannot enable autorelay; relay is not enabled")
-		}
+		fxopts = append(fxopts,
+			fx.Invoke(func(h *bhost.BasicHost, lifecycle fx.Lifecycle) (*autorelay.AutoRelay, error) {
+				ar, err := autorelay.NewAutoRelay(h, cfg.AutoRelayOpts...)
+				if err != nil {
+					return nil, err
+				}
+				lifecycle.Append(fx.StartStopHook(ar.Start, ar.Close))
```
MarcoPolo, line 378: autorelay depends on Identify to start first, how can we enforce that?

marten-seemann: I _think_ this is guaranteed because we construct the host first, then autorelay (since the autorelay constructor depends on the host). Fx executes startup hooks in the order they were added, so this determines that `host.Start` is called before `autorelay.Start`. Now I admit that this is pretty subtle.

The correct solution is to remove the `IDService()` from the host, and instead pass a reference to identify to the autorelay constructor. That would make the dependency tree more obvious. Let's do that in a follow-up PR?

config/config.go
MarcoPolo: How would users get the access to the peerstore or eventbus then?

p2p/host/autorelay/autorelay_test.go
```diff
@@ -444,9 +444,8 @@ func TestMinInterval(t *testing.T) {
 	)
 	defer h.Close()
 
-	cl.Add(500 * time.Millisecond)
```
MarcoPolo, line 447: why this change?

p2p/transport/quic/virtuallistener.go
```diff
@@ -70,7 +70,7 @@ func (r *acceptLoopRunner) RmAcceptForVersion(v quic.VersionNumber) {
 
 	ch, ok := r.muxer[v]
 	if !ok {
-		panic("expected chan in accept muxer")
```
MarcoPolo, line 73: Why this change? Did you hit this panic?

p2p/host/autorelay/autorelay_test.go
marten-seemann: Removing it.

p2p/transport/quic/virtuallistener.go
marten-seemann: I think I accidentally committed unrelated changes. Discarding this.

config/config.go
```diff
@@ -264,47 +256,38 @@ func (cfg *Config) addTransports(h host.Host) error {
...
-	eventBus := eventbus.NewBus(eventbus.WithMetricsTracer(eventbus.NewMetricsTracer(eventbus.WithRegisterer(cfg.PrometheusRegisterer))))
-	swrm, err := cfg.makeSwarm(eventBus, !cfg.DisableMetrics)
-	if err != nil {
-		return nil, err
-	}
-
-	if !cfg.DisableMetrics {
-		rcmgr.MustRegisterWith(cfg.PrometheusRegisterer)
-	}
-
```
sukunrt, line 307: Is this intentional? We need to register prometheus metrics with the libp2p configured prometheus registerer, no?

p2p/test/transport/rcmgr_test.go
```diff
@@ -9,13 +9,14 @@ import (
...
 
-	gomock "github.com/golang/mock/gomock"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	mocknetwork "github.com/libp2p/go-libp2p/core/network/mocks"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
 	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
+
+	"github.com/golang/mock/gomock"
```
sukunrt, line 19: I'm guessing diff will be removed with the move to uber/gomock?

sukunrt:
Thanks for removing AutoRelayHost 🎉

config/config.go
```diff
@@ -322,45 +308,98 @@ func (cfg *Config) NewNode() (host.Host, error) {
...
+	fxopts := []fx.Option{
+		fx.Provide(func() event.Bus {
+			return eventbus.NewBus(eventbus.WithMetricsTracer(eventbus.NewMetricsTracer(eventbus.WithRegisterer(cfg.PrometheusRegisterer))))
+		}),
+		fx.Provide(func(eventBus event.Bus, lifecycle fx.Lifecycle) (*swarm.Swarm, error) {
+			sw, err := cfg.makeSwarm(eventBus, !cfg.DisableMetrics)
+			if err != nil {
+				return nil, err
+			}
+			lifecycle.Append(fx.StopHook(sw.Close))
```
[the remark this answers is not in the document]
MarcoPolo: As mentioned elsewhere, we are but that's okay since the swarm close is guarded to only close once. The closing twice is useful because each close is logically for a different reason.
1. Here we are closing because we created the swarm.
2. Below, in decorate we are closing to ensure we close before the quicreuse conn manager.

By closing twice we prevent these different procedures from being tangled.

MarcoPolo (changes requested):
We need to add a couple more tests. At least:

1. A test that asserts the connmanager is closed after we close the host/swarm.
2. A test that asserts their are no dangling goroutines after starting and stopping a host.

Stebalien:
I've added a dangling goroutine test which _should_ be sufficient. It implicitly asserts that the connection manager is closed as the connection manager launches a background goroutine. Ideally we'd have some way to also assert that there are no scheduled timers, but I don't think we can.

When doing that, I noticed two issues:

1. #2743.
2. We don't shut anything down if we fail to start. But, IMO, that's not something to solve in this patch.

(I'm trying to push this forward because I don't want it to develop a bunch of merge conflicts)

MarcoPolo:
Thanks @Stebalien! I didn't know about uber/goleak. Looks helpful. I'll try to fix the current leaks.

Stebalien:
Hm. That goroutine should have been excluded...

Stebalien:
Oh, different nat goroutine.

Stebalien:
I'd add `github.com/jackpal/go-nat-pmp.(*Client).GetExternalAddress` to the exclusion list.

MarcoPolo:
> A test that asserts the connmanager is closed after we close the host/swarm.

This is effectively covered by checking for dangling goroutines.

MarcoPolo (approved):

MarcoPolo (approved):

PR #2120, opened 2023-02-21
autorelay: Refactor relay_finder and start autorelay after identify

MarcoPolo:
Fixes two flaky issues:

1. autrelay may start making connections before the identify goroutines are ready.
2. autorelay_tests are flaky because the ticker channels may be received in the wrong order.

Also uses a fork of the mock clock library to fix a data race in the library. Upstream patch coming soon.


p2p/host/autorelay/relay_finder.go
```diff
@@ -128,6 +142,15 @@ func (rf *relayFinder) background(ctx context.Context) {
 	oldCandidateTicker := rf.conf.clock.Ticker(rf.conf.maxCandidateAge / 5)
 	defer oldCandidateTicker.Stop()
 
+	scheduledWork := &scheduledWorkTimes{
+		nextRefresh:                 rf.conf.clock.Now().Add(rsvpRefreshInterval),
+		nextBackoff:                 rf.conf.clock.Now().Add(rf.conf.backoff / 5),
+		nextOldCandidateCheck:       rf.conf.clock.Now().Add(rf.conf.maxCandidateAge / 5),
+		nextAllowedCallToPeerSource: rf.conf.clock.Now().Add(-time.Second), // allow immediately
+		peerSourceTimer:             peerSourceTimer,
+		needMorePeers:               true,
```
marten-seemann, line 151: Having this bool feels a bit wonky, as it's used to communicate state between this function `runScheduledWork`. I think I'd prefer to
1. Make this a member variable.
1. Pass this struct by value to make it clear that no modification is happening.

p2p/host/autorelay/relay_finder.go
```diff
@@ -156,31 +179,18 @@ func (rf *relayFinder) background(ctx context.Context) {
...
-					log.Debugw("deleting candidate due to age", "id", id)
-					delete(rf.candidates, id)
-				}
-			}
-			rf.candidateMx.Unlock()
-			if deleted {
-				rf.notifyMaybeNeedNewCandidates()
-			}
+		case <-refreshTicker.C:
+			push = rf.runScheduledWork(ctx, rf.conf.clock.Now(), scheduledWork, canCallPeerSourceChan)
```
marten-seemann, line 183: You could move this call to after the `select` statement instead of repeating it 5 times.

p2p/host/autorelay/relay_finder.go
```diff
@@ -194,41 +204,91 @@ func (rf *relayFinder) background(ctx context.Context) {
 	}
 }
 
+func (rf *relayFinder) runScheduledWork(ctx context.Context, now time.Time, scheduledWork *scheduledWorkTimes, canCallPeerSourceChan chan<- struct{}) bool {
```
marten-seemann, line 207: Nit: Any reason to pass `now` to this function instead of just calling `Now()` first thing here?

p2p/host/autorelay/relay_finder.go
```diff
@@ -194,41 +204,91 @@ func (rf *relayFinder) background(ctx context.Context) {
...
+	var push bool
+
+	if now.After(scheduledWork.nextRefresh) {
+		scheduledWork.nextRefresh = now.Add(rsvpRefreshInterval)
+		push = rf.refreshReservations(ctx, now)
+	}
+
+	if now.After(scheduledWork.nextBackoff) {
+		scheduledWork.nextBackoff = now.Add(rf.conf.backoff / 5)
```
marten-seemann, line 216: Where is the 5 coming from?

p2p/host/autorelay/relay_finder.go
```diff
@@ -119,6 +131,8 @@ func (rf *relayFinder) background(ctx context.Context) {
 	}
 	defer subConnectedness.Close()
 
+	peerSourceTimer := rf.conf.clock.Timer(rf.conf.minInterval)
```
marten-seemann, line 134: Would it make sense to have a single timer that gets set to `min(nextRefresh, nextBackoff, ...)`? You could then deal with resetting this timer exclusively in this function.

marten-seemann (commented):
This looks a lot cleaner than what we had. Some suggestions to make it even cleaner.

go.mod
```diff
@@ -115,3 +115,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/blake3 v1.1.7 // indirect
 )
+
+replace github.com/benbjohnson/clock => github.com/marcopolo/clock v1.4.0
```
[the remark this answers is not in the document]
MarcoPolo: Agreed with the upstream fix

p2p/host/autorelay/autorelay_test.go
```diff
@@ -121,7 +119,7 @@ func TestSingleCandidate(t *testing.T) {
 	)
 	defer h.Close()
 
-	require.Eventually(t, func() bool { return numRelays(h) > 0 }, 3*time.Second, 100*time.Millisecond)
+	require.Eventually(t, func() bool { return numRelays(h) > 0 }, 10*time.Second, 100*time.Millisecond)
```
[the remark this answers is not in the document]
MarcoPolo: Should be, but as a rule of thumb I just made these all 10 seconds since if that fails in CI you know for sure it isn't a timing issue. I think I saw some cases where 3s was not enough

p2p/host/autorelay/relay_finder.go
MarcoPolo: pureness

p2p/host/autorelay/relay_finder.go
MarcoPolo: Original code, I think you wrote this :)

p2p/host/autorelay/relay_finder.go
MarcoPolo: We can try and see if that's simpler to understand.

p2p/host/autorelay/relay_finder.go
marten-seemann: https://miro.medium.com/v2/resize:fit:1400/format:webp/1*66rSQZJnG_K1d9jYHdxDRg.png

p2p/host/autorelay/relay_finder.go
MarcoPolo: I don't want to make it a member variable because I don't want any other method to access it. It would be tempting to have two goroutines try to modify this and then you introduce a lock for what should be a scoped var. I also don't want to pollute the struct with this field that only this scope uses.

Modification is happening. All the next checks are replaced. We can have the function return a scheduledWorkTimes, not sure if that's much better though. wdyt?

MarcoPolo: I'll refactor this differently. Coming soon :)

MarcoPolo: Refactored this away.

p2p/host/autorelay/relay_finder.go
MarcoPolo: I fixed this btw :)

p2p/host/autorelay/relay_finder.go
MarcoPolo: I think it's better. It has the nice side effect of avoiding the `/ 5` timers. We also have a single timer now. And a single channel to rate limit findNodes.

go.mod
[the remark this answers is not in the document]
MarcoPolo: Nice! with the refactor I don't hit the data race codepath! (I was stopping a timer before, now I'm just resetting). So we don't _need_ this.

p2p/host/autorelay/relay_finder.go
```diff
@@ -150,85 +177,156 @@ func (rf *relayFinder) background(ctx context.Context) {
...
+	if now.After(scheduledWork.nextBackoff) {
+		scheduledWork.nextBackoff = rf.clearBackoff(now)
+	}
+
+	if now.After(scheduledWork.nextOldCandidateCheck) {
+		scheduledWork.nextOldCandidateCheck = rf.clearOldCandidates(now)
+	}
+
+	if now.After(scheduledWork.nextAllowedCallToPeerSource) {
+		scheduledWork.nextAllowedCallToPeerSource = now.Add(rf.conf.minInterval)
```
marten-seemann, line 226: Why is this adding to `now`? Should this add to `lastCallToPeerSource`?

p2p/host/autorelay/relay_finder.go
```diff
@@ -173,102 +177,137 @@ func (rf *relayFinder) background(ctx context.Context) {
...
-			rf.cachedAddrs = nil
-			rf.relayMx.Unlock()
-			rf.host.SignalAddressChange()
-		}
 	}
 }
 
-func (rf *relayFinder) runScheduledWork(ctx context.Context, now time.Time, scheduledWork *scheduledWorkTimes, canCallPeerSourceChan chan<- struct{}) bool {
-	var push bool
+func (rf *relayFinder) clearCachedAddrsAndIdentifyPush() {
```
marten-seemann, line 200:
```suggestion
func (rf *relayFinder) clearCachedAddrsAndSignalAddressChange() {
```

marten-seemann (approved):

PR #2029, opened 2023-01-30
Update circuitv2 to proto3

ckousik:
Updates circuit v2 to use proto3. Currently awaiting issue to be resolved and spec to be updated https://github.com/libp2p/specs/issues/509

PR #2121, opened 2023-02-22
circuitv2: Update proto files to proto3

MarcoPolo:
closes #2029 

I just copied the pb definition from specs directly and ran go generate. 

The gist is that this goes from required to optional.


marten-seemann:
How is this different from https://github.com/libp2p/go-libp2p/pull/2029?

MarcoPolo:
Same idea, but:

1. I copied the comments from the specs so they show up in the field comments.
2. I created this before @ckousik updated the voucher proto.
3. I kept the name `ReservationVoucher` to not change any existing code.

Full diff here: https://gist.github.com/MarcoPolo/9fafd391036f7805e5815b62cbe5d407

marten-seemann (approved):

PR #2122, opened 2023-02-22
quic: fix stalled virtual listener

MarcoPolo:
Fixes a case that only really shows up in CI where a listener can be stalled. This lead to some test flakiness.

### DIFF
--- a/p2p/transport/quic/virtuallistener.go
+++ b/p2p/transport/quic/virtuallistener.go
@@ -146,9 +146,19 @@ func (r *acceptLoopRunner) innerAccept(l *listener, expectedVersion quic.Version
 
 func (r *acceptLoopRunner) Accept(l *listener, expectedVersion quic.VersionNumber, bufferedConnChan chan acceptVal) (tpt.CapableConn, error) {
 	for {
-		r.acceptSem <- struct{}{}
-		conn, err := r.innerAccept(l, expectedVersion, bufferedConnChan)
-		<-r.acceptSem
+		var conn tpt.CapableConn
+		var err error
+		select {
+		case r.acceptSem <- struct{}{}:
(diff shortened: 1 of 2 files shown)


marten-seemann:
How is this different from #2029?

p2p/transport/quicreuse/reuse.go
```diff
@@ -110,7 +110,8 @@ func (r *reuse) gc() {
 		select {
 		case <-r.closeChan:
 			return
-		case now := <-ticker.C:
+		case <-ticker.C:
+			now := time.Now()
```
marten-seemann, line 114: This seems unrelated?

marten-seemann (approved):

p2p/transport/quic/virtuallistener.go
```diff
@@ -146,9 +146,19 @@ func (r *acceptLoopRunner) innerAccept(l *listener, expectedVersion quic.Version
...
 	for {
-		r.acceptSem <- struct{}{}
-		conn, err := r.innerAccept(l, expectedVersion, bufferedConnChan)
-		<-r.acceptSem
+		var conn tpt.CapableConn
+		var err error
+		select {
+		case r.acceptSem <- struct{}{}:
+			conn, err = r.innerAccept(l, expectedVersion, bufferedConnChan)
+			<-r.acceptSem
```
marten-seemann, line 154: Should the `conn == nil && err == nil` be moved into this `case` for clarity?

marten-seemann (approved):

p2p/transport/quicreuse/reuse.go
MarcoPolo: I'll back this out

PR #2125, opened 2023-02-22
circuitv2: add a relay option to disable limits

### LINKED ISSUE #2079: add a config option to allow all protocols via circuit v2
We should do this if we want to drop v1 support

marten-seemann:
Fixes #2079.

### DIFF
--- a/p2p/protocol/circuitv2/relay/options.go
+++ b/p2p/protocol/circuitv2/relay/options.go
@@ -18,6 +18,14 @@ func WithLimit(limit *RelayLimit) Option {
 	}
 }
 
+// WithInfiniteLimits is a Relay option that disables limits.
+func WithInfiniteLimits() Option {
+	return func(r *Relay) error {
+		r.rc.Limit = nil
+		return nil
+	}
+}
+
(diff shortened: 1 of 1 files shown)


lidel (approved):

p2p/protocol/circuitv2/relay/options.go
```diff
@@ -18,6 +18,14 @@ func WithLimit(limit *RelayLimit) Option {
 	}
 }
 
+// WithInfiniteLimits is a Relay option that disables limits.
+func WithInfiniteLimits() Option {
```
p-shahi, line 22:
```suggestion
// WithInfiniteLimits is a Relay option that disables limits.
// Enabling this option doesn't apply any filter for which protocols can be used.
func WithInfiniteLimits() Option {
```

Would it be valuable to add this additional comment?

MarcoPolo (approved):
Much easier than expected!

p2p/protocol/circuitv2/relay/options.go
marten-seemann: It's not the relay that limits the protocols, this is entirely a decision that the two endpoints using the relay make.

PR #2126, opened 2023-02-22
identify: add metrics

sukunrt:
added metrics for:
address count
protocol count
peer push support


dashboards/identify/identify.json
```diff
@@ -0,0 +1,800 @@
...
+          "editorMode": "code",
+          "exemplar": false,
+          "expr": "libp2p_identify_addrs_count",
+          "legendFormat": "address count",
+          "range": true,
+          "refId": "A"
+        }
+      ],
+      "title": "Outgoing Address Count",
+      "type": "gauge"
```
sukunrt, line 334: This looks a bit misleading since the gauge is never full. Any suggestions here?

dashboards/identify/identify.json
```diff
@@ -0,0 +1,800 @@
...
+          "editorMode": "code",
+          "expr": "libp2p_identify_protocols_count",
+          "legendFormat": "protocols count",
+          "range": true,
+          "refId": "A"
+        }
+      ],
+      "title": "Outgoing Protocols Count",
+      "type": "gauge"
+    },
```
sukunrt, line 411: Same here the gauge may be a bit empty and so a bit misleading.

p2p/protocol/identify/id.go
```diff
@@ -581,6 +586,11 @@ func (ids *idService) createBaseIdentifyResponse(conn network.Conn, snapshot *id
 		}
 		mes.ListenAddrs = append(mes.ListenAddrs, addr.Bytes())
 	}
+	if ids.metricsTracer != nil {
+		ids.metricsTracer.NumProtocols(len(mes.Protocols))
+		ids.metricsTracer.NumAddrs(len(mes.ListenAddrs))
+	}
```
sukunrt, line 592: we can simplify this a bit if we are fine with counting snaphot.Addrs as Num Address Sent.

sukunrt:
@marten-seemann left comments at a few places where I'm confused. Suggestions would be very helpful.

p2p/protocol/identify/id.go
```diff
@@ -491,13 +491,18 @@ func (ids *idService) handleIdentifyResponse(s network.Stream, isPush bool) erro
...
 	}
+	prevPushSupport := e.PushSupport
 	sup, err := ids.Host.Peerstore().SupportsProtocols(c.RemotePeer(), IDPush)
 	if supportsIdentifyPush := err == nil && len(sup) > 0; supportsIdentifyPush {
 		e.PushSupport = identifyPushSupported
 	} else {
 		e.PushSupport = identifyPushUnsupported
 	}
 	ids.conns[c] = e
+	if ids.metricsTracer != nil && e.PushSupport != prevPushSupport {
```
marten-seemann, line 502: Why is this conditioned on the previous push support? Why can't we just count all connections identified by push supported (yes / no)?

p2p/protocol/identify/id.go
```diff
@@ -894,8 +913,12 @@ func (nn *netNotifiee) Disconnected(_ network.Network, c network.Conn) {
 
 	// Stop tracking the connection.
 	ids.connsMu.Lock()
+	e, ok := ids.conns[c]
 	delete(ids.conns, c)
 	ids.connsMu.Unlock()
+	if ok && ids.metricsTracer != nil {
+		ids.metricsTracer.DecrementPushSupport(e.PushSupport)
```
marten-seemann, line 920: This could be removed if we just count newly identified connection, as suggested above.

p2p/protocol/identify/id.go
marten-seemann: I'm wondering if we should rework the API a bit. Might be nice to have a `Identify{Push}Received(numProtos, numAddrs int)`. wdyt?

p2p/protocol/identify/id.go
sukunrt: I was keeping it a gauge counting support on current connections. This change will make this metric only work for new connections.

p2p/protocol/identify/id.go
sukunrt: agreed. will fix this.

sukunrt:
addressed review comments. This looks much cleaner now.

dashboards/identify/identify.json
marten-seemann: What about using a Stat? That way you'd also if a change brought number up or down: 
image

sukunrt: fixed this.

marten-seemann (approved):
lgtm

PR #2129, opened 2023-02-22
examples: Add sleep to fix flaky test

MarcoPolo:
If we end the test right away we may miss logging the "Established Connection" line that this test checks for.

### DIFF
--- a/examples/chat/chat_test.go
+++ b/examples/chat/chat_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"testing"
+	"time"
 
 	"github.com/libp2p/go-libp2p/core/network"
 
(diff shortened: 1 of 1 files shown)


examples/chat/chat_test.go
```diff
@@ -43,6 +44,8 @@ func TestMain(t *testing.T) {
 
 		go startPeer(ctx, h1, func(network.Stream) {
 			log.Println("Got a new stream!")
+			// Sleep a bit to let h2 print the logs we're waiting for
+			time.Sleep(500 * time.Millisecond)
```
marten-seemann, line 48: This would be nicer with a proper synchronization mechanism, but I'm fine with merging this for now.

marten-seemann (approved):

PR #2130, opened 2023-02-22
swarm: remove parallel tests from swarm tests

MarcoPolo:
Doing these tests in parallel causes flakiness in CI.

### DIFF
--- a/p2p/net/swarm/dial_test.go
+++ b/p2p/net/swarm/dial_test.go
@@ -29,8 +29,6 @@ func closeSwarms(swarms []*swarm.Swarm) {
 }
 
 func TestBasicDialPeer(t *testing.T) {
-	t.Parallel()
-
 	swarms := makeSwarms(t, 2)
 	defer closeSwarms(swarms)
 	s1 := swarms[0]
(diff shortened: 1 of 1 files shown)


p2p/net/swarm/dial_test.go
```diff
@@ -552,7 +538,7 @@ func newSilentListener(t *testing.T) ([]ma.Multiaddr, net.Listener) {
 }
 
 func TestDialSimultaneousJoin(t *testing.T) {
-	const dialTimeout = 250 * time.Millisecond
+	const dialTimeout = 3 * time.Second
```
marten-seemann, line 541: Is this necessary? This one test now takes 3s.

MarcoPolo: Yes. I'm pretty sure I saw this test flake here because of the dialTimeout. 1s might be okay as well but 3s definitely is.

marten-seemann: Fine for now, but long term, we'll need to find a faster solution for this.

marten-seemann (approved):

p2p/net/swarm/dial_test.go
MarcoPolo: There's much much lower hanging fruit than this. 

1. We can use a build cache
2. We can split the 32bit tests/race tests into a separate job so they can run in parallel. With a build cache, these won't waste too much time in duplicated efforts.

PR #2136, opened 2023-02-23
basic_host: Fix flaky tests

MarcoPolo:
Fixes flaky `TestHostProtoPreknowledge` test.

### DIFF
--- a/p2p/host/basic/basic_host_test.go
+++ b/p2p/host/basic/basic_host_test.go
@@ -250,7 +250,7 @@ func getHostPair(t *testing.T) (host.Host, host.Host) {
 	require.NoError(t, err)
 	h2.Start()
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
 	defer cancel()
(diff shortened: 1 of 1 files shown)


p2p/host/basic/basic_host_test.go
```diff
@@ -359,15 +359,23 @@ func TestHostProtoPreknowledge(t *testing.T) {
...
 
 	h2.SetStreamHandler("/super", handler)
-	// Prevent pushing identify information so this test actually _uses_ the super protocol.
-	h1.RemoveStreamHandler(identify.IDPush)
 
 	h1.Start()
 	h2.Start()
 
+	// Prevent pushing identify information so this test actually _uses_ the super protocol.
+	h1.RemoveStreamHandler(identify.IDPush)
```
marten-seemann, line 367: Why after the `Start`? This will cause an Identify Push to be sent.

MarcoPolo: Start in identify is what sets the stream handler. Removing the stream handler before start is no-op since it's not even registered.

Identify won't do anything until these nodes are connected.

marten-seemann (approved):

PR #2137, opened 2023-02-23
examples: chat-with-rendezvous: Bootstrap after connecting to bootstrappers

MarcoPolo:
I think this is the right thing to do? The other way would work eventually because of the background timer.

### DIFF
--- a/examples/chat-with-rendezvous/chat.go
+++ b/examples/chat-with-rendezvous/chat.go
@@ -119,13 +119,6 @@ func main() {
 		panic(err)
 	}
 
-	// Bootstrap the DHT. In the default configuration, this spawns a Background
-	// thread that will refresh the peer table every five minutes.
-	logger.Debug("Bootstrapping the DHT")
-	if err = kademliaDHT.Bootstrap(ctx); err != nil {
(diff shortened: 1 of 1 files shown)


examples/chat-with-rendezvous/chat.go
```diff
@@ -119,13 +119,6 @@ func main() {
...
 
-	// Bootstrap the DHT. In the default configuration, this spawns a Background
-	// thread that will refresh the peer table every five minutes.
-	logger.Debug("Bootstrapping the DHT")
-	if err = kademliaDHT.Bootstrap(ctx); err != nil {
-		panic(err)
-	}
-
 	// Let's connect to the bootstrap nodes first. They will tell us about the
 	// other nodes in the network.
```
Jorropo, line 123: This shouldn't be needed, we do it in Kubo too but I think this is wrong.

We could pass [`BootstrapPeers`](https://pkg.go.dev/github.com/libp2p/go-libp2p-kad-dht#BootstrapPeers) option so the DHT handles this itself.

Jorropo (commented):
I would remove the connection thing completely.

PR #2141, opened 2023-02-24
swarm: no parallel dials in CI

MarcoPolo:
Leaving this in draft until we decide we need this.

The pros:
1. Dialing in CI becomes more deterministic. I'm not sure it's fully deterministic since addresses aren't strictly sorted before being dialed. 
2. We don't worry about race conditions around two connections showing up at the same time. This solves a whole class of racey flakes.
3. We do less work overall which makes the runtime faster and avoids the class of bugs around slow runtimes.

The cons:
1. It's a slightly different dial path than what users use. It would be better to make the stock dial path work here.
  a. The counter argument is that unless users are constantly dialing localhost, the network latency will mean that in practice it's uncommon to get multiple connections succeeding in parallel.
2. Some tests relay on parallel dials. While I think I've identified them all, it still requires an extra line of code to make them work.

### DIFF
--- a/p2p/host/autonat/autonat_test.go
+++ b/p2p/host/autonat/autonat_test.go
@@ -193,7 +193,7 @@ func TestAutoNATIncomingEvents(t *testing.T) {
 
 	require.Eventually(t, func() bool {
 		return an.Status() != network.ReachabilityUnknown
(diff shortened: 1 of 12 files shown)


marten-seemann:
Closing this PR since with smart dialing, we now don't dial in parallel any more (except in some corner cases).

PR #2143, opened 2023-02-24
relay: initialize the ASN util when starting the service

marten-seemann:


MarcoPolo (approved):
Looks good, thanks for the proper fix 🙏

PR #2144, opened 2023-02-25
core: remove LocalPrivateKey method from network.Conn interface

### LINKED ISSUE #1695: core: consider removing the LocalPrivateKey method from secure connections
It's really not necessary and generally not a good idea to expose private keys like this.

marten-seemann:
Fixes #1695.


MarcoPolo (approved):

marten-seemann:
Rebased and added a changelog entry. I also confirmed that this doesn't lead to any problems when bubbling up into Kubo, apparently none of Kubo's dependencies have been using this method.

Issue #2102, opened 2023-02-15
autonat: don't use AutoNAT to discover addresses


marten-seemann:
> Is there any reason to keep the `address` field on the `autoNATresult`?

It's used in the PublicAddr method which is used by host.AllAddrs. 
This seems to me that it is mostly redundant information with information received from the ObservedAddressManager.

_Originally posted by @sukunrt in https://github.com/libp2p/go-libp2p/issues/2092#issuecomment-1430651078_

marten-seemann:
According to the address validation logic (see [spec](https://github.com/libp2p/specs/tree/master/autonat#autonat-protocol)), an AutoNAT implementation will only accept dial-back requests originating from IP addresses that don't match the remote IP address of the connection that the request was received on.

Since the node will only dial us back on addresses we send in the dial-back request, this means that assuming a spec-compliant peer, it should be impossible to discover new addresses via AutoNAT. At best, the logic we have is worthless.

This mechanism however could be used by a malicious peer to feed us incorrect addresses. This is not a very interesting attack, since 1. the node will only advertise a single such address, and only until we've obtained a new AutoNAT observation (15 minutes refresh interval) and 2. it's not replacing any addresses, we're just adding this address to our list of addresses.
The worst scenario I can think of is that this leads to increased handshake load on that incorrect address.

sukunrt:
STUN RFC suggests an attack [here](https://www.rfc-editor.org/rfc/rfc3489#section-12.1.1)
> In this case, the attacker provides a large number of clients with
   the same faked MAPPED-ADDRESS that points to the intended target.
   This will trick all the STUN clients into thinking that their
   addresses are equal to that of the target.  The clients then hand out
   that address in order to receive traffic on it (for example, in SIP
   or H.323 messages).  However, all of that traffic becomes focused at
   the intended target.  The attack can provide substantial
   amplification, especially when used with clients that are using STUN
   to enable multimedia applications.
   
Could this mechanism of using autonat for address discovery have been used to launch such attacks. The attacker runs a malicious autonat server and to all dialback requests it responds with the ip of the target. These dialback clients will then possibly advertise this address as their address for 15 minutes and this would lead to increased traffic on the target. 

Metrics on a kubo node show that a particular node doesn't get many dial requests so this would have been mostly fine. However we should keep this scenario in mind for future changes. 

@marten-seemann thoughts?

marten-seemann:
[...]

PR #2148, opened 2023-02-27
autonat: don't use autonat for address discovery

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2102

### DIFF
--- a/p2p/host/autonat/autonat.go
+++ b/p2p/host/autonat/autonat.go
@@ -2,7 +2,6 @@ package autonat
 
 import (
 	"context"
-	"errors"
 	"math/rand"
 	"sync/atomic"
 	"time"
@@ -20,6 +19,8 @@ import (
 
 var log = logging.Logger("autonat")
 
+const maxConfidence = 3
+
 // AmbientAutoNAT is the implementation of ambient NAT autodiscovery
 type AmbientAutoNAT struct {
 	host host.Host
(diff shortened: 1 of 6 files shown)


p2p/host/autonat/autonat_test.go
```diff
@@ -209,46 +208,43 @@ func TestAutoNATObservationRecording(t *testing.T) {
...
+	// Don't emit events if reachability hasn't changed
+	an.recordObservation(network.ReachabilityPublic)
 	if an.Status() != network.ReachabilityPublic {
 		t.Fatalf("reachability should stay public")
 	}
 	select {
 	case <-s.Out():
 		t.Fatal("received event without state transition")
-	case <-time.After(300 * time.Millisecond):
+	case <-time.After(1 * time.Second):
```
marten-seemann, line 247: Do we really need to increase the timeout here?

p2p/host/autonat/autonat.go
```diff
@@ -199,16 +181,15 @@ func (as *AmbientAutoNAT) background() {
 		case e := <-subChan:
 			switch e := e.(type) {
 			case event.EvtLocalAddressesUpdated:
-				if !lastAddrUpdated.Add(time.Second).After(time.Now()) {
-					lastAddrUpdated = time.Now()
-					if as.confidence > 1 {
-						as.confidence--
-					}
+				// On local address update, reduce confidence from maximum so that we schedule
```
marten-seemann, line 184: What is a local address update?

p2p/host/autonat/autonat_test.go
sukunrt: I'll change this back to 300ms. I haven't seen this failing so far.

p2p/host/autonat/autonat.go
sukunrt: this is any change in the host address list(host.AllAddrs()). I used the phrase from the event EvtLocalAddressesUpdated.

I didn't change this much since it was not related to this PR, but the correct thing to do here would be to only reduce confidence if there are any addresses removed from the host address list and our status is public or if there are addresses added and our status is private.

marten-seemann: Yeah, this is fine. Sorry, I messed this up, didn't realize this was literally in the name of the event ;)

PR #2154, opened 2023-03-01
relay service: add metrics

sukunrt:
Metrics Added:

ReservationRequest: Opened, Closed, Renewed
ReservationRequestResponseStatus
ReservationRejectionReason

ConnectionRequest: Opened, Closed
ConnectionRequestResponseStatus
ConnectionRejectionReason
ConnectionDuration

BytesTransferred

RelayStatus

### DIFF
--- a/dashboards/relaysvc/relaysvc.json
+++ b/dashboards/relaysvc/relaysvc.json
@@ -136,7 +136,7 @@
       },
       "id": 2,
       "options": {
-        "alignValue": "left",
+        "alignValue": "center",
         "legend": {
           "displayMode": "list",
           "placement": "bottom",
@@ -1138,14 +1138,12 @@
             "mode": "absolute",
             "steps": [
               {
-                "color": "green"
-              },
-              {
-                "color": "red",
-                "value": 80
+                "color": "green",
+                "value": null
               }
             ]
-          }
+          },
+          "unit": "s"
         },
         "overrides": []
       },
@@ -1175,22 +1173,11 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "histogram_quantile(0.5, sum(rate(libp2p_relaysvc_connection_duration_seconds_bucket[$__rate_interval])) by (le))",
-          "legendFormat": "50th percentile",
+          "expr": "rate(libp2p_relaysvc_connection_duration_seconds_sum[$__range])/rate(libp2p_relaysvc_connection_duration_seconds_count[$__range])\n",
+          "hide": false,
+          "legendFormat": "avg",
           "range": true,
           "refId": "A"
-        },
-        {
-          "datasource": {
-            "type": "prometheus",
-            "uid": "${DS_PROMETHEUS}"
-          },
-          "editorMode": "code",
-          "expr": "histogram_quantile(0.99, sum(rate(libp2p_relaysvc_connection_duration_seconds_bucket[$__rate_interval])) by (le))",
-          "hide": false,
-          "legendFormat": "99th percentile",
-          "range": true,
-          "refId": "C"
         }
       ],
       "title": "Connection Duration",
@@ -1212,6 +1199,6 @@
   "timezone": "",
   "title": "libp2p Relay Service",
   "uid": "C6RUfAx4z",
-  "version": 3,
+  "version": 4,
   "weekStart": ""
 }
--- a/dashboards/relaysvc/relaysvc.json
+++ b/dashboards/relaysvc/relaysvc.json
@@ -643,7 +643,7 @@
           },
           "editorMode": "code",
           "expr": "increase(libp2p_relaysvc_data_transferred_bytes_total[$__range])",
-          "legendFormat": "__auto",
+          "legendFormat": "data transferred",
           "range": true,
           "refId": "A"
         }
@@ -1138,8 +1138,7 @@
             "mode": "absolute",
             "steps": [
               {
-                "color": "green",
-                "value": null
+                "color": "green"
               },
               {
                 "color": "red",
@@ -1211,8 +1210,8 @@
   },
   "timepicker": {},
   "timezone": "",
-  "title": "Relay Service",
+  "title": "libp2p Relay Service",
   "uid": "C6RUfAx4z",
-  "version": 2,
+  "version": 3,
   "weekStart": ""
 }
--- a/dashboards/relaysvc/relaysvc.json
+++ b/dashboards/relaysvc/relaysvc.json
@@ -1175,7 +1175,7 @@
           "editorMode": "code",
           "expr": "rate(libp2p_relaysvc_connection_duration_seconds_sum[$__range])/rate(libp2p_relaysvc_connection_duration_seconds_count[$__range])\n",
           "hide": false,
-          "legendFormat": "avg",
+          "legendFormat": "rolling average",
           "range": true,
           "refId": "A"
         }
@@ -1199,6 +1199,6 @@
   "timezone": "",
   "title": "libp2p Relay Service",
   "uid": "C6RUfAx4z",
-  "version": 4,
+  "version": 5,
   "weekStart": ""
 }
--- a/dashboards/relaysvc/relaysvc.json
+++ b/dashboards/relaysvc/relaysvc.json
@@ -72,6 +72,19 @@
   "liveNow": false,
   "panels": [
     {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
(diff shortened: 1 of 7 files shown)


p2p/protocol/circuitv2/relay/metrics.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+const metricNamespace = "libp2p_relaysvc"
+
+var (
+	status = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: metricNamespace,
+			Name:      "status",
+			Help:      "Relay Current Status",
```
marten-seemann, line 17:
```suggestion
			Help:      "Relay Status",
```

p2p/protocol/circuitv2/relay/metrics.go
```diff
@@ -0,0 +1,262 @@
...
+		},
+	)
+
+	reservationTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metricNamespace,
+			Name:      "reservation_total",
```
marten-seemann, line 24: I think the naming convention says this should be plural.
```suggestion
			Name:      "reservations_total",
```

p2p/protocol/circuitv2/relay/metrics.go
```diff
@@ -0,0 +1,262 @@
...
+			Namespace: metricNamespace,
+			Name:      "reservation_request_status_total",
+			Help:      "Relay Reservation Request Status",
+		},
+		[]string{"status"},
+	)
+	reservationRejectedTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metricNamespace,
+			Name:      "reservation_rejected_total",
```
marten-seemann, line 40:
```suggestion
			Name:      "reservations_rejected_total",
```

p2p/protocol/circuitv2/relay/relay.go
```diff
@@ -556,6 +678,8 @@ func (r *Relay) background() {
 		case <-ticker.C:
 			r.gc()
 		case <-r.ctx.Done():
+			r.gc()
+			r.wg.Done()
```
marten-seemann, line 682: Was this a bug in our implementation? If so, would you mind opening a separate PR?

p2p/protocol/circuitv2/relay/relay.go
```diff
@@ -467,6 +520,47 @@ func (r *Relay) relayUnlimited(src, dest network.Stream, srcID, destID peer.ID,
...
+			if ew != nil {
+				err = ew
+				break
+			}
+			if nr != nw {
+				err = io.ErrShortWrite
+				break
+			}
+			if r.metricsTracer != nil {
+				r.metricsTracer.BytesTransferred(nr + nw)
```
marten-seemann, line 551: Aren't you counting every byte twice here?

p2p/protocol/circuitv2/relay/metrics.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// ConnectionRequestReceived tracks a new relay connect request
+	ConnectionRequestReceived()
+	// ConnectionOpened tracks metrics on opening a relay connection
+	ConnectionOpened()
+	// ConnectionClosed tracks metrics on closing a relay connection
+	ConnectionClosed(d time.Duration)
+	// ConnectionRequestHandled tracks metrics on handling a relay connection request
+	// rejectionReason is ignored for status other than `requestStatusRejected`
+	ConnectionRequestHandled(status string, rejectionReason string)
```
marten-seemann, line 134: Would it make sense to split this up in `ConnectionRequestHandled` (for success) and `ConnectionRequestRejected`? What would be the difference between `ConnectionRequestReceived` and `ConnectionRequestHandled`?

p2p/protocol/circuitv2/relay/relay.go
```diff
@@ -164,30 +172,45 @@ func (r *Relay) handleStream(s network.Stream) {
...
+		r.metricsTracer.ReservationRequestReceived()
+	}
+
 	p := s.Conn().RemotePeer()
 	a := s.Conn().RemoteMultiaddr()
 
 	if isRelayAddr(a) {
 		log.Debugf("refusing relay reservation for %s; reservation attempt over relay connection")
-		r.handleError(s, pbv2.Status_PERMISSION_DENIED)
+		r.handleErrorAndTrackMetrics(s, pbv2.HopMessage_RESERVE, pbv2.Status_PERMISSION_DENIED,
```
marten-seemann, line 184: I'm not a big fan of the `handleError` function. I know you didn't introduce it, so this is not your fault :)

What about return an error and / or a status code, and handle it in the caller?

sukunrt: lemme see how that looks.

p2p/protocol/circuitv2/relay/relay.go
sukunrt: okay true. bytes transferred should only be one of them. I did it because bandwidth used would be incoming + outgoing. But that should be handled in the dashboard. Will fix.

p2p/protocol/circuitv2/relay/relay.go
sukunrt: it's a race condition. yes. Will do. https://github.com/libp2p/go-libp2p/issues/2162

sukunrt: This is the fix for the bug. I added the gc call there since it simplified counting closed connections. 
https://github.com/libp2p/go-libp2p/pull/2164/files

p2p/protocol/circuitv2/relay/relay.go
sukunrt: this looks much better! thanks!
I've left handleError as it is, if you want to remove that I can open a separate PR.

p2p/protocol/circuitv2/relay/metrics.go
sukunrt: yeah received was the same. I've removed ```ConnectionRequestReceived``` and ```ReservationRequestReceived``` and kept ```ReservationRequestHandled``` and ```ConnectionRequestHandled```

p2p/protocol/circuitv2/relay/relay.go
marten-seemann: Yeah, let's do this in a follow-up PR.

dashboards/relaysvc/relaysvc.json
```diff
@@ -0,0 +1,1217 @@
...
+          },
+          "editorMode": "code",
+          "expr": "histogram_quantile(0.99, sum(rate(libp2p_relaysvc_connection_duration_seconds_bucket[$__rate_interval])) by (le))",
+          "hide": false,
+          "legendFormat": "99th percentile",
+          "range": true,
+          "refId": "C"
+        }
+      ],
+      "title": "Connection Duration",
```
marten-seemann, line 1196: This diagram seems to be missing the unit.

I'm also not sure how useful it is, the graph looks kind of wild on my instance:
image


Maybe it's easier if we just track the average:
```
rate(libp2p_relaysvc_connection_duration_seconds_sum[$__range])/rate(libp2p_relaysvc_connection_duration_seconds_count[$__range])
```
(Please double-check if `$__range` is the appropriate variable here)

image

marten-seemann (commented):
It's really cool to see _how_ cheap it is to run the relay service. This was recorded over a 3h time frame: 
image

dashboards/relaysvc/relaysvc.json
sukunrt: I'm keeping it range because I think it is most informative, while rate_interval will give you average in that time period the graph is very spiky and it is also an average over all connections in that period so the spike doesn't necessarily mean it's something is wrong. 
On the other hand increasing the dashboard range will change this graph which seems wrong. 
Screenshot 2023-03-07 at 12 32 24 PM

dashboards/relaysvc/relaysvc.json
```diff
@@ -0,0 +1,1217 @@
...
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "code",
+          "expr": "libp2p_relaysvc_status",
```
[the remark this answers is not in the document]
sukunrt: fixed this.

dashboards/relaysvc/relaysvc.json
sukunrt: changed the label to rolling average.

marten-seemann (approved):

Issue #2162, opened 2023-03-02
circuit v2: relay service not cleaned up properly on close


sukunrt:
There is a race condition between 
```

relay.Close()

```
and 
```

relay.handleReserve()

```

if relay.Close() is called around handleReserve execution at: 
https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/relay.go#L181
if relay.Close acquires the lock it'll untag the peers and cancel the background goroutine running the gc. 
Then handleReserve will add the reservation and tag the peer with conn manager.

PR #2164, opened 2023-03-02
circuitv2: cleanup relay service properly

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2162


marten-seemann (commented):
I think this fix is correct, but having both a mutex and an atomic makes me feel a bit uncomfortable.

Can we make `closed` a normal bool and move it under the mutex (in `Close`)? `Close` is not a code path that we need to micro-optimize.

sukunrt:
that would be better. I'll change this.

marten-seemann (approved):

PR #2165, opened 2023-03-02
connmgr: reduce log level for closing connections

richard-ramos:
This PR changes the log level for closing connections in connmgr.go from INFO to DEBUG, since as reported in https://github.com/status-im/status-go/issues/3249 these logs add too much noise.

### DIFF
--- a/p2p/net/connmgr/connmgr.go
+++ b/p2p/net/connmgr/connmgr.go
@@ -375,7 +375,7 @@ func (cm *BasicConnMgr) doTrim() {
 func (cm *BasicConnMgr) trim() {
 	// do the actual trim.
 	for _, c := range cm.getConnsToClose() {
-		log.Infow("closing conn", "peer", c.RemotePeer())
+		log.Debugw("closing conn", "peer", c.RemotePeer())
 		c.Close()
 	}
 }


jakubgs:
I'd argue that the `closing conn` one should probably be a `trace` level log.

p2p/net/connmgr/connmgr.go
```diff
@@ -173,7 +173,7 @@ func (cm *BasicConnMgr) memoryEmergency() {
 
 	// Trim connections without paying attention to the silence period.
 	for _, c := range cm.getConnsToCloseEmergency(target) {
-		log.Infow("low on memory. closing conn", "peer", c.RemotePeer())
+		log.Debugw("low on memory. closing conn", "peer", c.RemotePeer())
```
marten-seemann, line 176: This definitely shouldn't be hidden as debug. Running low on memory is a pretty serious condition.

p2p/net/connmgr/connmgr.go
```diff
@@ -375,7 +375,7 @@ func (cm *BasicConnMgr) doTrim() {
 func (cm *BasicConnMgr) trim() {
 	// do the actual trim.
 	for _, c := range cm.getConnsToClose() {
-		log.Infow("closing conn", "peer", c.RemotePeer())
+		log.Debugw("closing conn", "peer", c.RemotePeer())
```
marten-seemann, line 378: This one is fine.

marten-seemann (changes requested):

p2p/net/connmgr/connmgr.go
jakubgs: Are you sure it's debug? Look at this:
```
admin@node-01.sm-eu-zur1.spiff.app-test:~ % ls -lh /var/log/docker/spiff-workflow-wakunode/docker.log
-rw-r--r-- 1 syslog syslog 13G Mar  2 13:13 /var/log/docker/spiff-workflow-wakunode/docker.log

admin@node-01.sm-eu-zur1.spiff.app-test:~ % grep 'closing conn' /var/log/docker/spiff-workflow-wakunode/docker.log | wc -l
57729782
```
This is from 1.5 day of not really that active of a node. If you have a trace level it should be trace.

marten-seemann (approved):

p2p/net/connmgr/connmgr.go
marten-seemann: I don't think we have a trace log level. Or if we do, we don't use it.

jakubgs: Fair enough.

PR #2169, opened 2023-03-03
fix(host/routed): return connection error instead of routing

Wondertan:
A peer might not have any connections, and failure to establish the first one will not give a meaningful error, but the `failed to find peers: no peers in the table`


marten-seemann (approved):

Wondertan:
Why wasn't this included in any of the recent releases? 

cc @marten-seemann or @MarcoPolo

MarcoPolo:
We didn’t explicitly not include this. Generally we try to keep patch releases as small as possible. This will make it into the next minor release.

Issue #2167, opened 2023-03-03
autorelay: deadlock in tests


marten-seemann:
https://github.com/libp2p/go-libp2p/actions/runs/4321376917/jobs/7542536126

sukunrt:
Sorry. I introduced this here: 
https://github.com/libp2p/go-libp2p/pull/2164

The deadlock is swarm.disconnect has swarm.notifee lock and calls disconnected on relay service which depends on relay service lock
The relay service Close has relay service lock and calls swarm.stopNotify which depends on swarm.notifee lock.

PR #2171, opened 2023-03-04
relay: fix deadlock when closing

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2167

### DIFF
--- a/p2p/protocol/circuitv2/relay/relay.go
+++ b/p2p/protocol/circuitv2/relay/relay.go
@@ -51,6 +51,7 @@ type Relay struct {
 	constraints *constraints
 	scope       network.ResourceScopeSpan
 	notifiee    network.Notifiee
+	wg          sync.WaitGroup
 
 	mx     sync.Mutex
 	rsvp   map[peer.ID]time.Time
@@ -98,6 +99,8 @@ func New(h host.Host, opts ...Option) (*Relay, error) {
 	h.SetStreamHandler(proto.ProtoIDv2Hop, r.handleStream)
 	r.notifiee = &network.NotifyBundle{DisconnectedF: r.disconnected}
 	h.Network().Notify(r.notifiee)
+
+	r.wg.Add(1)
 	go r.background()
 
 	return r, nil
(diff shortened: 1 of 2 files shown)


p2p/protocol/circuitv2/relay/relay.go
```diff
@@ -105,17 +105,24 @@ func New(h host.Host, opts ...Option) (*Relay, error) {
...
 	if !r.closed {
 		r.closed = true
+		r.mx.Unlock()
+
 		r.host.RemoveStreamHandler(proto.ProtoIDv2Hop)
 		r.host.Network().StopNotify(r.notifiee)
 		r.scope.Done()
 		r.cancel()
+
+		r.mx.Lock()
```
marten-seemann, line 117: Would it make sense to move the untagging a few lines up? That way you'd only need to lock the mutex once.

sukunrt: That line actually has a race condition :( 
if both r.gc and r.Close(waiting on lock before untag peer) happen at the same time we'd untag the peer twice with the conn manager since in close we don't delete the entry from the map. 

I updated it to also cleanup the background goroutine like it happens in other services. If you want I'll make the exact change you asked.

p2p/protocol/circuitv2/relay/relay.go
```diff
@@ -98,24 +99,27 @@ func New(h host.Host, opts ...Option) (*Relay, error) {
...
 		r.closed = true
+		r.mx.Unlock()
+
 		r.host.RemoveStreamHandler(proto.ProtoIDv2Hop)
 		r.host.Network().StopNotify(r.notifiee)
 		r.scope.Done()
 		r.cancel()
-		for p := range r.rsvp {
-			r.host.ConnManager().UntagPeer(p, "relay-reservation")
-		}
```
marten-seemann, line 117: Where does the untagging happen now?

sukunrt: https://github.com/sukunrt/go-libp2p/blob/relay-test/p2p/protocol/circuitv2/relay/relay.go#L584-L590
here in ```r.gc```

marten-seemann (approved):

PR #2173, opened 2023-03-06
identify: Fix IdentifyWait when Connected events happen out of order

### LINKED ISSUE #2163: holepunch: race condition with identifyWait
Currently holepunch calls hs.ids.IdentifyWait(conn) inside the netNotifee.Connected call. 
But IdentifyWait requires netNotifee.Connected call to identify to be completed. 

https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/holepunch/holepuncher.go#L262

Swarm doesn't guarantee any ordering on notifcations. The notifees are obtained from map keys.

MarcoPolo:
fixes #2163 

Adds an `IsClosed` method to `network.Conn`

### DIFF
--- a/core/network/conn.go
+++ b/core/network/conn.go
@@ -33,6 +33,10 @@ type Conn interface {
 
 	// GetStreams returns all open streams over this conn.
 	GetStreams() []Stream
+
+	// IsClosed returns whether a connection is fully closed, so it can
+	// be garbage collected.
+	IsClosed() bool
 }
 
 // ConnectionState holds information about the connection.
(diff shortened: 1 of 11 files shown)


marten-seemann (commented):
Is it guaranteed that `Closed` will return `true` **before** any of the notifications will be sent?

marten-seemann (approved):

MarcoPolo:
This fixes a lot of the flakiness in the autorelay tests. I'm having a much harder time reproducing some earlier flakes. Thanks @sukunrt :)

PR #2176, opened 2023-03-08
circuitv2: remove handleError helper from relaysvc

sukunrt:
The problem with handleError was that it closed or reset the stream. This is not a problem in itself but there are cases where you don't want to call handle error but still reset the stream like https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/relay.go#L234 leading to stream.Reset calls in a bunch of places. 

This change makes all reset and close call on the stream happen in handleStream call itself.


sukunrt:
closing this. changes too much for little benefit.

PR #2179, opened 2023-03-09
identify: fix stale comment

sukunrt:
https://github.com/libp2p/go-libp2p/pull/2173 removed the requirement that swarm notifies identify before any new streams are started on a connection


MarcoPolo (approved):
While it is still the case we get the notification before accepting streams, it isn't crucial like this comment makes it sound. So it makes sense to remove this comment.

Ty!

PR #2190, opened 2023-03-16
relay_finder: don't busyloop on min interval 0

sukunrt:


p2p/host/autorelay/autorelay_test.go
```diff
@@ -224,6 +224,7 @@ func TestBackoff(t *testing.T) {
 	defer h.Close()
 
 	require.Eventually(t, func() bool {
+		cl.Add(1)
```
MarcoPolo, line 227: We shouldn't do this. It's at best a hack. Instead #2187 should fix this.

p2p/host/autorelay/autorelay_test.go
```diff
@@ -440,13 +443,14 @@ func TestMinInterval(t *testing.T) {
 		autorelay.WithMinCandidates(2),
 		autorelay.WithNumRelays(1),
 		autorelay.WithBootDelay(time.Hour),
-		autorelay.WithMinInterval(500*time.Millisecond),
+		autorelay.WithMinInterval(2*time.Second),
 	)
 	defer h.Close()
 
-	cl.Add(500 * time.Millisecond)
```
MarcoPolo, line 447: Why these changes?

p2p/host/autorelay/relay_finder.go
```diff
@@ -222,14 +234,6 @@ func (rf *relayFinder) runScheduledWork(ctx context.Context, now time.Time, sche
 		scheduledWork.nextOldCandidateCheck = rf.clearOldCandidates(now)
 	}
 
-	if now.After(scheduledWork.nextAllowedCallToPeerSource) {
-		scheduledWork.nextAllowedCallToPeerSource = scheduledWork.nextAllowedCallToPeerSource.Add(rf.conf.minInterval)
-		select {
-		case peerSourceRateLimiter <- struct{}{}:
```
MarcoPolo, line 228: Instead of these changes, we should simply set the `scheduledWork.nextAllowedCallToPeerSource` when we put a struct on this channel. I think that will have the same effect while being a smaller change and not introduce anything new.

MarcoPolo (changes requested):

p2p/host/autorelay/relay_finder.go
MarcoPolo: I guess it's not exactly the same, we still need to know when the peer source fn was called.

It might make sense for the findNodes method to signal via the new channel [here](https://github.com/libp2p/go-libp2p/pull/2187/files#diff-4fabce8a14510095c124d7159db6ae68844f99c52092ee12b05c1334323fa741R94) that we should run the background tasks.

sukunrt: yeah the problem is knowing when peerSource was called. I'll check the changes happening in marco/test-max-age-flaky

p2p/host/autorelay/autorelay_test.go
sukunrt: so previously the initial call to peer source was triggered without waiting for any time to progress on the clock by the scheduledWork function which runs right before we go into the for loop in background go routine. 

With this change with a mock clock we need to increment the clock once after the background go routine has initialised the timer for peerSourceCallback.

p2p/host/autorelay/autorelay_test.go
sukunrt: I'm fine with rebasing these changes on that branch. 

Another thing we can do would be to expose a channel which is closed when the background go routine has initialised. This cl.Add is required because we need to increment the timer once after creating a timer for it to trigger.

For these changes we can also fix this in the library itself. We should fire the timer when the timer is created.

PR #1999, opened 2023-01-17
add WebRTC transport

GlenDC:
Successor of original PR: https://github.com/libp2p/go-libp2p/pull/1655 (hand over of developer)

This PR implements the WebRTC transport spec according to <https://github.com/libp2p/specs/tree/cfcf0230b2f5f11ed6dd060f97305faa973abed2/webrtc>.

## Benchmarks

Benchmarks of how the WebRTC transport compares in terms of resource usage and continuous R/W throughput can be found at <https://github.com/libp2p/libp2p-go-webrtc-benchmarks#2-benchmarks> (submitted in function of this PR)

Comparison Quic Vs WebRTC transport:

image
image

## Interoperability

Compatibility of go-libp2p's WebRTC transport with the Rust and JS implementations are to be validated using the [libp2p-testplans](https://github.com/glendc/libp2p-test-plans):

- as the WebRTC implementation of go-libp2p lives in this PR's branch it is for now not automatically tested
- instead you can test it by running an interop ping test manually using the fork at <https://github.com/glendc/libp2p-test-plans>;

### WebRTC Interoperability Matrix

|                    | libp2p-go (listener) | libp2p-rs (listener) | libp2p-js (listener) |
|--------------------|----------------------|----------------------|----------------------|
| libp2p-go (dialer) | ✅                   | ✅                    | /                    |
| libp2p-rs (dialer) | ✅                    | ✅                    | /                    |
| libp2p-js (dialer) | ✅                    | ✅                    | /                    |

Note that `noise` is used for the encryption and `yamux` for the multiplexer, and this for all of the tests above.

### How to run the WebRTC interoperability tests 

Executables used for interop tests below:
- libp2p-go: <https://github.com/GlenDC/libp2p-test-plans/tree/draft/go-webrtc/multidim-interop/go/v0.24> (patched webrtc support into the ping test)
- libp2p-js: <https://github.com/GlenDC/libp2p-test-plans/tree/draft/go-webrtc/multidim-interop/js/v0.41> (patched webrtc support into the ping test)
  - To be able to debug  problems in `libp2p-js` while testing I added a PR <https://github.com/libp2p/test-plans/pull/119> to help with that
- libp2p-rs: <https://github.com/GlenDC/libp2p-test-plans/tree/draft/go-webrtc/multidim-interop/rust> (identical to libp2p/test-plans master)

First make sure to have a redis server running locally, as this is what used in the new interop tests, e.g. you can do this using the command:

```
[...]

PR #2191, opened 2023-03-16
[RFC] WASM

Jorropo:
I would like to run code in the browser however I hate JS and JS hates me.

It would be nice if `go-libp2p` would *magically* work in the browser (with gopherjs and or `GOOS=js GOARCH=wasm`).

IO would be provided in a degrated mode supporting websocket. ~~(https://pkg.go.dev/nhooyr.io/websocket#hdr-Wasm)~~
~~We could also do webtransport and webrtc but we would need to write the `syscall/js` wrapping code ourselves which might be more or less complex.~~

One potential issue which makes this less straight forward than it may seems, is that even if the big required elements like WS are reachable, the options that can be applied are way less diverse making some low level config impossible.

---

Current status of this PR:
- Todo:
  - [X] Websocket implementation brought back from the dead.
    - I don't like it, the code isn't good because the websocket API really sucks (no backpressure, callback based).
    It's probably worth removing it and relying exclusively on webtransport as long as https://developer.chrome.com/articles/websocketstream/ is not a thing. (we could also use a JS polyfill mapping websocket to the streaming API that webtransport is also using).
  - [X] New Webtransport `syscall/js` implementation
    - [ ] Tests
  - [ ] ~~WebRTC~~
    - I don't want to start writing the shim before #1999 is merged.
- Examples:
  - [X] Chat example !
  - [ ] Kad DHT
    - It compiles and runs on my machine but very quickly crashes chrome but the DHT doesn't work properly due to the lack of publicly announced webtransport addresses in the wild + a limit of 6 concurrent webtransport connections imposed by the browser.
- Browser support:
  - :yellow_circle: Chromium (SIGSEV a lot for whatever reason)
  - :red_circle: Firefox, lacks support for `serverCertificateHashes` https://bugzilla.mozilla.org/show_bug.cgi?id=1806693
  - :red_circle: Safari, no webtransport


Jorropo:
Because I'm bad at rust.

Wondertan:
Celestia would support these efforts. We have a notion of Light Nodes that can access the chain state in trust minimized manner, and we want to experiment with running them in the browser over WebTransport/RTC. We don't have a rust implementation of the protocol and do not plan to have it soon, so Golang would be appreciated, even considering how inefficient it is.

MarcoPolo:
@Jorropo how much of go-libipfs compiles to wasm? What about kad and pubsub?

I have some thoughts I've been drafting about a need to be able to compile protocols into a wasm module. Then have plumbing to allow every implementation to use the same wasm protocol module. I'll link them here when they're ready.

As for this, it's definitely cool but I'm expecting not practical besides for demos. If you do get it to compile, my guess is that the bundled size will be pretty big (I hope I'm wrong!). So I don't expect core go-libp2p maintainers to spend effort maintaining this anytime soon.

Jorropo:
> What about kad and pubsub?

I would guess they work out of the box, given they don't call any io themselves.

In general, anything compiles to wasm, the language is fully implemented, however the API you have access to are pretty limited.
In theory if we fix direct io based stuff (transports, local interfaces discovery, ...) then any libp2p would work out of the box.

> the bundled size will be pretty big

You are probably right, however I don't think the bundle size maters much for let's say a browser application that implements a gateway, you download it once and then keep it cached.

PR #2192, opened 2023-03-16
relay_finder: fix busy loop

### LINKED ISSUE #2189: autorelay: Handle WithMinInterval(0) without constantly looping
If we initialise autorelay with WithMinInterval(0) that causes the background go routine to go into a spin loop. 
https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autorelay/relay_finder.go#L225-L230
here the scheduledTime for next allowed call will always be current time and the scheduledWorkTimer fires immediately. 

kubo currently uses MinInterval(0)
https://github.com/ipfs/kubo/blob/a3366c522a578db610d8633c5a15ebddb6e96e39/core/node/libp2p/relay.go#L101

MarcoPolo:
Fixes #2189. Closes #2190

### DIFF
--- a/p2p/host/autorelay/relay_finder.go
+++ b/p2p/host/autorelay/relay_finder.go
@@ -168,14 +168,19 @@ func (rf *relayFinder) background(ctx context.Context) {
 
 	// This is the least frequent event. It's our fallback timer if we don't have any other work to do.
 	leastFrequentInterval := rf.conf.minInterval
-	if rf.conf.backoff > leastFrequentInterval {
+	// Check if leastFrequentInterval is 0 to avoid busy looping
+	if rf.conf.backoff > leastFrequentInterval || leastFrequentInterval == 0 {
 		leastFrequentInterval = rf.conf.backoff
 	}
-	if rf.conf.maxCandidateAge > leastFrequentInterval {
+	if rf.conf.maxCandidateAge > leastFrequentInterval || leastFrequentInterval == 0 {
 		leastFrequentInterval = rf.conf.maxCandidateAge
 	}
-	if rsvpRefreshInterval > leastFrequentInterval {
-		leastFrequentInterval = rf.conf.maxCandidateAge
+	if rsvpRefreshInterval > leastFrequentInterval || leastFrequentInterval == 0 {
+		leastFrequentInterval = rsvpRefreshInterval
+	}
+	if leastFrequentInterval == 0 {
+		// fallback to avoid busy looping in case the config is set to all 0s
+		leastFrequentInterval = time.Hour
 	}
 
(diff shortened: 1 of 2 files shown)


p2p/host/autorelay/relay_finder.go
```diff
@@ -261,9 +266,12 @@ func (rf *relayFinder) runScheduledWork(ctx context.Context, now time.Time, sche
...
-		scheduledWork.nextAllowedCallToPeerSource = scheduledWork.nextAllowedCallToPeerSource.Add(rf.conf.minInterval)
 		select {
 		case peerSourceRateLimiter <- struct{}{}:
+			scheduledWork.nextAllowedCallToPeerSource = now.Add(rf.conf.minInterval)
+			if scheduledWork.nextAllowedCallToPeerSource.Before(nextTime) {
+				nextTime = scheduledWork.nextAllowedCallToPeerSource
+			}
 		default:
 		}
 	}
```
sukunrt, line 277: You will need to update scheduledWork.nextAllowedCallToPeerSource in the else clause to this if condition like this: 

```
if now.After(scheduledWork.nextAllowedCallToPeerSource) {
		scheduledWork.nextAllowedCallToPeerSource = scheduledWork.nextAllowedCallToPeerSource.Add(rf.conf.minInterval)
		select {
		case peerSourceRateLimiter <- struct{}{}:
			scheduledWork.nextAllowedCallToPeerSource = now.Add(rf.conf.minInterval)
			if scheduledWork.nextAllowedCallToPeerSource.Before(nextTime) {
				nextTime = scheduledWork.nextAllowedCallToPeerSource
			}
		default:
		}
}
else {
	if scheduledWork.nextAllowedCallToPeerSource.Before(nextTime) {
				nextTime = scheduledWork.nextAllowedCallToPeerSource
			}
}
```

sukunrt: We need this because consider the following scenario:

minInterval = 70s
nextRefresh = 60s

Say the call to peerSource returns 1 candidate. Closes the channel and we want to call peer source again. 
This call won't be allowed immediately because minInterval is 70s. 
The nextTime set in scheduledWork will be 60s for nextRefresh. 

After 60s for nextRefresh scheduledWork this function will skip updating nextTime to scheduledWork.nextAllowedCallToPeerSource(10 seconds from now) and schedule the next run for 60s later. 

Am I missing something here?

sukunrt (changes requested):

p2p/host/autorelay/relay_finder.go
MarcoPolo: > Am I missing something here?

Nope! excellent point :)

PR #2193, opened 2023-03-16
websocket: Don't limit message sizes in the websocket reader

### LINKED ISSUE #2188: Bitswap problems over websocket transport since switch to nhooyr websocket
Bug: attempting to fetch UnixFS files over bitswap with peers that only expose websockets will fail on large blocks. Note: this is all of Elastic IPFS/Web3.storage

Steps To Reproduce:

Build Kubo v0.18.0:

Start the daemon:
```
./cmd/ipfs/ipfs daemon
```

```
./cmd/ipfs/ipfs get bafybeiarntxetfbzpyp7baj7kqvqljrn5uyoobq2y4b3zegbbem5joazgy 
```
This is a CID I stored on web3.storage. It works.

Now switch to Kubo master and do the same. It now hangs.

Heavy debugging identifies the following error generated in the websocket conn: `failed to read: read limited at 32769 bytes` -- this comes from here I believe: https://github.com/nhooyr/websocket/blob/master/read.go#L385

Note that actually everything works fine until we get to the first 1MB raw block, I assume because it's over that limit.

This was originally discovered in Lassie and in debugging why it couldn't fetch while Kubo 0.18.0 could, I found that Kubo master couldn't either.

MarcoPolo:
Fixes #2188 

Adds a regression test and makes sure the websocket library isn't allocating the whole message in a byte slice somewhere.

### DIFF
--- a/p2p/transport/websocket/listener.go
+++ b/p2p/transport/websocket/listener.go
@@ -111,6 +111,7 @@ func (l *listener) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Set an arbitrarily large read limit since we don't actually want to limit the message size here.
+	// See https://github.com/nhooyr/websocket/issues/382 for details.
 	c.SetReadLimit(math.MaxInt64 - 1) // -1 because the library adds a byte for the fin frame
 
 	select {
--- a/p2p/transport/websocket/listener.go
+++ b/p2p/transport/websocket/listener.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"math"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


hannahhoward (approved):
LGTM from person who is very excited for this to be fixed but not actually capable of being a real reviewer.

p2p/transport/websocket/listener.go
```diff
@@ -109,8 +110,12 @@ func (l *listener) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Set an arbitrarily large read limit since we don't actually want to limit the message size here.
+	c.SetReadLimit(math.MaxInt64 - 1) // -1 because the library adds a byte for the fin frame
```
marten-seemann, line 114:
```suggestion
	// Set an arbitrarily large read limit since we don't actually want to limit the message size here.
	// See https://github.com/nhooyr/websocket/issues/382 for details.
	c.SetReadLimit(math.MaxInt64 - 1) // -1 because the library adds a byte for the fin frame
```

p2p/transport/websocket/listener.go
```diff
@@ -109,8 +110,12 @@ func (l *listener) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
 	select {
 	case l.incoming <- conn{
+
```
marten-seemann, line 118:
```suggestion
```

marten-seemann (approved):

PR #2196, opened 2023-03-16
basichost: prevent duplicate dials

### LINKED ISSUE #2195: basic host / swarm: duplicate dials when the handshake fails
On certain transports (e.g. QUIC), dialing a connection (`Connect`) will succeed if the server rejects the connection attempt based on the client's 2nd flight. For example, if it rejects the client certificate.

When we call `NewStream`, the swarm will try to establish a new connection again, leading to two connection attempts.
https://github.com/libp2p/go-libp2p/blob/ed96a64de69014fbbe48e9da99583087dd784161/p2p/host/basic/basic_host.go#L627-L641

This is the sequence of events:
1. `Connect` (seems to) succeed, the conn is added to the swarm
2. Identify tries to open a stream, that fails, the swarm recognizes that the connection is closed
3. `NewStream` is called. The swarm sees that we don’t have a connection, dials a new one

marten-seemann:
Fixes #2195.

This fix is not perfect. The error returned will be a `network.ErrNoConn`. Ideally, the swarm would not count a connection as "fully connected" before opening the first stream (and receiving _something_ from the peer) succeeds. This will also be necessary when we implement 0-RTT.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -635,8 +635,13 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
 		}
 	}
 
-	s, err := h.Network().NewStream(ctx, p)
+	s, err := h.Network().NewStream(network.WithNoDial(ctx, "already dialed"), p)
 	if err != nil {
+		// TODO: It would be nicer to get the actual error from the swarm,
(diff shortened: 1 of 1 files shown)


MarcoPolo (approved):

MarcoPolo:
We should catch the errnocon error and return a better error

PR #2199, opened 2023-03-17
websocket: don't set a WSS multiaddr for accepted unencrypted conns

### LINKED ISSUE #2198: websocket: accepted connections claim to be WSS even if they are not
Any WebSocket connection will have WSS address, even if we didn't set a `tls.Config`. They should have a WS address instead.

marten-seemann:
Fixes #2198.

### DIFF
--- a/p2p/transport/websocket/listener.go
+++ b/p2p/transport/websocket/listener.go
@@ -19,6 +19,9 @@ import (
 type listener struct {
 	nl     net.Listener
 	server http.Server
+	// The Go standard library sets the http.Server.TLSConfig no matter if this is a WS or WSS,
+	// so we can't rely on checking if server.TLSConfig is set.
+	isWss bool
 
 	laddr ma.Multiaddr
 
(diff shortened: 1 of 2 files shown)


p2p/transport/websocket/listener.go
```diff
@@ -96,7 +100,7 @@ func (l *listener) serve() {
 
 func (l *listener) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	scheme := "ws"
-	if l.server.TLSConfig != nil {
+	if l.isWss {
```
MarcoPolo, line 103: Should we fix the other checks as well?

MarcoPolo (approved):

p2p/transport/websocket/listener.go
MarcoPolo: I guess there's just the other one in `serve`

marten-seemann: Done.

PR #2200, opened 2023-03-17
tests: add comprehensive end-to-end tests for connection gating

marten-seemann:


BigLep:
What is needed to land this?  Review from @MarcoPolo ?  I'm asking because this is one of the blockers for https://github.com/libp2p/go-libp2p/pull/1999 right?

BigLep:
What is needed to land this?  Review from @MarcoPolo ?  I'm asking because this is one of the blockers for https://github.com/libp2p/go-libp2p/pull/1999 right?

p2p/test/transport/gating_test.go
```diff
@@ -0,0 +1,234 @@
...
+			defer cancel()
+			gomock.InOrder(
+				connGater.EXPECT().InterceptPeerDial(h2.ID()).Return(true),
+				connGater.EXPECT().InterceptAddrDial(h2.ID(), gomock.Any()).Return(true),
+				connGater.EXPECT().InterceptSecured(network.DirOutbound, h2.ID(), gomock.Any()).Do(func(_ network.Direction, _ peer.ID, addrs network.ConnMultiaddrs) {
+					// remove the certhash component from WebTransport addresses
+					require.Equal(t, stripCertHash(h2.Addrs()[0]).String(), addrs.RemoteMultiaddr().String())
+				}),
+			)
+			require.Error(t, h1.Connect(ctx, peer.AddrInfo{ID: h2.ID(), Addrs: h2.Addrs()}))
```
MarcoPolo, line 95: What error are we expecting?

p2p/test/transport/gating_test.go
```diff
@@ -0,0 +1,234 @@
...
+				connGater.EXPECT().InterceptUpgraded(gomock.Any()).Do(func(c network.Conn) {
+					// remove the certhash component from WebTransport addresses
+					require.Equal(t, stripCertHash(h2.Addrs()[0]), c.RemoteMultiaddr())
+					require.Equal(t, h1.ID(), c.LocalPeer())
+					require.Equal(t, h2.ID(), c.RemotePeer())
+				}))
+			require.Error(t, h1.Connect(ctx, peer.AddrInfo{ID: h2.ID(), Addrs: h2.Addrs()}))
+			// There's a bug in the WebSocket library, making Close block for up to 5s.
+			// See https://github.com/nhooyr/websocket/issues/355 for details.
+			if _, err := h2.Addrs()[0].ValueForProtocol(ma.P_WS); err == nil {
```
MarcoPolo, line 132: Same comments as above here.

MarcoPolo (changes requested):

MarcoPolo:
> What is needed to land this? Review from @MarcoPolo ? I'm asking because this is one of the blockers for https://github.com/libp2p/go-libp2p/pull/1999 right?

Was reviewing as you wrote that!

p2p/test/transport/gating_test.go
marten-seemann: Depends on the transport. All we can assert here is that some kind of error occurred.

MarcoPolo: Can we do better? Should we return a well known error? Callers may want to check if this error case happened and it doesn’t make sense for them to check against all error types depending on underlying transport.

MarcoPolo (approved):

PR #2206, opened 2023-03-20
basichost: don't allocate when deduplicating multiaddrs

marten-seemann:
This is a major source of allocations. We can live with a `O(N*log(N))` scaling here.

We probably should move this function to the multiaddr package at some point.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -817,7 +817,7 @@ func dedupAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
 	if len(addrs) == 0 {
 		return addrs
 	}
-	sort.Slice(addrs, func(i, j int) bool { return bytes.Compare(addrs[i].Bytes(), addrs[j].Bytes()) <= 0 })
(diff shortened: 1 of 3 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -763,18 +765,24 @@ func (h *BasicHost) Addrs() []ma.Multiaddr {
...
-		k := string(addr.Bytes())
-		if exists[k] {
-			continue
+// dedupAddrs deduplicates addresses in place, leave only unique addresses.
+// It doesn't allocate.
+func dedupAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
+	if len(addrs) == 0 {
+		return addrs
+	}
+	sort.Slice(addrs, func(i, j int) bool { return bytes.Compare(addrs[i].Bytes(), addrs[j].Bytes()) <= 0 })
```
MarcoPolo, line 774: This should be less, not less or equal.
```suggestion
	sort.Slice(addrs, func(i, j int) bool { return bytes.Compare(addrs[i].Bytes(), addrs[j].Bytes()) < 0 })
```

MarcoPolo (commented):
small change, then looks good

MarcoPolo (approved):

PR #2208, opened 2023-03-21
autorelay: fix busy loop bug and flaky tests in relay finder

### LINKED ISSUE #2189: autorelay: Handle WithMinInterval(0) without constantly looping

MarcoPolo:
Also adds our own custom mock clock that fits a better interface for this module.

The new mock clock implements the following interface:
```go
// InstantTimer is a timer that triggers at some instant rather than some duration
type InstantTimer interface {
	Reset(d time.Time) bool
	Stop() bool
	Ch() <-chan time.Time
}

// ClockWithInstantTimer is a clock that can create timers that trigger at some
// instant rather than some duration
type ClockWithInstantTimer interface {
	Now() time.Time
	Since(t time.Time) time.Duration
	InstantTimer(when time.Time) InstantTimer
}
```

This lets us avoid the issue of https://github.com/libp2p/go-libp2p/pull/2187 altogether.

The new mock clock is about 100 LOC and should behave logically close to a real clock (time shouldn't jump from 0 to 1000 without work being done in between).

Closes #2187. Closes #2192. Fixes #2189. Closes #2183 

I chose to incorporate those PRs into this one to avoid the merge conflicts that would ensue.

### DIFF
--- a/core/test/mockclock.go
+++ b/core/test/mockclock.go
@@ -89,6 +89,10 @@ func (c *mockClock) AdvanceBy(dur time.Duration) {
 	// sort timers by when
 	if len(c.timers) > 1 {
 		sort.Slice(c.timers, func(i, j int) bool {
+			c.timers[i].mu.Lock()
+			c.timers[j].mu.Lock()
+			defer c.timers[i].mu.Unlock()
+			defer c.timers[j].mu.Unlock()
 			return c.timers[i].when.Before(c.timers[j].when)
 		})
 	}
--- a/core/test/mockclock.go
+++ b/core/test/mockclock.go
@@ -1,0 +1,119 @@
+package test
+
+import (
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/libp2p/go-libp2p/p2p/host/autorelay"
+)
+
+type mockClock struct {
+	mu     sync.Mutex
+	now    time.Time
+	timers []*mockInstantTimer
+}
+
+type mockInstantTimer struct {
+	c      *mockClock
+	mu     sync.Mutex
+	when   time.Time
+	active bool
+	ch     chan time.Time
+}
+
+func (t *mockInstantTimer) Ch() <-chan time.Time {
+	return t.ch
+}
+
+func (t *mockInstantTimer) Reset(d time.Time) bool {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	wasActive := t.active
+	t.active = true
+	t.when = d
+
+	// Schedule any timers that need to run. This will run this timer if t.when is before c.now
+	go t.c.AdvanceBy(0)
+
+	return wasActive
+}
+
+func (t *mockInstantTimer) Stop() bool {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	wasActive := t.active
+	t.active = false
+	return wasActive
+}
+
+var _ autorelay.InstantTimer = &mockInstantTimer{}
+var _ autorelay.ClockWithInstantTimer = &mockClock{}
+
+func NewMockClock() *mockClock {
+	return &mockClock{now: time.Unix(0, 0)}
+}
+
+// InstantTimer implements autorelay.ClockWithInstantTimer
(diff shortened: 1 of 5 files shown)


p2p/host/autorelay/autorelay_test.go
```diff
@@ -294,8 +296,122 @@ func TestConnectOnDisconnect(t *testing.T) {
...
 
+type mockClock struct {
+	mu     sync.Mutex
+	now    time.Time
+	timers []*mockInstantTimer
+}
+
+var _ autorelay.ClockWithInstantTimer = &mockClock{}
+
+type mockInstantTimer struct {
```
marten-seemann, line 307: Move this to a separate file? We might need this at other places as well, and that makes it easier to expose later without losing Git history.

p2p/host/autorelay/autorelay_test.go
```diff
@@ -294,8 +296,122 @@ func TestConnectOnDisconnect(t *testing.T) {
...
+	return c.now.Sub(t)
+}
+
+func (c *mockClock) Now() time.Time {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.now
+}
+
+func (c *mockClock) advanceBy(dur time.Duration) {
```
marten-seemann, line 376:
```suggestion
func (c *mockClock) AdvanceBy(dur time.Duration) {
```

p2p/host/autorelay/relay_finder.go
```diff
@@ -128,32 +132,36 @@ func (rf *relayFinder) background(ctx context.Context) {
...
+	if rf.conf.maxCandidateAge > leastFrequentInterval || leastFrequentInterval == 0 {
 		leastFrequentInterval = rf.conf.maxCandidateAge
 	}
-	if rsvpRefreshInterval > leastFrequentInterval {
-		leastFrequentInterval = rf.conf.maxCandidateAge
+	if rsvpRefreshInterval > leastFrequentInterval || leastFrequentInterval == 0 {
+		leastFrequentInterval = rsvpRefreshInterval
+	}
+	if leastFrequentInterval == 0 {
+		// fallback to avoid busy looping in case the config is set to all 0s
```
marten-seemann, line 152: `rsvpRefreshInterval` is a constant (1 minute), so this is not possible, right?

core/test/mockclock.go
marten-seemann, line 1: Oh nice, you already extracted it to a separate package.

p2p/host/autorelay/relay_finder.go
```diff
@@ -128,32 +132,32 @@ func (rf *relayFinder) background(ctx context.Context) {
...
 
-	now := rf.conf.clock.Now()
-
 	scheduledWork := &scheduledWorkTimes{
 		leastFrequentInterval:       leastFrequentInterval,
 		nextRefresh:                 now.Add(rsvpRefreshInterval),
-		nextBackoff:                 now.Add(rf.conf.backoff / 5),
-		nextOldCandidateCheck:       now.Add(rf.conf.maxCandidateAge / 5),
+		nextBackoff:                 now.Add(rf.conf.backoff),
+		nextOldCandidateCheck:       now.Add(rf.conf.maxCandidateAge),
```
marten-seemann, line 156: Is this correct? I thought the 5 was so we don't add up with lists that are completely outdated. Reminder to self: document every magic number.

marten-seemann: I've convinced myself that this isn't needed any more.

marten-seemann (approved):

PR #2215, opened 2023-03-24
relaysvc: flaky TestReachabilityChangeEvent

sukunrt:


sukunrt:
this line is expensive: https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/relay.go#L629

But we can run gc separately in the Close call itself. 

should we add this file to the repo? this was helpful
https://github.com/libp2p/go-libp2p/blob/e86584ffb2c17f05a494334acb7cef99b048cc00/docs/flaky-tests.md

MarcoPolo:
> this line is expensive: https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/relay.go#L629


Yeah, that's expected. It should be only needed once per process start though.

MarcoPolo (approved):
Looks good

MarcoPolo:
I'm guessing you removed the waitgroup since the gc method is safe to run multiple times, and you run the gc once more during close which effectively waits for the background goroutine to finish. Is that correct?

sukunrt:
> I'm guessing you removed the waitgroup since the gc method is safe to run multiple times, and you run the gc once more during close which effectively waits for the background goroutine to finish. Is that correct?

Yes. We want to run the gc once before closing to remove all the tagged connections. 

The source of flakiness in the test is that the background go routine hasn't finished ```asnutil.Store.Init``` by the time we've called Close which leads to a timeout [here](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/relaysvc/relay_test.go#L34) 
On calling Close it is waiting for the background goroutine to end. 

It is fine to return before the background goroutine has finished as we've removed this relay out of the rest of the system by removing the streamhandlers and network.Notifee in Close and untagged all connections by running the gc once.

PR #2218, opened 2023-03-25
Add initial prototype of http over libp2p

MarcoPolo:
This doesn't handle all the edge cases (like expect continue) and trailers. Maybe less code than implementing a net.Listener? At least it's one goroutine.

Might switch to a net.Listener later.

### DIFF
--- a/p2p/http/responsewriter.go
+++ b/p2p/http/responsewriter.go
@@ -1,0 +1,146 @@
+package p2phttp
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+	"sync"
+
+	logging "github.com/ipfs/go-log/v2"
+)
+
+var bufWriterPool = sync.Pool{
+	New: func() any {
+		return bufio.NewWriterSize(nil, 4<<10)
+	},
+}
+
+var bufReaderPool = sync.Pool{
+	New: func() any {
+		return bufio.NewReaderSize(nil, 4<<10)
+	},
+}
+
+var log = logging.Logger("p2phttp")
+
+var _ http.ResponseWriter = (*httpResponseWriter)(nil)
+
+type httpResponseWriter struct {
+	w                     *bufio.Writer
+	directWriter          io.Writer
+	header                http.Header
+	wroteHeader           bool
+	inferredContentLength bool
+}
+
+// Header implements http.ResponseWriter
+func (w *httpResponseWriter) Header() http.Header {
+	return w.header
+}
+
+// Write implements http.ResponseWriter
+func (w *httpResponseWriter) Write(b []byte) (int, error) {
+	if !w.wroteHeader {
+		if w.header.Get("Content-Type") == "" {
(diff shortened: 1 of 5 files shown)


p2p/http/responsewriter.go
```diff
@@ -0,0 +1,146 @@
...
+	New: func() any {
+		return bufio.NewWriterSize(nil, 4<<10)
+	},
+}
+
+var bufReaderPool = sync.Pool{
+	New: func() any {
+		return bufio.NewReaderSize(nil, 4<<10)
+	},
+}
```
marten-seemann, line 24: Probably doesn't matter (and definitely not for a prototype): What about creating a new reader / writer, but taking the underlying the byte slice from our shared buffer pool? That's an additional alloc for the `bufio.{Reader/Writer}`, but we get better byte slice reuse.

p2p/http/responsewriter.go
```diff
@@ -0,0 +1,146 @@
...
+			return w.w.Write(b)
+		}
+	}
+
+	// Ran out of buffered space, We should check if we need to write the headers.
+	if !w.wroteHeader {
+		// Be nice for small things
+		if w.header.Get("Content-Length") == "" {
+			w.inferredContentLength = true
+			w.header.Set("Content-Length", strconv.Itoa(len(b)+w.w.Buffered()))
```
marten-seemann, line 61: This would set a wrong content length if `Write` is called multiple times.

p2p/http/responsewriter.go
```diff
@@ -0,0 +1,146 @@
...
+		return
+	}
+	w.wroteHeader = true
+	w.writeStatusLine(statusCode)
+	w.header.Write(w.directWriter)
+	w.directWriter.Write([]byte("\r\n"))
+}
+
+// Copied from Go stdlib https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/net/http/server.go;drc=ea4631cc0cf301c824bd665a7980c13289ab5c9d;l=1533
+func (w *httpResponseWriter) writeStatusLine(code int) {
```
marten-seemann, line 104: Text-based protocols 🤢. This is so much nicer in H3!

marten-seemann (commented):
LGTM, except for the content length issue. Maybe creating a `net.Listener` isn't that bad after all, at least we wouldn't have to deal that much with HTTP internals.

p2p/http/responsewriter.go
MarcoPolo: We get to reuse the whole thing here which is probably net better for the GC.

p2p/http/responsewriter.go
MarcoPolo: It doesn't. It's subtle. I added a test to show this.

p2p/http/responsewriter.go
MarcoPolo: Glass houses...

p-shahi:
closing this in favor of https://github.com/libp2p/go-libp2p/pull/2438

PR #2224, opened 2023-03-26
fix: fix externally defined webtransport addresses

### LINKED ISSUE #2223: Cannot advertise WebTransport with externally known addresses (external static IP + port forwarding, dns, etc.)

aschmahmann:
fixes #2223

Support externally defined (e.g. with externally known ports from port forwarding, dns, etc.)  webtransport addresses. This mechanism is very hacky and should be removed as expediently as possible.

There are definitely better wasy to do this, but wanted to at least put out an option that exists so that:
1. We at least have something we're looking at/comparing against
2. There's code in case this fix is deemed "good enough" for now
3. If downstream consumers feel they need this fix sooner than it can be delivered in go-libp2p they can copy this hack into their codebases

Maintainers can feel free to close or modify this PR.


SgtPooki:
Can confirm this populates my certhash properly when adding 

```
    "AppendAnnounce": [
      "/ip4/x.x.x.x/udp/4001/quic-v1/webtransport"
    ],
```

to my ~/.ipfs/config

marten-seemann (commented):
Thanks for the fix! This looks correct to me, but it does bake in some assumptions which might not be true in the future (e.g. there'll always be exactly 2 certhash components).

We really need to refactor our API and get rid of this `AddrsFactory`. Having external applications (and internal components as well) mess with the addresses is not ideal...

p-shahi:
Can we close this in favor of https://github.com/libp2p/go-libp2p/pull/2227 ?

aschmahmann:
Yep, they're fairly equivalent and Marco has made more reusable exports in his. Thanks for pushing on this 🙏

PR #2227, opened 2023-03-27
webtransport: Add helpers to transform webtransport multiaddrs in AddrsFactory

### LINKED ISSUE #2223: Cannot advertise WebTransport with externally known addresses (external static IP + port forwarding, dns, etc.)

MarcoPolo:
I think the AddrsFactory is a bit hard to use. This PR doesn't address that.

This PR adds some helper functions to allow users to do the correct thing inside AddrsFactory. Fixes #2223 

Users can add their own webtransport mappings by copying certhashes from an existing multiaddr. For example,

```go
AddrsFactory(func(multiaddrs []ma.Multiaddr) []ma.Multiaddr {
	outAddrs := make([]ma.Multiaddr, 0, len(multiaddrs)+1)
	for _, addr := range multiaddrs {
		// Keep the original addrs.
		outAddrs = append(outAddrs, addr)
		if webtransport.IsWebtransportMultiaddrWithCerthash(addr) {
			// Add our custom addr with certhashes from an existing addr.
			outAddrs = append(outAddrs, webtransport.CopyCerthashes(addr, customAddr))
		}
	}
	return outAddrs
})
```

This also allows users to do this only for specifics addresses in case they are listening on multiple addresses.

### DIFF
--- a/libp2p_test.go
+++ b/libp2p_test.go
@@ -17,6 +17,7 @@ import (
 	tls "github.com/libp2p/go-libp2p/p2p/security/tls"
 	quic "github.com/libp2p/go-libp2p/p2p/transport/quic"
 	"github.com/libp2p/go-libp2p/p2p/transport/tcp"
+	webtransport "github.com/libp2p/go-libp2p/p2p/transport/webtransport"
 
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/stretchr/testify/require"
@@ -288,4 +289,49 @@ func TestSecurityConstructor(t *testing.T) {
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "failed to negotiate security protocol")
 	require.NoError(t, h2.Connect(context.Background(), ai))
+}
+
+func TestTransportConstructorWebTransport(t *testing.T) {
+	h, err := New(
+		Transport(webtransport.New),
+		DisableRelay(),
+	)
+	require.NoError(t, err)
+	defer h.Close()
+	require.NoError(t, h.Network().Listen(ma.StringCast("/ip4/127.0.0.1/udp/0/quic-v1/webtransport")))
+	err = h.Network().Listen(ma.StringCast("/ip4/127.0.0.1/udp/0/quic-v1/"))
+	require.Error(t, err)
+	require.Contains(t, err.Error(), swarm.ErrNoTransport.Error())
+}
+
+func TestTransportCustomAddressWebTransport(t *testing.T) {
+	customAddr, err := ma.NewMultiaddr("/ip4/1.2.3.4/udp/1234/quic-v1/webtransport")
+	if err != nil {
+		t.Fatal(err)
+	}
+	h, err := New(
+		Transport(webtransport.New),
+		DisableRelay(),
+		AddrsFactory(func(multiaddrs []ma.Multiaddr) []ma.Multiaddr {
+			customAddrWithCertHash := customAddr
+			for _, addr := range multiaddrs {
+				if webtransport.IsWebtransportMultiaddrWithCerthash(addr) {
+					customAddrWithCertHash = webtransport.CopyCerthashes(addr, customAddrWithCertHash)
+					break
+				}
+			}
+			return []ma.Multiaddr{customAddrWithCertHash}
+		}),
+	)
+	require.NoError(t, err)
+	defer h.Close()
+	require.NoError(t, h.Network().Listen(ma.StringCast("/ip4/127.0.0.1/udp/0/quic-v1/webtransport")))
+	addrs := h.Addrs()
+	require.Len(t, addrs, 1)
+	require.NotEqual(t, addrs[0], customAddr)
+	restOfAddr, lastComp := ma.SplitLast(addrs[0])
+	restOfAddr, secondToLastComp := ma.SplitLast(restOfAddr)
+	require.Equal(t, lastComp.Protocol().Code, ma.P_CERTHASH)
+	require.Equal(t, secondToLastComp.Protocol().Code, ma.P_CERTHASH)
+	require.True(t, restOfAddr.Equal(customAddr))
 }
--- a/libp2p_test.go
+++ b/libp2p_test.go
@@ -313,14 +313,7 @@ func TestTransportCustomAddressWebTransport(t *testing.T) {
 		Transport(webtransport.New),
 		DisableRelay(),
 		AddrsFactory(func(multiaddrs []ma.Multiaddr) []ma.Multiaddr {
-			customAddrWithCertHash := customAddr
-			for _, addr := range multiaddrs {
-				if webtransport.IsWebtransportMultiaddrWithCerthash(addr) {
-					customAddrWithCertHash = webtransport.CopyCerthashes(addr, customAddrWithCertHash)
-					break
-				}
-			}
-			return []ma.Multiaddr{customAddrWithCertHash}
+			return []ma.Multiaddr{customAddr}
 		}),
 	)
 	require.NoError(t, err)
--- a/options.go
+++ b/options.go
@@ -271,13 +271,11 @@ func AddrsFactory(factory config.AddrsFactory) Option {
 		var log = logging.Logger("addrs-factory")
 
 		cfg.AddrsFactory = func(in []ma.Multiaddr) []ma.Multiaddr {
-			// This wrapper does the following:
-			// - If the user returns a webtransport multiaddr that doesn't have
-			//   a certhash, and all our known addrs have a certhash, copy the
-			//   certhash over.
-			//   - If we only have a webtransport multiaddr without a certhash
-			//     (client passed a static tls config), don't copy.
-			//   - If we have both, log a warning and don't copy the certhash.
+			// This wrapper checks fills in any missing webtransport certhashes
+			// in user provided webtransport addresses. It does this by getting
+			// the certhashes from the input multiaddrs. It's an error if we get
+			// see a webtransport multiaddr without a certhash since we don't
+			// support static TLS configs currently.
 
 			out := factory(in)
 			var multiaddrWithCerthash ma.Multiaddr
@@ -303,8 +301,8 @@ func AddrsFactory(factory config.AddrsFactory) Option {
 				}
 			}
 
-			if multiaddrWithCerthash != nil && foundInMultiaddrWithoutCerthash {
-				log.Warn("Ambigous mapping in AddrsFactory. Not sure we if should fill in a certhash in output multiaddrs since we have both a multiaddr with and without a certhash in the input.")
+			if foundInMultiaddrWithoutCerthash {
+				log.Error("Found a webtransport multiaddr without a certhash in the input to AddrsFactory. This shouldn't happen")
 			}
 			return out
 		}
--- a/options.go
+++ b/options.go
@@ -11,6 +11,7 @@ import (
 	"reflect"
 	"time"
 
+	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p/config"
 	"github.com/libp2p/go-libp2p/core/connmgr"
 	"github.com/libp2p/go-libp2p/core/crypto"
@@ -27,6 +28,7 @@ import (
 	relayv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
 	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
 	"github.com/libp2p/go-libp2p/p2p/transport/quicreuse"
+	libp2pwebtransport "github.com/libp2p/go-libp2p/p2p/transport/webtransport"
 	"github.com/prometheus/client_golang/prometheus"
 
 	ma "github.com/multiformats/go-multiaddr"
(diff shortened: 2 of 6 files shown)


p2p/transport/webtransport/multiaddr.go
```diff
@@ -78,3 +75,59 @@ func addrComponentForCert(hash []byte) (ma.Multiaddr, error) {
...
 }
+
+// IsWebtransportMultiaddrWithCerthash returns true if the given multiaddr is a
+// well formed webtransport multiaddr. Returns the number of certhashes found.
+func IsWebtransportMultiaddr(multiaddr ma.Multiaddr) (bool, int) {
+	const (
+		init              = iota
+		foundUDP          = iota
+		foundQuicV1       = iota
+		foundWebTransport = iota
```
marten-seemann, line 86: One `iota` should be enough. I don't know how multiple of them behave, does it reset to 0?

p2p/transport/webtransport/multiaddr.go
```diff
@@ -78,3 +75,59 @@ func addrComponentForCert(hash []byte) (ma.Multiaddr, error) {
...
+	ma.ForEach(multiaddr, func(c ma.Component) bool {
+		if c.Protocol().Code == ma.P_QUIC_V1 && state == init {
+			state = foundUDP
+		}
+		if c.Protocol().Code == ma.P_QUIC_V1 && state == foundUDP {
+			state = foundQuicV1
+		}
+		if c.Protocol().Code == ma.P_WEBTRANSPORT && state == foundQuicV1 {
+			state = foundWebTransport
+		}
```
marten-seemann, line 100: Why not use an address matcher? We shouldn't be rewriting basic multiaddr functions here.

p2p/transport/webtransport/multiaddr.go
```diff
@@ -78,3 +75,59 @@ func addrComponentForCert(hash []byte) (ma.Multiaddr, error) {
...
+	ok, n := IsWebtransportMultiaddr(multiaddr)
+	return ok && n > 0
+}
+
+// CopyCerthashes copies the certificate hashes from the first multiaddr to the
+// second. If there are no multiaddrs in the first, the second is returned
+// unchanged.  Otherwise a new multiaddr is returned with certhashes appeneded
+// to it
+func CopyCerthashes(existingMultiaddrWithCerthashes, multiaddrWithoutCerthashes ma.Multiaddr) ma.Multiaddr {
+	var certhashComponents []ma.Component
```
marten-seemann, line 121: Pre-allocate 2, as this is what we'd usually have.

p2p/host/basic/basic_host.go
```diff
@@ -760,6 +760,7 @@ func (h *BasicHost) ConnManager() connmgr.ConnManager {
 // Addrs returns listening addresses that are safe to announce to the network.
 // The output is the same as AllAddrs, but processed by AddrsFactory.
 func (h *BasicHost) Addrs() []ma.Multiaddr {
+	h.Network()
```
marten-seemann, line 763: What's this?

marten-seemann (commented):
It's unfortunate that we need this. It's a lot of additional API for something that "should just work". `AddrsFactory` really is a terrbile API, and we should work on cleaning that up soon.

p2p/transport/webtransport/multiaddr.go
MarcoPolo: I was following the [docs](https://go.dev/ref/spec#Iota). It's the same thing.

p2p/transport/webtransport/multiaddr.go
MarcoPolo: Address matcher doesn't do exactly what I want which is:

1. I don't care if it's ipv4/dns4/ipv6/dns6
2. I don't care if it has certhashes or not, but if it does tell me how many.

I know about address matcher, it just didn't do what I needed. See that hack that I removed in `CanDial`. This is pretty straightforward, but it would be good to expand the address matcher to handle these use cases. Doing that in this PR is out of scope though.

p2p/host/basic/basic_host.go
MarcoPolo: a typo, thanks.

libp2p_test.go
```diff
@@ -289,3 +290,48 @@ func TestSecurityConstructor(t *testing.T) {
...
+		AddrsFactory(func(multiaddrs []ma.Multiaddr) []ma.Multiaddr {
+			customAddrWithCertHash := customAddr
+			for _, addr := range multiaddrs {
+				if webtransport.IsWebtransportMultiaddrWithCerthash(addr) {
+					customAddrWithCertHash = webtransport.CopyCerthashes(addr, customAddrWithCertHash)
+					break
+				}
+			}
+			return []ma.Multiaddr{customAddrWithCertHash}
+		}),
```
aschmahmann, line 324: What's the reasoning behind surfacing this requirement to any libp2p applications who require something like webtransport + port forwarding rather than just handling it internally (e.g. applying this automatically in the host constructor in config.go)?

MarcoPolo: I don't think we could do the right thing in 100% of cases. For example, what if the server is listening on two webtransport addresses, which one are you forwarding to (the cerhashes need to line up)? This is something the application using go-libp2p has insight, but not go-libp2p itself. Another example, maybe you also want to return both multiaddrs instead of replacing it.

At a high level though, I think this option `AddrsFactory` is hard to use, and I'd like a better api that covers 90% of use cases in a simpler way.

aschmahmann: > the cerhashes need to line up

Aren't the certhashes always exactly the same on every webtransport interface you're listening on, so it doesn't matter what interface's certhashes you choose they should all be the same.

https://github.com/libp2p/go-libp2p/blob/8c466ab7f8d0e1eb12ea8a23933e77d25190a81a/p2p/transport/webtransport/cert_manager.go#L63 only takes the private key and time and there's one private key per host and one time globally 😄.

An alternative approach to copy certhashes here could've been to just recalculate the deterministic certhashes and use them, but those functions are unexported in the `webtransport` package (as they probably should be since they're definitely internal implementation details). The way you've implemented this (and me in #2224) happens to be slightly less abstraction breaking.

> At a high level though, I think this option AddrsFactory is hard to use, and I'd like a better api that covers 90% of use cases in a simpler way.

Better APIs sound😄. However, I don't want to rely on future refactors to fix existing issues especially if they could take a while to materialize.

MarcoPolo: > Aren't the certhashes always exactly the same on every webtransport interface you're listening on, so it doesn't matter what interface's certhashes you choose they should all be the same.

Good point. But this may not always be true. Maybe this is an argument for trying to do this ourselves?

You could have an address that doesn't have certhashes (I don't think we support this right now, but we want to) because it has a valid TLS cert. In that case you would want to forward something without a certhash. Automatically appending a certhash would be wrong.

#2224 was helpful, I took the tests from it. But I didn't like the change where config.go now has special knowledge about WebTransport addresses. I want to keep AddrsFactory as simple as possible.

> However, I don't want to rely on future refactors to fix existing issues especially if they could take a while to materialize.

This change should unblock kubo, no?

MarcoPolo: Happy to come up with a better change that keeps AddrsFactory simple and makes things better for the user if you have any suggestions!

aschmahmann: > Good point. But this may not always be true. Maybe this is an argument for trying to do this ourselves?

Yeah, my hope is to ask fewer people to worry about the complexity here assuming it's not much worse in the central place we're dealing with it (e.g. in go-libp2p).

> This change should unblock kubo, no?

Yeah, it will. Technically no changes were needed to unblock kubo as none of the changes made here access private fields/functions, these functions do make things easier though 😄. However, needing to add those extra lines of code seems like to surprise people.

If we went the currently proposed route I'd probably try to save other people, at least in the IPFS ecosystem, the hassle and also add it in a boxo helper so other IPFS implementations don't need to worry too much about it (e.g. https://github.com/ipfs/boxo/pull/230). Maybe that'll help enough 🤷.

> You could have an address that doesn't have certhashes (I don't think we support this right now, but we want to) because it has a valid TLS cert.

Yeah that's a good point. We could certainly do this and that would complicate things, although IIUC it would only cause a problem in a case where:
1. You have an externally provided TLS cert for that address
2. The TLS cert lives outside of go-libp2p (where the constructor couldn't access it) and that part of the QUIC connection establishment happens outside of go-libp2p

Maybe this a reasonable thing. I have no idea how common such a QUIC proxy setup is in general or would be for go-libp2p users. I don't know if people would complain too much if they had to pass extra options to deal with externally established connections.

My preference is to try and eat the complexity so the users don't have to. Given the cases not handled by automatic fixing in the constructor seem to not exist quite yet it seems like we can protect go-libp2p users from thinking about this for a little while longer which is 👍.

---

> Happy to come up with a better change that keeps AddrsFactory simple and makes things better for the user if you have any suggestions!

Unfortunately, nothing better than what I've suggested is coming to mind at the moment. I can see the advantages to both views here, although I have my preference 😄. I'll take whatever the maintainers decide and incorporate with it in the downstreams I help maintain though.

MarcoPolo: > 2. The TLS cert lives outside of go-libp2p (where the constructor couldn't access it) and that part of the QUIC connection establishment happens outside of go-libp2p

No, we could pass a static tls config to the constructor and that would be fine. This isn't like websockets.

Okay. I think eating this complexity/messiness for the user is the nice thing to do. I'll update this PR.

Concretely:
- If the user returns a webtransport multiaddr that doesn't have a certhash, and all our known addrs have a certhash, copy the certhash over.
  - If we only have a webtransport multiaddr without a certhash (client passed a static tls config), don't copy.
  - If we have both, log a warning and don't copy the certhash.

Does that seem right?

aschmahmann: Yeah that seems right to me. I suspect if we reach a scenario where someone is trying to instantiate two or more webtransport libp2p transports (e.g. 1+ with CA certs and 1 with the autogenerated ones) and they're having trouble specifying to the `AddrsFactory` how to signal "please give me autogenerated certs" that we can deal with it then.

I couldn't find an issue for enabling being a server with CA certs for webtransport (since they're not currently supported https://github.com/libp2p/go-libp2p/blob/8c466ab7f8d0e1eb12ea8a23933e77d25190a81a/p2p/transport/webtransport/transport.go#L316). If not we should probably start one and then flag this edge case there since this is the kind of thing that could regress or surprise people if the feature isn't implemented in the near future (when any of us still remember this issue 😅).

MarcoPolo: https://github.com/libp2p/go-libp2p/issues/2230 is the issue.

I've updated the PR. Thanks for your input @aschmahmann ☺️

options.go
```diff
@@ -262,7 +264,51 @@ func AddrsFactory(factory config.AddrsFactory) Option {
 		if cfg.AddrsFactory != nil {
 			return fmt.Errorf("cannot specify multiple address factories")
 		}
-		cfg.AddrsFactory = factory
+		if factory == nil {
+			return fmt.Errorf("cannot specify a nil address factory")
```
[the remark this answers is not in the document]
MarcoPolo: The idea is that this is just middleware. It just wraps the function and the rest of the code doesn't care that this changed.

Another reason to not put it in config.go is that function is already a bit unwieldy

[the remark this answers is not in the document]
marten-seemann: > Technically users can also instantiate directly with `config.NewNode()`
> 
> https://github.com/libp2p/go-libp2p/blob/8c466ab7f8d0e1eb12ea8a23933e77d25190a81a/config/config.go#L288
> 
> although I couldn't find any examples of anyone who did.

Oh no, please don't do that, @aschmahmann! `config` will be removed soon-ish.

options.go
```diff
@@ -262,7 +264,51 @@ func AddrsFactory(factory config.AddrsFactory) Option {
...
+		var log = logging.Logger("addrs-factory")
+
+		cfg.AddrsFactory = func(in []ma.Multiaddr) []ma.Multiaddr {
+			// This wrapper does the following:
+			// - If the user returns a webtransport multiaddr that doesn't have
+			//   a certhash, and all our known addrs have a certhash, copy the
+			//   certhash over.
+			//   - If we only have a webtransport multiaddr without a certhash
+			//     (client passed a static tls config), don't copy.
+			//   - If we have both, log a warning and don't copy the certhash.
```
marten-seemann, line 280: I usually don't review whitespace, but this looks really messy ;)

options.go
```diff
@@ -262,7 +264,51 @@ func AddrsFactory(factory config.AddrsFactory) Option {
...
+
+		var log = logging.Logger("addrs-factory")
+
+		cfg.AddrsFactory = func(in []ma.Multiaddr) []ma.Multiaddr {
+			// This wrapper does the following:
+			// - If the user returns a webtransport multiaddr that doesn't have
+			//   a certhash, and all our known addrs have a certhash, copy the
+			//   certhash over.
+			//   - If we only have a webtransport multiaddr without a certhash
+			//     (client passed a static tls config), don't copy.
```
marten-seemann, line 279: We don't support static TLS configs. As this is just a hack until we have a proper fix, we can reduce complexity here by hardcoding that.

p2p/transport/webtransport/multiaddr.go
```diff
@@ -78,3 +75,59 @@ func addrComponentForCert(hash []byte) (ma.Multiaddr, error) {
...
+func IsWebtransportMultiaddrWithCerthash(multiaddr ma.Multiaddr) bool {
+	ok, n := IsWebtransportMultiaddr(multiaddr)
+	return ok && n > 0
+}
+
+// CopyCerthashes copies the certificate hashes from the first multiaddr to the
+// second. If there are no multiaddrs in the first, the second is returned
+// unchanged.  Otherwise a new multiaddr is returned with certhashes appeneded
+// to it
+func CopyCerthashes(existingMultiaddrWithCerthashes, multiaddrWithoutCerthashes ma.Multiaddr) ma.Multiaddr {
```
marten-seemann, line 120: Would it make more sense to just have a `AddCerthashes(ma.Multiaddr) ma.Multiaddr` that adds current set of certhashes?

MarcoPolo: It would for the other thing, but not for this. It would need a reference to the
transport which would be annoying. This is a pure function that can be used in
the AddrsFactory with no other context.

options.go
MarcoPolo: I'll warn instead since it's unexpected.

options.go
MarcoPolo: reworded.

options.go
MarcoPolo: log.error instead. We shouldn't hit that case, but if we do it's good to highlight it.

options.go
```diff
@@ -262,7 +264,49 @@ func AddrsFactory(factory config.AddrsFactory) Option {
...
+			for i, a := range out {
+				// Any webtransport addresses that need us to fill in a certhash?
+				if ok, n := libp2pwebtransport.IsWebtransportMultiaddr(a); ok && n == 0 {
+					if multiaddrWithCerthash == nil {
+						for _, inMultiaddr := range in {
+							if ok, n := libp2pwebtransport.IsWebtransportMultiaddr(inMultiaddr); ok && n > 0 {
+								// This will give us the certhash. All certhashes across all listening addresses.
+								multiaddrWithCerthash = inMultiaddr
+							} else if ok && n == 0 {
+								// This is possible if we are listening with a valid static TLS cert.
```
marten-seemann, line 293:
```suggestion
								// This shouldn't be possible, since we don't support static TLS certs
```

marten-seemann (approved):
LGTM, modulo one suggestion for the comment.

options.go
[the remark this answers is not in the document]
aschmahmann: Lol no problem will keep not using it 😜. Was just flagging the code path here as exposed/not having receiving this fix. If you just want to leave it alone your call though.

PR #2239, opened 2023-04-05
correctly handle WebTransport addresses without certhashes

### LINKED ISSUE #2223: Cannot advertise WebTransport with externally known addresses (external static IP + port forwarding, dns, etc.)

### LINKED ISSUE #2233: Cannot advertise WebTransport with public IP addresses discovered by identify (e.g. port forwarding behind a NAT with a dynamic public IP)
I haven't repro'd this myself yet, but we should investigate and add a test case for this.

MarcoPolo:
Fixes #2233. Closes #2227. Fixes #2223.

### DIFF
--- a/options.go
+++ b/options.go
@@ -262,7 +262,6 @@ func AddrsFactory(factory config.AddrsFactory) Option {
 		if cfg.AddrsFactory != nil {
 			return fmt.Errorf("cannot specify multiple address factories")
 		}
-
 		cfg.AddrsFactory = factory
 		return nil
 	}
--- a/options.go
+++ b/options.go
@@ -271,13 +271,11 @@ func AddrsFactory(factory config.AddrsFactory) Option {
 		var log = logging.Logger("addrs-factory")
 
 		cfg.AddrsFactory = func(in []ma.Multiaddr) []ma.Multiaddr {
-			// This wrapper does the following:
-			// - If the user returns a webtransport multiaddr that doesn't have
-			//   a certhash, and all our known addrs have a certhash, copy the
-			//   certhash over.
-			//   - If we only have a webtransport multiaddr without a certhash
-			//     (client passed a static tls config), don't copy.
-			//   - If we have both, log a warning and don't copy the certhash.
+			// This wrapper checks fills in any missing webtransport certhashes
+			// in user provided webtransport addresses. It does this by getting
+			// the certhashes from the input multiaddrs. It's an error if we get
+			// see a webtransport multiaddr without a certhash since we don't
+			// support static TLS configs currently.
 
 			out := factory(in)
 			var multiaddrWithCerthash ma.Multiaddr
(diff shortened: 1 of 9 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -760,7 +763,48 @@ func (h *BasicHost) ConnManager() connmgr.ConnManager {
...
+		}
+	}
+	return addrs
+}
+
+// NormalizeMultiaddr returns a multiaddr suitable for equality checks.
+// If the multiaddr is a webtransport component, it removes the certhashes.
+func (h *BasicHost) NormalizeMultiaddr(addr ma.Multiaddr) ma.Multiaddr {
+	if ok, n := libp2pwebtransport.IsWebtransportMultiaddr(addr); ok && n > 0 {
+		var firstCerthash ma.Multiaddr
```
MarcoPolo, line 797: I don't like this. @marten-seemann If you have a better idea here, I'm all ears.

options.go
```diff
@@ -262,6 +262,7 @@ func AddrsFactory(factory config.AddrsFactory) Option {
 		if cfg.AddrsFactory != nil {
 			return fmt.Errorf("cannot specify multiple address factories")
 		}
+
```
marten-seemann, line 265:
```suggestion

```

p2p/host/basic/basic_host.go
MarcoPolo: nvm, I forgot about SplitLast.

p2p/host/basic/basic_host.go
```diff
@@ -760,7 +763,48 @@ func (h *BasicHost) ConnManager() connmgr.ConnManager {
...
+			}
+			addrs[i] = tpt.AddCertHashes(addr)
+		}
+	}
+	return addrs
+}
+
+// NormalizeMultiaddr returns a multiaddr suitable for equality checks.
+// If the multiaddr is a webtransport component, it removes the certhashes.
+func (h *BasicHost) NormalizeMultiaddr(addr ma.Multiaddr) ma.Multiaddr {
```
marten-seemann, line 794: I assume this would move into the address pipeline, once we have it, right?

p2p/host/basic/basic_host.go
marten-seemann: `ma.SplitFunc`?

p2p/host/basic/basic_host.go
```diff
@@ -760,7 +762,44 @@ func (h *BasicHost) ConnManager() connmgr.ConnManager {
 // Addrs returns listening addresses that are safe to announce to the network.
 // The output is the same as AllAddrs, but processed by AddrsFactory.
 func (h *BasicHost) Addrs() []ma.Multiaddr {
-	return h.AddrsFactory(h.AllAddrs())
+	type transportForListeninger interface {
```
marten-seemann, line 765: Add a comment that this is a hack, linking to the respective issue?

marten-seemann (approved):

p2p/host/basic/basic_host.go
MarcoPolo: Sure! Anywhere that isn't adding another method to BasicHost

marten-seemann (approved):

PR #2245, opened 2023-04-06
autonat: fix flaky TestAutoNATDialRefused

sukunrt:


marten-seemann (commented):
Can we just shut down the host / remove the stream handler, so it doesn't get called again?

sukunrt:
That won't solve it. 

Here we wait for 3 seconds to wait for a dialback request to finish. When ```require.Never``` exits some probes may still be in progress. When we close the host it resets all the open streams and that leads to the stream reset error [here](https://github.com/libp2p/go-libp2p/blob/8e1f7b13e1681daf93cb94f61f6ba90c011c2539/p2p/host/autonat/autonat_test.go#L64) 

Closing host or removing the streamhandler will not fix this. 

The cleaner solution is to fix ```autonat.Close``` to clean up outstanding probes too. I'm not sure if that's the right thing since a dialback request has no timeout so we might have to wait a while.

marten-seemann (approved):

sukunrt:
correction. the timeout for autonat dialback is 1 minute.

PR #2247, opened 2023-04-08
swarm: fix multiaddr comparison in ListenClose

marten-seemann:
Multiaddresses need to be compared with `Equal`, not with `==`, since `ma.StringCast(addr) != ma.StringCast(addr)`, but `ma.StringCast(addr).Equal(ma.StringCast(addr))`.


sukunrt (approved):
Nice catch!

Issue #2229, opened 2023-03-27
introduce a proper address pipeline


marten-seemann:
We need to get a lot smarter about address management. Wrapping / replacing addresses via the `AddrsFactory` doesn't scale.

I'm imagining a new service that's responsible for the entire process of address management. 

1. It ingests in new "address candidates" from various sources: user provided address hints, transports claiming to listen on certain addresses, Identify, relay service etc.
2. It verifies these addresses using AutoNAT v2.
3. It groups and categorizes these addresses (public / private / relayed).
4. It provides interfaces to make these different categories of addresses available to consumers.

sukunrt:
We need something like #2293 to test this.

sukunrt:
The flow here is:
1. Get listen addresses from the swarm. 
2. Different modules take these listen addresses and provide an estimate of the external address corresponding to these listen addrs.
 
So we have an object that's something like this:
```
type AddrDetail struct {
    ListenAddr ma.Multiaddr // Swarm listen addr
    InterfaceAddr ma.Multiaddr // The interface for this address
    ExternalAddr ma.Multiaddr // The public facing address for this address
    Sources map[Source]struct{} // sources are machine interface, Identify, UserProvided(addr factory?), NAT Port Map, Autorelay
}
```

We then test these via autonat v2 and send reachability events.

sukunrt:
https://github.com/libp2p/go-libp2p/pull/3075/files

Refactors the existing logic enough to make it readable.

PR #2248, opened 2023-04-08
move NAT mapping logic out of the host, add tests for NAT handling

marten-seemann:
In preparation for #2229. 

There's a few packages involved in NAT handling here:
1. [go-nat](https://github.com/libp2p/go-nat): Wraps UPnP and PMP libraries and provides functions to add, query and remove NAT mappings.
2. `p2p/net/nat`: Uses `go-nat` periodically renew NAT mappings and query for IP address changes.
3. `p2p/host/basic/natmgr.go`: Listens for `Listen` / `ListenClose` notifications from the swarm and tries to obtain NAT mappings for our listen addresses. Provides a `GetMapping(ma.Multiaddr) ma.Multiaddr` method that applies NAT mappings to existing multiaddresses.


p2p/host/basic/basic_host.go
```diff
@@ -849,99 +848,19 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
...
-			naddr, err := manet.ToNetAddr(transport)
-			if err != nil {
-				log.Error("error parsing net multiaddr %q: %s", transport, err)
+			extMaddr := h.natmgr.GetMapping(listen)
+			if extMaddr == nil {
+				// not mapped
 				continue
 			}
 
-			var (
```
MarcoPolo, line 904: 🙏 . Thanks for getting rid of this

p2p/host/basic/mockgen_private.sh
```diff
@@ -0,0 +1,49 @@
...
+# These interfaces might be used as embedded interfaces,
+# so we need to pass them to mockgen as aux_files.
+AUX=()
+for f in *.go; do
+  if [[ -z ${f##*_test.go} ]]; then
+    # skip test files
+    continue;
+  fi
+  if $(egrep -qe "type (.*) interface" $f); then
+    AUX+=("github.com/quic-go/quic-go=$f")
```
MarcoPolo, line 20: Do we need to change this?

p2p/host/basic/mockgen_private.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+DEST=$2
```
MarcoPolo, line 3: Did you copy this from quic-go? I'm curious how deeply this needs to be reviewed

MarcoPolo (commented):
Couple comments, but looks good. Thanks for the tests!

p2p/host/basic/mockgen_private.sh
marten-seemann: I did, and it's super ugly. I figured out a better solution though (thanks ChatGPT): We can just export the type in a file that has a `gomock` build flag, and then run `mockgen` with the `-build_flags="-tags=gomock"` command line parameter.
That way, `gomock` will think that it's an exported type, whereas the final build (and, very importantly, your IDE) will not.

marten-seemann: In fact, I should do the same in quic-go. This bash script has been annoying me for a long time!

p2p/net/nat/nat.go
```diff
@@ -40,7 +53,20 @@ func DiscoverNAT(ctx context.Context) (*NAT, error) {
...
+		nat:       natInstance,
+		extAddr:   extAddr,
+		mappings:  make(map[entry]int),
+		ctx:       ctx,
+		ctxCancel: cancel,
+	}
+	nat.refCount.Add(1)
+	go func() {
+		defer nat.refCount.Done()
+		nat.background()
```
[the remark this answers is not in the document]
marten-seemann: I fixed the double decrement.

p2p/host/basic/mockgen_private.sh
marten-seemann: Done! It's so much nicer now!

MarcoPolo (approved):
Thanks for getting rid of the bash script :)

Issue #2235, opened 2023-04-03
webtransport: put custom WebTransport code in the (basic)host and observed address manager


p-shahi:
We could do the most minimal thing here, which is a bit hacky:
                 1. Add a helper func `RemoveCerthashes`.
                 2. ObservedAddrManager Calls `RemoveCerthashes` before comparing things.
                 3. basic_host will copy cerrthashes to webtransport addrs before returning multiaddrs from `Addrs`. This way conusmers of an addr get the correct and up to date webtransport addrs.
              It's hacky because 1 & 3 involve these components having special knowledge of the whole system. But it would probably be the fastest and smallest short term fix. And we could write a test for the behavior to protect against a future regression.

_Originally posted by @MarcoPolo in https://github.com/libp2p/go-libp2p/issues/2233#issuecomment-1491238670_

PR #2250, opened 2023-04-10
identify: fix normalization of interface listen addresses

MarcoPolo:
The previous fix in #2235 was so close, but didn't catch this bug since ListenAddrs == InterfaceListeneAddrs in the [mocknet](https://github.com/libp2p/go-libp2p/blob/master/p2p/net/mock/mock_peernet.go#L307).

### DIFF
--- a/p2p/protocol/identify/obsaddr.go
+++ b/p2p/protocol/identify/obsaddr.go
@@ -360,26 +360,39 @@ type normalizeMultiaddrer interface {
 	NormalizeMultiaddr(addr ma.Multiaddr) ma.Multiaddr
 }
 
-func (oas *ObservedAddrManager) maybeRecordObservation(conn network.Conn, observed ma.Multiaddr) {
+type addrsProvider interface {
+	Addrs() []ma.Multiaddr
+}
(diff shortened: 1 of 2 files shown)


marten-seemann:
What’s the issue that this PR intends to solve?

p2p/protocol/identify/obsaddr.go
```diff
@@ -360,26 +360,39 @@ type normalizeMultiaddrer interface {
...
+}
+type listenAddrsProvider interface {
+	ListenAddresses() []ma.Multiaddr
+	InterfaceListenAddresses() ([]ma.Multiaddr, error)
+}
+
+type connMultiaddrProvider interface {
+	RemoteMultiaddr() ma.Multiaddr
+	LocalMultiaddr() ma.Multiaddr
+}
```
marten-seemann, line 374: Reuse this interface? https://github.com/libp2p/go-libp2p/blob/952f7cbc5cf5d3e16d9171f1151b20c2a227d484/core/network/conn.go#L70-L80

marten-seemann (approved):

PR #2251, opened 2023-04-11
Infer public webtransport addrs from quic-v1 addrs.

MarcoPolo:
[edited]

Adds a pure function to infer public webtransport addrs from quic addrs if we see that we are using the same port for both quic and webtransport addrs.

For example say we know we are listening on:

- `/ip4/0.0.0.0/udp/1234/quic-v1`
- `/ip4/0.0.0.0/udp/1234/quic-v1/webtransport`

And we observe that our public IP addresses include
- `/ip4/2.3.4.5/udp/1234/quic-v1`

We see that we're listening on the same address for quic-v1 and webtransport, so we should be able to say pretty confidently that we are also reachable via `/ip4/2.3.4.5/udp/1234/quic-v1/webtransport`.

That's what this PR does.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -761,15 +761,15 @@ func (h *BasicHost) ConnManager() connmgr.ConnManager {
 	return h.cmgr
 }
 
-// This is a temporary workaround/hack that fixes #2233. Once we have a
-// proper address pipeline, rework this. See the issue for more context.
-type transportForListeninger interface {
-	TransportForListening(a ma.Multiaddr) transport.Transport
-}
-
 // Addrs returns listening addresses that are safe to announce to the network.
 // The output is the same as AllAddrs, but processed by AddrsFactory.
 func (h *BasicHost) Addrs() []ma.Multiaddr {
+	// This is a temporary workaround/hack that fixes #2233. Once we have a
+	// proper address pipeline, rework this. See the issue for more context.
+	type transportForListeninger interface {
+		TransportForListening(a ma.Multiaddr) transport.Transport
+	}
+
 	type addCertHasher interface {
 		AddCertHashes(m ma.Multiaddr) ma.Multiaddr
 	}
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -1029,15 +1029,10 @@ func inferWebtransportAddrsFromQuic(in []ma.Multiaddr) []ma.Multiaddr {
 			quicAddrCount++
 		}
 	}
-	seenAddr := make(map[string]struct{}, len(in))
 	quicOrWebtransportAddrs := make(map[string]struct{}, quicAddrCount)
 	webtransportAddrs := make(map[string]struct{}, quicAddrCount)
 	foundSameListeningAddr := false
 	for _, addr := range in {
-		if _, ok := seenAddr[addr.String()]; ok {
-			continue
-		}
-		seenAddr[addr.String()] = struct{}{}
 		isWebtransport, numCertHashes := libp2pwebtransport.IsWebtransportMultiaddr(addr)
 		if isWebtransport {
 			for i := 0; i < numCertHashes; i++ {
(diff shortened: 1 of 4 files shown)


marten-seemann (changes requested):
Strongly opposed to this PR. While I appreciate the problem it’s trying to solve, I don’t think we should solve it in this way.

I’ve started early work on the address pipeline, and part of that is moving cruft _out_ of the basic host (speaking of, can I get a review on https://github.com/libp2p/go-libp2p/pull/2248 please?). This PR adds more technical debt to the basic host, and will make implementing the address pipeline harder.

MarcoPolo:
Summarizing sync convo:

We both agreed the previous approach was too hacky. We also agreed there was value in having this fix out so that we can start seeing webtransport addrs in the wild. We came up with an alternate approach that makes this a pure function that is pretty well contained. When we fix this with the address pipeline, users won't notice the difference. Things will just keep working.

I've updated the PR to use the less hacky approach, and added some table tests. I've also verified this surfaces my public webtransport addrs on Kubo.

p2p/host/basic/basic_host.go
```diff
@@ -997,10 +998,83 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
...
 	return dedupAddrs(finalAddrs)
 }
 
+var wtComponent = ma.StringCast("/webtransport")
+
+// inferWebtransportAddrsFromQuic infers more webtransport addresses from QUIC addresses.
+// This is useful when we discover our public QUIC address, but haven't discovered our public WebTransport addrs.
+// If we see that we are listening on the same port for QUIC and WebTransport,
+// we can be pretty sure that the WebTransport addr will be reachable if the
+// QUIC one is.
```
marten-seemann, line 1012: Great explanation, I wish every function we have was that well documented!

p2p/host/basic/basic_host.go
```diff
@@ -997,10 +998,83 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
...
+func inferWebtransportAddrsFromQuic(in []ma.Multiaddr) []ma.Multiaddr {
+	// We need to check if we are listening on the same ip+port for QUIC and WebTransport.
+	// If not, there's nothing to do since we can't infer anything.
+
+	// Count the number of QUIC addrs, this will let us allocate just once at the beginning.
```
marten-seemann, line 1017: Not sure if the plan to reduce allocs actually works out, `SplitLast` also allocates pretty heavily. No need to change it though, as this is temporary anyway.

p2p/host/basic/basic_host.go
```diff
@@ -997,10 +998,83 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
...
+				addr, _ = ma.SplitLast(addr)
+			}
+			webtransportAddrs[addr.String()] = struct{}{}
+			// Remove webtransport component, now it's a multiaddr that ends in /quic-v1
+			addr, _ = ma.SplitLast(addr)
+		}
+
+		if _, lastComponent := ma.SplitLast(addr); lastComponent.Protocol().Code == ma.P_QUIC_V1 {
+			addrStr := addr.String()
+			if _, ok := quicOrWebtransportAddrs[addrStr]; ok {
```
marten-seemann, line 1041: If I understand correctly, this will make the function misbehave if the input contains duplicates. We should probably also have a test case that covers that case.

p2p/host/basic/basic_host.go
```diff
@@ -997,10 +998,83 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
...
+	if len(webtransportAddrs) == 0 {
+		// No webtransport addresses, we aren't listening on any webtransport
+		// address, so we shouldn't add any.
+		return in
+	}
+
+	out := make([]ma.Multiaddr, 0, len(in)+(quicAddrCount-len(webtransportAddrs)))
+	for _, addr := range in {
+		// Add all the original addresses
+		out = append(out, addr)
```
marten-seemann, line 1062: Why not just append to `in`?

marten-seemann (commented):
I'm struggling a bit to wrap my head around the logic here. I think there's a bug here if the input contains duplicates.

By the way, the code would be sooo much cleaner with the new multiaddr API.

p2p/host/basic/basic_host.go
MarcoPolo: I didn't want to mutate the in list in case someone else kept a reference to it (A similar bug bit me with a related PR in this area of the code).

MarcoPolo: Also if we append to in, the function is no longer pure

p2p/host/basic/basic_host.go
MarcoPolo: I'll dedup

p2p/host/basic/basic_host.go
MarcoPolo: Does it? I thought it just pointed to slices?

p2p/host/basic/basic_host.go
marten-seemann: Resolved by running deduping on the input of this function.

p2p/host/basic/basic_host.go
marten-seemann: Kind of? The caller's slice wouldn't change if we're only appending, we'd only be modifying slice elements between `len` and `cap`. But I'm fine either way, this will heavily change with the new multiaddr design anyway.

marten-seemann (approved):

PR #2257, opened 2023-04-20
swarm: remove unused variable in addrDial

sukunrt:


marten-seemann (approved):
Haven't looked through the logic, but provided that CI is passing, this is fine.

Issue #1435, opened 2018-02-16
Use reuseport


Stebalien:

chaitanyaprem:
@marten-seemann , would like to pick this up as my first contribution for libp2p.

PR #2261, opened 2023-04-26
websocket: add reuseport support

chaitanyaprem:
Implemented reuse-port capability for websocket transport #1435  using the existing [net/reuseport](https://github.com/libp2p/go-libp2p/tree/master/p2p/net/reuseportt) transport and [libp2p/go-reuseport](https://github.com/libp2p/go-reuseport/tree/master).

In order to utilize REUSEPORT, reuseport option has to be passed while creating the websocket transport. 
This would allow for multiple threads/routines to listen on same IP/port combination. 

Included a test that would verify reuseport by running 2 go-routines that would listen on same address and accept connections from 4 different clients.

Reran the existing test-suite for the websocket package to ensure no other tests fail.

Note: Tested the new changes in ubuntu-20.04 and windows-11 environment.


p2p/transport/websocket/websocket.go
```diff
@@ -188,7 +200,14 @@ func (t *WebsocketTransport) maDial(ctx context.Context, raddr ma.Multiaddr) (ma
 
 	transport := &http.Transport{
 		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
-			conn, err := net.Dial(network, addr)
+			var conn manet.Conn
+			var err error
+			if t.UseReuseport() {
+				conn, err = t.reuse.Dial(raddr)
```
marten-seemann, line 206: Should this use the `DialContext` function?

p2p/transport/websocket/websocket.go
```diff
@@ -188,7 +200,14 @@ func (t *WebsocketTransport) maDial(ctx context.Context, raddr ma.Multiaddr) (ma
...
+			} else {
+				var d manet.Dialer
+				conn, err = d.Dial(raddr)
```
marten-seemann, line 209: Same here? By the way, what's the advantage of replacing `net.Dial` here? Not saying this is wrong, just curious.

p2p/transport/websocket/websocket_test.go
```diff
@@ -548,3 +549,64 @@ func TestResolveMultiaddr(t *testing.T) {
...
+				buf := make([]byte, 6)
+				n, err := conn.Read(buf)
+				if n != 6 {
+					t.Errorf("read %d bytes, expected 2", n)
+				}
+				require.NoError(t, err)
+			}
+		}(i)
+	}
+	time.Sleep(2 * time.Second)
```
marten-seemann, line 588: Can we test this without sleeping for a long time?

marten-seemann (commented):
Thank you for your PR! This mostly looks good, just a few comments.

p2p/transport/websocket/websocket.go
chaitanyaprem: Sure, will change to DialContext. 
No functional advantage. Since there is already a wrapper for the dialer that uses multiaddr, wanted to use the same. Only minor advantage of use this wrapped dialer is in future if dialer is enhanced to add more capabilities, usage need not be changed here to take advatange of it.

p2p/transport/websocket/websocket.go
chaitanyaprem: Right, that makes more sense..will update it.

p2p/transport/websocket/websocket_test.go
chaitanyaprem: Took a shortcut here :). Will update the test script without sleep.

chaitanyaprem:
> Thank you for your PR! This mostly looks good, just a few comments.

You are welcome. Pushed new commits addressing your review comments.

p2p/transport/websocket/websocket_test.go
```diff
@@ -548,3 +548,71 @@ func TestResolveMultiaddr(t *testing.T) {
...
+	for i := 0; i < 4; i++ {
+		go func(index int, ch chan int) {
+			c, err := tpt.maDial(context.Background(), laddr)
+			if err != nil {
+				t.Error(err)
+				return
+			}
+			require.NoError(t, err)
+			defer c.Close()
+			msg := fmt.Sprintf("Hello%d", i)
```
MarcoPolo, line 604: you probably want `index` here. Thanks linter!

chaitanyaprem: Ah..my bad. Pushing an update.

chaitanyaprem:
@marten-seemann please approve the workflows post which this can be merged.

p2p/transport/websocket/websocket_test.go
```diff
@@ -548,3 +548,71 @@ func TestResolveMultiaddr(t *testing.T) {
 		})
 	}
 }
+
+func TestListenerResusePort(t *testing.T) {
+	laddr := ma.StringCast("/ip4/127.0.0.1/tcp/5002/ws")
```
marten-seemann, line 553: Can you listen on an unspecified port instead?
```go
ma.StringCast("/ip4/127.0.0.1/tcp/0/ws")
```

That way, this test won't fail it some other application is listening on port 5002.

p2p/transport/websocket/websocket_test.go
```diff
@@ -548,3 +548,71 @@ func TestResolveMultiaddr(t *testing.T) {
...
+	var opts []Option
+	opts = append(opts, EnableReuseport())
+	_, u := newUpgrader(t)
+	tpt, err := New(u, &network.NullResourceManager{}, opts...)
+	require.NoError(t, err)
+	c := make(chan int)
+
+	for i := 0; i < 2; i++ {
+		go func(index int, ch chan int) {
+			l, err := tpt.maListen(laddr)
```
marten-seemann, line 564: I don't understand this test. You're opening two listeners on the same local address. How do you now make sure that each of them receives exactly 4 incoming connections.

marten-seemann (commented):
Thank you @chaitanyaprem! The code looks good to me, but I'm a bit confused with the test case.

Unfortunately, we just switched back to the Gorilla WebSocket library (https://github.com/libp2p/go-libp2p/pull/2280). This is creating merge conflicts with this PR. Would you mind rebasing this PR on top of the current master?

p2p/transport/websocket/websocket_test.go
chaitanyaprem: That is a challenge, because whenever i ran the test..it is not gauranteed to distribute connections equally..I have noticed 1,3 or 2,2 happen. Hence i did not add an additional check for validating exactly equal connections in the test. Should've added comments in the test regarding this to make it more clear.

p2p/transport/websocket/websocket_test.go
chaitanyaprem: Sure, will make this change.

chaitanyaprem:
> Thank you @chaitanyaprem! The code looks good to me, but I'm a bit confused with the test case.
> 
> Unfortunately, we just switched back to the Gorilla WebSocket library (#2280). This is creating merge conflicts with this PR. Would you mind rebasing this PR on top of the current master?

Will resolve and push updated ones.

chaitanyaprem:
Pushed updated changes as per review comments.

p2p/transport/websocket/websocket_test.go
```diff
@@ -548,3 +549,122 @@ func TestResolveMultiaddr(t *testing.T) {
 		})
 	}
 }
+
+func startListeners(tpt *WebsocketTransport) (ma.Multiaddr, []*manet.Listener, error) {
```
Maelkum, line 553: Perhaps this function can be a `t.Helper()` and simplify some of the error handling?

Also, what is your opinion to using `fmt.Print*` family of functions in tests? I myself rarely use them, and use `t.Logf`. Those only get printed if the test fails or if the `verbose` output is enabled.

p2p/transport/websocket/websocket_test.go
```diff
@@ -548,3 +549,122 @@ func TestResolveMultiaddr(t *testing.T) {
...
+	noOfClientConns := 4
+	var opts []Option
+	opts = append(opts, EnableReuseport())
+	_, u := newUpgrader(t)
+	tpt, err := New(u, &network.NullResourceManager{}, opts...)
+	require.NoError(t, err)
+	c := make(chan int, noOfClientConns)
+
+	raddr, listeners, err := startListeners(tpt)
+	//fmt.Println("Started Listeners")
```
Maelkum, line 587: Perhaps these commented lines of code could be removed..?

p2p/transport/websocket/websocket_test.go
```diff
@@ -548,3 +549,122 @@ func TestResolveMultiaddr(t *testing.T) {
...
+				By contrast, the SO_REUSEPORT implementation distributes connections evenly across all of the threads (or processes)
+			 	that are blocked in accept() on the same port. */
+			for j := 0; j < noOfClientConns; j++ {
+				//j := 0
+				conn, err := l.Accept()
+				if err != nil {
+					fmt.Println("Server Routine-", index, " Failed accepting connection ", j, " due to error ", err)
+					return
+				}
+				require.NoError(t, err)
```
Maelkum, line 606: Perhaps use `require.NoErrorf` and remove the `if` statement?

p2p/transport/websocket/websocket_test.go
```diff
@@ -548,3 +549,122 @@ func TestResolveMultiaddr(t *testing.T) {
...
+				//fmt.Println("Server Routine-", index, " accepting connection-", j)
+				defer conn.Close()
+				buf := make([]byte, 6)
+				n, err := conn.Read(buf)
+				if n != 6 {
+					t.Errorf("read %d bytes, expected 6", n)
+				}
+				require.NoError(t, err)
```
Maelkum, line 614: What do you think about something like this? Same goes to some of the other cases too, e.g. `conn.Write` and others.

```suggestion
				n, err := conn.Read(buf)
				require.NoError(t, err)
				require.Equal(t, 6, n)
```

Maelkum (commented):
Nice PR! (commenting as an outsider)

Q: This PR adds the ability to _listen_ on the same port by multiple parties. The TCP implementation for `reuseport` will also try to use _listening_ ports also for _dialing_. This is not handled in this PR, is that right?

p2p/transport/websocket/websocket_test.go
chaitanyaprem: Thanks, was not aware of this. Will make changes accordingly. 
I think t.Logf makes more sense in comparison to fmt.Print*. 
I started using fmt functions for debugging and then continued with them as is.

chaitanyaprem:
> Nice PR! (commenting as an outsider)
> 
> Q: This PR adds the ability to _listen_ on the same port by multiple parties. The TCP implementation for `reuseport` will also try to use _listening_ ports also for _dialing_. This is not handled in this PR, is that right?

Thanks @Maelkum for your comments. Addressed all the others.
You are right that `reuseport` is not added for dialing as it doesn't seem straight-forward with gorilla websocket.
Will see how to get around that.

chaitanyaprem:
Wow, this is surprising that load balancing across multiple listeners is not functioning on windows as well as Mac even-though 2 sockets are allowed to listen on same address+port. 

Wrt Windows, i found this [ref](https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse?redirectedfrom=MSDN) which does not guarantee the behaviour of SO_REUSEADDR. 
Quote from windows ref is not very clear wrt load balancing. 
`
For example, if all of the sockets on the same port provide TCP service, any incoming TCP connection requests over the port cannot be guaranteed to be handled by the correct socket — the behavior is non-deterministic.`

Has the TCP reuse port been tested for load balancing? 

_Note: While going through the web searching for expected behaviour for these socket options, noticed that for FreeBSD based systems in order for load balancing to be supported an additional option has to be set SO_REUSEPORT_LB, [Ref](https://reviews.freebsd.org/D11003). This is something to be explored and added to the reuseport package_ 

Will update once i investigate further.

chaitanyaprem:
Update on further investigation, that macOS also doesn't support load balancing across listeners.
Few refs: 
  [Ref1](https://stackoverflow.com/questions/41247790/so-reuseport-on-macos-with-libuv)
  [Ref2](https://github.com/uNetworking/uWebSockets/issues/1194)

So, for now I have skipped checking for load balancing of connections for any platform other than linux in the test script. 
Shall also raise an improvement PR onto the go-reuseport repo in order to enable LB functionality for FREEBSD based systems by setting SO_REUSEPORT_LB as indicated above.

chaitanyaprem:
Pushed changes to reuse port in case of dailing. 
Note that in case of wss with SNI, reuse port cannot be used as the dail function has already been replaced in order to not resolve address and directly do a TCP Dial. 
@MarcoPolo , hope it is ok to leave it as is.

chaitanyaprem:
@marten-seemann  @MarcoPolo   All required changes and review comments have been addressed. 
Please go ahead and merge this PR if nothing else is required.

marten-seemann:
@chaitanyaprem The WebSocket tests are failing on CI. Can you please take a look?

chaitanyaprem:
Surprisingly the test is successful now, i ran it on a local ubuntu vm (Ubuntu 22.04.3 LTS) as well and there also tests are successful.

It is hard to determine whether some other change fixed the issue or is the test just flaky.

marten-seemann:
I reran the test suite and it's failing again. We can't introduce any flaky tests, please make sure it consistently passes (you can run the test suite with `-count=1000` locally).

chaitanyaprem:
> I reran the test suite and it's failing again. We can't introduce any flaky tests, please make sure it consistently passes (you can run the test suite with `-count=1000` locally).

Ran it 1000 times and it fails sometimes. 
Commenting the additional load balancing check for now. Looks like linux is doesn't gaurantee it.

p-shahi:
@sukunrt is gonna help take this over the finish line in time for the v0.31 release @chaitanyaprem

p2p/transport/websocket/websocket_test.go
```diff
@@ -548,3 +549,101 @@ func TestResolveMultiaddr(t *testing.T) {
...
+			 Refer https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse?redirectedfrom=MSDN for windows
+			 References for MACOS
+				Ref - (https://stackoverflow.com/questions/41247790/so-reuseport-on-macos-with-libuv)
+				Ref -(https://github.com/uNetworking/uWebSockets/issues/1194)
+		if runtime.GOOS == "linux" {
+			for i := 0; i < 2; i++ {
+				/*Not checking for equal distribution of connections due to above explanation.*/
+	/*require.NotEqualf(t, 0, connsHandled[i], "No connections handled by listener %d", i)
+			t.Logf("Listener %d handled %d connections.", i, connsHandled[i])
+		}
```
sukunrt, line 646: This test is fine. The distribution will never be exactly equal, the allocation is based on the hash of `clientIP:Port` so it's not round robin. You can make 20 connections and assert both the listeners get at least 1.

p2p/transport/websocket/listener.go
```diff
@@ -54,11 +56,20 @@ func newListener(a ma.Multiaddr, tlsConf *tls.Config) (*listener, error) {
 	if err != nil {
 		return nil, err
 	}
-	nl, err := net.Listen(lnet, lnaddr)
-	if err != nil {
-		return nil, err
+	if reuseportAvailable {
+		nl, err = reuseport.Listen(lnet, lnaddr)
+		if err != nil {
```
sukunrt, line 61: We want to use the transport's `reuseTransport` field. Otherwise this will never reuse the port we're listening on when dialing. I guess you can pass them in to `newListener`. 

Also add a test that we are reusing port when dialing.

p2p/transport/websocket/websocket.go
```diff
@@ -187,7 +198,13 @@ func (t *WebsocketTransport) maDial(ctx context.Context, raddr ma.Multiaddr) (ma
...
+
 	dialer := ws.Dialer{HandshakeTimeout: 30 * time.Second}
+	if t.UseReuseport() {
+		dialer.NetDial = func(network, address string) (net.Conn, error) {
+			return t.reuseTransport.DialContext(ctx, raddr)
+		}
+	}
 	if isWss {
 		sni := ""
 		sni, err = raddr.ValueForProtocol(ma.P_SNI)
```
sukunrt, line 210: This will also need to have the reuse logic: https://github.com/libp2p/go-libp2p/blob/master/p2p/transport/websocket/websocket.go#L205

sukunrt (changes requested):
I have a couple of other nits, but let's get the dial side to reuse the port we are listening on first.

p2p/transport/websocket/websocket_test.go
sukunrt: On this test, can we use a separate transport for dialing? Calling dial to connect to a listener on the same transport is uncommon.

chaitanyaprem: > This test is fine. The distribution will never be exactly equal, the allocation is based on the hash of `clientIP:Port` so it's not round robin. You can make 20 connections and assert both the listeners get at least 1.

That is what i initially thought, but if you look at the CI tests no distribution is happening sometimes which is making the test flaky. Hence i had commented it out to avoid failure of the test in linux. e.g Refer here https://github.com/libp2p/go-libp2p/actions/runs/5885440443/job/15969152974

sukunrt: I think that's because all the ports selected for dialing are hashing to the same slot. If we increase the number of ports(clients in this case) this should go away.

sukunrt:
@chaitanyaprem just checking in. Let me know if you have any questions here. I'm available in filecoin slack(link in repo README).

chaitanyaprem:
> @chaitanyaprem just checking in. Let me know if you have any questions here. I'm available in filecoin slack(link in repo README).

Got little busy with other work and hence not able to take a look at this. 
Thanks for providing a channel to reach out, will do once i start working on this again.

sukunrt:
Thanks for the update @chaitanyaprem

Maelkum:
Hi guys! I hope this is not me being rude.. At a similar time when this PR was opened, I played around with implementing a similar thing. I rebased those changes to sync with the latest master (at the time this repo used nhooyr.io/websocket). 

Tested it mostly in the listen => dial scenario, compared to multiple listens.

Since this PR is open for a while now, I figured I'd throw the link here. Feel free to cherry pick any piece of code out of there if it's useful. Hope this isn't bad manners!

https://github.com/Maelkum/go-libp2p/pull/1

sukunrt:
Thanks @Maelkum. This PR has been dormant for a while so I think it's okay. Do you want to open a new PR or take over this one? 
@chaitanyaprem is it fine if @Maelkum continues the work here?

chaitanyaprem:
> Thanks @Maelkum. This PR has been dormant for a while so I think it's okay. Do you want to open a new PR or take over this one? @chaitanyaprem is it fine if @Maelkum continues the work here?

Thanks @Maelkum 
Feel free to takeover, have not been able to get to this one myself.

Maelkum:
@sukunrt @chaitanyaprem 
Thanks guys! I'll open a new PR and link this one there, I think it will be tidier that way.

Issue #1534, opened 2019-08-06
quic: listen on connections that we dialed from before


marten-seemann:
See https://github.com/libp2p/go-libp2p-quic-transport/pull/63#issue-284337030 and https://github.com/libp2p/go-libp2p-quic-transport/pull/63#discussion_r310583866.

marten-seemann:
Fixed by #2262.

PR #2262, opened 2023-04-26
quic: prioritise listen connections for reuse

sukunrt:
Currently there is a race in kubo between the first dial made by a node on startup and starting to listen on a quic address. 

If the dial happens first, this places an entry in the ```reuse.global``` map. This entry will stay in the map till we get a period of 30 seconds where we make no new dials. In my experiments this entry is removed after 15 minutes of starting the node. This dial happens before listening 5-10% of the times on my machine. 

The impact of this is on holepunching. While this entry is in the map, the function ```reuse.Dial``` in ```p2p/transport/quicreuse/reuse.go``` can return either this connection that we dialed out of but are not listening on, or it can return the correct connection the one on which we are listening.

### DIFF
--- a/p2p/transport/quicreuse/reuse.go
+++ b/p2p/transport/quicreuse/reuse.go
@@ -75,13 +75,15 @@ type reuse struct {
 	routes  routing.Router
 	unicast map[string] /* IP.String() */ map[int] /* port */ *reuseConn
 	// global contains connections that are listening on 0.0.0.0 / ::
-	global map[int]*reuseConn
+	global     map[int]*reuseConn
+	globalDial map[int]*reuseConn
 }
 
 func newReuse() *reuse {
 	r := &reuse{
 		unicast:    make(map[string]map[int]*reuseConn),
 		global:     make(map[int]*reuseConn),
+		globalDial: make(map[int]*reuseConn),
 		closeChan:  make(chan struct{}),
 		gcStopChan: make(chan struct{}),
 	}
@@ -93,6 +95,9 @@ func (r *reuse) gc() {
 	defer func() {
 		r.mutex.Lock()
 		for _, conn := range r.global {
+			conn.Close()
+		}
+		for _, conn := range r.globalDial {
 			conn.Close()
 		}
 		for _, conns := range r.unicast {
@@ -117,6 +122,12 @@ func (r *reuse) gc() {
 				if conn.ShouldGarbageCollect(now) {
 					conn.Close()
 					delete(r.global, key)
+				}
+			}
+			for key, conn := range r.globalDial {
+				if conn.ShouldGarbageCollect(now) {
+					conn.Close()
+					delete(r.globalDial, key)
 				}
 			}
 			for ukey, conns := range r.unicast {
@@ -189,6 +200,11 @@ func (r *reuse) dialLocked(network string, source *net.IP) (*reuseConn, error) {
 		return conn, nil
 	}
 
+	// Use a connection we've previously dialed from
+	for _, conn := range r.globalDial {
+		return conn, nil
+	}
+
 	// We don't have a connection that we can use for dialing.
 	// Dial a new connection from a random port.
 	var addr *net.UDPAddr
@@ -203,29 +219,43 @@ func (r *reuse) dialLocked(network string, source *net.IP) (*reuseConn, error) {
 		return nil, err
 	}
 	rconn := newReuseConn(conn)
-	r.global[conn.LocalAddr().(*net.UDPAddr).Port] = rconn
+	r.globalDial[conn.LocalAddr().(*net.UDPAddr).Port] = rconn
 	return rconn, nil
 }
 
 func (r *reuse) Listen(network string, laddr *net.UDPAddr) (*reuseConn, error) {
-	conn, err := net.ListenUDP(network, laddr)
-	if err != nil {
-		return nil, err
-	}
-	localAddr := conn.LocalAddr().(*net.UDPAddr)
-
-	rconn := newReuseConn(conn)
-	rconn.IncreaseCount()
-
 	r.mutex.Lock()
 	defer r.mutex.Unlock()
+
+	var rconn *reuseConn
+	var localAddr *net.UDPAddr
+
+	// reuse the connection if we've dialed out of this port already
+	if laddr.IP.IsUnspecified() {
+		if _, ok := r.globalDial[laddr.Port]; ok {
+			rconn = r.globalDial[laddr.Port]
+			localAddr = rconn.UDPConn.LocalAddr().(*net.UDPAddr)
+		}
+	}
+	if rconn == nil {
+		conn, err := net.ListenUDP(network, laddr)
+		if err != nil {
+			return nil, err
+		}
+		localAddr = conn.LocalAddr().(*net.UDPAddr)
+		rconn = newReuseConn(conn)
+	}
+
+	rconn.IncreaseCount()
 
 	// Deal with listen on a global address
 	if localAddr.IP.IsUnspecified() {
 		// The kernel already checked that the laddr is not already listen
 		// so we need not check here (when we create ListenUDP).
 		r.global[localAddr.Port] = rconn
-		return rconn, err
+		// delete the entry from dial map in case we are reusing this connection
+		delete(r.globalDial, localAddr.Port)
+		return rconn, nil
 	}
 
 	// Deal with listen on a unicast address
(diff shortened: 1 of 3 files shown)


marten-seemann:
Does this PR fix https://github.com/libp2p/go-libp2p/issues/1534?

marten-seemann (commented):
What happens in the following scenario?
1. We dial a QUIC connection to a remote node. We  don't have any existing connections, so we listen on :0, and get assinged `:port1`.
2. Now we start a listener on `0.0.0.0:port2`.

With this PR, we'd ignore `port2`, and just reuse the connection that's listening on `port1`, right?

sukunrt:
>We dial a QUIC connection to a remote node. We don't have any existing connections, so we listen on :0, and get assinged :port1.

The :port1 will go in to ```globalDial``` map

>Now we start a listener on 0.0.0.0:port2.

The :port2 will go in to ```global``` map


>With this PR, we'd ignore port2, and just reuse the connection that's listening on port1, right?

No, we will prioritise :port2 because it is in the ```global``` map which is checked before ```globalDial``` map 

I added a test case for this. [`8efdc6a` (#2262)](https://github.com/libp2p/go-libp2p/pull/2262/commits/8efdc6a48cfd71cd34fa30b643f65a42691fbdfb)

sukunrt:
>Does this PR fix https://github.com/libp2p/go-libp2p/issues/1534?

I think so. comments below

> Ideally, we'd mark this as a fallback connection so we can stop using it when we get a real global connection.
https://github.com/libp2p/go-libp2p-quic-transport/pull/63#discussion_r309489026

Yes this is exactly what i've implemented. The fallback map is ```globalDial```

>The listening part of this is still a bit racy, if the user decides to listen on an ephemeral port that was picked for a Dial call before. We should reuse connections when Listen is called after Dial (and the port happens to be the same) as well. Planning to fix this in a separate PR.

In the original description
https://github.com/libp2p/go-libp2p-quic-transport/pull/63#issue-284337030

This is fixed by this PR. First we check globalDial to see if there's a connection. If there is we reuse that for listening.

p2p/transport/quicreuse/reuse.go
```diff
@@ -76,14 +76,19 @@ type reuse struct {
 	unicast map[string] /* IP.String() */ map[int] /* port */ *reuseConn
 	// global contains connections that are listening on 0.0.0.0 / ::
 	global map[int]*reuseConn
```
marten-seemann, line 78: Should this be called `globalListeners`?
```suggestion
	// globalListeners contains all listeners that are listening on 0.0.0.0 / ::
	// When dialing new connection, we always prefer to dial from a connection we're listening on.
	globalListeners map[int]*reuseConn
```

p2p/transport/quicreuse/reuse.go
```diff
@@ -76,14 +76,19 @@ type reuse struct {
...
+	// globalFallback contains connections that we've dialed out from. connections
+	// are reused from this map if no connection is available in the global
+	// map. These connections are listening on 0.0.0.0 / ::
+	globalFallback map[int]*reuseConn
```
marten-seemann, line 82: Maybe better: `globalDialers`?

p2p/transport/quicreuse/reuse.go
```diff
@@ -203,29 +222,43 @@ func (r *reuse) dialLocked(network string, source *net.IP) (*reuseConn, error) {
...
-
 	// Deal with listen on a global address
 	if localAddr.IP.IsUnspecified() {
 		// The kernel already checked that the laddr is not already listen
 		// so we need not check here (when we create ListenUDP).
 		r.global[localAddr.Port] = rconn
-		return rconn, err
+		// delete the entry from fallback map in case we are reusing this connection
+		delete(r.globalFallback, localAddr.Port)
+		return rconn, nil
```
marten-seemann, line 261: Why doesn't this happen up in the `laddr.IP.IsUnspecified()` code block?

p2p/transport/quicreuse/reuse.go
```diff
@@ -203,29 +222,43 @@ func (r *reuse) dialLocked(network string, source *net.IP) (*reuseConn, error) {
...
-		return nil, err
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	var rconn *reuseConn
+	var localAddr *net.UDPAddr
+
+	// reuse the fallback connection if we've dialed out from this port already
+	if laddr.IP.IsUnspecified() {
+		if _, ok := r.globalFallback[laddr.Port]; ok {
```
marten-seemann, line 238: Do we need to handle listening on port 0?

marten-seemann (commented):
I've tried this out using the following code: https://gist.github.com/marten-seemann/c221eed357fc14e15f26d066380c5c44

It seems to work! Just left some comments regarding naming, and a question regarding listening on port 0.

p2p/transport/quicreuse/reuse.go
sukunrt: I don't think so. An entry in the map(listeners or dialers) will never have port 0 because we add the finally assigned listener port to the map and not 0. So if a listen call happens with port 0 and there is an entry in globalDialers map, we will not find a match and then net.ListenUDP will give use a new port which is not in the globalDialers map. 

Do you think this makes sense?

p2p/transport/quicreuse/reuse.go
sukunrt: Have used both your suggestions.

p2p/transport/quicreuse/reuse.go
marten-seemann: > So if a listen call happens with port 0 and there is an entry in globalDialers map, we will not find a match and then net.ListenUDP will give use a new port which is not in the globalDialers map.

That's what I meant. If the port is 0, the user doesn't care which port is assigned. So we might as well reuse a connection that we used before, right?

sukunrt: oh, that's a nice idea. will fix.

sukunrt: fixed this. updated the test `TestReuseConnectionWhenDialBeforeListen`

p2p/transport/quicreuse/reuse.go
```diff
@@ -74,16 +74,21 @@ type reuse struct {
 
 	routes  routing.Router
 	unicast map[string] /* IP.String() */ map[int] /* port */ *reuseConn
-	// global contains connections that are listening on 0.0.0.0 / ::
-	global map[int]*reuseConn
+	// globalListeners contains connections that are listening on 0.0.0.0 / ::
+	globalListeners map[int]*reuseConn
+	// globalDialers contains connections that we've dialed out from. These connections are listening on 0.0.0.0 / ::
+	// On Dial, connections are reused from this map if no connection is available in the globalListeners
+	// On Listen, connections are reused from this map if the requested port is 0
```
marten-seemann, line 81:
```suggestion
	// On Listen, connections are reused from this map if the requested port is 0, and then moved to globalListeners
```

p2p/transport/quicreuse/reuse.go
```diff
@@ -203,29 +222,50 @@ func (r *reuse) dialLocked(network string, source *net.IP) (*reuseConn, error) {
...
+	var localAddr *net.UDPAddr
+
+	// reuse the fallback connection if we've dialed out from this port already
+	if laddr.IP.IsUnspecified() {
+		if laddr.Port == 0 {
+			// use any connection we have dialed out of
+			for _, conn := range r.globalDialers {
+				rconn = conn
+				localAddr = rconn.UDPConn.LocalAddr().(*net.UDPAddr)
+				delete(r.globalDialers, localAddr.Port)
```
marten-seemann, line 243: Add a comment that it will be added to the `globalListeners` map below.

p2p/transport/quicreuse/reuse.go
```diff
@@ -203,29 +222,50 @@ func (r *reuse) dialLocked(network string, source *net.IP) (*reuseConn, error) {
...
+				break
+			}
+		} else if _, ok := r.globalDialers[laddr.Port]; ok {
+			rconn = r.globalDialers[laddr.Port]
+			localAddr = rconn.UDPConn.LocalAddr().(*net.UDPAddr)
+			delete(r.globalDialers, localAddr.Port)
```
marten-seemann, line 249: Here too.

marten-seemann (approved):
Logic LGTM. Just suggesting to add another comment, to make the code easier to understand when we come back to it later.

p2p/transport/quicreuse/reuse.go
sukunrt: does this look fine?

marten-seemann: Actually, I think it would be clearer if you added it to the `globalListeners` map here and returned early.

sukunrt: I agree. that's much better.

sukunrt: fixed.

PR #2271, opened 2023-05-05
webtransport: only add cert hashes if we already started listening

marten-seemann:


p2p/transport/webtransport/transport.go
```diff
@@ -405,9 +404,10 @@ func (t *transport) Resolve(_ context.Context, maddr ma.Multiaddr) ([]ma.Multiad
 	return []ma.Multiaddr{beforeQuicMA.Encapsulate(quicComponent).Encapsulate(sniComponent).Encapsulate(afterQuicMA)}, nil
 }
 
+// AddCertHashes adds the current certificate hashes to a multiaddress.
+// If called before Listen, it's a no-op.
```
aschmahmann, line 408: Wouldn't returning an error or boolean here be better? Then callers, like the BasicHost could emit errors if they wanted to so that users would know they've done something wrong.

Not a strongly held view here, just seems like someone calling this before the certmanager has been initialized probably did so erroneously and so they should be informed of their error.

p2p/host/basic/basic_host.go
```diff
@@ -792,7 +792,7 @@ func (h *BasicHost) Addrs() []ma.Multiaddr {
 			if !ok {
 				continue
 			}
-			addrs[i] = tpt.AddCertHashes(addr)
+			addrs[i], _ = tpt.AddCertHashes(addr)
```
aschmahmann, line 795: @MarcoPolo should we add a log error here if this comes back false?

marten-seemann (commented):
Thank you all!

unicomp21:
When running automated integration tests, is there a way to lower the udp buffer size @marten-seemann ?

PR #2272, opened 2023-05-05
swarm: add a scheduler for smart dialing

sukunrt:
This PR implements smart dialing using a separate scheduler to keep the changes to the worker loop minimal. 

I prefer the other smart-dialing PR. It is difficult to separate the scheduling logic and the worker loop because they both depends on a lot of shared context.


sukunrt:
merged these changes into #2260

Issue #2264, opened 2023-05-01
Log error in accept


MarcoPolo:
https://github.com/libp2p/go-libp2p/blob/8313917571fec7d26813ff437013fde6f48659f7/p2p/net/swarm/swarm_listen.go#L130

ignore swarm close errors.

nikumbgo12345:
@MarcoPolo i am new to libp2p and trying to contribute , not sure what you mean by the description , have created a PR could you please review and guide

marten-seemann:
@nikumbgo12345 This issue was already assigned to @MarcoPolo. Maybe pick one of the [help wanted](https://github.com/libp2p/go-libp2p/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) issues instead?

PR #2274, opened 2023-05-08
Log error in accept #2264

nikumbgo12345:
added log message in case err when accepting connections


marten-seemann (changes requested):
This is not correct. As it says in #2264, we need to ignore errors that come from the swarm being closed.

nikumbgo12345:
discarding PR , issue assigned to somebody else and missed the description to ignore swarm errors

PR #2276, opened 2023-05-08
quic virtual listener: don't panic when quic-go's accept call errors

### LINKED ISSUE #2259: quic: panic on error in listener Accept
originally reported here: 
https://github.com/ipfs/kubo/issues/9693

The sequence of events that causes this is:

```list.Accept``` errors here https://github.com/libp2p/go-libp2p/blob/v0.26.4/p2p/net/swarm/swarm_listen.go#L129

and error in list.Accept has closed the listener accept loop already here:
https://github.com/libp2p/go-libp2p/blob/v0.26.4/p2p/transport/quic/virtuallistener.go#L108-L120
with `sendErrAndClose`

The defer cleanup in swarm_listen closes ```virtuallistener``` here https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/swarm_listen.go#L117

which causes panic here: https://github.com/libp2p/go-libp2p/blob/v0.26.4/p2p/transport/quic/virtuallistener.go#L73
as it checks for entry from ```muxer``` map which was removed by `sendErrAndClose`

### PR DESCRIPTION
**Pull request by MarcoPolo**
closes #2259 

This doesn't add any extra logging. That's a bit bigger of a change so I'll do that in another PR.

### DIFF
--- a/p2p/transport/quic/virtuallistener.go
+++ b/p2p/transport/quic/virtuallistener.go
@@ -46,8 +46,9 @@ type acceptVal struct {
 type acceptLoopRunner struct {
 	acceptSem chan struct{}
 
-	muxerMu sync.Mutex
-	muxer   map[quic.VersionNumber]chan acceptVal
+	muxerMu     sync.Mutex
+	muxer       map[quic.VersionNumber]chan acceptVal
+	muxerClosed bool
 }
 
 func (r *acceptLoopRunner) AcceptForVersion(v quic.VersionNumber) chan acceptVal {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: p2p/transport/quic/virtuallistener.go [hunk 1]
```diff
@@ -68,6 +69,11 @@ func (r *acceptLoopRunner) RmAcceptForVersion(v quic.VersionNumber) {
 	r.muxerMu.Lock()
 	defer r.muxerMu.Unlock()
 
+	if r.muxerClosed {
+		// Already closed, all versions are removed
```
aschmahmann, line 73: Is this an error-specific closing that basically indicates a failed start? If so maybe document/name it like that.

If so, does this mean that `Accept` should be modified as well? Want to make sure we don't end up in a position where we can keep calling accept but not RmAccept.

marten-seemann (approved):

p2p/transport/quic/virtuallistener.go
marten-seemann: We see the situation that quic-go's `Accept` sometimes errors on Windows machines. It might or might not be related to https://github.com/quic-go/quic-go/issues/1737. I don't have a Windows machine, so I can't really debug this. Also, I care more about other OSes ;)

There should probably be some logging, but I'm fine adding this in a later PR.

Issue #2201, opened 2023-03-17
websocket: closing a connection may block for 5s


marten-seemann:
See https://github.com/nhooyr/websocket/issues/355 for details. This might block the connection manager loop and the `swarm.Close` for a long time, if we have a lot of WebSocket connections and this happens multiple times.
Not sure how realistic this is in practice, but I wouldn't want to risk it.

MarcoPolo:
Could we read the rest of the frame and discard before calling close as a temporary workaround?

PR #2280, opened 2023-05-09
websocket: switch back to the gorilla library

marten-seemann:
Turns out that @Jorropo was right in https://github.com/libp2p/go-libp2p/pull/1982#issuecomment-1373925002. Switching to [nyhoor's websocket library](https://github.com/nhooyr/websocket) has caused us nothing but pain (examples: #1982, https://github.com/libp2p/go-libp2p/issues/2201, and most recently, ugly workaround in the [gating tests](#2200)).

Although the author claims to fix critical issues, nothing is happening when actual issues are reported (examples: https://github.com/nhooyr/websocket/issues/367, https://github.com/nhooyr/websocket/issues/355 and https://github.com/nhooyr/websocket/issues/384), and there hasn't even been a patch release April 2021).

This PR reverts the changes made to the WebSocket transport in reverse order they were made:
* #2199 
* #2193 
* #1982

After that, it reapplies the one fix that was unrelated to the library used: #2199.


p2p/transport/websocket/conn.go
```diff
@@ -1,44 +1,156 @@
...
-	return c.remoteAddr
+func (c *Conn) LocalAddr() net.Addr {
+	return NewAddrWithScheme(c.Conn.LocalAddr().String(), c.secure)
 }
 
-func (c conn) Read(b []byte) (int, error) {
-	n, err := c.Conn.Read(b)
-	if err == nil && n == 0 && c.readAttempts < maxReadAttempts {
-		c.readAttempts++
-		// Nothing happened, let's read again. We reached the end of the frame
```
MarcoPolo, line 33: Just flagging that we're losing this comment, but keeping this behavior. Maybe a similar comment would be helpful?

MarcoPolo (approved):
No strong opinion here. 
I agree fixing the 5s stall alone is worth the revert.

p2p/transport/websocket/conn.go
marten-seemann: Are we? This was introduced as part of https://github.com/libp2p/go-libp2p/pull/1982, wasn't it?

nhooyr:
Sorry you had a bad experience with my library. My delay in dealing with issues was unfortunate and I do sincerely apologize. I just wanted to let you know that all the issues you cited are now fixed for v1.8.8 which will be released tomorrow at noon. https://github.com/nhooyr/websocket/pull/256

Your feedback was very much appreciated even if I was too late to get things in for you.

Jorropo:
@nhooyr I see you handled:
> examples: https://github.com/nhooyr/websocket/issues/367, https://github.com/nhooyr/websocket/issues/355 and https://github.com/nhooyr/websocket/issues/384

Thx that is great, however my opinion is still the same, while gorilla is working (no bug or security issue known) I don't think go-libp2p should try switching away from it. However I'm not a maintainer on go-libp2p.

nhooyr:
By all means. I would suggest the same. The WebSocket protocol is stable and so gorilla is fine now and will be fine for a long time. I just wanted to note that the issues cited are now fixed.

paralin:
I like the WebAssembly support in @nhooyr library, among other things. It would be worthwhile to have a look at including it, imo

talentlessguy:
Came here after trying to write a custom ws transport that would support wasm using coder/websocket lib but hitting weird handshake issues from libp2p... would be cool to keep a browser version that would use pure JS syscalls and continue using gorilla/websocket for native

Issue #2283, opened 2023-05-11
swarm: duplicate tracking in dial worker loop doesn't work


marten-seemann:
In the dial worker loop, we use maps of `ma.Multiaddr`s: https://github.com/libp2p/go-libp2p/blob/8a3fafdaaf4ccb9b18c9d2c10347d90af5d9818c/p2p/net/swarm/dial_worker.go#L32

and https://github.com/libp2p/go-libp2p/blob/8a3fafdaaf4ccb9b18c9d2c10347d90af5d9818c/p2p/net/swarm/dial_worker.go#L49

This is incorrect, since `ma.Multiaddr` is (until we fix https://github.com/multiformats/go-multiaddr/issues/198) an interface, and equality only checks equality of the pointer, not of the actual multiaddress. This has bitten us before, for example in https://github.com/libp2p/go-libp2p/pull/2247.

marten-seemann:
The impact of this is that we dial the same multiaddr multiple times. On TCP, due to reuseport, this should lead to failed dial right away, whereas on QUIC, we'd actually dial twice. I instrumented my Kubo node, and this happens from time to time.

This probably doesn't warrant another patch release, but we should include a fix in the next release.

PR #2284, opened 2023-05-11
swarm: change maps with multiaddress keys to use strings

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2283

`string(addr.Bytes())` is already used in peerstore. https://github.com/libp2p/go-libp2p/issues/2283

Fixes one more instance in `routed.go` https://github.com/libp2p/go-libp2p/blob/master/p2p/host/routed/routed.go#L122


marten-seemann (approved):

sukunrt:
@marten-seemann 
I've fixed the test

marten-seemann (approved):

PR #2287, opened 2023-05-14
net/mock: mimic Swarm's event and notification behavior in MockNet

### LINKED ISSUE #2267: Connectedness events are not submitted by mocknet
After https://github.com/libp2p/go-libp2p/pull/1574, the `Network` is now responsible for emitting such events. However, only `Swarm` was updated to emit them, and the only second implementation, `mocknet` was not. There are projects in the ecosystem that uses `mockent` and for whom updating to v0.27 will break tests because of this.

Additionally, the tests that depend on `swarm/testing` will break as well due to the reason @vyzo mentioned in [here](https://github.com/libp2p/go-libp2p/pull/1574#discussion_r884115720). The user now has two instantiate the EventBus themselves and pass two both the host constructor and the `GenSwarm`, but if I understand correctly this is part of a larger issue https://github.com/libp2p/go-libp2p/issues/1993.

Wondertan:
Fixes #2267 + synchronously notifies as Swarm after #1562

### DIFF
--- a/p2p/net/mock/mock_conn.go
+++ b/p2p/net/mock/mock_conn.go
@@ -86,20 +86,12 @@ func (c *conn) Close() error {
 	return nil
 }
 
-func (c *conn) teardown() error {
+func (c *conn) teardown() {
 	for _, s := range c.allStreams() {
 		s.Reset()
 	}
-	c.net.removeConn(c)
 
-	go func() {
-		c.notifLk.Lock()
-		defer c.notifLk.Unlock()
-		c.net.notifyAll(func(n network.Notifiee) {
-			n.Disconnected(c.net, c)
-		})
-	}()
-	return nil
+	c.net.removeConn(c)
 }
 
 func (c *conn) addStream(s *stream) {
--- a/p2p/net/mock/mock_net.go
+++ b/p2p/net/mock/mock_net.go
@@ -14,6 +14,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+	"github.com/libp2p/go-libp2p/p2p/host/eventbus"
 	"github.com/libp2p/go-libp2p/p2p/host/peerstore/pstoremem"
 
 	ma "github.com/multiformats/go-multiaddr"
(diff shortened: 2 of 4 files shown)


p2p/net/mock/mock_peernet.go
```diff
@@ -210,6 +221,22 @@ func (pn *peernet) removeConn(c *conn) {
 		panic(fmt.Sprintf("attempting to remove a conn that doesnt exist %v", c.remote))
 	}
 	delete(cs, c)
+	pn.Unlock()
+
+	// notify asynchronously to mimic Swarm
+	// FIXME: IIRC, we wanted to make notify for Close synchronous
```
Wondertan, line 227: > and yes, this means that ideally, we'd also get rid of the Go routine for the Disconnected notification.

https://github.com/libp2p/go-libp2p/pull/2027#issuecomment-1407475172

p2p/net/mock/mock_conn.go
```diff
@@ -86,20 +86,12 @@ func (c *conn) Close() error {
...
-	c.net.removeConn(c)
 
-	go func() {
-		c.notifLk.Lock()
-		defer c.notifLk.Unlock()
-		c.net.notifyAll(func(n network.Notifiee) {
-			n.Disconnected(c.net, c)
-		})
-	}()
-	return nil
```
Wondertan, line 102: Moved for nicer logic grouping.

p2p/net/mock/mock_net.go
```diff
@@ -108,14 +109,16 @@ func (mn *mocknet) AddPeer(k ic.PrivKey, a ma.Multiaddr) (host.Host, error) {
...
 		return nil, err
 	}
 
 	opts := &bhost.HostOpts{
 		NegotiationTimeout:      -1,
 		DisableSignedPeerRecord: true,
+		EventBus: bus,
 	}
 
 	h, err := bhost.NewHost(n, opts)
```
Wondertan, line 124: Any reason _not_ to use lightweight blankhost here isntead?
Doesn't seem like basichost gives anything here

MarcoPolo: Generally I'd prefer this to behave as similar to a normal node as possible. So I'd lean towards keeping this as the basic host. Is there something in basichost that's too heavy for you? I wonder if it would make more sense to make that lighter instead.

MarcoPolo (approved):

MarcoPolo:
Thank you @Wondertan 🙏

p2p/net/mock/mock_net.go
Wondertan: Nothing specific, just wonderting. BasicHost also works

Issue #2194, opened 2023-03-16
Add integration tests for transports


MarcoPolo:
We should have some integration tests for transports in general to:

1. Codify what well-behaved transports need to do.
2. Make it easier to write new transports.

The tests should at least exercise:

- [x] One stream some data.
- [x] One stream more data (e.g. 1MiB).
- [x] Stream resets.
- [x] connection gater: https://github.com/libp2p/go-libp2p/pull/2200
- [x] resource manager #2285 
- [x] tests for deadlines https://github.com/libp2p/go-libp2p/pull/2286
    - [ ] deadline on streams
    - [ ] deadline on dials
- [x] Multiple streams (concurrently). The sum of the stream data transferred should exceed 20 MB, for the same reason as above. https://github.com/libp2p/go-libp2p/pull/2296
- [x] Many streams, more than the muxer's stream limit (`N = 1000`?). Or alternatively, setting a lower rcmgr limit on the receiver side. The client would need to back off on stream resets then (wait for 50ms, then try again). Eventually, both server and client should have handled exactly `N` streams. https://github.com/libp2p/go-libp2p/pull/2296
- [x] p2p/transport/testsuite https://github.com/libp2p/go-libp2p/pull/2295

marten-seemann:
> - [ ] One stream more data (e.g. 1MiB).

Ideally, more than 1 MB. 1 MB still might fit in the initial flow control window, depending on the transport. 20 MB sounds like a safer choice, while still being reasonably fast.

> - [ ]  Stream resets.

That's 2 tests: The sender resetting the write side and the receiver resetting the read side.

> - [ ]  connection gater. @marten-seemann has a WIP somewhere

Not a WIP any more after today's work: https://github.com/libp2p/go-libp2p/pull/2200.
Ideally, the transport tests would live in the same package, so we only have to define the address slice (maybe refactored to be a slice of `libp2p.Option`s) once: https://github.com/libp2p/go-libp2p/blob/03a37d32fcea81344ff3f95e088db303cf3d3c91/p2p/test/connectiongating/gating_test.go#L23-L29

---

Adding to the list in the original post (feel free to copy-paste):
- [ ] Multiple streams (concurrently). The sum of the stream data transferred should exceed 20 MB, for the same reason as above.
[...]

PR #2295, opened 2023-05-16
Transport integration tests: transport suite

MarcoPolo:
Related to #2194 

This runs the `p2p/transport/testsuite` on all transports. There is one change made to avoid starting a new listener for every connection, which would fail for QUIC listeners since multiple calls to listen on port 0 will use the same port and fail.

### DIFF
--- a/p2p/test/transport/suite_test.go
+++ b/p2p/test/transport/suite_test.go
@@ -1,0 +1,46 @@
+package transport_integration
+
+import (
+	"regexp"
+	"testing"
+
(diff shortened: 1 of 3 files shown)


marten-seemann (commented):
What's the intention behind this PR? The existing transport suite has caused us nothing but pain.

MarcoPolo:
I wanted to make use of the existing tests and code we had. I didn't know this caused us pain. If the feeling is that this is more effort than it's worth, I can close.

MarcoPolo:
Closing for lack of interest.

PR #2299, opened 2023-05-18
identify: avoid spuriously triggering pushes

marten-seemann:
This definitely happens during node startup: We receive a notification for adding the Identify Push protocol, as well as for starting to listen on our addresses. It might also happen later (maybe?).

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -1,11 +1,15 @@
 package identify
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
+	"sort"
 	"sync"
 	"time"
+
+	"golang.org/x/exp/slices"
 
 	"github.com/libp2p/go-libp2p/core/crypto"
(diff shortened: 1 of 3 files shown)


p2p/protocol/identify/id.go
```diff
@@ -62,6 +66,31 @@ type identifySnapshot struct {
...
+	if hasRecord != otherHasRecord {
+		return false
+	}
+	if hasRecord && !s.record.Equal(other.record) {
+		return false
+	}
+	if !slices.Equal(s.protocols, other.protocols) {
+		return false
+	}
+	if len(s.protocols) != len(other.protocols) {
```
sukunrt: I think you want, `s.addrs` and `other.addrs`

sukunrt (approved):

p2p/protocol/identify/id.go
sukunrt: I see you fixed it :)

PR #2300, opened 2023-05-18
identify: filter received addresses based on the node's remote address

marten-seemann:
Unfortunately, this is impossible to test. I tried using mocknet, but of course that thing is so horribly broken that it can't even reliably set the correct remote address of a connection. It fails at the most easy thing it was designed to do.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -769,7 +769,7 @@ func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn, isPush bo
 			log.Debugf("error adding signed addrs to peerstore: %v", err)
 		}
 	} else {
-		ids.Host.Peerstore().AddAddrs(p, lmaddrs, ttl)
+		ids.Host.Peerstore().AddAddrs(p, filterAddrs(lmaddrs, c.RemoteMultiaddr()), ttl)
 	}
 
 	// Finally, expire all temporary addrs.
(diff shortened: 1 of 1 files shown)


sukunrt:
If you want a connection with a remote addr on private IP and a remote addr on loopback ip, it should be possible to do this by listening on "0.0.0.0" address and dialing only the private or the loopback address. Testing public ips does look impossible.

p2p/protocol/identify/id.go
```diff
@@ -767,12 +767,11 @@ func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn, isPush bo
 	// add signed addrs if we have them and the peerstore supports them
 	cab, ok := peerstore.GetCertifiedAddrBook(ids.Host.Peerstore())
 	if ok && signedPeerRecord != nil {
-		_, addErr := cab.ConsumePeerRecord(signedPeerRecord, ttl)
-		if addErr != nil {
-			log.Debugf("error adding signed addrs to peerstore: %v", addErr)
+		if _, err := cab.ConsumePeerRecord(signedPeerRecord, ttl); err != nil {
+			log.Debugf("error adding signed addrs to peerstore: %v", err)
```
sukunrt, line 771: for this to be effective, these addresses need to be filtered too, right?

marten-seemann: Right. I've done some digging and convinced myself that the `ConsumePeerRecord` API is flawed. Will open a separate issue for that.

marten-seemann:
@sukunrt Should we merge this PR, and do the fixes for signed peer records when we tackle #2304?

sukunrt (approved):

sukunrt:
Yes. We can take signed peer records separately.

PR #2301, opened 2023-05-19
metrics: add instance filter to all dashboards

sukunrt:
This provides an instance filter on dashboards like this:

Screenshot 2023-05-19 at 6 51 46 PM

The script I used to update these is at 
https://github.com/sukunrt/go-libp2p/blob/add-instance-script/dashboards/script/script.go


marten-seemann (commented):
Change looks good, and it works fine in the PL Grafana instance.

Any reason to change the file permissions though?
image

sukunrt:
oh no. I didn't notice I'd updated them. Changed them back to 644 as before.

marten-seemann (approved):

PR #2305, opened 2023-05-20
webtransport: close the challenge stream after the Noise handshake

Jorropo:
Once the challenge succeeded there is no reason to keep the challenge stream open, it will never be used again.

### DIFF
--- a/p2p/transport/webtransport/transport.go
+++ b/p2p/transport/webtransport/transport.go
@@ -233,6 +233,7 @@ func (t *transport) upgrade(ctx context.Context, sess *webtransport.Session, p p
 	if err != nil {
 		return nil, err
 	}
+	defer str.Close()
 
 	// Now run a Noise handshake (using early data) and get all the certificate hashes from the server.
 	// We will verify that the certhashes we used to dial is a subset of the certhashes we received from the server.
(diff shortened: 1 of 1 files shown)


p2p/transport/webtransport/transport.go
```diff
@@ -264,6 +265,9 @@ func (t *transport) upgrade(ctx context.Context, sess *webtransport.Session, p p
 	if err != nil {
 		return nil, err
 	}
+	if err = c.Close(); err != nil {
```
MarcoPolo, line 268: Do we care about this error at all? Can we do `defer c.Close()` instead?

p2p/transport/webtransport/transport.go
```diff
@@ -264,6 +265,9 @@ func (t *transport) upgrade(ctx context.Context, sess *webtransport.Session, p p
...
+		return nil, err
+	}
```
marten-seemann, line 270:
```suggestion
	defer c.Close()
```

p2p/transport/webtransport/transport.go
marten-seemann: I made that change.

marten-seemann (approved):

PR #2310, opened 2023-05-24
adding UserAgent to Nat for description of port mapping

master255:


MarcoPolo (commented):
Nice, thank you. Could we include some tests so that we don't break this in the future.

master255:
I don't know how this can be tested. It's just input data. I don't have any ideas.

p2p/host/basic/natmgr.go
```diff
@@ -25,8 +25,8 @@ type NATManager interface {
 }
 
 // NewNATManager creates a NAT manager.
-func NewNATManager(net network.Network) NATManager {
-	return newNATManager(net)
+func NewNATManager(net network.Network, userAgent string) NATManager {
```
marten-seemann, line 28: Nicer: use the option pattern.

master255: @marten-seemann What does that mean? In what place? Where is it better to take the variable from?

master255: @marten-seemann Ok. Maybe this is the right way?

Wondertan: Marten likely meant defining a new options type for the `NATManager` instead of extending Swarm's options set and Network interface with UserAgant getter.

master255: @Wondertan Why not?

Wondertan: Managing UserAgent is logically not Network's responsibility.

master255: @Wondertan Whose responsibility is that?

Wondertan: IIRC, it's just a param that's passed from BasicHost down to Identify, so you can intercept it and pass it to discoverNAT as an option.

master255: @marten-seemann Where is the answer from Marten?

master255: @Wondertan 

> // Network is the interface used to connect to the outside world.
// It dials and listens for connections.

I use Network to store my UserAgent so that Network can tell the outside world my UserAgent. Isn't that logical?

marten-seemann: I was talking about the option pattern, like here: https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/circuitv2/relay/options.go

master255: @marten-seemann And you think [putting the UserAgent in the Swarm](https://github.com/libp2p/go-libp2p/pull/2310/files#diff-562d33c4dff7fb735c38df99f47ca60a7ec4043be9199d5ca081745f6239a09cR36) - Network is not right?

master255: @marten-seemann I don't know. I just read the comments about what the swarm was doing and looked at which direction the data was going. If the Network is establishing connections, it needs a UserAgent.

Okay, if this idea of mine is wrong, I'll fix the code. I'll try to make an options.

master255: @marten-seemann The new version with options is ready. But it seems to me it is too much for one parameter.

p2p/host/basic/basic_host.go
```diff
@@ -131,7 +131,7 @@ type HostOpts struct {
 
 	// NATManager takes care of setting NAT port mappings, and discovering external addresses.
 	// If omitted, this will simply be disabled.
-	NATManager func(network.Network) NATManager
+	NATManager func(network.Network, Option) (NATManager, error)
```
marten-seemann, line 134: The beauty of the option pattern is that the options are optional. You need to make them variadic though.

```suggestion
	NATManager func(network.Network, ...Option) (NATManager, error)
```

master255: @marten-seemann Ready. But I still think it's redundant to add an Option for a single parameter. Perhaps there will be more parameters in the future?

master255: Fixed the error

p2p/net/nat/nat.go
```diff
@@ -34,7 +34,7 @@ type entry struct {
 var discoverGateway = nat.DiscoverGateway
 
 // DiscoverNAT looks for a NAT device in the network and returns an object that can manage port mappings.
-func DiscoverNAT(ctx context.Context) (*NAT, error) {
+func DiscoverNAT(ctx context.Context, userAgent string) (*NAT, error) {
```
MarcoPolo, line 37: Do we want to change the interface of this public method?

master255: What are the alternatives?

MarcoPolo: Introducing a new func.

master255: If we create a new function, we need to designate a default userAgent somewhere in this file. The default value is set in another file.
So this constructor has no sense without the userAgent string.

MarcoPolo: The default should be in this file. That was the old behavior. If users want to change it they can call the new function (maybe call it `DiscoverNATWithUserAgent`).

This change as is is a breaking API change. I like to avoid them unless necessary. I don't think it's necessary here.

master255: @MarcoPolo Okay. Then what to do with this one?
https://github.com/libp2p/go-libp2p/pull/2310/files#diff-a5aa28aae96d02e4316747f7950909f36d447f7833e5a70121c3808852e5e440R80

What to pass in this line? What will be the default value here?

master255: And then what will be in this place? "if" and two constructors? 
https://github.com/libp2p/go-libp2p/pull/2310/files#diff-a5aa28aae96d02e4316747f7950909f36d447f7833e5a70121c3808852e5e440R117

master255: @MarcoPolo Okay. Why shouldn't there be an incoming userAgent parameter in this function? After all, we cannot communicate with NAT without userAgent.

sukunrt: >Why shouldn't there be an incoming userAgent parameter in this function?

Because it's a breaking API change. Users of this function, when they update go-libp2p will need to update their usage of this function and pass in a userAgent. 

It'd be nicer if we add another function `DiscoverNATWithUserAgent` and `NatManager` switches based on whether it has a user configured UserAgent or not

master255: @sukunrt Okay. I fixed it.

MarcoPolo:
Happy Spring. I'm doing some spring cleaning and closing old PRs. There doesn't seem to be sufficient maintainer interest to keep moving this forward. Closing this for now, but if it comes up again we know where to find this PR.

Issue #2231, opened 2023-03-29
Peerstore interfaces should accept contexts


guseggert:
The [peerstore](https://github.com/libp2p/go-libp2p/blob/master/core/peerstore/peerstore.go) family of interfaces currently don't accept any Go contexts, despite performing IO. This means that, for example, the "pstoreds" package which implements a datastore-backed peerstore can't propagate contexts into the backing datastore.  Example: 

https://github.com/libp2p/go-libp2p/blob/8c466ab7f8d0e1eb12ea8a23933e77d25190a81a/p2p/host/peerstore/pstoreds/addr_book.go#L54

This means that the datastore can't respect context cancelations and any values stored in the context are also lost (ex. tracing metadata).

marten-seemann:
Didn't we have a PR for that at some point?

PR #2312, opened 2023-05-30
peerstore: add contexts to all methods

chaitanyaprem:
Addressing #2231 

Brief of changes:
- Included an argument to accept contexts for all PeerStore interfaces. 
- Wherever there is a background worker or a loop running in the flow, chosen to use the context.Background(). 
- All test code is updated to use context.Background(). 
- Areas that are unclear, used context.TODO()

Ran complete test-suite in ubuntu-20.04 after changes successfully.


chaitanyaprem:
@marten-seemann  It would be better to merge this PR if there are no issues, otherwise changes keep happening and require lot of merges from master.

marten-seemann:
You're right. Sorry, didn't mean to cause you extra trouble. Will merge as soon as CI passes.

PR #2319, opened 2023-06-01
update go-nat to v0.2.0, use context on AddMapping and RemoveMapping

marten-seemann:


sukunrt (approved):
Looks fine. Needs go mod tidy though.

PR #2320, opened 2023-06-02
swarm: implement blackhole detection

sukunrt:


sukunrt:
For blackhole detection if we want just a binary check like it’s a black hole if the last n consecutive calls are failures we can do away with the sliding window.

This doesn't implement happy eyeballs for TCP since the delay required for that to be effective is about 700ms. That feels too long to me. We can implement happy eyeballs for TCP separately where we introduce a mechanism to notify the worker loop that a connection has been established and security and muxer selection is in progress.

p2p/net/swarm/swarm_dial.go
```diff
@@ -468,6 +468,8 @@ func (s *Swarm) filterKnownUndialables(p peer.ID, addrs []ma.Multiaddr) []ma.Mul
 		func(addr ma.Multiaddr) bool {
 			return s.gater == nil || s.gater.InterceptAddrDial(p, addr)
 		},
+		s.udpBHD.IsDialAllowed,
+		s.ipv6BHD.IsDialAllowed,
```
marten-seemann, line 472: Would be nice to have just a single black hole detector. This detector would then hand off to one IPv6 and UDP blackhole detector.

marten-seemann:
Should we make the changes to the dial prioritization logic in a follow-up PR? Would be nice to disentangle it.

p2p/net/swarm/swarm_dial.go
marten-seemann: To clarify:
```go
type blackHoleDetector struct {
    udp, ipv6 singleBlackHoleDetector // TODO: better naming
}

func (d *blackHoleDetector) RecordResult(addr ma.Multiaddr, result bool) {
    if d.udp != nil & isUDPAddr(addr) {
         d.udp.RecordResult(addr, result)
    }
    if d.ipv6 != nil & isIPv6Addr(addr) {
         d.ipv6.RecordResult(addr, result)
    }
}

**Comment by sukunrt** [comment]
Completely agree with the metrics suggestion. 


>if behind a blackhole, the next time we'll probe

might be difficult since we probe based on number of requests(every nth request is allowed). It should be fine to make this number of requests till next probe.

**Comment by sukunrt** [comment]
This is what the dashboard looks like for me: 

Screenshot 2023-06-07 at 6 47 40 PM

File: p2p/net/swarm/black_hole_detector.go [hunk 2]
```diff
@@ -0,0 +1,188 @@
...
+	// outcomeIdx is the index of the next outcome in the sliding window
+	outcomeIdx int
+	// successes is the count of successful dials in outcomes
+	successes int
+	// failures is the count of failed dials in outcomes
+	failures int
+	// full is true when we have a full sliding window worth of outcomes.
+	// Keeping this as a separate variable helps avoid clearing out the entire sliding window on
+	// reset.
+	full bool
```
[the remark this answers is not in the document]
sukunrt: I've removed other parameters and made n serve the dual purpose of being the sliding window size and the minimum number of dials required to start blocking.

p2p/net/swarm/black_hole_detector.go
```diff
@@ -0,0 +1,188 @@
...
+		},
+		"UDP",
+		100,
+		100,
+		1000,
+		0.01,
+	)
+}
+
+func (b *blackHoleDetector) CompletedDial(addr ma.Multiaddr, success bool) {
```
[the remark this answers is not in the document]
sukunrt: `RecordOutcome`
The Dial is implicit anyway in everything related to black hole detection.

p2p/net/swarm/black_hole_detector.go
```diff
@@ -0,0 +1,208 @@
...
+	if d.ipv6 != nil && isProtocolAddr(addr, ma.P_IP6) {
+		ipv6State, ipv6Allowed = d.ipv6.IsAllowed()
+	}
+
+	// Allow all probes irrespective of the state of the other filter
+	if (udpState == blackHoleStateBlocked && udpAllowed) ||
+		(ipv6State == blackHoleStateBlocked && ipv6Allowed) {
+		return true
+	}
+	return (udpAllowed && ipv6Allowed)
```
marten-seemann, line 183: Nit:
```suggestion
	return udpAllowed && ipv6Allowed
```

p2p/net/swarm/black_hole_detector.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+}
+
+func newBlackHoleDetector(detectUDP, detectIPv6 bool, mt MetricsTracer) *blackHoleDetector {
+	d := &blackHoleDetector{}
+	if detectUDP {
+		d.udp = &blackHoleFilter{n: 100, minSuccessFraction: 0.01, name: "UDP", metricsTracer: mt}
+	}
+	if detectIPv6 {
+		d.ipv6 = &blackHoleFilter{n: 100, minSuccessFraction: 0.01, name: "IPv6", metricsTracer: mt}
```
marten-seemann, line 205: This would lead to problems in the pathological case where peers never listened on QUIC on IPv6, right? Then the QUIC dials on IPv4 would succeed, canceling the TCP dials on IPv6, leading us to conclude that we're blackholed.

I think this is fine, but we should add some justification in a comment.

p2p/net/swarm/black_hole_detector.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func (b *blackHoleFilter) IsAllowed() (state blackHoleState, isAllowed bool) {
+	if b == nil {
+		return blackHoleStateAllowed, true
+	}
+
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	b.requests++
```
marten-seemann, line 104: Why are you incrementing the request counter here? Just the fact that a dial is allowed doesn't mean that it actually happened. It might have been delayed by the smart dialing logic for example.

p2p/net/swarm/black_hole_detector.go
```diff
@@ -0,0 +1,208 @@
...
+	if b.requests == b.n {
```
marten-seemann, line 105: I don't understand this.

sukunrt: b.requests is the counter used to allow every nth request. We reset the counter when it reaches n.

p2p/net/swarm/black_hole_detector.go
sukunrt: This is why I moved the IsAllowed check to [here](https://github.com/sukunrt/go-libp2p/blob/blackhole-detector/p2p/net/swarm/swarm_dial.go#L399-L406) See the comment. 
If I have this check with filtering we run in to the issue you mentioned here.

p2p/net/swarm/black_hole_detector.go
```diff
@@ -0,0 +1,208 @@
...
+	ipv6State, ipv6Allowed := blackHoleStateAllowed, true
+	if d.ipv6 != nil && isProtocolAddr(addr, ma.P_IP6) {
+		ipv6State, ipv6Allowed = d.ipv6.IsAllowed()
+	}
+
+	// Allow all probes irrespective of the state of the other filter
+	if (udpState == blackHoleStateBlocked && udpAllowed) ||
+		(ipv6State == blackHoleStateBlocked && ipv6Allowed) {
+		return true
+	}
```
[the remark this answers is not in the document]
sukunrt: I'm using the term probe to mean we are blocked and we want to reevaluate the black hole state by allowing the request.
This is a composite check. We want to allow when either of the blocked black hole filters decide that it is time to probe.  
Say UDP and IPv6 are both blocked and UDP black hole filter allows the probe but IPv6 black hole filter blocks it. Just a single check `udpAllowed && ipv6Allowed` will block this probe. I think it is better to allow all probes in case the black hole filter state has changed.

p2p/net/swarm/black_hole_detector.go
marten-seemann: What exactly does resetting mean? I would have assumed that a reset occurs when we're transitioning from blocked -> probing state, because a dial unexpectedly succeeded.

In all other cases, it's not really necessary to reset the counter, there's no way it'll overflow.

p2p/net/swarm/black_hole_detector.go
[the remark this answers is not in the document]
sukunrt: I am now returning a `blackHoleResult` type which will be `blackHoleResultProbe` in this case.

p2p/net/swarm/black_hole_detector.go
```diff
@@ -0,0 +1,208 @@
...
+	if d.ipv6 != nil && isProtocolAddr(addr, ma.P_IP6) {
+		d.ipv6.RecordOutcome(success)
+	}
+}
+
+func newBlackHoleDetector(detectUDP, detectIPv6 bool, mt MetricsTracer) *blackHoleDetector {
+	d := &blackHoleDetector{}
+	if detectUDP {
+		d.udp = &blackHoleFilter{n: 100, minSuccessFraction: 0.01, name: "UDP", metricsTracer: mt}
```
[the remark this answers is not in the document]
sukunrt: I've added comment. Network black hole is a binary property. We only want to block when all calls would fail.

p2p/net/swarm/black_hole_detector.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+		b.outcomes = b.outcomes[1 : b.n+1]
+	}
+
+	b.updateState()
+	b.trackMetrics()
+}
+
+func (b *blackHoleFilter) IsAllowed() (state blackHoleState, isAllowed bool) {
+	if b == nil {
```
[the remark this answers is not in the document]
sukunrt: removed.

p2p/net/swarm/black_hole_detector.go
sukunrt: Yes in this case we will be incorrect for a few probes. Another alternative is to allow 3-4 probes which reduces the chances that this would happen.

p2p/net/swarm/black_hole_detector.go
sukunrt: We need to increment the counter here. There is no other way to handle this. If we are blocking request there will be no call to RecordOutcome, so this counter needs to be reset in this method.

p2p/net/swarm/black_hole_detector.go
sukunrt: I think that is simpler. Both have the same effect, but yours is clearer. I've moved this check to (b.requests % b.n == 0).

sukunrt:
@marten-seemann 
An alternative is to allow 5-10 probes every n requests. If we do this we can move the black hole can dial check to `swarm.filterKnownUndialables`. 
This is useful because eventually I'd like to get rid of the backoff check here too(opening a separate issue for that). https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/swarm_dial.go#L392

sukunrt:
Updated the PR to do address filtering and probing for reevaluating black hole state at a peer level.

p2p/net/swarm/black_hole_detector.go
```diff
@@ -0,0 +1,257 @@
...
+	if success {
+		b.successes++
+	}
+	b.dialResults = append(b.dialResults, success)
+
+	if len(b.dialResults) > b.n {
+		if b.dialResults[0] {
+			b.successes--
+		}
+		b.dialResults = b.dialResults[1 : b.n+1]
```
marten-seemann, line 99: This is equivalent to
```suggestion
		b.dialResults = b.dialResults[1:]
```
right?

p2p/net/swarm/swarm.go
```diff
@@ -209,8 +211,12 @@ func NewSwarm(local peer.ID, peers peerstore.Peerstore, eventBus event.Bus, opts
 	}
 
 	s.dsync = newDialSync(s.dialWorkerLoop)
+
 	s.limiter = newDialLimiter(s.dialAddr)
 	s.backf.init(s.ctx)
+
+	s.bhd = newBlackHoleDetector(true, true, s.metricsTracer)
```
marten-seemann, line 218: We need a config option for this. Ideally, the user could configure the thresholds. Doesn't need to be part of this PR, but we'd need this for the next release.

p2p/net/swarm/swarm_dial.go
```diff
@@ -484,6 +491,9 @@ func (s *Swarm) dialAddr(ctx context.Context, p peer.ID, addr ma.Multiaddr) (tra
 
 	start := time.Now()
 	connC, err := tpt.Dial(ctx, addr, p)
+
+	s.bhd.RecordResult(addr, err == nil)
```
marten-seemann, line 495:
```suggestion
	// We're recording any error as a failure here.
	// Notably, this also applies to cancelations (i.e. if another dial attempt was faster).
	// This is ok since the black hole detector uses a very low threshold (5%).
	s.bhd.RecordResult(addr, err == nil)
```

p2p/net/swarm/dial_worker.go
```diff
@@ -358,7 +358,8 @@ loop:
 			}
 
 			// it must be an error -- add backoff if applicable and dispatch
-			if res.Err != context.Canceled && !w.connected {
+			// ErrDialRefusedBlackHole shouldn't end up here, just a safety check
+			if res.Err != ErrDialRefusedBlackHole && res.Err != context.Canceled && !w.connected {
```
marten-seemann, line 362: Should this go in error log? If we're not noisy about it, we'd miss this bug if we ever introduce it.

p2p/net/swarm/black_hole_detector_test.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+			got := bhf.HandleRequest()
+			if got != tc.result {
+				t.Fatalf("expected %d got %d", tc.result, got)
+			}
+		})
+	}
+}
+
+func areAddrsEqual(a, b []ma.Multiaddr) bool {
```
marten-seemann, line 76: Should we just use `require.ElementsMatch`?

p2p/net/swarm/swarm.go
sukunrt: you mean with libp2p.New?

marten-seemann: Same as for the dial ranker.

p2p/net/swarm/black_hole_detector_test.go
sukunrt: TIL this does deep equal. Thanks

p2p/net/swarm/swarm.go
sukunrt: will raise a separate PR

p2p/net/swarm/dial_worker.go
sukunrt: Added this

marten-seemann (approved):

p-shahi:
Add the description in the release issue highlights while its fresh https://github.com/libp2p/go-libp2p/issues/2326?

PR #2322, opened 2023-06-02
swarm: Dedup addresses to dial

MarcoPolo:
I moved the DedupAddrs function to `core/network` since that seemed like the most relevant place for it to be so that basic_host and swarm can use it. Really this should probably be part of the multiaddr package. Although I would say we make that change later.

This change makes `swarm.addrsForDial` return only unique addresses. This is assumed by they dialWorker. I also added some existence checks to avoid any potential panics in case the key was missing from the pending map.

Most of this diff is a test or moving code around. The real change is the one line in swarm_dial.go.

### DIFF
--- a/core/network/network.go
+++ b/core/network/network.go
@@ -6,8 +6,10 @@
 package network
 
 import (
+	"bytes"
 	"context"
 	"io"
+	"sort"
 	"time"
 
 	"github.com/libp2p/go-libp2p/core/peer"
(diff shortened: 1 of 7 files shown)


core/network/network.go
```diff
@@ -184,3 +186,23 @@ type Dialer interface {
...
 	StopNotify(Notifiee)
 }
+
+// DedupAddrs deduplicates addresses in place, leave only unique addresses.
+// It doesn't allocate.
+func DedupAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
+	if len(addrs) == 0 {
+		return addrs
+	}
+	sort.Slice(addrs, func(i, j int) bool { return bytes.Compare(addrs[i].Bytes(), addrs[j].Bytes()) < 0 })
```
sukunrt, line 196: Apologies for the drive by comment. Feel free to ignore. 

Sort.Slice will allocate unfortunately :( 
https://github.com/golang/go/issues/17332

```
func BenchmarkDedupAddrs(b *testing.B) {
	b.ReportAllocs()
	tcpAddr := ma.StringCast("/ip4/127.0.0.1/tcp/1234")
	quicAddr := ma.StringCast("/ip4/127.0.0.1/udp/1234/quic-v1")
	wsAddr := ma.StringCast("/ip4/127.0.0.1/tcp/1234/ws")
	addrs := []ma.Multiaddr{tcpAddr, quicAddr, tcpAddr, quicAddr, wsAddr}
	for i := 0; i < b.N; i++ {
		addrs = DedupAddrs(addrs)
	}
}
```

MarcoPolo: Unfortunate. Given this is pre-existing code, I'll leave it. But I'll open an issue

PR #2331, opened 2023-06-05
basichost / blankhost: wrap errors

Wondertan:
The goal here is to improve the experience of working with errors returned by the Host. With the current state, it's PITA to debug exactly what went wrong during a stream open and this feat attempts to solve it.

* Inspired by errors in std `net`.
* I wanted to improve the RoutedHost as well, but I will wait until you migrate to v1.20, as I need `errors.Join`
* I am open to any other suggestions on how to make this even better
* Can we include this in 0.28? 🙏🏻 🙏🏻 🙏🏻 

Closes https://github.com/libp2p/go-libp2p/issues/2334

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -641,7 +641,7 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
 		if errors.Is(err, network.ErrNoConn) {
 			return nil, errors.New("connection failed")
 		}
-		return nil, err
+		return nil, fmt.Errorf("failed to open stream: %w", err)
 	}
 
 	// Wait for any in-progress identifies on the connection to finish. This
@@ -653,7 +653,7 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
 	case <-h.ids.IdentifyWait(s.Conn()):
 	case <-ctx.Done():
 		_ = s.Reset()
-		return nil, ctx.Err()
+		return nil, fmt.Errorf("identify failed to complete: %w", ctx.Err())
 	}
 
 	pref, err := h.preferredProtocol(p, pids)
@@ -682,13 +682,13 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
 	case err = <-errCh:
 		if err != nil {
 			s.Reset()
-			return nil, err
+			return nil, fmt.Errorf("failed to negotiate protocol: %w", err)
 		}
 	case <-ctx.Done():
 		s.Reset()
 		// wait for `SelectOneOf` to error out because of resetting the stream.
 		<-errCh
-		return nil, ctx.Err()
+		return nil, fmt.Errorf("failed to negotiate protocol: %w", ctx.Err())
 	}
 
 	s.SetProtocol(selected)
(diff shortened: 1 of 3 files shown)


marten-seemann:
I’m ok with the error wrapping changes introduced here. I’d like to be a little bit more careful about introducing a new error type here. Not saying that it’s a bad idea, but I’d prefer to have an issue and an API discussion first.

Wondertan:
>  I’d prefer to have an issue and an API discussion first.

My toxic trait is coming up with contributions out of nowhere 😄. 

Ok, let's make an issue!

Wondertan:
Any chance you can take a look at this?

marten-seemann:
> I’m ok with the error wrapping changes introduced here.

I already reviewed this PR. Let me know once you've updated it.

Wondertan:
Ok, it wasn't clear whether changes were requested or not

Wondertan:
Updated to keep wrappings only

p2p/host/basic/basic_host.go
```diff
@@ -641,7 +641,7 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
 		if errors.Is(err, network.ErrNoConn) {
 			return nil, errors.New("connection failed")
 		}
-		return nil, err
+		return nil, fmt.Errorf("opening Network stream: %w", err)
```
marten-seemann, line 644:
```suggestion
		return nil, fmt.Errorf("failed to open stream: %w", err)
```

p2p/host/basic/basic_host.go
```diff
@@ -653,7 +653,7 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
 	case <-h.ids.IdentifyWait(s.Conn()):
 	case <-ctx.Done():
 		_ = s.Reset()
-		return nil, ctx.Err()
+		return nil, fmt.Errorf("awating identify: %w", ctx.Err())
```
marten-seemann, line 656:
```suggestion
		return nil, fmt.Errorf("identify failed to complete: %w", ctx.Err())
```

p2p/host/basic/basic_host.go
```diff
@@ -682,13 +682,13 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
 	case err = <-errCh:
 		if err != nil {
 			s.Reset()
-			return nil, err
+			return nil, fmt.Errorf("negotiating protocol: %w", err)
```
marten-seemann, line 685:
```suggestion
			return nil, fmt.Errorf("failed to negotiate protocol: %w", err)
```

p2p/host/basic/basic_host.go
```diff
@@ -682,13 +682,13 @@ func (h *BasicHost) NewStream(ctx context.Context, p peer.ID, pids ...protocol.I
...
 		}
 	case <-ctx.Done():
 		s.Reset()
 		// wait for `SelectOneOf` to error out because of resetting the stream.
 		<-errCh
-		return nil, ctx.Err()
+		return nil, fmt.Errorf("negotiating protocol: %w", ctx.Err())
```
marten-seemann, line 691:
```suggestion
		return nil, fmt.Errorf("failed to negotiate protocol: %w", ctx.Err())
```

marten-seemann (changes requested):
... and so on for the other errors.

Wondertan:
A user would usually wrap the error with "failed" or the log would also say that error, but I don't feel strongly about it, so applied.

marten-seemann (changes requested):
@Wondertan Any reason you only changed ~half of the error messages?

Wondertan:
> @Wondertan Any reason you only changed ~half of the error messages?

I simply applied all the comments

Wondertan:
Fixed now

PR #2333, opened 2023-06-05
swarm: cleanup address filtering logic

sukunrt:


sukunrt:
This would introduce a bug if merged as is. Converting it to draft.

PR #2345, opened 2023-06-09
Clear stream scope if we error

MarcoPolo:
We should still clean up the resources tracked by the resource manager even if we error here.


MarcoPolo:
This is tricky to test, but I'll add a test in a follow up PR.

sukunrt (approved):

PR #2353, opened 2023-06-12
swarm: fix stale DialBackoff comment

sukunrt:
Removes comment suggesting we dial an address via one goroutine when it's on backoff. We do not do so. 
Removes old code snippets.

### DIFF
--- a/p2p/net/swarm/swarm_dial.go
+++ b/p2p/net/swarm/swarm_dial.go
@@ -75,33 +75,12 @@ const ConcurrentFdDials = 160
 // per peer
 var DefaultPerPeerRateLimit = 8
 
-// dialbackoff is a struct used to avoid over-dialing the same, dead peers.
-// Whenever we totally time out on a peer (all three attempts), we add them
-// to dialbackoff. Then, whenevers goroutines would _wait_ (dialsync), they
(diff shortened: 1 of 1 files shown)


p2p/net/swarm/swarm_dial.go
```diff
@@ -75,33 +75,12 @@ const ConcurrentFdDials = 160
...
-//  }
-//
-
-// DialBackoff is a type for tracking peer dial backoffs.
+// DialBackoff is a type for tracking peer dial backoffs. Dialbackoff is used to
+// avoid over-dialing the same, dead peers. Whenever we totally time out on all
+// addresses of a peer, we add the addresses to DialBackoff. Then, whenever we
+// attempt to dial the peer again, we check each address for backoff. If it's on
+// backoff, we don't dial the address and exit promptly. If a dial is
+// successful, the peer and all its addresses are removed from backoff.
```
marten-seemann, line 83: Do we? That's probably something we should clean up at some point...

sukunrt: I'm assuming you mean if we clear the backoffs on success. We remove the peer from the backoff map effectively clearing all addresses. 
https://github.com/libp2p/go-libp2p/blob/v0.27.0/p2p/net/swarm/swarm.go#L320

PR #2356, opened 2023-06-12
docs: fix some comments errors

sysvm:


config/config.go
```diff
@@ -69,7 +69,7 @@ type Security struct {
 // This is *not* a stable interface. Use the options defined in the root
 // package.
 type Config struct {
-	// UserAgent is the identifier this node will send to other peers when
+	// UserAgent is the identifier that node will send to other peers when
```
marten-seemann, line 72: This is wrong.

core/crypto/secp256k1.go
```diff
@@ -12,10 +12,10 @@ import (
 	"github.com/minio/sha256-simd"
 )
 
-// Secp256k1PrivateKey is an Secp256k1 private key
+// Secp256k1PrivateKey is a Secp256k1 private key
```
marten-seemann, line 15: I don't think that's correct.

core/event/network.go
```diff
@@ -22,7 +22,7 @@ import (
 //   - It's possible to have _multiple_ connections to a given peer.
 //   - Both libp2p and networks are asynchronous.
 //
-// This means that all of the following situations are possible:
+// This means that all the following situations are possible:
```
marten-seemann, line 25: That's not correct.

core/network/network.go
```diff
@@ -130,7 +130,7 @@ type Network interface {
 	io.Closer
 
 	// SetStreamHandler sets the handler for new streams opened by the
-	// remote side. This operation is threadsafe.
+	// remote side. This operation is thread safe.
```
marten-seemann, line 133: Above you were using a -.

core/network/rcmgr.go
```diff
@@ -102,7 +102,7 @@ type ResourceManager interface {
 // ResourceScopeViewer is a mixin interface providing view methods for accessing top level
 // scopes.
 type ResourceScopeViewer interface {
-	// ViewSystem views the system wide resource scope.
+	// ViewSystem views the system-wide resource scope.
```
marten-seemann, line 105: This is not correct.

core/record/envelope.go
```diff
@@ -271,8 +271,8 @@ func makeUnsigned(domain string, payloadType []byte, payload []byte) ([]byte, er
 	var (
 		fields = [][]byte{[]byte(domain), payloadType, payload}
 
-		// fields are prefixed with their length as an unsigned varint. we
-		// compute the lengths before allocating the sig buffer so we know how
+		// fields are prefixed with their length as an unsigned variant. we
```
marten-seemann, line 274: This is not correct.

marten-seemann (changes requested):
Lots of incorrect suggestions here.

core/record/envelope.go
sysvm: fixed

core/crypto/secp256k1.go
sysvm: This is a non-vowel word.

core/event/network.go
sysvm: I don't understand this. Why?

marten-seemann:
@p-shahi Could you help review this PR please?

core/crypto/secp256k1.go
p-shahi: The comment is a very circular definition, but using `a` is right since `Secp256k1` is a name

core/network/rcmgr.go
p-shahi: Apparently it is either `systemwide` or `system-wide` https://en.wiktionary.org/wiki/system-wide#English so `system-wide` is fine here

p-shahi (approved):

p-shahi:
> What does this sentence mean? I think it use incorrect words. [Link](https://github.com/libp2p/go-libp2p/blob/master/core/network/rcmgr.go#L35) image

Modus operandi is a Latin phrase: https://en.wikipedia.org/wiki/Modus_operandi
The comment in the screenshot is fine as is.

Thanks for doing this, not the most critical change but grammatical corrections are appreciated

PR #2358, opened 2023-06-12
nat: add HasDiscoveredNAT method for checking NAT environments

sukunrt:


MarcoPolo:
For context, this fixes a regression from V0.27.

MarcoPolo (approved):

Issue #2349, opened 2023-06-12
Handle dial prioritisation for well known prefix IPv4 embedded IPv6 addresses


sukunrt:
When implementing happy eyeballs logic to prefer QUIC IPv6 over QUIC IPv4 addresses, I've observed that a lot of the dial failures to IPv6 addresses are for addresses of this form `64:ff9b::192.0.2.33`. These are https://datatracker.ietf.org/doc/html/rfc6052#section-3.1 addresses. 

In most cases these nodes also report `192.0.2.33` as their address and the subsequent dial to that address succeeds. 
Can we just skip the dial to `64:ff9b::192.0.2.33`? 

If we cannot skip them, we should probably still rank them the same as IPv4 addresses and not prefer them to other IPv4 addresses.

Jorropo:
It doesn't make sense for peoples to advertise `64:ff9b::192.0.2.33` on windows, linux, android, ios and macos (I havn't checked the rest) the os takes care of rewriting IPv4 addresses to `64:ff9b::` ones if you are using NAT64,
I think what should happen is that when adding addresses to the peerstore if an `ip6` component is in the `64:ff9b::/96` range multiaddr should actually store or load it as an `ip4` component.

Note: I don't know who advertise this in the first place, might be a bug somewhere (golang encode ipv4 as 16 bytes `64:ff9b::/96`).

sukunrt:
What's happening here is: 
_A_ is behind a NAT64 router _R_. _B_ is a publicly reachable IPv4 only node with IPv4 addr 192.0.2.33

_A_ connects to _B_ via NAT64 router _R_
From _A_s perspective it sees the address of B as 64:ff9b::192.0.2.33. It informs B about this address via identify. B will believe A that its address is this IPv6 address.

Note: B shouldn't believe this, but that's a separate issue in identify. Will open a separate issue for that.

@Jorropo Can you explain how the OS takes care of the rewrite? Let's only consider linux and go-libp2p.
My understanding so far is that this requires some DNS64 like system for _A_ to get a IPv4-converted IPv6 address like 64:ff9b::192.0.2.33

I ask because according to my understanding, _A_ does need the address in the converted form to make the call to _B_. However if the system can be configured to do this automatically it is safe to drop this.

sukunrt:
Not taking this up for #2365 . We can handle this case in a follow up PR.

Jorropo:
[...]

PR #2365, opened 2023-06-14
swarm: implement Happy Eyeballs ranking

sukunrt:
Depends on #2320 

Ignores https://github.com/libp2p/go-libp2p/issues/2349 We can take that up after this.

### DIFF
--- a/p2p/net/swarm/dial_ranker.go
+++ b/p2p/net/swarm/dial_ranker.go
@@ -12,11 +12,11 @@ import (
 
 // The 250ms value is from happy eyeballs RFC 8305. This is a rough estimate of 1 RTT
 const (
-	// duration by which TCP dials are delayed relative to QUIC dial
+	// duration by which TCP dials are delayed relative to the last QUIC dial
 	PublicTCPDelay  = 250 * time.Millisecond
 	PrivateTCPDelay = 30 * time.Millisecond
 
-	// duration by which QUIC dials are delayed relative to first QUIC dial
+	// duration by which QUIC dials are delayed relative to previous QUIC dial
 	PublicQUICDelay  = 250 * time.Millisecond
 	PrivateQUICDelay = 30 * time.Millisecond
 
@@ -31,11 +31,10 @@ func NoDelayDialRanker(addrs []ma.Multiaddr) []network.AddrDelay {
 
 // DefaultDialRanker determines the ranking of outgoing connection attempts.
 //
-// Addresses are grouped into four distinct groups:
+// Addresses are grouped into three distinct groups:
 //
 //   - private addresses (localhost and local networks (RFC 1918))
-//   - public IPv4 addresses
-//   - public IPv6 addresses
+//   - public addresses
 //   - relay addresses
 //
 // Within each group, the addresses are ranked according to the ranking logic described below.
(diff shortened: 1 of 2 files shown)


p2p/net/swarm/dial_ranker.go
```diff
@@ -77,70 +75,92 @@ func DefaultDialRanker(addrs []ma.Multiaddr) []network.AddrDelay {
...
+// documentation for defaultDialRanker.
 // offset is used to delay all addresses by a fixed duration. This is useful for delaying all relay
-// addresses relative to direct addresses
+// addresses relative to direct addresses.
 func getAddrDelay(addrs []ma.Multiaddr, tcpDelay time.Duration, quicDelay time.Duration,
 	offset time.Duration) []network.AddrDelay {
+
 	sort.Slice(addrs, func(i, j int) bool { return score(addrs[i]) < score(addrs[j]) })
 
+	// If the first address is (QUIC, IPv6), make the second address (QUIC, IPv4).
```
marten-seemann, line 94: Why does this only apply to QUIC? If a peer has QUIC disabled, we should still do Happy Eyeballs for it's TCP6 and TCP4 address, no?

sukunrt: We can, provided we add a delay of 750ms between the two. I was not sure if we wanted that so I've done it just for QUIC.

marten-seemann: Ok, let's do this once https://github.com/libp2p/go-libp2p/issues/2394 is resolved then.

p2p/net/swarm/dial_ranker.go
```diff
@@ -77,70 +75,92 @@ func DefaultDialRanker(addrs []ma.Multiaddr) []network.AddrDelay {
...
+	happyEyeballs := false
+	if len(addrs) > 0 {
+		if isQUICAddr(addrs[0]) && isProtocolAddr(addrs[0], ma.P_IP6) {
+			for i := 1; i < len(addrs); i++ {
+				if isQUICAddr(addrs[i]) && isProtocolAddr(addrs[i], ma.P_IP4) {
+					addrs[i], addrs[1] = addrs[1], addrs[i]
```
marten-seemann, line 100: Is swapping the right thing to do here, or should the address be inserted at position 1, shifting all other addresses?

p2p/net/swarm/dial_ranker.go
```diff
@@ -77,70 +75,92 @@ func DefaultDialRanker(addrs []ma.Multiaddr) []network.AddrDelay {
...
+			}
+		}
+	}
+
 	res := make([]network.AddrDelay, 0, len(addrs))
-	quicCount := 0
-	for _, a := range addrs {
-		delay := offset
+
+	var totalTCPDelay time.Duration = 0
```
marten-seemann, line 110:
```suggestion
	var totalTCPDelay time.Duration
```

p2p/net/swarm/dial_ranker.go
```diff
@@ -77,70 +75,92 @@ func DefaultDialRanker(addrs []ma.Multiaddr) []network.AddrDelay {
...
+	for i, addr := range addrs {
+		var delay time.Duration = 0
```
marten-seemann, line 112:
```suggestion
		var delay time.Duration
```

p2p/net/swarm/dial_ranker.go
```diff
@@ -77,70 +75,92 @@ func DefaultDialRanker(addrs []ma.Multiaddr) []network.AddrDelay {
...
+// QUICv1 IPv6 > QUICdraft29 IPv6 > QUICv1 IPv4 > QUICdraft29 IPv4 >
+// WebTransport IPv6 > WebTransport IPv4 > TCP IPv6 > TCP IPv4
 func score(a ma.Multiaddr) int {
-	// the lower 16 bits of the result are the relavant port
-	// the higher bits rank the protocol
-	// low ports are ranked higher because they're more likely to
-	// be listen addresses
+	ip4Weight := 0
+	if isProtocolAddr(a, ma.P_IP4) {
+		ip4Weight = (1 << 18)
```
marten-seemann, line 143:
```suggestion
		ip4Weight = 1 << 18
```

marten-seemann (approved):
lgtm, modulo the question regarding swapping vs shifting

p2p/net/swarm/dial_ranker.go
sukunrt: There's no impact because we have three tiers: 
IPv6 dialed first, IPv4 dialed second, rest of the addresses. 
I'll change it to shifting though because that seems more correct in case we eventually want to dial only a single address at a time.

sukunrt: updated

PR #2370, opened 2023-06-15
rcmgr: register prometheus metrics with the libp2p registerer

sukunrt:
If the caller forgot to call rcmgrObs.MustRegisterWith, we will still get the metrics. If they have a StatsTraceReporter, they're anyway updating the counters. If they didn't provide a StatsTraceReporter, we will have registered a few extra metrics with prometheus. 

fixes: https://github.com/libp2p/go-libp2p/issues/2371


marten-seemann (commented):
Do we need to update in the rcmgr dashboard README that the user doesn't have to call `MustRegisterWith` any more?

marten-seemann (approved):

MarcoPolo (approved):

PR #2373, opened 2023-06-16
fix: in the swarm move Connectedness emit after releasing conns

Jorropo:
go-libp2p-kad-dht now listen to both EvtPeerIdentificationCompleted and EvtPeerConnectednessChanged and EvtPeerIdentificationCompleted calls .ConnsToPeer inorder to do some filtering.

However it happens that it deadlocks because if the swarm is trying to emit a EvtPeerConnectednessChanged while the subscriber is trying to process an EvtPeerIdentificationCompleted, the subscriber is stuck on s.conns.RLock() while the swarm wont release it before having sent EvtPeerConnectednessChanged. Deadlock !

I havn't confirmed this fixes my bug given this takes time to reproduce, I'll startup a new experiment later.

### DIFF
--- a/p2p/net/swarm/swarm.go
+++ b/p2p/net/swarm/swarm.go
@@ -353,8 +353,8 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
 	c.notifyLk.Lock()
 	s.conns.Unlock()
 
-	// Emit event after releasing s.conns else a subscriber that listen boths on
-	// event.EvtPeerConnectednessChanged can't call any of the swarm method that read or write.
+	// Emit event after releasing `s.conns` lock so that a consumer can still
+	// use swarm methods that need the `s.conns` lock.
 	if isFirstConnection {
 		s.emitter.Emit(event.EvtPeerConnectednessChanged{
 			Peer:          p,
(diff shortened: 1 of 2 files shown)


Jorropo:
Yes, this has happened to me

sukunrt:
I agree this should be fixed, just want to get the scenario right. 

- You have a subscription for both `EvtPeerIdentificationCompleted` and `EvtPeerConnectednessChanged`. 
- You receive the identification completed event and try to obtain a swarm.conns.RLock
- The goroutine cannot acquire the lock because another goroutine has the lock which is going to push connectedness changed event. This goroutine cannot push the event because the queue of the same subscriber is full. 

Does this sound correct?

Jorropo:
@sukunrt what you described is exactly what happens right now.
With my patch the goroutine will first release conns before emiting, so the RLock is taken, the consumer process events and then consume events unlocking the second goroutine.
The only way my patch fail is if there are 2 goroutines trying to emit an event on the same stream, the first goroutine is stuck on emiting on a full channel the second one is blocked on notifylk and the consumer of the channel is blocked on conns (which hasn't been released yet) however we can confirm experimentally this does not happen because the dht v0.23.0 and v0.24.0 used notifies which already is emited synchronously inside the notifylk region and yet everything is fine and thoses versions to not show this behaviour.

MarcoPolo:
Great work @Jorropo! I didn't consider how a consumer could try to grab a lock that the publisher has and thus create a deadlock.

Publishers should only grab locks to guarantee the ordering of connect/disconnect events. I don't think it needs the swarm.conns lock, though I'll double check the code path.

I can help here and add some tests.

Jorropo:
@MarcoPolo I don't have time to finish any of this, feel free to fix ci and add tests thx

MarcoPolo:
Test added and fixed a related deadlock (on the disconnect side of things).

MarcoPolo (approved):

MarcoPolo:
Once CI passes I'll merge a patch release

PR #2374, opened 2023-06-16
Release v0.27.7

MarcoPolo:
Fixes a deadlock that can occur when consumers don't pull from an event channel because they're waiting for the swarm conns lock.


marten-seemann:
Wrong order of commits, not sure if it matters. Would be safer to put the release commit at the end.

MarcoPolo:
Will squash

MarcoPolo:
We agreed on this over a sync call.

PR #2375, opened 2023-06-16
Release v0.28.1

MarcoPolo:
Fixes a deadlock that can occur when consumers don't pull from an event channel because they're waiting for the swarm conns lock.


marten-seemann:
Same here, would be safer to put the version commit last.

marten-seemann:
Fixed the commit order, and updated the release notes.

PR #2382, opened 2023-06-19
identify: set stream deadlines for Identify and Identify Push streams

### LINKED ISSUE #2379: identify: stuck at reading multistream header
I recently updated [`Nebula`](https://github.com/dennis-tra/nebula) to use `go-libp2p` v0.28.0. Further, when I crawl another peer, I started to explicitly wait for the Identify exchange to complete before extracting data from the peerstore.

Over the weekend, I saw that two crawls, which usually take ~5m, didn't terminate after >12h. I extracted a `goroutine` dump which is attached below.

From that dump, I found out that one of the 1,000 crawl-workers is waiting to receive an event from a channel:

image

I dug deeper into what it's waiting for and found that it should be [this `select`- statement](https://github.com/dennis-tra/nebula/blob/2317f90c8fbdba9107b2f36503df842b0e1d9532/pkg/crawl/crawl_p2p.go#L229). Here the excerpt:

```go
select {
case <-ctx.Done():
   // ...
case <-c.host.IDService().IdentifyWait(conn):
   // ...
}
```

So, it's waiting for the `identify` exchange to complete. If I interpret the goroutine dump correctly, it's hanging at:

- [`SelectProtoOrFail`](https://github.com/libp2p/go-libp2p/blob/5e3fc8aedfdba9679921f9bb2d12f953d9e7a873/p2p/protocol/identify/id.go#L425)
- [`readMultistreamHeader`](https://github.com/multiformats/go-multistream/blob/5b2f48e2db22aa1db90bb5a21aa4e6702d4ee33c/client.go#L66) (`go-multistream`)
- [`ReadNextToken`](https://github.com/multiformats/go-multistream/blob/5b2f48e2db22aa1db90bb5a21aa4e6702d4ee33c/client.go#LL331C14-L331C27) (`go-multistream`)

It seems like it can't read the multistream header from the identify stream. My hypotheses: 1) the remote peer is just super slow to respond, or 2) something hung internally, or 3) something else?

I had expected some of the various timeouts across the stack to kick in (transport dial timeouts, security handshake timeouts, connection timeouts, etc.) and cancel the exchange.

My fix, for now, is to time out after 15s:

```go
timeoutCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
defer cancel()

select {
case <-timeoutCtx.Done():
   // ...
case <-c.host.IDService().IdentifyWait(conn):
   // ...
}
```

However, I think this leaks resources if the above happens again.

Hyper speculative: If this proves to be not and issue on my end, could this be abused in a slowloris-like fashion?

---

[goroutine.zip](https://github.com/libp2p/go-libp2p/files/11782197/goroutine.zip)

marten-seemann:
Fixes #2379.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -47,8 +47,7 @@ const ServiceName = "libp2p.identify"
 
 const maxPushConcurrency = 32
 
-// StreamReadTimeout is the read timeout on all incoming Identify family streams.
-var StreamReadTimeout = 60 * time.Second
+var Timeout = 60 * time.Second // timeout on all incoming Identify interactions
 
 const (
 	legacyIDSize = 2 * 1024 // 2k Bytes
@@ -410,11 +409,14 @@ func (ids *idService) IdentifyWait(c network.Conn) <-chan struct{} {
 }
 
 func (ids *idService) identifyConn(c network.Conn) error {
-	s, err := c.NewStream(network.WithUseTransient(context.TODO(), "identify"))
+	ctx, cancel := context.WithTimeout(context.Background(), Timeout)
+	defer cancel()
+	s, err := c.NewStream(network.WithUseTransient(ctx, "identify"))
 	if err != nil {
 		log.Debugw("error opening identify stream", "peer", c.RemotePeer(), "error", err)
 		return err
 	}
+	s.SetDeadline(time.Now().Add(Timeout))
 
 	if err := s.SetProtocol(ID); err != nil {
 		log.Warnf("error setting identify protocol for stream: %s", err)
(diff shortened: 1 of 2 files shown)


p2p/protocol/identify/id.go
```diff
@@ -47,13 +47,11 @@ const ServiceName = "libp2p.identify"
...
-var StreamReadTimeout = 60 * time.Second
-
 const (
-	legacyIDSize = 2 * 1024 // 2k Bytes
-	signedIDSize = 8 * 1024 // 8K
-	maxMessages  = 10
+	legacyIDSize    = 2 * 1024 // 2k Bytes
+	signedIDSize    = 8 * 1024 // 8K
+	maxMessages     = 10
+	identifyTimeout = 60 * time.Second // imeout on all incoming Identify interactions
```
dennis-tra, line 54:
```suggestion
	identifyTimeout = 60 * time.Second // timeout on all incoming Identify interactions
```

PR #2383, opened 2023-06-20
metrics: make it possible to spin up Grafana using docker-compose

marten-seemann:
As a last step to finish our metrics story, here's an easy way to make them more useful for local development.Just run
```bash
docker-compose up
```

and your local Grafana instance is up and running, collecting metrics from your libp2p node! See README for more instructions.

### DIFF
--- a/dashboards/README.md
+++ b/dashboards/README.md
@@ -25,7 +25,7 @@ Dashboards can be inspected locally by running
 docker-compose up
 ```
 
-and opening Grafana at http://localhost:3000 (user: admin, password: admin).
+and opening Grafana at http://localhost:3000.
 
 
 ### Making Dashboards usable with Provisioning
--- a/dashboards/README.md
+++ b/dashboards/README.md
@@ -33,4 +33,3 @@ and opening Grafana at http://localhost:3000 (user: admin, password: admin).
 The following section is only relevant for creators of dashboards.
 
 Due to a bug in Grafana, it's not possible to provision dashboards shared for external use directly. We need to apply the workaround described in https://github.com/grafana/grafana/issues/10786#issuecomment-568788499 (adding the a few lines in the dashboard JSON file).
->>>>>>> Stashed changes
--- a/dashboards/README.md
+++ b/dashboards/README.md
@@ -2,3 +2,35 @@
 
 This directory contains prebuilt dashboards (provided as JSON files) for various components.
 For steps on how to import and use them [please read the official Grafana documentation.](https://grafana.com/docs/grafana/latest/dashboards/export-import/#import-dashboard)
+
+## Using locally
+
+For local development and debugging, it can be useful to spin up a local Prometheus and Grafana instance.
+
+To expose metrics, we first need to expose a metrics collection endpoint. Add this to your code:
+
+```go
+import "github.com/prometheus/client_golang/prometheus/promhttp"
+
+go func() {
+    http.Handle("/debug/metrics/prometheus", promhttp.Handler())
+    log.Fatal(http.ListenAndServe(":5001", nil))
+}()
+```
+
+This exposes a metrics collection endpoint at http://localhost:5001/debug/metrics/prometheus. Note that this is the same endpoint that [Kubo](https://github.com/ipfs/kubo) uses, so if you want to gather metrics from Kubo, you can skip this step.
+
+Dashboards can be inspected locally by running
+```bash
+docker-compose up
+```
+
+and opening Grafana at http://localhost:3000 (user: admin, password: admin).
+
+
+### Making Dashboards usable with Provisioning
+
+The following section is only relevant for creators of dashboards.
+
+Due to a bug in Grafana, it's not possible to provision dashboards shared for external use directly. We need to apply the workaround described in https://github.com/grafana/grafana/issues/10786#issuecomment-568788499 (adding the a few lines in the dashboard JSON file).
+>>>>>>> Stashed changes
--- a/dashboards/autonat/autonat.json
+++ b/dashboards/autonat/autonat.json
@@ -691,6 +691,16 @@
   "tags": [],
   "templating": {
     "list": [
+       {
+        "hide": 0,
+        "label": "datasource",
+        "name": "DS_PROMETHEUS",
+        "options": [],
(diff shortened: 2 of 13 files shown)


dashboards/README.md
```diff
@@ -2,3 +2,35 @@
...
+
+and opening Grafana at http://localhost:3000 (user: admin, password: admin).
+
+
+### Making Dashboards usable with Provisioning
+
+The following section is only relevant for creators of dashboards.
+
+Due to a bug in Grafana, it's not possible to provision dashboards shared for external use directly. We need to apply the workaround described in https://github.com/grafana/grafana/issues/10786#issuecomment-568788499 (adding the a few lines in the dashboard JSON file).
+>>>>>>> Stashed changes
```
p-shahi, line 36:
```suggestion
```

p-shahi (approved):

dashboards/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+      - "3000:3000"
+    volumes:
+      - ./dashboard.yml:/etc/grafana/provisioning/dashboards/main.yml
+      - ./datasources.yml:/etc/grafana/provisioning/datasources/prom.yml
+      - ./autonat/autonat.json:/var/lib/grafana/dashboards/autonat.json
+      - ./autorelay/autorelay.json:/var/lib/grafana/dashboards/autorelay.json
+      - ./eventbus/eventbus.json:/var/lib/grafana/dashboards/eventbus.json
+      - ./identify/identify.json:/var/lib/grafana/dashboards/identify.json
+      - ./relaysvc/relaysvc.json:/var/lib/grafana/dashboards/relaysvc.json
+      - ./swarm/swarm.json:/var/lib/grafana/dashboards/swarm.json
```
sukunrt, line 23: If you want to allow anonymous viewing without login you can use: 
```
    environment:
      - GF_AUTH_DISABLE_LOGIN_FORM=true
      - GF_AUTH_ANONYMOUS_ENABLED=true
      - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
```
as done here: 
https://github.com/libp2p/go-libp2p/blob/master/examples/metrics-and-dashboards/compose.yml

I think it'd be nice to have this.

marten-seemann: Added.

marten-seemann:
@sukunrt Have you tried running this on your machine? Would be good to get a check that this actually works. Ideally, running a libp2p node with the code snippet in the README (not a Kubo node).

dashboards/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
+    volumes:
+      - ./dashboard.yml:/etc/grafana/provisioning/dashboards/main.yml
+      - ./datasources.yml:/etc/grafana/provisioning/datasources/prom.yml
+      - ./autonat/autonat.json:/var/lib/grafana/dashboards/autonat.json
+      - ./autorelay/autorelay.json:/var/lib/grafana/dashboards/autorelay.json
+      - ./eventbus/eventbus.json:/var/lib/grafana/dashboards/eventbus.json
+      - ./identify/identify.json:/var/lib/grafana/dashboards/identify.json
+      - ./relaysvc/relaysvc.json:/var/lib/grafana/dashboards/relaysvc.json
+      - ./swarm/swarm.json:/var/lib/grafana/dashboards/swarm.json
```
sukunrt, line 27: Add hole punch metrics too here.

sukunrt (commented):
I just ran it. It works find. Picked up metrics for my locally running Kubo instance. We should also add the hole punch metrics dashboard. 

There's one thing though, this also happened when using the metrics and dashboard example. If you are logged in to a grafana running locally(i was running one via homebrew), you'll get an "Unauthorized" error and you'll need to clear your browser cache to make it work. In hindsight, it's obvious what's happening but the first time this happened to me, I spent about half an hour figuring out what's wrong.

dashboards/docker-compose.yml
marten-seemann: Oops, looks like I forgot that one.

sukunrt (approved):

dashboards/autonat/autonat.json
```diff
@@ -691,6 +691,16 @@
   "tags": [],
   "templating": {
     "list": [
+       {
```
MarcoPolo, line 694: Didn’t know you could do this. Cool!

dashboards/prometheus.yml
```diff
@@ -0,0 +1,30 @@
...
+    - localhost:9090
+- job_name: libp2p
+  honor_timestamps: true
+  scrape_interval: 15s
+  scrape_timeout: 10s
+  metrics_path: /debug/metrics/prometheus
+  scheme: http
+  static_configs:
+  - targets:
+    - host.docker.internal:5001
```
MarcoPolo, line 30: This only works on macos. This will fail on linux

MarcoPolo:
Should we remove the old example then?

dashboards/prometheus.yml
marten-seemann: How would this work on Linux?

marten-seemann:
> Should we remove the old example then?

Makes sense. Can you open a PR?

dashboards/datasources.yml
```diff
@@ -0,0 +1,13 @@
...
+deleteDatasources:
+  - name: Prometheus
+    orgId: 1
+
+datasources:
+  - name: Prometheus
+    orgId: 1
+    type: prometheus
+    access: proxy
+    url: http://host.docker.internal:9090
```
MarcoPolo, line 12: This should be `prometheus:9090`

dashboards/prometheus.yml
MarcoPolo: there's no easy way unfortunately. You can use https://github.com/qoomon/docker-host but that requires binding the host thing to 0.0.0.0 (or at least the bridge interface).

You can do `network_mode=host`, but then that doesn't work for MacOS environmnets.

marten-seemann: I opened #2387.

PR #2390, opened 2023-06-23
webtransport: fix flaky TestListenAddrValidity

sukunrt:
This fails because some other process/goroutine is listening on 1234


marten-seemann:
We're now using 0 (#2426).

PR #2391, opened 2023-06-24
docs: fix some comments errors

sysvm:
1. fix some comments errors such as: misspelling, syntax error and so on.
2. extract a string const.

### DIFF
--- a/core/pnet/codec.go
+++ b/core/pnet/codec.go
@@ -31,7 +31,7 @@ func expectHeader(r *bufio.Reader, expected []byte) error {
 		return err
 	}
 	if !bytes.Equal(header, expected) {
-		return fmt.Errorf("expected file header %s, got: %s", pathPSKv1, header)
+		return fmt.Errorf("expected file header %s, got: %s", expected, header)
 	}
 	return nil
 }
--- a/examples/multipro/node.go
+++ b/examples/multipro/node.go
@@ -35,7 +35,7 @@ func NewNode(host host.Host, done chan bool) *Node {
 }
 
 // Authenticate incoming p2p message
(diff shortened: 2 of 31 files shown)


examples/multipro/node.go
```diff
@@ -35,15 +35,15 @@ func NewNode(host host.Host, done chan bool) *Node {
...
+// message: a protobuf go data object
 // data: common p2p message data
 func (n *Node) authenticateMessage(message proto.Message, data *p2p.MessageData) bool {
 	// store a temp ref to signature and remove it from message data
 	// sign is a string to allow easy reset to zero-value (empty string)
 	sign := data.Sign
 	data.Sign = nil
 
-	// marshall data without the signature to protobufs3 binary format
+	// marshall data without the signature to protobuf binary format
```
p-shahi, line 46: haven't looked but if it's actually marshaling to proto3 we may want to keep that distinction

sysvm: fixed

core/pnet/codec.go
```diff
@@ -31,7 +31,7 @@ func expectHeader(r *bufio.Reader, expected []byte) error {
 		return err
 	}
 	if !bytes.Equal(header, expected) {
-		return fmt.Errorf("expected file header %s, got: %s", pathPSKv1, header)
+		return fmt.Errorf("expected file header %s, got: %s", expected, header)
```
p-shahi, line 34: Makes sense to change this to expected but I am not the expert here, cc @marten-seemann

p-shahi (approved):
Thank you

marten-seemann (approved):

PR #2392, opened 2023-06-26
identify: filter nat64 well-known prefix ipv6 addresses

### LINKED ISSUE #2349: Handle dial prioritisation for well known prefix IPv4 embedded IPv6 addresses

sukunrt:
fixes #2349 

For details see https://github.com/libp2p/go-libp2p/issues/2349#issuecomment-1608033872

### DIFF
--- a/p2p/protocol/identify/obsaddr.go
+++ b/p2p/protocol/identify/obsaddr.go
@@ -3,7 +3,6 @@ package identify
 import (
 	"context"
 	"fmt"
-	"strings"
 	"sync"
 	"time"
 
@@ -377,7 +376,7 @@ func shouldRecordObservation(host addrsProvider, network listenAddrsProvider, co
 	}
 
 	// Provided by NAT64 peers, these addresses are specific to the peer and not publicly routable
-	if isNAT64IPv4ConvertedIPv6Addr(observed) {
+	if manet.IsNAT64IPv4ConvertedIPv6Addr(observed) {
 		return false
 	}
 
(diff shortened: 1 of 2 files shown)


p2p/protocol/identify/obsaddr.go
```diff
@@ -435,6 +441,20 @@ func shouldRecordObservation(host addrsProvider, network listenAddrsProvider, co
 	return true
 }
 
+// isNAT64IPv4ConvertedIPv6Addr returns whether addr is an IPv6 address that begins with
+// the well known prefix "64:ff9b" used for NAT64 Translation
+// see RFC 6052
+func isNAT64IPv4ConvertedIPv6Addr(addr ma.Multiaddr) bool {
```
marten-seemann, line 447: Should this be part of the `manet` package?

p2p/protocol/identify/obsaddr.go
```diff
@@ -435,6 +441,20 @@ func shouldRecordObservation(host addrsProvider, network listenAddrsProvider, co
...
+	ip, err := addr.ValueForProtocol(ma.P_IP6)
+	if err != nil {
+		return false
+	}
+	if strings.HasPrefix(ip, "64:ff9b::") {
```
marten-seemann, line 452: Is string matching the best way to do this? Would be nicer to match on the underlying bytes (and probably also causes fewer allocs?).

p2p/protocol/identify/obsaddr.go
sukunrt: will move.

p2p/protocol/identify/obsaddr.go
marten-seemann: Ok, let's keep it for now, but we should keep this is mind for our multiaddrs API redesign. It's absurd that we return IPs as strings.

Jorropo (approved):
NAT64 adresses shouldn't be advertised, this makes sense to me.

sukunrt:
blocked on https://github.com/multiformats/go-multiaddr/pull/205

sukunrt:
updated.

marten-seemann (approved):

PR #2393, opened 2023-06-26
swarm: rank nat64 well-known prefix addresses lower

sukunrt:
depends on #2365 

fixes https://github.com/libp2p/go-libp2p/issues/2349

I'm fine with dropping this if this seems too complex/arbitrary. With #2392 we'll stop receiving these addresses. For details see https://github.com/libp2p/go-libp2p/issues/2349#issuecomment-1608033872

Update: Converting to draft. Will open again once #2394 is merged.

### DIFF
--- a/dashboards/swarm/swarm.json
+++ b/dashboards/swarm/swarm.json
@@ -37,6 +37,18 @@
     },
     {
       "type": "panel",
+      "id": "stat",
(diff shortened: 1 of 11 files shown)


sukunrt:
Not a pressing issues. Correctly filtering NAT64 has already been handled.

PR #2403, opened 2023-06-29
swarm: make black hole detection configurable

### LINKED ISSUE #2398: swarm: add config options for blackhole detection
We need a config option for this. Ideally, the user could configure the thresholds. Doesn't need to be part of this PR, but we'd need this for the next release.

_Originally posted by @marten-seemann in https://github.com/libp2p/go-libp2p/pull/2320#discussion_r1243969672_

sukunrt:
fixes #2398

### DIFF
--- a/options.go
+++ b/options.go
@@ -23,6 +23,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/transport"
 	"github.com/libp2p/go-libp2p/p2p/host/autorelay"
 	bhost "github.com/libp2p/go-libp2p/p2p/host/basic"
+	"github.com/libp2p/go-libp2p/p2p/net/swarm"
 	tptu "github.com/libp2p/go-libp2p/p2p/net/upgrader"
 	relayv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
 	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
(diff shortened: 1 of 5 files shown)


p2p/net/swarm/swarm.go
```diff
@@ -108,6 +108,28 @@ func WithDialRanker(d network.DialRanker) Option {
 	}
 }
 
+// WithUDPBlackHoleConfig configures swarm to use c as the config for UDP black hole detection
+func WithUDPBlackHoleConfig(c *BlackHoleConfig) Option {
```
marten-seemann, line 112: Alternatively, you could just make these 3 function parameters:
```go
func WithUDPBlackHoleConfig(enabled bool, n, min int)
```
and then make `BlackHoleConfig` an unexported struct.

options.go
```diff
@@ -587,3 +588,25 @@ func DialRanker(d network.DialRanker) Option {
...
+// IPv6BlackHoleConfig configures libp2p to use c as the config for IPv6 black hole detection
+func IPv6BlackHoleConfig(c *swarm.BlackHoleConfig) Option {
+	return func(cfg *Config) error {
+		if c == nil {
+			return errors.New("ipv6 black hole config cannot be nil")
+		}
+		cfg.IPv6BlackHoleConfig = c
+		return nil
+	}
+}
```
marten-seemann, line 612: What about making this more general by having a `SwarmOpts(opts ...swarm.Option) Option` here?

sukunrt: Your suggestion is right. I did it this way because none of the other swarm.Options were exposed here. We should also fix the other cases like this one https://github.com/libp2p/go-libp2p/blob/master/config/config.go#L176

marten-seemann (approved):

PR #2385, opened 2023-06-21
Revert "metrics: add example for metrics and dashboard (#2232)"

MarcoPolo:
This reverts commit 22c45a11bb15139a0e3e18eb1e3cdb10c8cd60b9. See https://github.com/libp2p/go-libp2p/pull/2383#issuecomment-1601681566 for context

PR #2404, opened 2023-06-29
metrics: deduplicate code in examples

MarcoPolo:
Closes https://github.com/libp2p/go-libp2p/pull/2385

This changes the example to reuse code defined in `dashboards/`. The example is still helpful because it's a full example on how metrics are setup and lets you see some metrics with a single command.


dashboards/datasources.yml
```diff
@@ -9,5 +9,5 @@ datasources:
     orgId: 1
     type: prometheus
     access: proxy
-    url: http://host.docker.internal:9090
+    url: http://prometheus:9090
```
marten-seemann, line 12: I don't think this works with host mode, does it? Have you tried it on Linux?

MarcoPolo: You're right. I've update the linux compose file to make this work. Tested on linux.

marten-seemann (approved):

PR #2410, opened 2023-07-06
transport integration tests: Make TestMoreStreamsThanOurLimits less flaky

MarcoPolo:
... by

1. Counting on the initiator side (less racy and we are already waiting on a waitgroup).
2. Slowly ramping up concurrency while we haven't seen an error, but if we see an error we'll stop increasing concurrency. Ideally the transport would block us on this but that's not how this currently works.

The goal of this test is to make sure that if a client tries deal with many streams in a reasonable way, the transport won't blow up.


p2p/test/transport/transport_test.go
```diff
@@ -368,10 +368,15 @@ func TestMoreStreamsThanOurLimits(t *testing.T) {
 			}))
 
 			var handledStreams atomic.Int32
+			var sawFirstErr atomic.Bool
```
marten-seemann, line 371: This `atomic.Bool` is only ever loaded, but never set. Or am I missing something?

marten-seemann (changes requested):

marten-seemann (approved):

Issue #2380, opened 2023-06-19
Dial worker loop simplification. Move backoff filtering together with other filters. 


sukunrt:
This is a small simplification that I believe will fix tricky points in the worker loop. 

We should do backoff filtering along with other address filtering in `swarm.filterKnownUndialables`
This will clean up some of the awkward error handling [here](https://github.com/libp2p/go-libp2p/blob/v0.27.0/p2p/net/swarm/dial_worker.go#L277) and [here](https://github.com/libp2p/go-libp2p/blob/v0.27.0/p2p/net/swarm/dial_worker.go#L229)

see https://github.com/libp2p/go-libp2p/pull/2320#issuecomment-1597009229 
for how black hole detection can be made to work with this. 

As a follow up I'd like to change the worker loop interface to just take the list of addresses to dial with the ranking and the address filtering and ranking all happens before the worker loop.

PR #2416, opened 2023-07-09
swarm: remove unnecessary reqno for pending request tracking

sukunrt:
First merge this to backoff cleanup and then we can merge both to master. 

Opening small PRs for individual items since this piece is complicated

### DIFF
--- a/p2p/net/swarm/dial_worker.go
+++ b/p2p/net/swarm/dial_worker.go
@@ -48,7 +48,7 @@ type pendRequest struct {
 	// At the time of creation addrs is initialised to all the addresses of the peer. On a failed dial,
 	// the addr is removed from the map and err is updated. On a successful dial, the dialRequest is
 	// completed and response is sent with the connection
-	addrs map[string]struct{}
+	addrs map[string]bool
 }
 
 // addrDial tracks dials to a particular multiaddress.
@@ -61,9 +61,6 @@ type addrDial struct {
 	conn *Conn
 	// err is the err on dialing the address
 	err error
-	// requests is the list of pendRequests interested in this dial
-	// the value in the slice is the request number assigned to this request by the dialWorker
-	requests []int
 	// dialed indicates whether we have triggered the dial to the address
 	dialed bool
 	// createdAt is the time this struct was created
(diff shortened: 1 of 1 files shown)


p2p/net/swarm/dial_worker.go
```diff
@@ -326,20 +315,14 @@ loop:
 					continue loop
 				}
 
-				// request succeeded, respond to all pending requests
-				for _, reqno := range ad.requests {
-					pr, ok := w.pendingRequests[reqno]
-					if !ok {
-						// some other dial for this request succeeded before this one
-						continue
+				for pr := range w.pendingRequests {
```
marten-seemann, line 318: This loop will now be executed in random order. Is that ok?

p2p/net/swarm/dial_worker.go
```diff
@@ -367,32 +350,25 @@ loop:
...
 func (w *dialWorker) dispatchError(ad *addrDial, err error) {
 	ad.err = err
-	for _, reqno := range ad.requests {
-		pr, ok := w.pendingRequests[reqno]
-		if !ok {
-			// some other dial for this request succeeded before this one
-			continue
-		}
-
+	for pr := range w.pendingRequests {
```
marten-seemann, line 353: Same here.

sukunrt: Yeah. Assuming we have two go routines waiting. 
1. We don't actually guarantee ordered delivery. This just publishes to a channel. This channel is being read [here](https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/dial_sync.go#L62). So the waiting go routines may be run in a different order.
2. This doesn't block. The channel has 1 capacity, so both the waiters are notified immediately. 

I don't know of any case where this might be an issue especially considering 1. 

Also [here](https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/dial_worker.go#L398) we do provide an out of order delivery. In this case the first requester will be informed of the connection after the second requester(which added a valid address) was informed.

p2p/net/swarm/dial_worker.go
sukunrt: see https://github.com/libp2p/go-libp2p/pull/2416/files/06199d786a8323515bd18ba3e3ccb7b0db5704f8#r1258151128

marten-seemann (approved):
LGTM, this looks like a nice simplification. I don't think it resolves #2380 though.

PR #2417, opened 2023-07-09
swarm: implement Happy Eyeballs for TCP

### LINKED ISSUE #2394: Implement Happy Eyeballs for TCP
We currently do not do any prioritisation among TCP addresses when dialing. This is because TCP connection establishment takes too long(3*RTT) and the delay required between dials for effective prioritisation would be roughly 750ms - 1s. This delay is too high in case we fail to establish a connection with the first tcp address we try. 

We can get around this if we setup a system where the worker loop is notified when a TCP connection is established and security negotiation is in progress.

sukunrt:
Closes #2394

### DIFF
--- a/core/transport/transport.go
+++ b/core/transport/transport.go
@@ -5,6 +5,7 @@ package transport
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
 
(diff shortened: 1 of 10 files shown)


sukunrt:
closing in favour of #2573

PR #2424, opened 2023-07-11
quic: Update to quic-go v0.36.2

MarcoPolo:
quic-go changed its [API in v0.35](https://github.com/quic-go/quic-go/releases/tag/v0.35.0), this PR updates our usage.

Most of the changes in this PR are due to renaming things from conn -> transport. Previously we could only share a underlying socket by sharing the "connection" between quic-go dialers/listeners. quic-go v0.35.0 introduces a `Transport` that represents QUIC on a socket. This `Transport` is what is shared between different dialers/listeners to reuse the same socket. I think that's a better abstraction since it's clear when something is getting shared vs not.

A lot of the renaming is to make clear what is happening. For example, `reuse.Listen` is now called `reuse.TransportForListen` since what we actually get back from the call is a `quic.Transport`. Similarly most things that were previously referencing a "conn" (as in packetconn) now reference a `transport`. Hopefully this also clears up any confusion about different "connection" like objects.


p2p/transport/quicreuse/connmgr.go
```diff
@@ -47,17 +49,23 @@ func NewConnManager(statelessResetKey quic.StatelessResetKey, opts ...Option) (*
...
 	var tracers []quiclogging.Tracer
-	if qlogTracer != nil {
-		tracers = append(tracers, qlogTracer)
-	}
+
 	if cm.enableMetrics {
-		tracers = append(tracers, newMetricsTracer())
+		cm.metricsTracer = newMetricsTracer()
 	}
 	if len(tracers) > 0 {
```
sukunrt, line 58: Is `len(tracers)` always 0?

p2p/transport/quicreuse/connmgr.go
```diff
@@ -27,6 +25,9 @@ type ConnManager struct {
 
 	connsMu sync.Mutex
 	conns   map[string]connListenerEntry
+
+	srk           quic.StatelessResetKey
+	metricsTracer *metricsTracer
```
sukunrt, line 30: NIT: Is `mt *metricsTracer` better? Given that it is not used anywhere other than being passed to underlying quic.Transport and this is similar to `srk`.

p2p/transport/quicreuse/connmgr.go
```diff
@@ -143,28 +151,28 @@ func (c *ConnManager) onListenerClosed(key string) {
 	}
 }
 
-func (c *ConnManager) listen(network string, laddr *net.UDPAddr) (pConn, error) {
+func (c *ConnManager) transportForListen(network string, laddr *net.UDPAddr) (refCountedQuicTransport, error) {
```
sukunrt, line 154: NIT: Do we need to rename this?

p2p/transport/quicreuse/reuse.go
```diff
@@ -7,23 +7,48 @@ import (
...
 
 	// count conn reference
 	DecreaseCount()
 	IncreaseCount()
 }
 
-type noreuseConn struct {
-	*net.UDPConn
+type singleOwnerTransport struct {
+	tr quic.Transport
```
sukunrt, line 26: Is it possible/useful to embed this?

p2p/transport/quicreuse/reuse.go
```diff
@@ -73,22 +114,27 @@ type reuse struct {
...
-	// globalDialers contains connections that we've dialed out from. These connections are listening on 0.0.0.0 / ::
-	// On Dial, connections are reused from this map if no connection is available in the globalListeners
-	// On Listen, connections are reused from this map if the requested port is 0, and then moved to globalListeners
-	globalDialers map[int]*reuseConn
+	unicast map[string] /* IP.String() */ map[int] /* port */ *refcountedTransport
+	// globalListeners contains transports that are listening on 0.0.0.0 / ::
+	globalListeners map[int]*refcountedTransport
+	// globalDialers contains transports that we've dialed out from. These transports are listening on 0.0.0.0 / ::
+	// On Dial, transports are reused from this map if no transport is available in the globalListeners
+	// On Listen, transport are reused from this map if the requested port is 0, and then moved to globalListeners
```
sukunrt, line 122: NIT: `transports are`

p2p/transport/quicreuse/connmgr.go
sukunrt: maybe you want: 
```
tracers = append(tracers, cm.metricsTracer)
```

Also if enableMetrics is false, we want to enable qLogTracer initialised from 
`qloggerForDir` given that we had these lines before

```
if qlogTracer != nil {
		tracers = append(tracers, qlogTracer)
	}
```
and qlogTracer was initialised in `init`

sukunrt (commented):
I don't think we should rename like
>reuse.Listen is now called reuse.TransportForListen

1. For our purposes the PacketConn is now replaced by Transport. This renaming makes things clearer now but given that this will be the API of quic-go going forward, these names will feel out of place 2-3 months from now. 
2. The type system already provides the type of the return value so it's clear what sort of object we are getting. Plus most(all?) of these are internal to `transport/quic` so the expected type of object is also clear from the context. 

IMO we should just add a long explanation in the changelog / commit message.

MarcoPolo:
> I don't think we should rename like
> 
> > reuse.Listen is now called reuse.TransportForListen
> 
> 1. For our purposes the PacketConn is now replaced by Transport. This renaming makes things clearer right now but given that this will be the API of quic-go going forward, these names will feel out of place 2-3 months from now.

What do you mean by the names will feel out of place? As long as quic-go gives us a transport and as long as we are getting a Transport for listening from this method, the new name makes sense.

> 2. The type system already provides the type of the return value so it's clear what sort of object we are getting. Plus most(all?) of these are internal to `transport/quic` so the expected type of object is also clear from the context.


It's really confusing for a `Listen` method to not return a listener, just like it's confusing for `Dial` to not return a connection to another node. This makes the API clearer because you aren't getting a Listener, you're getting the thing that can `Listen` (Transport).

> IMO we should just add a long explanation in the changelog / commit message.

While better than nothing, I think this isn't enough. We are already breaking the API by no longer returning a `PacketConn` we should update the API to make it clear what we are doing and returning. A method called `Listen` should return a Listener like thing that you can call `Accept` on.

I'm open to changing the name from `TransportForListen` to something else as long as it's not `Listen`. Maybe `ListenQuicTransport`? It would be in line with Go std's `ListenTCP/ListenUDP/ListenPacket/ListenX`

p2p/transport/quicreuse/reuse.go
```diff
@@ -217,16 +263,20 @@ func (r *reuse) dialLocked(network string, source *net.IP) (*reuseConn, error) {
...
-	conn, err := net.ListenUDP(network, addr)
+	conn, err := listenAndOptimize(network, addr)
 	if err != nil {
 		return nil, err
 	}
-	rconn := newReuseConn(conn)
+	rconn := &refcountedTransport{tr: quic.Transport{
+		Conn:              conn,
+		StatelessResetKey: r.statelessResetKey,
+		Tracer:            r.metricsTracer,
```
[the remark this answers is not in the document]
MarcoPolo: The classic typed nil bug, thanks.

marten-seemann:
quic-go v0.36.2 just shipped: https://github.com/quic-go/quic-go/releases/tag/v0.36.2.

sukunrt:
>A method called Listen should return a Listener-like thing that you can call Accept on.

I see your point. Thanks for the explanation. 
My point was the `pConn` had the same interface as transport and the method wasn't named `pConnForListen`. But maybe it should have. 

>These names will feel out of place 2-3 months from now.
I meant Transport in the name was redundant with the returned type

But I am convinced about the renaming. I suggest we either keep it as is or rename struct reuse to `reuseTransport` and name the methods `ForListen` and `ForDial` so it reads like `reuseTransport.ForListen`

I do not like naming along the lines `ListenTCP` because the returned object is a listener of type TCP which doesn't fit in with what we're doing. So `TransportForListen` is better.

p2p/transport/quicreuse/connmgr_test.go
```diff
@@ -92,91 +92,85 @@ func testListenOnSameProto(t *testing.T, enableReuseport bool) {
...
 func TestConnectionPassedToQUICForListening(t *testing.T) {
-	origQuicListen := quicListen
-	t.Cleanup(func() { quicListen = origQuicListen })
-
-	var conn net.PacketConn
-	quicListen = func(c net.PacketConn, _ *tls.Config, _ *quic.Config) (quic.Listener, error) {
-		conn = c
-		return nil, errors.New("listen error")
+	if runtime.GOOS == "windows" {
+		t.Skip("skipping on windows. Not sure why this fails")
```
MarcoPolo, line 96: @marten-seemann Any clue off the top of your head why this fails? https://github.com/libp2p/go-libp2p/actions/runs/5537794470/jobs/10107022242#step:8:4371. 

Presumably this means that on windows after calling `OptimizeConn` the returned conn isn't a `(quic.OOBCapablePacketConn)`

marten-seemann: Yes, that’s right. It’s using a `basicConn` on Windows: https://github.com/quic-go/quic-go/blob/418b866e3260b65496eb0d4420465e9d11b07e5f/sys_conn.go#L104

I’m becoming more and more convinced that `OptimizeConn` is not the right API. We should change that in a future release.

p2p/transport/quicreuse/connmgr.go
MarcoPolo: `github.com/quic-go/quic-go/qlog` no longer includes a `NewTracer`. This is only a connection tracer so we don't need it here anymore. It's now done below in the connection tracers.

MarcoPolo: ah sorry I see what I mean, I think this is a half finished refactor actually.

p2p/transport/quicreuse/connmgr_test.go
MarcoPolo: Yeah. I think if Transport has a `WriteTo` that would be better. The problem is I can't treat the conn from OptimizeConn as a normal PacketConn once I give it to the Transport.

MarcoPolo: No, this is fine for now. And the change will be very straightforward

p2p/transport/quicreuse/reuse.go
```diff
@@ -97,15 +143,15 @@ func newReuse() *reuse {
...
+		for _, tr := range r.globalDialers {
+			tr.Close()
 		}
-		for _, conns := range r.unicast {
-			for _, conn := range conns {
-				conn.Close()
+		for _, trs := range r.unicast {
+			for _, tr := range trs {
+				tr.Close()
 			}
```
sukunrt, line 155: Not sure how to comment on unchanged code. Can you fix naming on Line 170. That's still using Conns.

p2p/transport/quicreuse/reuse.go
```diff
@@ -1,29 +1,56 @@
...
 
 	"github.com/google/gopacket/routing"
 	"github.com/libp2p/go-netroute"
+	"github.com/quic-go/quic-go"
 )
 
-type pConn interface {
-	net.PacketConn
+type refCountedQuicTransport interface {
+	WriteTo([]byte, net.Addr) (int, error)
```
sukunrt, line 16: Is this only used for hole punching? If so, should we add a comment?

p2p/transport/webtransport/transport.go
```diff
@@ -295,10 +295,13 @@ func (t *transport) CanDial(addr ma.Multiaddr) bool {
...
 
 func (t *transport) Listen(laddr ma.Multiaddr) (tpt.Listener, error) {
-	isWebTransport, _ := IsWebtransportMultiaddr(laddr)
+	isWebTransport, certhashCount := IsWebtransportMultiaddr(laddr)
 	if !isWebTransport {
 		return nil, fmt.Errorf("cannot listen on non-WebTransport addr: %s", laddr)
 	}
+	if certhashCount > 0 {
+		return nil, fmt.Errorf("cannot listen on a specific certhash non-WebTransport addr: %s", laddr)
+	}
```
sukunrt, line 304: Thanks! 😅

sukunrt (approved):
Just a couple of nits.

p2p/transport/webtransport/transport.go
MarcoPolo: Not strictly necessary for this PR, I should undo this and open it in a separate PR.

sukunrt: oh no. It is the right thing. Plus a minor change. I think it is fine as is.

MarcoPolo: #2426

p2p/transport/quicreuse/reuse.go
```diff
@@ -13,9 +13,11 @@ import (
 )
 
 type refCountedQuicTransport interface {
-	WriteTo([]byte, net.Addr) (int, error)
 	LocalAddr() net.Addr
 
+	// Used to send packets directly around QUIC. Useful for hole punching.
+	WriteTo([]byte, net.Addr) (int, error)
```
sukunrt, line 19: 👍

p2p/transport/quicreuse/connmgr.go
```diff
@@ -104,22 +104,22 @@ func (c *ConnManager) ListenQUIC(addr ma.Multiaddr, tlsConf *tls.Config, allowWi
...
-	entry, ok := c.conns[key]
+	entry, ok := c.quicListeners[key]
 	if !ok {
-		conn, err := c.transportForListen(netw, laddr)
+		tr, err := c.transportForListen(netw, laddr)
 		if err != nil {
 			return nil, err
 		}
-		ln, err := newConnListener(conn, c.serverConfig, c.enableDraft29)
+		ln, err := newQuicListener(tr, c.serverConfig, c.enableDraft29)
```
sukunrt, line 117: can you rename this function's argument at [definition](https://github.com/libp2p/go-libp2p/blob/7773888b80bdab3c1e6292941898cd3348122575/p2p/transport/quicreuse/listener.go#L41) to tr? 
```
func newQuicListener(c refCountedQuicTransport, quicConfig *quic.Config, enableDraft29 bool) (*quicListener, error) {
```

sukunrt (approved):
This is nicer. Thanks. I think this one comment is the last renaming left.

PR #2431, opened 2023-07-14
Release v0.29.0

MarcoPolo:


marten-seemann:
Any idea why gorelease is not working?

MarcoPolo:
I'm thinking it has something to do with the manual dependency mucking. Here's what I see locally: https://gist.github.com/MarcoPolo/d5e079cb740aa3bf89bbb2fdcce9b8a9

marten-seemann:
Thank you. That doesn’t really make much sense either. We’ll need to take a closer look into why gorelease is failing here at some time, but that’s not a release blocker.

MarcoPolo:
This error:
```
❯ ~/go/bin/gorelease -base=v0.28.1
# github.com/libp2p/go-libp2p/p2p/transport/quicreuse
## errors in release version:
/Users/marco/go/pkg/mod/github.com/libp2p/go-libp2p@v0.28.1/p2p/transport/quicreuse/connmgr.go:16:28: DialContext not declared by package quic
```

Makes me think that this is related to how Go sees v0.36 of quic-go as compatible with v0.33. Since it seems to be that it's trying to build go-libp2p v0.28.1 using quic-go v0.36.

I made sure I could build Kubo using the latest go-libp2p version. So I think we're okay here. We just lose the ability to identify breaking APIs. We should figure out a better solution here.

Issue #2436, opened 2023-07-19
swarm.DialError should wrap DialError.DialErrors


sukunrt:
Almost all of the users of swarm.DialError either expect swarm.DialError to wrap over errors of individual dials. To summarise this should return true
```go
errors.Is(
		&DialError{
			Peer:       "pid",
			DialErrors: []TransportError{
			{ 
              Address: ma.StringCast("/ip4/1.2.3.4/tcp/1234"), 
			  Cause: ErrDialBackoff}
			},
		},
		ErrDialBackoff,
	)
```
I think this is a reasonable thing to fix. Some previous discussion here: https://github.com/libp2p/go-libp2p/pull/2414#discussion_r1260163707

There are a lot of usecases here:
https://github.com/search?q=swarm.DialError&type=code

A few examples:

https://github.com/ipni/cassette/blob/ca08e5af7376dc249cbb3a1173bbc6b37e38ceea/metrics.go#L301-L309
```
		case *swarm.DialError:
			errKind = "dial-error"
			for _, de := range e.DialErrors {
				if de.Cause.Error() == "dial backoff" {
					errKind = "dial-backoff"
					break
				}
			}
```

https://github.com/libp2p/go-libp2p-kad-dht/blob/ee95d1ab03c47aab03218e18d07483f51e7787e4/fullrt/dht.go#L315
```
dialErr, ok := err.(*swarm.DialError)
				if ok {
					for _, transportErr := range dialErr.DialErrors {
						if errors.Is(transportErr.Cause, network.ErrResourceLimitExceeded) {
							limitErrOnce.Do(func() { logger.Errorf(rtRefreshLimitsMsg) })
						}
					}
				}
```

https://github.com/smartcontractkit/libocr/blob/82b910bef5c1c95cc7a886091ccfc1895bde76f5/networking/ocr_endpoint_v1.go#L428
```
		if !errors.Is(err, swarm.ErrDialBackoff) {
```

PR #2437, opened 2023-07-19
swarm: fix Unwrap for DialError, implement Unwrap for TransportError

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2436

Depends on #2460  #2461 
Blocked on go1.21 release.

### DIFF
--- a/p2p/net/swarm/dial_error.go
+++ b/p2p/net/swarm/dial_error.go
@@ -30,10 +30,7 @@ func (e *DialError) recordErr(addr ma.Multiaddr, err error) {
 		e.Skipped++
 		return
 	}
-	e.DialErrors = append(e.DialErrors, TransportError{
-		Address: addr,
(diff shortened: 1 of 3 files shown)


marten-seemann (changes requested):
There was a change in Go 1.20 that allows wrapping multiple errors. Should we use that new API?

We could simply wait until we drop Go 1.19 in a month, then we don't need to deal with any build flags now.

sukunrt:
I don't mind waiting. This is not blocking anything.

marten-seemann:
We can still get it into the next release. v0.30 won't happen before the Go 1.21 transition (quic-go forces us to do so, if we want to include v0.37).

marten-seemann:
#2460 and #2461. @sukunrt can you rebase please?

marten-seemann (approved):

PR #2438, opened 2023-07-20
feat: Implement HTTP spec

MarcoPolo:
A MVP for libp2p HTTP spec https://github.com/libp2p/specs/pull/508. 

This adds an `Host` for doing request-response with HTTP semantics. The libp2phttp.Host contrasts with a "stream host" (basichost or `host.Host` interface). It is capable of doing HTTP over an HTTP transport and/or a stream transport (provided by the stream host).

What works in this MVP:
- `.well-known/libp2p`
- A helper func to give you the best roundtripper given a list of multiaddrs and an optional stream host.
- A way of listening for HTTP on stream transports as well as a native HTTP transport.

My thoughts after implementing the spec is that the spec is a nice minimal set of constraints that are relatively easy to implement. Even making a compliant node using the the stock HTTP library is fairly straightforward.

If this looks good after a review, I would suggest merging this as an experimental feature to allow folks to start using it and getting feedback. I may try to integrate this into the IPNI stack to see how it feels.

Note: This branch targets `marco/gostream` which pulls in https://github.com/libp2p/go-libp2p-gostream into the monorepo.

---

Update Aug 2:
I've removed the libp2p-noise auth from the http spec, and replaced it with this spec focused only on auth: https://github.com/libp2p/specs/pull/564.

I think for the mvp we can merge this without doing any peer id auth, since peer id auth is optional anyways.


BigLep:
@MarcoPolo : for the "libp2p and HTTP" effort, what will the testing story be?   I'm not saying that all needs to be in this PR, but I'm curious about where we're aiming to end up.

MarcoPolo:
> @MarcoPolo : for the "libp2p and HTTP" effort, what will the testing story be? I'm not saying that all needs to be in this PR, but I'm curious about where we're aiming to end up.

This PR adds basic tests that test end-to-end communication on top of stream transports or an HTTP transport. It tests that a stock HTTP client/server can interoperate. It also adds tests around peer id authentication. It doesn't currently, but will include a test that tests the SNI+TLS flow for peer id authentication, and tests that incorrect auth steps correctly fail (these are noted in before merge list in the PR body).

Once we have multiple implementations of the spec, we will leverage the multidim interop tester to test this as well. It should just work, with no changes needed to the test runner.

Was there a test you had in mind that was missing here? Or something that you feel should be highlighted?

BigLep:
> Was there a test you had in mind that was missing here? Or something that you feel should be highlighted?

Thanks for explaining.  I didn't have anything else in mind.  I had missed the tests you described when I took a quick scan earlier.  My bad for not catching those.

Concerning multidimen interop running, we'll also need that for testing version compatibility as well right (even if go-libp2p is the only implementation to support this functionality for awhile)?

MarcoPolo:
I've removed the libp2p-noise auth from the http spec, and replaced it with this spec focused only on auth: https://github.com/libp2p/specs/pull/564. I've also removed auth from this PR.

I think for the mvp we can merge this without doing any peer id auth, since peer id auth is optional anyways.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,525 @@
...
+			p := c.Protocol()
+			if p.Code == ma.P_IP4 || p.Code == ma.P_IP6 || p.Code == ma.P_DNS || p.Code == ma.P_DNS4 || p.Code == ma.P_DNS6 {
+				host = c.Value()
+			} else if p.Code == ma.P_TCP || p.Code == ma.P_UDP {
+				port = c.Value()
+			} else if p.Code == ma.P_TLS {
+				useHTTPS = true
+			} else if p.Code == ma.P_SNI {
+				sni = c.Value()
+			}
```
gammazero, line 396:
```suggestion
			switch p.Code {
			case ma.P_IP4, ma.P_IP6, ma.P_DNS, ma.P_DNS4, ma.P_DNS6:
				host = c.Value()
			case ma.P_TCP, ma.P_UDP:
				port = c.Value()
			case ma.P_TLS:
				useHTTPS = true
			case ma.P_SNI:
				sni = c.Value()
			}
```

p2p/http/ping.go
```diff
@@ -0,0 +1,67 @@
...
+	body := [pingSize]byte{}
+	_, err := io.ReadFull(r.Body, body[:])
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/octet-stream")
+	w.Header().Set("Content-Length", strconv.Itoa(pingSize))
+	w.WriteHeader(http.StatusOK)
```
gammazero, line 31: Unnecessary
```suggestion
```

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,525 @@
...
+
+	// Return a JSON object with the well-known protocols
+	h.wellknownMapMu.Lock()
+	mapping, err := json.Marshal(h.wellKnownMapping)
+	h.wellknownMapMu.Unlock()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
```
gammazero, line 74: [Unnecessary if status is `http.StatusOK`](https://pkg.go.dev/net/http@go1.20.7#ResponseWriter)
```suggestion
```

p2p/http/ping.go
```diff
@@ -0,0 +1,67 @@
...
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
+
+	rBody := [pingSize]byte{}
+	_, err = io.ReadFull(resp.Body, rBody[:])
+	if err != nil {
```
gammazero, line 59:
```suggestion
	defer resp.Body.Close()
	if err != nil {
```
[It is the caller's responsibility to close Body](https://pkg.go.dev/net/http@go1.20.7#Response)

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,525 @@
...
+	}
+
+	body := [PeerMetadataLimit]byte{}
+	bytesRead := 0
+	for {
+		n, err := resp.Body.Read(body[bytesRead:])
+		bytesRead += n
+		if err == io.EOF {
+			break
+		} else if err != nil {
```
gammazero, line 510:
```suggestion
		}
		if err != nil {
```

gammazero (commented):
This looks great. I plan on using it soon.
Left some trivial suggestions.

MarcoPolo:
@gammazero I have a surprise for you: https://github.com/ipni/go-libipni/pull/102

p2p/http/libp2phttp.go
MarcoPolo: True, although there is no harm in being explicit.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,675 @@
...
+}
+
+// HTTPHost is a libp2p host for request/responses with HTTP semantics. This is
+// in contrast to a stream-oriented host like the host.Host interface. Warning,
+// this is experimental. The API will likely change.
+type HTTPHost struct {
+	rootHandler      http.ServeMux
+	wk               WellKnownHandler
+	httpRoundTripper *http.Transport
+	recentHTTPAddrs  *lru.Cache[peer.ID, httpAddr]
```
MarcoPolo, line 108: Let's document these and there purpose (both lru cache). Why aren't we using a peer store?

p2p/security/noise/pb/payload.proto
```diff
@@ -4,6 +4,8 @@ package pb;
 message NoiseExtensions {
 	repeated bytes webtransport_certhashes = 1;
 	repeated string stream_muxers = 2;
+	optional string SNI = 3;
```
MarcoPolo, line 7: undo this

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,674 @@
...
+// this is experimental. The API will likely change.
+type HTTPHost struct {
+	rootHandler      http.ServeMux
+	wk               WellKnownHandler
+	httpRoundTripper *http.Transport
+	// recentHTTPAddrs is an lru cache of recently used HTTP addresses. This
+	// lets us know if we've recently connected to an HTTP endpoint and might
+	// have a warm idle connection for it (managed by the underlying HTTP
+	// roundtripper). In some cases, this lets us reuse our existing custom roundtripper (i.e. SNI != host).
+	recentHTTPAddrs *lru.Cache[peer.ID, httpAddr]
```
MarcoPolo, line 106: This should probably be keyed by the addr+scheme+sni combo.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,688 @@
...
+	if h.wellKnownMapping != nil {
+		delete(h.wellKnownMapping, p)
+	}
+	h.wellknownMapMu.Unlock()
+}
+
+// HTTPHost is a libp2p host for request/responses with HTTP semantics. This is
+// in contrast to a stream-oriented host like the host.Host interface. Warning,
+// this is experimental. The API will likely change.
+type HTTPHost struct {
```
marten-seemann, line 101: In analogy with `net/http`, should this be called `Server`?

I'm also wondering if we should get rid of the constructor (`net/http.Server` doesn't have one either). My first attempt at setting up a server was:
```go
host := libp2phttp.HTTPHost{}
host.SetHttpHandler("/website", handler)
log.Fatal(host.Serve())
```
which panicked immediately.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,688 @@
...
+		case ma.P_TCP, ma.P_UDP:
+			out.port = c.Value()
+		case ma.P_TLS:
+			out.useHTTPS = true
+		case ma.P_SNI:
+			out.sni = c.Value()
+
+		}
+		return out.host == "" || out.port == "" || !out.useHTTPS || out.sni == ""
+	})
```
marten-seemann, line 588: Please don't reimplement multiaddr parsing. `manet.ToNetAddr` should provide half of what you need here.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,688 @@
...
+	return err
+}
+
+func (h *HTTPHost) Close() error {
+	close(h.httpTransport.closeListeners)
+	return nil
+}
+
+// SetHttpHandler sets the HTTP handler for a given protocol. Automatically
+// manages the .well-known/libp2p mapping.
```
marten-seemann, line 281: Add an example here. It's not clear if the protocol ID contains a leading / or not, for example.

marten-seemann (commented):
Lots of my comments come from looking at the GoDoc (`godoc -http=:6060`), and putting myself into the shoes of a user of this package.

Should we split up libp2phttp.go into two files, one for the client and one for the server?

---

The following code doesn't work:
```go
host, err := libp2phttp.New(libp2phttp.ListenAddrs([]ma.Multiaddr{ma.StringCast("/ip4/127.0.0.1/tcp/8080/http")}))
if err != nil {
	log.Fatal(err)
}
host.SetHttpHandler("/website", http.FileServer(http.Dir("/tmp")))
log.Fatal(host.Serve())
```

Haven't debugged it yet. It creates the well-known handler, but http://localhost:8080/website/ returns a 404.

p2p/http/ping.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// SendPing send an ping request over HTTP. The provided client should be namespaced to the Ping protocol.
+func SendPing(client http.Client) error {
+	body := [32]byte{}
+	_, err := io.ReadFull(rand.Reader, body[:])
+	if err != nil {
+		return err
+	}
+	req, err := http.NewRequest("POST", "/", bytes.NewReader(body[:]))
```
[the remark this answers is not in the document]
MarcoPolo: This is namespaced.

MarcoPolo:
> Lots of my comments come from looking at the GoDoc (`godoc -http=:6060`), and putting myself into the shoes of a user of this package.

Nice! Thanks!


> The following code doesn't work:
> 
> ```go
> host, err := libp2phttp.New(libp2phttp.ListenAddrs([]ma.Multiaddr{ma.StringCast("/ip4/127.0.0.1/tcp/8080/http")}))
> if err != nil {
> 	log.Fatal(err)
> }
> host.SetHttpHandler("/website", http.FileServer(http.Dir("/tmp")))
> log.Fatal(host.Serve())
> ```

When you go to "/website" the fileserver is seeing the full path "/website" and you get a 404 because you don't have a `/tmp/website` folder. This is the exact same behavior you get with a servemux (`m.Handle("/website/", http.FileServer(http.Dir("/tmp")))`). 

You should strip the prefix of the path:
```
host.SetHttpHandler("/website", http.StripPrefix("/website", http.FileServer(http.Dir("/tmp"))))
```

Maybe this should be the default behavior for SetHttpHandler?

p2p/http/libp2phttp.go
MarcoPolo: > In analogy with net/http, should this be called Server?

This can be used by the client as well (just like the stream host).

> I'm also wondering if we should get rid of the constructor

Not opposed. Should options be public fields as well then?

MarcoPolo: I've made this change. I've removed the constructor entirely and options are set by setting the public fields.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,688 @@
...
+				srv := http.Server{
+					Handler:   &h.rootHandler,
+					TLSConfig: h.httpTransport.tlsConfig,
+				}
+				errCh <- srv.ServeTLS(l, "", "")
+			}()
+		} else {
+			go func() {
+				errCh <- http.Serve(l, &h.rootHandler)
+			}()
```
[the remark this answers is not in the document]
MarcoPolo: I'm thinking about how we'd support a proxy (like nginx) that terminating SSL and communicating over HTTP.

Maybe it's better to make that a bit harder to avoid users broadcasting unencrypted HTTP by accident.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,688 @@
...
+	h.wellknownMapMu.Lock()
+	mapping, err := json.Marshal(h.wellKnownMapping)
+	h.wellknownMapMu.Unlock()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	w.Header().Add("Content-Type", "application/json")
+	w.Header().Add("Content-Length", strconv.Itoa(len(mapping)))
+	w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
MarcoPolo: https://github.com/libp2p/go-libp2p/pull/2438#discussion_r1284629402

I guess I'm out voted. I'll change it. For the record, I like the consistency of always writing the header, not just in error/informational cases.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,688 @@
...
+const PeerMetadataLimit = 8 << 10 // 8KB
+const PeerMetadataLRUSize = 256   // How many different peer's metadata to keep in our LRU cache
+
+// The scheme we use internally to name a stream based roundtripper. This is only used locally.
+const libp2pStreamScheme = "libp2p-stream"
+
+// TODOs:
+// - integrate with the conn gater and resource manager
+
+type WellKnownProtocolMeta struct {
```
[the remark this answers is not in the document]
MarcoPolo: Using a struct gives us more options for backwards compatibility in the future. If we add more to this besides a path, we don't introduce a breaking api change.

Maybe just ProtocolMeta? It's not configuration for a protocol, it's metadata about where this protocol is mounted on this HTTP server.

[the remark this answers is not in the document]
MarcoPolo: I've renamed to `ProtocolMeta`

MarcoPolo:
I've changed the behavior of SetHTTPHandler to strip the prefix for the handler automatically. I can't think of use cases where you wouldn't want that. You can still avoid it by setting a handler on the underlying servemux directly.

p2p/http/libp2phttp.go
MarcoPolo: It shouldn't matter. But I agree that an example would be good.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,688 @@
...
+type WellKnownProtoMap map[protocol.ID]WellKnownProtocolMeta
+
+// WellKnownHandler is an http.Handler that serves the .well-known/libp2p resource
+type WellKnownHandler struct {
+	wellknownMapMu   sync.Mutex
+	wellKnownMapping WellKnownProtoMap
+}
+
+// StreamHostListen retuns a net.Listener that listens on libp2p streams for HTTP/1.1 messages.
+func StreamHostListen(streamHost host.Host) (net.Listener, error) {
```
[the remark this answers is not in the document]
MarcoPolo: I'm hiding this function

p2p/http/ping/ping.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	w.Header().Set("Content-Type", "application/octet-stream")
+	w.Header().Set("Content-Length", strconv.Itoa(pingSize))
+	w.Write(body[:])
+}
+
+// SendPing send an ping request over HTTP. The provided client should be namespaced to the Ping protocol.
+func SendPing(client http.Client) error {
+	body := [32]byte{}
```
Wondertan, line 36:
```suggestion
	body := [pingSize]byte{}
```

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,744 @@
...
+	resp, err := http.ReadResponse(bufio.NewReader(s), r)
+	if err != nil {
+		return nil, err
+	}
+	resp.Body = &streamReadCloser{resp.Body, s}
+
+	return resp, nil
+}
+
+// roundTripperForSpecificHost is an http.RoundTripper targets a specific server. Still reuses the underlying RoundTripper for the requests.
```
Wondertan, line 383:
```suggestion
// roundTripperForSpecificServer is an http.RoundTripper targets a specific server. Still reuses the underlying RoundTripper for the requests.
```

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,744 @@
...
+
+			if parsedAddr.useHTTPS {
+				go func() {
+					srv := http.Server{
+						Handler:   &h.ServeMux,
+						TLSConfig: h.TLSConfig,
+					}
+					errCh <- srv.ServeTLS(l, "", "")
+				}()
+				h.httpTransport.listenAddrs = append(h.httpTransport.listenAddrs, listenAddr)
```
gammazero, line 262: This causes a race with `Addrs() `

If `Serve` is called in a separate goroutine, then there needs to be some synchronization to know when `Addrs()` cannot be called. Or, `h.httpTransport.listenAddrs` needs to be set in some function that is called before calling `Serve`. This worked previously because the constructor served that purpose.

Seems that a constructor is needed. Or, maybe a `Start` function that does this setup and calls `Serve` in a separate goroutine and returns.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,744 @@
...
+func (h *HTTPHost) SetHTTPHandler(p protocol.ID, handler http.Handler) {
+	h.SetHTTPHandlerAtPath(p, string(p), handler)
+}
+
+// SetHTTPHandlerAtPath sets the HTTP handler for a given protocol using the
+// given path. Automatically manages the .well-known/libp2p mapping.
+// http.StripPrefix is called on the handler, so the handler will be unaware of
+// it's prefix path.
+func (h *HTTPHost) SetHTTPHandlerAtPath(p protocol.ID, path string, handler http.Handler) {
+	if path[len(path)-1] != '/' {
```
gammazero, line 336: If an empty `path` is passed in, this will panic. Caller will need to check for empty `path`. Maybe safer to handle the `len(path) == 0` case?

MarcoPolo: Also typo on it’s

gammazero (approved):
I am currently using this here: https://github.com/ipni/go-libipni/pull/113

This looks great and I cannot wait until it gets merged.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,754 @@
...
+	h.wellknownMapMu.Unlock()
+}
+
+// HTTPHost is a libp2p host for request/responses with HTTP semantics. This is
+// in contrast to a stream-oriented host like the host.Host interface. Its
+// zero-value (&HTTPHost{}) is usable. Do not copy by value.
+// See examples for usage.
+//
+//	Warning, this is experimental. The API will likely change.
+type HTTPHost struct {
```
MarcoPolo, line 107: Should this just be called `Host`? Users will use this as `libp2phttp.Host` so it should be easy to differentiate against the core stream-based `host.Host` type.

gammazero: Yes. Avoids stutter. If everything were defined in a single package, then the type would be `Libp2pHTTPHost`, so `libp2phttp.Host` is appropriate.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,756 @@
...
+// you have out-of-band knowledge of a peer's protocol mapping.
+func (h *Host) AddPeerMetadata(server peer.ID, meta PeerMeta) {
+	if h.peerMetadata == nil {
+		h.peerMetadata = newPeerMetadataCache()
+	}
+	h.peerMetadata.Add(server, meta)
+}
+
+// GetPeerMetadata gets a peer's cached protocol metadata from the http host.
+func (h *Host) GetPeerMetadata(server peer.ID) (PeerMeta, bool) {
```
marten-seemann, line 743: Is there any reason to return the whole map here? Alternative: `SupportsProtocol(peer.ID) (path string, bool)`. It would be nice if we didn't have to make the map an exported type.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,756 @@
...
+	if h.peerMetadata == nil {
+		return nil, false
+	}
+	return h.peerMetadata.Get(server)
+}
+
+// RemovePeerMetadata removes a peer's protocol metadata from the http host
+func (h *Host) RemovePeerMetadata(server peer.ID, meta PeerMeta) {
```
marten-seemann, line 751: Unused `meta`?

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,756 @@
...
+		return http.Client{}, err
+	}
+
+	return http.Client{Transport: nrt}, nil
+}
+
+// NewRoundTripper returns an http.RoundTripper that can fulfill and HTTP
+// request to the given server. It may use an HTTP transport or a stream based
+// transport. It is valid to pass an empty server.ID.
+func (h *Host) NewRoundTripper(server peer.AddrInfo, opts ...RoundTripperOption) (http.RoundTripper, error) {
```
marten-seemann, line 536: I'm struggling with this abstraction. An `http.RoundTripper` is supposed to be able to issue requests to any HTTP host. This is more than just a theoretical concern: The `http.Client` has a `Transport` field that takes an `http.RoundTripper`, and would then be able to follow redirects. If we return a `RoundTripper` here that's scoped a single host, it won't be possible to redirect from one libp2p+HTTP host to another.

I don't have a good solution for this now, I'll have to think about the correct abstraction more.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,756 @@
...
+
+const ProtocolIDForMultistreamSelect = "/http/1.1"
+const peerMetadataLimit = 8 << 10 // 8KB
+const peerMetadataLRUSize = 256   // How many different peer's metadata to keep in our LRU cache
+
+// TODOs:
+// - integrate with the conn gater and resource manager
+
+// ProtocolMeta is metadata about a protocol.
+type ProtocolMeta struct {
```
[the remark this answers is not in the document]
MarcoPolo: Config implies that this is a configurable, but this is metadata about a peer's supported protocols.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,756 @@
...
+	// Path defines the HTTP Path prefix used for this protocol
+	Path string `json:"path"`
+}
+
+type PeerMeta map[protocol.ID]ProtocolMeta
```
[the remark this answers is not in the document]
MarcoPolo: No strong preference here, but I do prefer that `PeerMeta` is consistent with `ProtocolMeta`.

p2p/http/libp2phttp.go
MarcoPolo: The reason to export the whole type is to give users flexibility in managing the Host's peerMetadata cache without exposing that cache. This is useful if they want to add out of band knowledge of a peer's metadata. This method along with Set and Remove should let users do anything they need to with this metadata cache.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,756 @@
...
+	// ListenAddrs are the requested addresses to listen on. Multiaddrs must be
+	// a valid HTTP(s) multiaddr. Only multiaddrs for an HTTP transport are
+	// supported (must end with /http or /https).
+	ListenAddrs []ma.Multiaddr
+	// TLSConfig is the TLS config for the server to use
+	TLSConfig *tls.Config
+	// ServeInsecureHTTP indicates if the server should serve unencrypted HTTP requests over TCP.
+	ServeInsecureHTTP bool
+	// ServeMux is the http.ServeMux used by the server to serve requests
+	ServeMux http.ServeMux
```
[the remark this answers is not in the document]
MarcoPolo: No it will not use http.DefaultServeMux. To me that's a default for the standard library, and I think it would be confusing if we referenced it here. Users would create a new host that automatically exposes things they weren't expecting.

The reason this is public is to allow a user to set a custom servemux, or to manually manage the mapping outside of `SetHTTPHandler`. I'll expand the comment to mention this.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,756 @@
...
+
+	// DefaultClientRoundTripper is the default http.RoundTripper for clients
+	DefaultClientRoundTripper *http.Transport
```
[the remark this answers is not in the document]
MarcoPolo: Documented

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,756 @@
...
+)
+
+var log = logging.Logger("libp2phttp")
+
+const ProtocolIDForMultistreamSelect = "/http/1.1"
+const peerMetadataLimit = 8 << 10 // 8KB
+const peerMetadataLRUSize = 256   // How many different peer's metadata to keep in our LRU cache
+
+// TODOs:
+// - integrate with the conn gater and resource manager
```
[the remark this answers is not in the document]
MarcoPolo: https://github.com/libp2p/go-libp2p/issues/2511

p2p/http/libp2phttp.go
MarcoPolo: > it won't be possible to redirect from one libp2p+HTTP host to another.

I agree this is an interesting use case we should try to support.

The thing I like about this api is that a user can pass in many multiaddrs and the best transport is chosen for them. There isn't anywhere else we could do this besides when we create a round tripper. But, thinking more about this, this doesn't have to be a property of the core round tripper. We can have a round tripper on top of some core round tripper that can do this logic. It would be great if the core round tripper could, as you say, fulfill any HTTP request regardless of the transport.

Here's one idea how to do this:

1. Change this function to accept no parameters. It just returns an http.RoundTripper.
2. This roundtripper can fulfill HTTP requests to the following URIs:
  a. `http://` and `https://`, just like the http.DefaultTransport does.
  b. `multiaddr:<peer-multiaddr>`. Which would be a new URI scheme for multiaddrs. The value after `multiaddr:` is simply the string encoded version of the multiaddr. The multiaddr can include a peer's transport information, or could include only the peer id `/p2p/<peer-id>` and rely on discovery or out-of-band knowledge of a peer's addresses. Note this multiaddr can also be an http transport multiaddr.

This would work but would lead to us having to define the [httppath component](https://github.com/libp2p/specs/issues/554) since we'll need a way to reference the HTTP path we want to use in the multiaddr.

We would probably also want to register the `multiaddr` uri scheme in IANA (but could defer).

Here's a branch of this idea: https://github.com/libp2p/go-libp2p/compare/marco/http...marco/newroundtripper?expand=1. It includes an httppath implementation, but we should standardize that (not use the private area) if we go that route. It also has a test demonstrating the redirect to other host (using multiaddrs).

MarcoPolo: For now let’s remove the NewRoundtripper and only expose the NamespacedRoundtripper and NamespacedClient. We can then add the NewRoundTripper later. 

The only thing we loose is redirects across hosts. I thinks that’s fine to omit for this initial version. And we have plan above on how to add it.

MarcoPolo: I've renamed this to `NewConstrainedRoundTripper`, I think that will unblock us here. In a future version we can introduce the `NewRoundTripper` that accepts a multiaddr uri and can handle redirects across hosts.

p2p/http/libp2phttp.go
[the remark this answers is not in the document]
MarcoPolo: This should probably be a pointer if we want to let users set a custom servemux

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+	// valid HTTP(s) multiaddr. Only multiaddrs for an HTTP transport are
+	// supported (must end with /http or /https).
+	ListenAddrs []ma.Multiaddr
+	// TLSConfig is the TLS config for the server to use
+	TLSConfig *tls.Config
+	// InsecureAllowHTTP indicates if the server is allowed to serve unencrypted
+	// HTTP requests over TCP.
+	InsecureAllowHTTP bool
+	// ServeMux is the http.ServeMux used by the server to serve requests. If nil,
+	// new serve mux will be allocated. Users may manually add handlers to this
```
marten-seemann, line 117:
```suggestion
	// a new serve mux will be created. Users may manually add handlers to this
```

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+	// DefaultClientRoundTripper is the default http.RoundTripper for clients to
+	// use when making requests over an HTTP transport. This must be an
+	// `*http.Transport` type so that the transport can be cloned and the
+	// `TLSClientConfig` field can be configured. If unset, it will create a new
+	// `http.Transport` on first use.
+	DefaultClientRoundTripper *http.Transport
+
+	// WellKnownHandler is the http handler for the `.well-known/libp2p`
+	// resource. It is responsible for sharing this node's protocol metadata
+	// with other nodes.
```
marten-seemann, line 132: Mention that users only need to care about it if they set their own `ServeMux`, and that it's handled automatically otherwise (if I understand correctly).

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+		}()
+	}
+
+	closeAllListeners := func() {
+		for _, l := range h.httpTransport.listeners {
+			l.Close()
+		}
+	}
+
+	err := func() error {
```
marten-seemann, line 248: Maybe cleaner: Make this a separate method that returns the listen addrs and an error.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+			var listenAddr ma.Multiaddr
+			if parsedAddr.useHTTPS && parsedAddr.sni != "" && parsedAddr.sni != host {
+				listenAddr = ma.StringCast(fmt.Sprintf("/ip4/%s/tcp/%s/tls/sni/%s/http", host, port, parsedAddr.sni))
+			} else {
+				scheme := "http"
+				if parsedAddr.useHTTPS {
+					scheme = "https"
+				}
+				listenAddr = ma.StringCast(fmt.Sprintf("/ip4/%s/tcp/%s/%s", host, port, scheme))
+
```
marten-seemann, line 279:
```suggestion
```

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+	GetPeerMetadata() (PeerMeta, error)
+}
+
+type streamRoundTripper struct {
+	server   peer.ID
+	h        host.Host
+	httpHost *Host
+}
+
+type streamReadCloser struct {
```
marten-seemann, line 372: Can you add a comment why you're introducing this type (as opposed to just using the underlying `network.Stream`).

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+		r.Write(s)
+		if r.Body != nil {
+			r.Body.Close()
+		}
+	}()
+
+	resp, err := http.ReadResponse(bufio.NewReader(s), r)
+	if err != nil {
+		return nil, err
+	}
```
marten-seemann, line 404: You're ignoring the request context here. If this causes too much complexity for the first iteration, feel free to ignore, but please add a TODO to the code.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+		if tr, ok := rt.RoundTripper.(closeIdler); ok {
+			tr.CloseIdleConnections()
+		}
+	}
+	// No-op, since we don't want users thinking they are closing idle
+	// connections for this server, when in fact they are closing all idle
+	// connections
+}
+
+type namespacedRoundTripper struct {
```
marten-seemann, line 473: Add a doc comment?

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+	if err != nil {
+		return http.Client{}, err
+	}
+
+	return http.Client{Transport: nrt}, nil
+}
+
+// NewConstrainedRoundTripper returns an http.RoundTripper that can fulfill and HTTP
+// request to the given server. It may use an HTTP transport or a stream based
+// transport. It is valid to pass an empty server.ID.
```
marten-seemann, line 550: Describe the logic that's used to use a stream based vs. an HTTP transport?

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+func (h *Host) NewConstrainedRoundTripper(server peer.AddrInfo, opts ...RoundTripperOption) (http.RoundTripper, error) {
```
marten-seemann, line 551: This function is a bit problematic, as it dials a new connection (which might take an extended period of time), without providing a way to cancel that. One option would be to add a context, but that would be a bit surprising for a function that look like a constructor. Another option would be to defer dialing.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		if bytesRead >= peerMetadataLimit {
+			return nil, fmt.Errorf("peer metadata too large")
+		}
+	}
```
marten-seemann, line 739: Consider using an `io.LimitedReader`. Avoids allocating 8 kB, which won't be needed in the vast majority of cases.

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,790 @@
...
+
+	meta := PeerMeta{}
+	json.Unmarshal(body[:bytesRead], &meta)
```
marten-seemann, line 742: Ignored error return value?

p2p/http/libp2phttp.go
MarcoPolo: This also adds to the listeners. Maybe I'll have this accept a pointer of http transport so it's clear what is getting mutated.

MarcoPolo: Actually moved this out into a separate method so it was clear what is in scope vs not.

p2p/http/libp2phttp.go
MarcoPolo: https://github.com/libp2p/go-libp2p/issues/2533

p2p/http/libp2phttp.go
```diff
@@ -0,0 +1,804 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
+
+	meta := PeerMeta{}
+	err = json.NewDecoder(&io.LimitedReader{
+		R: resp.Body,
+		N: 8 << 10,
```
gammazero, line 752:
```suggestion
		N: peerMetadataLimit,
```

marten-seemann (approved):
Ship it!

Issue #2440, opened 2023-07-24
identify/id.go: `filterAddrs` should not be used


wlynxg:
In the Identify service code, I noticed that the filterAddrs function was added in the June update, which is used in the following context:
```go
func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn, isPush bool) {
        ......
	ids.Host.Peerstore().AddAddrs(p, filterAddrs(addrs, c.RemoteMultiaddr()), ttl)
        ......
}
```
About the filterAddrs function:
```go
// filterAddrs filters the address slice based on the remove multiaddr:
// * if it's a localhost address, no filtering is applied
// * if it's a local network address, all localhost addresses are filtered out
// * if it's a public address, all localhost and local network addresses are filtered out
func filterAddrs(addrs []ma.Multiaddr, remote ma.Multiaddr) []ma.Multiaddr {
	if manet.IsIPLoopback(remote) {
		return addrs
	}
	if manet.IsPrivateAddr(remote) {
		return ma.FilterAddrs(addrs, func(a ma.Multiaddr) bool { return !manet.IsIPLoopback(a) })
	}
	return ma.FilterAddrs(addrs, manet.IsPublicAddr)
}
```

I believe that this function should not be used to filter out private network addresses, as it can lead to the following problem:
When two nodes are both behind the same NAT with a public IP address, using this filter function will cause the DHT nodes to only store the public IP addresses of the two nodes. When the two nodes establish a connection, they will send packets to each other's public IP addresses. However, most routers do not support [NAT loopback](https://en.wikipedia.org/wiki/Network_address_translation#NAT_hairpinning) ( also known as NAT loopback or NAT reflection), which can cause the packets sent between the two nodes to be discarded by the router, resulting in a failed connection between the nodes.

marten-seemann:
Your analysis is correct, but this is an exceedingly rare occasion. On the other hand, we were seeing tons and tons of spurious dials to localhost and local addresses from nodes that are not on the same network.

wlynxg:
> Your analysis is correct, but this is an exceedingly rare occasion. On the other hand, we were seeing tons and tons of spurious dials to localhost and local addresses from nodes that are not on the same network.

Thank you for your response.

First of all, I think your purpose of filtering intranet addresses is completely correct. It helps reduce invalid data packets sent by clients and also helps conceal the details of the intranet where the node resides.

[...]

PR #2441, opened 2023-07-24
fix: remove `filterAddrs` func

wlynxg:
fix: https://github.com/libp2p/go-libp2p/issues/2440
In order to solve the NAT loopback problem, I removed the `filterAddrs` function and related calls


marten-seemann:
Let’s keep the discussion on the issue.

PR #2454, opened 2023-08-02
core/crypto: restrict RSA keys to <= 8192 bits

MarcoPolo:
At a certain point these become too much effort to compute. This limits the max size.

### DIFF
--- a/core/crypto/rsa_common.go
+++ b/core/crypto/rsa_common.go
@@ -11,10 +11,12 @@ import (
 const WeakRsaKeyEnv = "LIBP2P_ALLOW_WEAK_RSA_KEYS"
 
 var MinRsaKeyBits = 2048
+var MaxRsaKeyBits = 8192
 
 // ErrRsaKeyTooSmall is returned when trying to generate or parse an RSA key
 // that's smaller than MinRsaKeyBits bits. In test
 var ErrRsaKeyTooSmall error
+var ErrRsaKeyTooBig error = fmt.Errorf("rsa keys must be <= %d bits", MaxRsaKeyBits)
 
 func init() {
 	if _, ok := os.LookupEnv(WeakRsaKeyEnv); ok {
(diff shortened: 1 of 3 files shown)


core/crypto/rsa_common.go
```diff
@@ -11,10 +11,12 @@ import (
 const WeakRsaKeyEnv = "LIBP2P_ALLOW_WEAK_RSA_KEYS"
 
 var MinRsaKeyBits = 2048
+var MaxRsaKeyBits = 8192
```
marten-seemann, line 14: Does this need to be configurable?

```suggestion
const maxRsaKeyBits = 8192
```

core/crypto/rsa_common.go
```diff
@@ -11,10 +11,12 @@ import (
...
 
 // ErrRsaKeyTooSmall is returned when trying to generate or parse an RSA key
 // that's smaller than MinRsaKeyBits bits. In test
 var ErrRsaKeyTooSmall error
+var ErrRsaKeyTooBig error = fmt.Errorf("rsa keys must be <= %d bits", MaxRsaKeyBits)
```
marten-seemann, line 19:
```suggestion
var ErrRsaKeyTooBig error = fmt.Errorf("RSA keys must be <= %d bits", maxRsaKeyBits)
```

marten-seemann (commented):
Could it be that you missed a few cases? I'm less worried about the code path where you generate the RSA key yourself (although we should prevent that footgun as well), but more about those where the peer sends us its pubkey. What could path would be taken during the TLS / Noise handshake when we receive the peer's key?
 
image

MarcoPolo:
> Could it be that you missed a few cases? I'm less worried about the code path where you generate the RSA key yourself (although we should prevent that footgun as well), but more about those where the peer sends us its pubkey. What could path would be taken during the TLS / Noise handshake when we receive the peer's key?
> 
> image

PublicKeyFromProto uses `UnmarshalRsaPublicKey` under the hood which is covered by this patch. I think only `UnmarshalRsaPublicKey` and `GenerateRSAKeyPair` give you an RSA public key.

PR #2461, opened 2023-08-07
swarm: return errors on filtered addresses when dialing

sukunrt:
Depends on #2460 

This introduces a small breaking change. On no good addresses we previously used to return `ErrNoGoodAddress` not we return a `*DialError` which returns true on `errors.Is(err, ErrNoGoodAddress)`
This along with #2437 will allow us to check for all the reasons why dialing an address failed. 

The minor breaking change is easy enough to remove but in this case it seems worth it since we are returning more information and also preserving existing behaviour, the only change required in existing code is using `errors.Is` instead of `==`

### DIFF
--- a/p2p/net/swarm/black_hole_detector.go
+++ b/p2p/net/swarm/black_hole_detector.go
@@ -178,7 +178,7 @@ type blackHoleDetector struct {
 }
 
 // FilterAddrs filters the peer's addresses removing black holed addresses
-func (d *blackHoleDetector) FilterAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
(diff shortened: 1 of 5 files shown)


marten-seemann:
Will review after #2460 has landed.

marten-seemann:
@sukunrt Can you rebase this PR, so I can review it?

sukunrt:
Fixed. Sorry, I didn't realise that the reqno removal PR changes caused conflicts here.

marten-seemann:
> This introduces a small breaking change. On no good addresses we previously used to return `ErrNoGoodAddress` not we return a `*DialError` which returns true on `errors.Is(err, ErrNoGoodAddress)`

This is fine. You're not supposed to compare errors with `==` anyway.

marten-seemann (approved):

Issue #2422, opened 2023-07-11
Implement AutoNATv2


sukunrt:
Implement https://github.com/libp2p/specs/pull/538

We want a light weight implementation that is just the protocol client and the protocol server. The protocol client will implement `/libp2p/autonat/2/dial` and the server will handle this. 

The logic of address management: which addresses to verify and when will live inside Address Pipeline Service(provisional name) as described [here](https://github.com/libp2p/go-libp2p/issues/2229). This Address Pipeline will have to implement a user of this AutoNATv2 client with logic similar to one described [here](https://github.com/libp2p/specs/issues/539#issuecomment-1505278563)

One important benefit of merging just the protocol client and the server into master first is, this protocol will be unusable till some threshold percentage(maybe 10?) of nodes upgrade and start supporting this because before reaching this threshold, AutoNATv2 nodes will not be able to find any AutoNATv2 peers.

sukunrt:
Deployment Consideration for kubo

When initially deployed a node won't be able to find any other servers and so it cannot use autonat v2 to verify its addresses. So initially we will need to enable the server and not use the client. 
There are two considerations for a subsequent deployment that uses the v2 client to verify its addresses. 

1. Finding a Peer

As the adoption of the version with the AutoNATv2 server increases, nodes will be able to find servers for AutoNATv2. This adoption fraction, when nodes can find at least 3 servers is some small multiple of `3 * 1/(dht routing table size)`. For the ipfs dht the routing table size is roughly 250 nodes which makes this number 1.2%. So once 5% nodes have adopted the version with the AutoNATv2 server, most nodes should be able to find enough autonat v2 servers to verify their addresses.

2. Server Rate limiting 

If the ratio of clients to servers is too high, AutoNATv2 will not be useful as clients won't be able to verify their addresses because of server rate limiting. 

In the steady state, when a large chunk(10%?) of the network runs AutoNATv2 servers, this is not a problem. The ratio of DHT nodes in client mode to server mode is roughly [10:1](https://probelab.io/ipfskpi/#ipfs-servers-vs-clients-plot). So we should have a ratio of 10 AutoNATv2 clients to 1 server. If all the clients have 10 addresses each and they verify these 10 addresses every 30 minutes(a very naive implementation), the server needs to verify 100 addresses in 30 minutes which is 4 addresses a minute.

[...]

PR #2469, opened 2023-08-10
autonatv2: implement autonatv2 spec

sukunrt:
closes: #2422 

Will add metrics for this in a separate PR.


marten-seemann (commented):
Partial review. I might be missing something, but where is the logic which address to check next? Or is that left for the implementation of the address pipeline?

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// cleanup removes stale requests.
+//
+// This is fast enough in rate limited cases and the state is small enough to
+// clean up quickly when blocking requests.
+func (r *rateLimiter) cleanup(p peer.ID, now time.Time) {
+	idx := len(r.reqs)
+	for i, t := range r.reqs {
+		if now.Sub(t).Minutes() <= 1 {
```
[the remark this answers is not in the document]
sukunrt: yes, they're equivalent.  Keeping a one minute sliding window to implement rate limiting.

p2p/protocol/autonatv2/client.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+//go:generate protoc --go_out=. --go_opt=Mpbv2/autonat.proto=./pbv2 pbv2/autonat.proto
+
+type Client struct {
+	host       host.Host
+	dialCharge []byte
+
+	mu            sync.Mutex
+	attemptQueues map[uint64]chan ma.Multiaddr
```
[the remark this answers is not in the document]
sukunrt: yes. it is.

sukunrt:
>Partial review. I might be missing something, but where is the logic which address to check next? Or is that left for the implementation of the address pipeline?

Yes, that part will go in the address pipeline.

p2p/protocol/autonatv2/server.go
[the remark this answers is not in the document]
marten-seemann: Can you use the suggested variant then? I think it's easier to understand, since you don't have to worry about rounding.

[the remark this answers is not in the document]
sukunrt: Yes, your suggested variant is better. I'll update.

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,275 @@
...
+		if err := w.WriteMsg(msg); err != nil {
+			log.Debugf("failed to write response: %s", err)
+			s.Reset()
+			return
+		}
+		return
+	}
+
+	if as.dataRequestPolicy(s, dialAddr) {
+		msg.Reset()
```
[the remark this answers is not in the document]
sukunrt: Nope. Fixed these. 
proto.Unmarshal does msg.Reset(): https://github.com/golang/protobuf/issues/890

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,275 @@
...
+	w := pbio.NewDelimitedWriter(s)
+	if err := w.WriteMsg(&pbv2.DialAttempt{Nonce: nonce}); err != nil {
+		s.Reset()
+		return pbv2.DialStatus_E_ATTEMPT_ERROR
+	}
+	// s.Close() here might discard the message
+	s.CloseWrite()
+	s.SetDeadline(as.now().Add(1 * time.Second))
+	b := make([]byte, 1)
+	s.Read(b)
```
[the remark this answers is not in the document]
sukunrt: Updated comment

Any clean way to close a stream with an ACK?
The problem here is, if we just do
```
stream.Write(msg)
stream.Close()
...
host.Network().ClosePeer(pid)
```
The last line will Reset the stream and discard the message.
We do want to remove the peer from the dialer host once the request is complete.

p2p/protocol/autonatv2/autonat.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func New(h host.Host, dialer host.Host, opts ...AutoNATOption) (*AutoNAT, error) {
+	s := defaultSettings()
+	for _, o := range opts {
+		if err := o(s); err != nil {
+			return nil, err
+		}
+	}
+	sub, err := h.EventBus().Subscribe(new(event.EvtLocalReachabilityChanged))
```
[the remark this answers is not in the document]
sukunrt: I copied this logic from AutoNATv1. Do we need to disable the server in case the node is private? 
We can provide this service over a relay connection too since private nodes can dial out and check reachability. Even with amplification attack prevention this only takes 100kb.

p2p/protocol/autonatv2/client.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	var attempt ma.Multiaddr
+	for _, st := range resp.GetDialStatuses() {
+		if st == pbv2.DialStatus_OK {
+			select {
+			case at := <-ch:
+				attempt = at
+			case <-ctx.Done():
+			case <-time.After(5 * time.Second):
```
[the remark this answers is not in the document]
sukunrt: fixed this.

p2p/protocol/autonatv2/client.go
```diff
@@ -0,0 +1,257 @@
...
+	mu            sync.Mutex
+	attemptQueues map[uint64]chan ma.Multiaddr
+}
+
+func NewClient(h host.Host) *Client {
+	return &Client{host: h, dialCharge: make([]byte, 4096), attemptQueues: make(map[uint64]chan ma.Multiaddr)}
+}
+
+func (ac *Client) CheckReachability(ctx context.Context, p peer.ID, highPriorityAddrs []ma.Multiaddr, lowPriorityAddrs []ma.Multiaddr) ([]Result, error) {
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(time.Minute))
```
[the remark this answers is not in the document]
sukunrt: Fixed.

p2p/protocol/autonatv2/autonat.go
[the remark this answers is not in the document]
sukunrt: added a comment explaining the transition period.

p2p/protocol/autonatv2/client.go
```diff
@@ -0,0 +1,257 @@
...
+
+const (
+	ServiceName     = "libp2p.autonatv2"
+	AttemptProtocol = "/libp2p/autonat/2/attempt"
+	maxMsgSize      = 8192
+)
+
+//go:generate protoc --go_out=. --go_opt=Mpbv2/autonat.proto=./pbv2 pbv2/autonat.proto
+
+type Client struct {
```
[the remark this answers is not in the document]
sukunrt: fixed.

p2p/protocol/autonatv2/autonat.go
```diff
@@ -0,0 +1,237 @@
...
+		ctx:           ctx,
+		cancel:        cancel,
+		sub:           sub,
+		srv:           newServer(host, dialerHost, s),
+		cli:           newClient(host),
+		allowAllAddrs: s.allowAllAddrs,
+		peers:         newPeersMap(),
+	}
+	an.cli.RegisterDialBack()
+	an.srv.Enable()
```
MarcoPolo, line 124: This should be part of the Server's Start method. And Disable should be the Stop method. Or Close if it's only expected to be called once.

p2p/protocol/autonatv2/autonat.go
```diff
@@ -0,0 +1,237 @@
...
+
+	an.wg.Add(1)
+	go an.background()
```
MarcoPolo, line 127: This should happen in a Start method.

p2p/protocol/autonatv2/autonat.go
```diff
@@ -0,0 +1,237 @@
...
+func New(host host.Host, dialerHost host.Host, opts ...AutoNATOption) (*AutoNAT, error) {
+	s := defaultSettings()
+	for _, o := range opts {
+		if err := o(s); err != nil {
+			return nil, fmt.Errorf("failed to apply option: %w", err)
+		}
+	}
+	// Listen on event.EvtPeerProtocolsUpdated, event.EvtPeerConnectednessChanged
+	// event.EvtPeerIdentificationCompleted to maintain our set of autonat supporting peers.
+	sub, err := host.EventBus().Subscribe([]interface{}{
```
MarcoPolo, line 103: This should also probably be in the start method. My thinking is that New creates the struct, and `.Start` starts any side-effects/background tasks.

p2p/protocol/autonatv2/client.go
```diff
@@ -0,0 +1,308 @@
...
+		return Result{}, fmt.Errorf("dial msg read failed: %w", err)
+	}
+
+	switch {
+	case msg.GetDialResponse() != nil:
+		break
+	// provide dial data if appropriate
+	case msg.GetDialDataRequest() != nil:
+		idx := int(msg.GetDialDataRequest().AddrIdx)
+		if idx >= len(reqs) { // invalid address index
```
MarcoPolo, line 95: Maybe all these checks should go in a `validateDialDataRequest()` method and we can then do s.Reset() once in a single place.

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,410 @@
...
+				AddrIdx:  uint32(addrIdx),
+				NumBytes: uint64(numBytes),
+			},
+		},
+	}
+	if err := w.WriteMsg(msg); err != nil {
+		return fmt.Errorf("dial data write: %w", err)
+	}
+	for remain := numBytes; remain > 0; {
+		if err := r.ReadMsg(msg); err != nil {
```
MarcoPolo, line 233: Unfortunately I think this will allocate on every loop iteration. Example: https://gist.github.com/MarcoPolo/37338f09fe97f3ce7300f0397da76e30

An alternative would be to just read expected num of bytes from the stream not wrapped it in a pb.

p2p/protocol/autonatv2/client.go
```diff
@@ -0,0 +1,308 @@
...
+
+	localProtos := local.Protocols()
+	externalProtos := external.Protocols()
+	if len(localProtos) != len(externalProtos) {
+		return false
+	}
+	for i := 0; i < len(localProtos); i++ {
+		if i == 0 {
+			if localProtos[i].Code == externalProtos[i].Code ||
+				localProtos[i].Code == ma.P_IP6 && externalProtos[i].Code == ma.P_IP4 /* NAT64 */ {
```
MarcoPolo, line 297: I don't understand what you mean here. Can you elaborate?

p2p/protocol/autonatv2/client.go
```diff
@@ -0,0 +1,308 @@
...
+	addr := reqs[idx].Addr
+
+	var rch network.Reachability
+	switch resp.DialStatus {
+	case pb.DialStatus_OK:
+		if !areAddrsConsistent(dialBackAddr, addr) {
+			// the server is misinforming us about the address it successfully dialed
+			// either we received no dialback or the address on the dialback is inconsistent with
+			// what the server is telling us
+			return Result{}, fmt.Errorf("invalid repsonse: dialBackAddr: %s, respAddr: %s", dialBackAddr, addr)
```
MarcoPolo, line 169: typo

p2p/protocol/autonatv2/options.go
```diff
@@ -0,0 +1,48 @@
+package autonatv2
+
+import "time"
+
+// autoNATSettings is used to configure AutoNAT
+type autoNATSettings struct {
+	allowAllAddrs     bool
```
MarcoPolo, line 7: I prefer the name you used elsewhere "allowPrivateAddrs"

MarcoPolo (commented):
First pass. I think this is looks good! I want to think about any vulnerabilities here.

MarcoPolo (commented):
First pass. I think this is looks good! I want to think about any vulnerabilities here.

core/network/network.go
```diff
@@ -161,6 +161,20 @@ type Network interface {
 	ResourceManager() ResourceManager
 }
 
+// Dialability indicates how dialable a (peer, addr) combination is.
+type Dialability int
+
+const (
+	// DialabilityUnknown indicates that the dialer cannot dial the (peer, addr) combination
```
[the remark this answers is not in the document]
sukunrt: I have removed this. We just use a bool for this now.

p2p/protocol/autonatv2/autonat.go
```diff
@@ -0,0 +1,237 @@
...
+	// Addr is the multiaddr to verify
+	Addr ma.Multiaddr
+	// SendDialData indicates whether to send dial data if the server requests it for Addr
+	SendDialData bool
+}
+
+// Result is the result of the CheckReachability call
+type Result struct {
+	// Idx is the index of the dialed address
+	Idx int
```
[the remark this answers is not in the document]
sukunrt: We can add this field later if needed. I mapped the autonat response to this struct.

p2p/protocol/autonatv2/options.go
sukunrt: incomplete renaming

p2p/protocol/autonatv2/client.go
sukunrt: I will remove this but this was to support NAT64 addresses. Locally they appear to be IPv6 and remotely they are IPv4. But we can add that support later when someone asks for it. I'm not sure if rest of the code works with NAT64 anyway.

p2p/protocol/autonatv2/server.go
sukunrt: We might need to parse the protobuf because in some cases we would be leaving 1 message unread in the stream(data request size 4k + 1). This is fine for now, but I prefer to not have it. 

We can also change the spec to send raw bytes for `DialDataRequest`. In case we add other charge types for dialing an address besides dial data we can support those without any breaking change.

p2p/net/swarm/black_hole_detector.go
```diff
@@ -29,35 +29,26 @@ func (st blackHoleState) String() string {
...
-	n int
-	// minSuccesses is the minimum number of Success required in the last n dials
+// BlackHoleFilter provides black hole filtering for dials. This filter should be used in concert
+// with a UDP or IPv6 address filter to detect UDP or IPv6 black hole. In a black holed environment,
+// dial requests are refused Requests are blocked if the number of successes in the last N dials is
+// less than MinSuccesses.
+// If a request succeeds in Blocked state, the filter state is reset and N subsequent requests are
+// allowed before reevaluating black hole state. Dials cancelled when some other concurrent dial
+// succeeded are counted as failures. A sufficiently large N prevents false negatives in such cases.
+type BlackHoleFilter struct {
```
sukunrt, line 39: Other possible names: 

BlackHoleSuccessCounter
BlackHoleTracker

p2p/protocol/autonatv2/server.go
MarcoPolo: I think changing the spec to just send and expect empty bytes is fine. I don't think we gain anything by wrapping it in a pb.

MarcoPolo:
Can we default to having autonatv2 disabled initially? I want folks to opt-in to it at the start until we gain some more confidence with experience.

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,394 @@
...
+
+type entry struct {
+	PeerID peer.ID
+	Time   time.Time
+}
+
+func (r *rateLimiter) Accept(p peer.ID) bool {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if r.peerReqs == nil {
```
MarcoPolo, line 305: Should this happen after cleanup and the checks below? Maybe right before the assignments on L321.

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,394 @@
...
+	r.ongoingReqs[p] = struct{}{}
+	r.reqs = append(r.reqs, entry{PeerID: p, Time: nw})
+	r.peerReqs[p] = append(r.peerReqs[p], nw)
+	return true
+}
+
+func (r *rateLimiter) AcceptDialDataRequest(p peer.ID) bool {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	if r.peerReqs == nil {
```
MarcoPolo, line 329: Same comment here

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+	r.dialDataReqs = append(r.dialDataReqs, nw)
+	return true
+}
+
+// cleanup removes stale requests.
+//
+// This is fast enough in rate limited cases and the state is small enough to
+// clean up quickly when blocking requests.
+func (r *rateLimiter) cleanup(now time.Time) {
```
MarcoPolo, line 346: Can we write a fuzz test for the rateLimiter? I think it looks right, but there might be some subtleties. One simple check we could do is to do a bunch of operations and ensure that after some time we end up with empty states.

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,394 @@
...
+			log.Debugf("failed to write dial refused response to %s: %s", p, err)
+			return
+		}
+		return
+	}
+
+	nonce := msg.GetDialRequest().Nonce
+
+	isDialDataRequired := as.dialDataRequestPolicy(s, dialAddr)
+	if !as.limiter.AcceptDialDataRequest(p) {
```
MarcoPolo, line 160: Should this only apply if `isDialDataRequired == true`?

MarcoPolo (commented):
Just a couple comments left. I think this is good after those are addressed and we change this to be opt-in.

p2p/protocol/autonatv2/server.go
MarcoPolo: can we do something better here? I don’t love the quadratic complexity for every call. 

Maybe we can track when the next cleanup time is and return early if we aren’t there yet

core/network/network.go
```diff
@@ -194,6 +194,9 @@ type Dialer interface {
 	// Notify/StopNotify register and unregister a notifiee for signals
 	Notify(Notifiee)
 	StopNotify(Notifiee)
+
+	// CanDial returns whether the dialer can dial peer p at addr
```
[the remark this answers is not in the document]
sukunrt: It means we _will_ dial when true and we _wont_ dial when false. 

In the case of autonatv2 swarm use a read only black hole detector(because there may be many false negatives) so whenever we are not sure about our connectivity we will not attempt to dial from the autonatv2 swarm.

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+	defer s.Scope().ReleaseMemory(maxMsgSize)
+
+	s.SetDeadline(as.now().Add(streamTimeout))
+	defer s.Close()
+
+	p := s.Conn().RemotePeer()
+
+	var msg pb.Message
+	w := pbio.NewDelimitedWriter(s)
```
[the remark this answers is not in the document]
sukunrt: We need to send the rate limited response if we block the request.

p2p/protocol/autonatv2/server.go
sukunrt: It's easier to reason if it's set upfront, right?

p2p/protocol/autonatv2/server.go
sukunrt: This should be O(N) at worst with total number of entries ~200. Do you think this is quadratic? 
For every request that is older than a minute, we check that peers specific requests in the `peerReq` map. We will need to clear the `peerReqs` slice only once.

p2p/protocol/autonatv2/server.go
sukunrt: Thanks!

p2p/protocol/autonatv2/server.go
sukunrt: Rust libp2p implementation prefers to have typed streams. I believe the idea is that you can wrap streams like: `Stream<AutoNATMsg>` so all read calls return object of the type `AutoNATMsg`

p2p/protocol/autonatv2/client.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+	w := pbio.NewDelimitedWriter(s)
+	res := pb.DialBackResponse{}
+	if err := w.WriteMsg(&res); err != nil {
+		log.Debugf("failed to write dialback response: %s", err)
+		s.Reset()
+	}
+}
+
+func areAddrsConsistent(local, external ma.Multiaddr) bool {
```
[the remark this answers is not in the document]
sukunrt: using `connLocalAddr, dialedAddr `

p2p/protocol/autonatv2/server.go
sukunrt: How about the test: `TestRateLimitStress`
https://github.com/libp2p/go-libp2p/blob/1219f9ea13b503721ee15661e1df987f947bd04d/p2p/protocol/autonatv2/server_test.go#L286

It does a bunch of opts in different go routines and we check whether the numbers are roughly correct. After your suggestion, modified it to also check the lengths of the internal data structures. 

We should improve this rate limiter to do token bucket rate limiting like https://pkg.go.dev/golang.org/x/time/rate But I think it's fine to take that up in a separate PR.

p2p/protocol/autonatv2/server.go
MarcoPolo: No strong opinion. We'll only do this ever once so it's fine.

p2p/protocol/autonatv2/server.go
MarcoPolo: > This should be O(N) at worst with total number of entries ~200. Do you think this is quadratic?

It's O(N reqs * M reqs/peer). Default settings for M is at most 2, and N is at most 20. So these are fine, my apologies. I get nervous when I see two for loops :)

MarcoPolo (approved):

p2p/protocol/autonatv2/server.go
sukunrt: I did custom protobuf parsing. Now the number of allocations are 4 for send side and 2 for read side for the entire dial data(100kB).

MarcoPolo: What were they before?

p2p/protocol/autonatv2/server.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+	// pbio.Reader that we used so far on this stream is buffered. But at this point
+	// there is nothing unread on the stream. So it is safe to use the raw stream to
+	// read, reducing allocations.
+	return readDialData(numBytes, s)
+}
+
+func readDialData(numBytes int, r io.Reader) error {
+	mr := &msgReader{R: r, Buf: make([]byte, maxMsgSize)}
+	for remain := numBytes; remain > 0; {
```
MarcoPolo, line 226: Using 	`pool "github.com/libp2p/go-buffer-pool"`. Saves you one alloc.
```suggestion
	buf := pool.Get(maxMsgSize)
	defer pool.Put(buf)
	mr := &msgReader{R: r, Buf: buf}
```

p2p/protocol/autonatv2/server.go
MarcoPolo: Answering my question, I think around 10 times more.

p2p/protocol/autonatv2/client.go
```diff
@@ -0,0 +1,356 @@
...
+	data, err := proto.Marshal(msg)
+	if err != nil {
+		return err
+	}
+	lenBuf := make([]byte, 5) // 5 bytes is enough for 4k
+	n := varint.PutUvarint(lenBuf, uint64(len(data)))
+	lenBuf = lenBuf[:n]
+
+	for remain := numBytes; remain > 0; {
+		_, err = w.Write(lenBuf)
```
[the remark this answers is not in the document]
sukunrt: I agree with the suggestion to remove this on the client. We can always add this side later on when we start using this from rest of the code. 

>won't this overshoot the amount of data to dial?

I think this is okay. Servers should handle this case.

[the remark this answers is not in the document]
MarcoPolo: We should update the spec if the client may send a bit more data.

MarcoPolo (approved):

MarcoPolo (approved):

PR #2472, opened 2023-08-11
transport tests: refactor workers in TestMoreStreamsThanOurLimits

MarcoPolo:
The goal of this test is to make a bunch of requests in parallel and hit a peer's limits and eventually make all our requests. The test starts with a certain number of workers and increases them exponentially as long as we haven't seen an error.

The previous code did this as well, but it was extremely subtle. I think this refactor is much clearer in its intent. Instead of having a semaphore that controls the number of workers allowed to run at once, we have dedicated worker goroutines that pull work from a queue. The initial worker (index=0) is special in that it can spawn more workers. This makes it clear who is increasing the worker count and how many parallel workers we have.

This also allows us to set initial worker counts and max worker counts easily, which is useful for debugging some transports.

The core logic of each worker hasn't changed, so it may be useful to review with `?w=1` github flag to ignore whitespace changes.


marten-seemann (approved):

marten-seemann (changes requested):
Code LGTM, but CI is unhappy.

marten-seemann (approved):

PR #2478, opened 2023-08-14
QUIC: drop draft-29

MarcoPolo:
This does the minimal amount of changes to drop draft29. Along the way I ran into this issue: https://github.com/quic-go/quic-go/issues/4029.

I'm torn between removing all the machinery of the multiversion quic transport or keeping it.

The reasons to keep it:
1. It'll make adding support for QUIC-v2 very easy.
2. It'll avoid us from reinventing this when we do. This is tricky code and this code works and has been deployed for a bit.

The reasons to remove it:
1. Some complexity around muxing between versions that we aren't benefitting from right now.

I'm less worried about code paths that are untrodden, because we are still exercising these codepaths. They are just used to multiplex a single option.

If we think we'll add QUIC-v2 soon (next 6 months), I think we should keep it. Open to hear other folks thoughts on this.


MarcoPolo:
I believe failures in Go 1.19 are expected.

marten-seemann (approved):

marten-seemann (commented):
It would probably make sense to split this into two PRs:
1. drop 29
2. update quic-go

That way, we'll be able to bisect later if we encounter any problems.

Neo2308:
@marten-seemann could we use https://github.com/libp2p/go-libp2p/pull/2485 for the quic go update? I have also added the updated it in the subdirs similar to this PR.

PR #2483, opened 2023-08-15
Added dependabot

Neo2308:
* Added dependabot to keep go dependencies & actions updated.


marten-seemann:
Please no, dependabot is very noisy.

Neo2308:
While dependabot can be very noisy, it can be handy for automatically keeping dependencies up to date. There is a feature to batch all the updates into a single PR. I can update the config if you would like.

marten-seemann:
We’ve tried it before and we have no intention of activating it again.

PR #2487, opened 2023-08-15
quic: drop support for QUIC draft-29

MarcoPolo:
This does the minimal amount of changes to drop draft29 without updating quic-go.

I'm thinking, for now at least, that we'll leave machinery of the multiversion QUIC for a couple of reasons:

1. It'll make adding support for QUIC-v2 very easy.
  a. This would mostly be greasing exercise to avoid ossification on `quic-v1`. There's nothing QUIC-v2 gives us that we don't have with QUIC-v1.
2. It'll avoid us from reinventing this if we re-introduce this. This is tricky code and this code works and has been deployed for a bit.
3. It's stable as is, and a refactor would introduce more variables to the next release (already quite big with quic-go changes).

### DIFF
--- a/defaults.go
+++ b/defaults.go
@@ -79,11 +79,9 @@ var RandomIdentity = func(cfg *Config) error {
 var DefaultListenAddrs = func(cfg *Config) error {
 	addrs := []string{
 		"/ip4/0.0.0.0/tcp/0",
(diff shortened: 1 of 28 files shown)


marten-seemann (approved):

marten-seemann:
Thank you for splitting up the PR @MarcoPolo!

PR #2488, opened 2023-08-15
quic-go: Update quic-go to v0.37.4

MarcoPolo:
This branch is targetting on https://github.com/libp2p/go-libp2p/pull/2487. Updates quic-go and our usage of it. I'd recommend reviewing this PR commit by commit because the hack to avoid running examples in Go 1.21 is noisy.

The workaround for examples is only temporary. The issue is that the examples use a released version of go-libp2p, but no released version supports Go 1.21 yet. This ignores Go 1.21 in the examples. This commit is designed to be reverted when we release the next version of go-libp2p.

Also note that this works around this issue: https://github.com/quic-go/quic-go/issues/4029. A later version of quic-go will fix this.


MarcoPolo:
Failing interop tests against zig-libp2p pretty reliably. Debugging this now.

MarcoPolo:
Interop issue will be fixed in the next version of quic-go

marten-seemann (approved):

marten-seemann:
Wrong merge target. Will recreate.

PR #2493, opened 2023-08-16
perf: do not copy the addresses in (*BasicHost).Addrs

Jorropo:
We do not need to copy the addresses here because AllAddrs already copies and returns a unique slice. We assume that AddrsFactory will do the same and add documentaion saying so.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -54,6 +54,7 @@ var (
 
 // AddrsFactory functions can be passed to New in order to override
 // addresses returned by Addrs.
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Tests fail because of data races around the returned addr slice. Presumably the caller of Addrs is no longer receiving a unique copy of the addr slices.

> AllAddrs already copies and returns a unique slice

Is that true? I don't see that looking at the code.

marten-seemann:
Let’s not spend too much effort on this. We’re implementing the address pipeline soon, and this will get rewritten anyway.

PR #2498, opened 2023-08-17
muxer: remove support for mplex

marten-seemann:
For https://github.com/libp2p/specs/issues/553.

Kubo recently removed mplex support: https://github.com/ipfs/kubo/issues/9958.


mxinden (approved):

sukunrt (approved):
🥳

nisdas:
For any other downstream users who need mplex, how would they go about using `go-libp2p` with it ? From what I understand we will be unable to update Prysm to the next go-libp2p release which contains this PR as we are still required to support mplex.

sukunrt:
@nisdas I'm not aware of past conversations regarding this, so apologies if this is incorrect.
The removed package looks self contained to me. Can you maintain a fork of that specific package?

nisdas:
@sukunrt That would require me to maintain a separate fork of go-libp2p. I would like to avoid maintaining a fork of go-libp2p if possible here

sukunrt:
I haven't looked at this closely, but this should be possible: 

1. Copy the files in package `github.com/libp2p/go-libp2p/p2p/muxer/mplex` to a new repo.
2. Rename the package to the repo path.
3. Pass in the new package mplex muxer to libp2p option `func Muxer(name string, muxer network.Multiplexer) Option`

So you should be able to only fork the package and not entire go-libp2p.

marten-seemann:
@sukunrt is right. We built libp2p in such a way that it’s modular: you can bring your own muxer implementation without the need to upstream it. 

If that doesn’t work, we did something wrong.

nisdas:
Thank you @sukunrt @marten-seemann , I will take a look at including mplex separately then in a separate repo/package

sukunrt:
@nisdas 
we've moved this package to https://github.com/libp2p/go-libp2p-mplex/
you can just change your import path for the muxer.

nisdas:
Perfect, thanks a lot ! @sukunrt

PR #2499, opened 2023-08-17
chore: remove unused and outdated package-list.json

marten-seemann:
I don't think this is needed anywhere. @p-shahi Can you check if this is imported anywhere (potentially from the libp2p website)?


p-shahi (approved):
It isn't imported anywhere

PR #2513, opened 2023-08-23
ci: limit interop tests runs to core folders

p-shahi:
Avoid running the interop tests on code changes to directories that do not need testing.
Ref: https://github.com/libp2p/test-plans/pull/272

### DIFF
--- a/.github/workflows/interop-test.yml
+++ b/.github/workflows/interop-test.yml
@@ -1,9 +1,22 @@
 name: Interoperability Testing
 on:
+  workflow_dispatch:
   pull_request:
+    paths:
+      - 'config/**'
+      - 'core/**'
+      - 'internal/**'
+      - 'p2p/**'
+      - 'test-plans/**'
   push:
     branches:
       - "master"
+    paths:
+      - 'config/**'
+      - 'core/**'
+      - 'internal/**'
+      - 'p2p/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/interop-test.yml
```diff
@@ -1,9 +1,22 @@
 name: Interoperability Testing
 on:
+  workflow_dispatch:
```
marten-seemann, line 3: What is this doing?

marten-seemann (commented):
Would it make sense to use a blacklist instead of a whitelist here? I assume you want to exclude runs for changes to the README and similar files.

.github/workflows/interop-test.yml
p-shahi: it enables a workflow to be triggered [manually](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow#configuring-a-workflow-to-run-manually).

p-shahi:
> Would it make sense to use a blacklist instead of a whitelist here? I assume you want to exclude runs for changes to the README and similar files.

The way its configured now with the allowlist, it won't run for top level directory changes either. if your preference is to have a deny list I can make that change.

MarcoPolo (approved):
No preference for allow vs deny.

PR #2518, opened 2023-08-24
metrics: fix resource manager dashboard name

sukunrt:
Introduced in https://github.com/libp2p/go-libp2p/pull/2512 where I prefixed all dashboard names with `go-`


sukunrt:
My bad, @p-shahi fixed this here https://github.com/libp2p/go-libp2p/pull/2512/commits/a0e8da1e70365d44b3779fe427448780f30159b0
I was only looking at my local branch.

PR #2519, opened 2023-08-24
quic: return a more meaningful error when dialing QUIC draft-29

### LINKED ISSUE #2515: Better error message when trying to dial `/quic` peer
Using `v0.30.0`:
Currently `.Error()` gives:
```
Error: failed to dial: failed to dial $PEERID:
  * [/ip4/$IP/udp/4001/quic] unknown QUIC version
```
I would rather see something along the lines of:
```
Error: failed to dial: failed to dial $PEERID:
  * [/ip4/$IP/udp/4001/quic] QUIC draft 29 has been removed, QUIC RFC9000 is accessible with /quic-v1
```
I think a self explanatory error message is warranted here because this used to work and now it don't.
Note that this will be shown if all dials fail.

marten-seemann:
Fixes #2515.

### DIFF
--- a/p2p/transport/quicreuse/quic_multiaddr.go
+++ b/p2p/transport/quicreuse/quic_multiaddr.go
@@ -9,9 +9,11 @@ import (
 	"github.com/quic-go/quic-go"
 )
 
-var (
-	quicV1MA = ma.StringCast("/quic-v1")
-)
+var quicV1MA = ma.StringCast("/quic-v1")
+
+const quicDraft29 quic.VersionNumber = 0xff00001d
+
+var ErrQUICDraft29 = errors.New("QUIC draft-29 has been removed, QUIC (RFC 9000) is accessible with /quic-v1")
 
 func ToQuicMultiaddr(na net.Addr, version quic.VersionNumber) (ma.Multiaddr, error) {
 	udpMA, err := manet.FromNetAddr(na)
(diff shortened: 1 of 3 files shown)


p2p/transport/quicreuse/quic_multiaddr.go
```diff
@@ -21,6 +23,8 @@ func ToQuicMultiaddr(na net.Addr, version quic.VersionNumber) (ma.Multiaddr, err
 	switch version {
 	case quic.Version1:
 		return udpMA.Encapsulate(quicV1MA), nil
+	case quicDraft29:
```
marten-seemann, line 26: Not sure if this is needed here, but it certainly doesn't hurt.

Jorropo (approved):
SGTM thx

marten-seemann:
I'm not sure if this fix will still work with #2520. But that's a bug that we need to fix.

MarcoPolo (approved):

marten-seemann:
With #2520, this won't work any more, because it's not the QUIC transport's responsibility any more to handle /quic addresses. Instead, the swarm would now return an `ErrNoGoodAddresses`. I'm not sure if we should change that, since this an exported error and users might do an error assertion.

@Jorropo, any thoughts?

marten-seemann:
Closing. We'll solve this in the swarm after #2437 has landed.

PR #2521, opened 2023-08-24
feat: Bring in gostream to the monorepo

MarcoPolo:
#2438 Builds on gostream. I think it would be good to bring it in to the monorepo.


marten-seemann (changes requested):
This doesn't preserve Git history. Do you want me to use my repo consolidation script to do that?

p2p/gostream/conn.go
marten-seemann, line 1: Should this go into `p2p/net` instead?

MarcoPolo: Done.

MarcoPolo:
I think this is simple enough that there isn't a ton to gain from the perspective of keeping history to help debug things.

marten-seemann:
> I think this is simple enough that there isn't a ton to gain from the perspective of keeping history to help debug things. But if you want to run your script, you are welcome to do so.

Probably not, I'd still prefer to preserve history. I don't see a good reason not to.

PR #2524, opened 2023-08-25
swarm: return a more meaningful error when dialing QUIC draft-29

marten-seemann:


p2p/net/swarm/swarm_dial.go
```diff
@@ -63,6 +65,9 @@ var (
 	// ErrGaterDisallowedConnection is returned when the gater prevents us from
 	// forming a connection with a peer.
 	ErrGaterDisallowedConnection = errors.New("gater disallows connection to peer")
+
+	// ErrQUICDraft29 is returned instead of an ErrNoTransport when attempting to dial a QUIC draft-29 address.
+	ErrQUICDraft29 = errors.New("QUIC draft-29 has been removed, QUIC (RFC 9000) is accessible with /quic-v1")
 )
```
Jorropo, line 71:
```suggestion
)

// ErrQUICDraft29 wraps [ErrNoTransport] and provide a more meaningful error message while
type ErrQUICDraft29 struct{}

func (ErrQUICDraft29) Error() string {
 return "QUIC draft-29 has been removed, QUIC (RFC 9000) is accessible with /quic-v1"
}

func (ErrQUICDraft29) Unwrap() error {
 return ErrNoTransport
}
```
This is enough to allow `errors.Is(ErrQUICDraft29{}, ErrNoTransport)` to return true.

p2p/net/swarm/dial_test.go
```diff
@@ -642,3 +642,11 @@ func TestDialSelf(t *testing.T) {
...
 	require.ErrorIs(t, err, swarm.ErrDialToSelf, "expected error from self dial")
 }
+
+func TestDialQUICDraft29(t *testing.T) {
+	s := makeDialOnlySwarm(t)
+	id := testutil.RandPeerIDFatal(t)
+	s.Peerstore().AddAddr(id, ma.StringCast("/ip4/127.0.0.1/udp/1234/quic"), time.Hour)
+	_, err := s.DialPeer(context.Background(), id)
+	require.ErrorIs(t, err, swarm.ErrQUICDraft29)
+}
```
Jorropo, line 652:
```suggestion
	require.ErrorIs(t, err, swarm.ErrNoTransport)
}
```

marten-seemann:
Nice one, thanks @Jorropo!

p2p/net/swarm/swarm_dial.go
```diff
@@ -65,6 +67,17 @@ var (
 	ErrGaterDisallowedConnection = errors.New("gater disallows connection to peer")
 )
 
+// ErrQUICDraft29 wraps [ErrNoTransport] and provide a more meaningful error message while
```
marten-seemann, line 70:
```suggestion
// ErrQUICDraft29 wraps ErrNoTransport and provide a more meaningful error message
```

Jorropo (approved):
LGTM, you need to `go fmt` since I used space instead of tabs in my suggestion.
Thx

p2p/net/swarm/swarm_dial.go
Jorropo: BTW the `[]` around `ErrNoTransport` makes the string `ErrNoTransport` on pkg.go.dev a clickable link to `ErrNoTransport`'s definition

Jorropo: Docs here https://golang.org/doc/comment

Jorropo: I use gopls which don't do that.
I think it would take an hour or two max to write a linter which finds all potentially applications of this.
Someone might already have done so.

sukunrt (approved):

PR #2527, opened 2023-08-26
config: warn if connmgr limits conflict with rcmgr

piersy:
closes: #1707 

If the high water mark of the connection manager exceeds the total connection limit of the resource manager then a warning will be issued at node startup.

@MarcoPolo, I actually think it would be neater to pass an int to `CheckLimit` as opposed to what I've currently implemented. 
```go
CheckLimit(systemLimit connmgr.GetterConnLimit)
```
Let me know what you think


MarcoPolo (commented):
>  I actually think it would be neater to pass an int to CheckLimit as opposed to what I've currently implemented.

That could make sense. Who would own the interface that the resourcemanager implements to get the int?

I think I prefer this model because the interface is owned by the conn manager, and the resource manager is implementing it so that it can be used to check against.

I almost would consider flipping these so that the resource manager is the one checking against other things, since it's the one that is the ultimate source of truth. But I don't think it really matters.

Besides a single comment to hopefully give a nicer warning, I think this is good. Thank you!

piersy:
Hi @MarcoPolo could you elaborate on how to improve the warning message?

Currently the warnings would take the following format:

>conn manager high water limit: 8, exceeds the system limit of: 4

config/config.go
```diff
@@ -295,6 +295,16 @@ func (cfg *Config) addTransports(h host.Host) error {
 //
 // This function consumes the config. Do not reuse it (really!).
 func (cfg *Config) NewNode() (host.Host, error) {
+	// If possible check that the resource manager conn limit is higer than the
+	// limit set in the conn manager.
```
sukunrt, line 299:
```suggestion
	// If possible check that the resource manager conn limit is higher than the
	// limit set in the conn manager.
```

config/config.go
```diff
@@ -295,6 +295,16 @@ func (cfg *Config) addTransports(h host.Host) error {
...
+	l, ok := cfg.ResourceManager.(connmgr.GetterConnLimit)
+	if ok {
+		err := cfg.ConnManager.CheckLimit(l)
+		if err != nil {
+			log.Warn(err)
+		}
+	}
```
sukunrt, line 306: NIT:
```
if l, ok := cfg.ResourceManager.(connmgr.GetterConnLimit); ok {
	err := cfg.ConnManager.CheckLimit(l)
	if err != nil {
		log.Warn(err)
	}
}
```

config/config.go
```diff
@@ -295,6 +295,16 @@ func (cfg *Config) addTransports(h host.Host) error {
 //
 // This function consumes the config. Do not reuse it (really!).
 func (cfg *Config) NewNode() (host.Host, error) {
+	// If possible check that the resource manager conn limit is higer than the
+	// limit set in the conn manager.
+	l, ok := cfg.ResourceManager.(connmgr.GetterConnLimit)
+	if ok {
+		err := cfg.ConnManager.CheckLimit(l)
+		if err != nil {
+			log.Warn(err)
```
sukunrt, line 304: It'd be nice to augment the warning with the fact that it comes from the resource manager:
something like this:
`rcmgr limit conflicts with connmgr limit: conn manager high water limit: 8, exceeds the system limit of: 4`

core/connmgr/manager.go
```diff
@@ -89,3 +93,9 @@ type TagInfo struct {
 	// Conns maps connection ids (such as remote multiaddr) to their creation time.
 	Conns map[string]time.Time
 }
+
+// GetterConnLimit provides access to a component's total connection limit.
+type GetterConnLimit interface {
+	// GetConnLimit returns the total connection limit of the implementing component.
```
sukunrt, line 99: I'd prefer `GetConnLimiter` or `ConnLimitGetter` over the current one. 
Between the two slight preference for `GetConnLimiter` since it is inline with [effective-go](https://go.dev/doc/effective_go) recommendation:
>By convention, one-method interfaces are named by the method name plus an -er suffix or similar modification to construct an agent noun: Reader, Writer, Formatter, CloseNotifier etc.

p2p/net/connmgr/connmgr.go
```diff
@@ -239,6 +240,13 @@ func (cm *BasicConnMgr) IsProtected(id peer.ID, tag string) (protected bool) {
 	return protected
 }
 
+func (cm *BasicConnMgr) CheckLimit(systemLimit connmgr.GetterConnLimit) error {
+	if cm.cfg.highWater > systemLimit.GetConnLimit() {
+		return fmt.Errorf("conn manager high water limit: %d, exceeds the system limit of: %d", cm.cfg.highWater, systemLimit.GetConnLimit())
```
sukunrt, line 245: NIT: I'd prefer: 
>conn manager high watermark limit: %d, exceeds the system connection limit of: %d

Though this is fine too since from connmgr it can be inferred that the issue is relating to connection limits.

piersy:
Hi @sukunrt, thanks for the review, I think I've addressed all your points.

sukunrt (approved):
Thank you, @piersy

PR #2529, opened 2023-08-27
host: disable black hole detection on autonat dialer

sukunrt:
If the autonat server receives too many requests from private nodes, this can cause the autonat server to incorrectly skip QUIC or IPv6 dial back requests. 

Fixes: #2530


marten-seemann (commented):
What if we're actually behind a blackhole? This means we'll report incorrect results, right?

sukunrt:
Yeah, but that's existing behaviour. This is handled in AutoNAT v2, we can port the same strategy here too.

marten-seemann (approved):

Issue #2510, opened 2023-08-23
Feature request: Add IPv6 tests to transport integration tests


MarcoPolo:
We should add IPv6 versions to the the `transportsToTest` var in `p2p/test/transport/transport_test.go` so that we test each transport with IPv6.

MarcoPolo:
Please do, Thanks! Let me know if you run into any questions

ajaygupta112:
@MarcoPolo is this issue open for taking?

sukunrt:
@ajaygupta112 Yes!
You can start here: https://github.com/libp2p/go-libp2p/pull/2531/files#diff-54f6226e2cf835d8f4f664f08e1e9a325aaa75d1bf9384c89f3c66fbd5ff8ae7

To begin with, you can enable just the Ping test for ip6.

ajaygupta112:
@sukunrt Alright, thanks

DarkLord017:
Is this issue still open , would like to take this up !

varun-r-mallya:
Interested in this! Please assign this to me

rishishanbhag:
@MarcoPolo, if this issue is still open, could you please assign it to me? I would be happy to work on it.

MarcoPolo:
No need to assign. Just give it a go.

rishishanbhag:
@MarcoPolo I’ve opened PR #3461 addressing this issue.
Would appreciate your review when you get a chance.

PR #2531, opened 2023-08-27
#2510: Adding IPv6 cases to transport tests

CtrlAltGiri:
Fixing #2510

### DIFF
--- a/p2p/test/transport/transport_test.go
+++ b/p2p/test/transport/transport_test.go
@@ -134,6 +134,80 @@ var transportsToTest = []TransportTestCase{
 			return h
 		},
 	},
+	{
+		Name: "TCP / Noise / Yamux - IP6",
+		HostGenerator: func(t *testing.T, opts TransportTestCaseOpts) host.Host {
+			libp2pOpts := transformOpts(opts)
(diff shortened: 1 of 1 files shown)


CtrlAltGiri:
Fixes #2510

marten-seemann:
This will double the run time of this test suite, which is already taking 10s, and we'll be adding more tests in the future. Should this instead be a single IPv6 test?

MarcoPolo:
Great to see no failures!

I think another 10s is an okay tradeoff because:

1. We don't really test IPv6 anywhere else.
2. We really want to test all the things you can do with a transport, this includes doing all those things with IPv6 as well.

> Should this instead be a single IPv6 test?

I think if IPv6 had been interoperable with IPv4, this would be something I would agree with. Sadly, we are living in a two stack world, so I think it makes sense to run all these tests with IPv6 as well.

marten-seemann:
> 2\. We really want to test all the things you can do with a transport, this includes doing all those things with IPv6 as well.

Lots of the tests test functionality that's going on _on top_ of an established connection, for example the stream deadline test, gating test, etc. There's not a lot of value in running those on IPv6, is there?

Issue #2321, opened 2023-06-02
Use CancelWithCause in swarm.dialPeer for better metrics reporting


sukunrt:
By using https://pkg.go.dev/context#CancelCauseFunc in swarm.dialPeer we can detect if we cancelled because we received a connection while tracking metrics. This will help us understand if we are cancelling because we received a connection or because the caller cancelled the call for some reason. 

Since this type is not available in go 1.19 we might have to wait till go1.21 to tackle this.

marten-seemann:
This is unblocked now, since we switched to Go 1.20. @sukunrt, should we get this into the v0.31 release?

sukunrt:
Yeah, it'd be nice to get this in. Thanks for the reminder. I've raise a PR.

PR #2532, opened 2023-08-27
swarm: track dial cancellation reason

sukunrt:
closes: #2321

### DIFF
--- a/p2p/net/swarm/dial_sync.go
+++ b/p2p/net/swarm/dial_sync.go
@@ -2,6 +2,8 @@ package swarm
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"sync"
 
 	"github.com/libp2p/go-libp2p/core/network"
@@ -10,6 +12,11 @@ import (
 
 // dialWorkerFunc is used by dialSync to spawn a new dial worker
 type dialWorkerFunc func(peer.ID, <-chan dialRequest)
+
+// errParentContextCanceled is used with cancelCause to cancel the activeDial when the parent
+// context is cancelled. This helps distinguish whether parallel dial succeeded or the entire dial
+// was canceled when inspecting the dial context with context.Cause
+var errParentContextCanceled = errors.New("parent context cancelled")
 
 // newDialSync constructs a new dialSync
 func newDialSync(worker dialWorkerFunc) *dialSync {
@@ -30,14 +37,17 @@ type dialSync struct {
 type activeDial struct {
 	refCnt int
 
-	ctx    context.Context
-	cancel func()
+	ctx         context.Context
+	cancelCause func(error)
 
 	reqch chan dialRequest
 }
 
-func (ad *activeDial) close() {
-	ad.cancel()
+func (ad *activeDial) close(err error) {
+	if errors.Is(err, context.Canceled) {
+		err = fmt.Errorf("%w: %w", errParentContextCanceled, context.Canceled)
+	}
+	ad.cancelCause(err)
 	close(ad.reqch)
 }
 
@@ -74,11 +84,11 @@ func (ds *dialSync) getActiveDial(p peer.ID) (*activeDial, error) {
 	if !ok {
 		// This code intentionally uses the background context. Otherwise, if the first call
 		// to Dial is canceled, subsequent dial calls will also be canceled.
-		ctx, cancel := context.WithCancel(context.Background())
+		ctx, cancel := context.WithCancelCause(context.Background())
 		actd = &activeDial{
-			ctx:    ctx,
-			cancel: cancel,
-			reqch:  make(chan dialRequest),
+			ctx:         ctx,
+			cancelCause: cancel,
+			reqch:       make(chan dialRequest),
 		}
 		go ds.dialWorker(p, actd.reqch)
 		ds.dials[p] = actd
@@ -101,9 +111,10 @@ func (ds *dialSync) Dial(ctx context.Context, p peer.ID) (*Conn, error) {
 		defer ds.mutex.Unlock()
 		ad.refCnt--
 		if ad.refCnt == 0 {
-			ad.close()
+			ad.close(err)
 			delete(ds.dials, p)
 		}
 	}()
-	return ad.dial(ctx)
+	conn, err := ad.dial(ctx) // updated err is used in defered func
+	return conn, err
 }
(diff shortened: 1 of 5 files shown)


p2p/net/swarm/swarm_metrics.go
```diff
@@ -216,18 +216,30 @@ func (m *metricsTracer) CompletedHandshake(t time.Duration, cs network.Connectio
...
-	} else if errors.Is(err, context.DeadlineExceeded) {
+	// dial deadline exceeded or the the parent contexts deadline exceeded
+	if errors.Is(dialErr, context.DeadlineExceeded) || errors.Is(cause, context.DeadlineExceeded) {
 		e = "deadline"
+	} else if errors.Is(dialErr, context.Canceled) {
+		// dial was cancelled.
+
+		if errors.Is(cause, errParentContextCanceled) {
+			// parent was canceled
+			e = "canceled"
```
marten-seemann, line 230: This necessarily means that the application canceled the dial, right? Should we say "application canceled" to make it clear that it wasn't libp2p's fault?

p2p/net/swarm/swarm_metrics.go
```diff
@@ -216,18 +216,30 @@ func (m *metricsTracer) CompletedHandshake(t time.Duration, cs network.Connectio
...
+		} else if errors.Is(cause, context.Canceled) {
```
marten-seemann, line 231: This looks brittle. Should we have a separate cause for this?

p2p/net/swarm/swarm_metrics.go
```diff
@@ -216,18 +216,30 @@ func (m *metricsTracer) CompletedHandshake(t time.Duration, cs network.Connectio
...
+			// another parallel dial succeeded
+			e = "canceled: dial successful"
```
marten-seemann, line 233:
```suggestion
			e = "canceled: concurrent dial successful"
```

p2p/net/swarm/swarm_metrics.go
```diff
@@ -216,18 +216,30 @@ func (m *metricsTracer) CompletedHandshake(t time.Duration, cs network.Connectio
...
+		} else {
+			// something else
+			e = "canceled: other"
```
marten-seemann, line 236: Do we expect this to happen at all?

p2p/net/swarm/dial_sync.go
```diff
@@ -30,14 +37,17 @@ type dialSync struct {
...
-	cancel func()
+	ctx         context.Context
+	cancelCause func(error)
 
 	reqch chan dialRequest
 }
 
-func (ad *activeDial) close() {
-	ad.cancel()
+func (ad *activeDial) close(err error) {
```
marten-seemann, line 46: Nit: it might be clearer if you just inline this function, as it's only used once.

p2p/net/swarm/dial_sync.go
```diff
@@ -101,9 +111,10 @@ func (ds *dialSync) Dial(ctx context.Context, p peer.ID) (*Conn, error) {
 		defer ds.mutex.Unlock()
 		ad.refCnt--
 		if ad.refCnt == 0 {
-			ad.close()
+			ad.close(err)
 			delete(ds.dials, p)
 		}
 	}()
-	return ad.dial(ctx)
+	conn, err := ad.dial(ctx) // updated err is used in defered func
```
marten-seemann, line 118: Are you sure this works? Isn't `:=` creating a new `err` here?

Why do we have this `defer` anyway, there's only a single return from this function.

p2p/net/swarm/swarm_metrics.go
sukunrt: Don't think so. Didnt happen on my kubo node. I ran it for about an hour

p2p/net/swarm/dial_sync.go
sukunrt: err is declared in the beginning so this will assign to it. 

Anyway, I will remove the defer. That is confusing.

p2p/net/swarm/swarm_metrics.go
sukunrt: Makes sense. That would be cleaner.

p2p/net/swarm/swarm_metrics.go
sukunrt: changed it to `application canceled`

p2p/net/swarm/swarm_metrics.go
sukunrt: I removed `errParentContextCanceled` and added `errConcurrentDialSuccessful`.

Now if the cause is `context.Canceled` it necessarily means the application cancelled the context and if the cause is `errConcurrentDialSuccessful` it means a concurrent dial succeeded.

p2p/net/swarm/dial_sync.go
sukunrt: removed it.

p2p/net/swarm/dial_sync.go
```diff
@@ -96,14 +95,21 @@ func (ds *dialSync) Dial(ctx context.Context, p peer.ID) (*Conn, error) {
...
-			delete(ds.dials, p)
+	conn, err := ad.dial(ctx)
+
+	ds.mutex.Lock()
+	defer ds.mutex.Unlock()
+
+	ad.refCnt--
+	if ad.refCnt == 0 {
+		if err == nil {
+			ad.cancelCause(errConcurrentDialSuccessful)
```
marten-seemann, line 106: Why? What would happen if we're just dialing one address, and that dial is successful?

sukunrt: That dial would have succeeded so the context cancellation has no effect on that dial. That method would have returned. 

There are two options:
1. We set call `cancelCause(nil)` and live with the brittle check that `errors.Is(Cause(err), context.Canceled)` signals concurrent successful dial. Here we override actual cancelations of parent context with errParentCanceled. This is the strategy in 055cec0c7c1e8341b2bd5dc7b31ee7ae591d9ad4 commit. 
2. Is the latest commit. We explicitly signal `errConcurrentDialSuccessful` when err is nil. In case the dial succeeds, this is not a problem since we will not call `FailedDial` for tracking metrics in that case.

marten-seemann (approved):

PR #2541, opened 2023-08-30
transport integration tests: use DialContext instead of Dial to fail quickly

sukunrt:
using `transport.Dial` will never timeout. I think this is useful for tests.


marten-seemann:
This is not the transport integration test, this is the reuseport test.

sukunrt:
mybad, fixed it.

marten-seemann (approved):

PR #2542, opened 2023-08-30
swarm: wait for transient connections to upgrade for NewStream

### LINKED ISSUE #1603: Connect should wait for an acceptable connection

sukunrt:
depends on #2547, This PR is just the last commit. This might help to review it independently https://github.com/libp2p/go-libp2p/pull/2542/commits/b2377e9b625ffa328047b0298daa71c1f5793059

closes #1603

### DIFF
--- a/p2p/net/swarm/swarm.go
+++ b/p2p/net/swarm/swarm.go
@@ -398,17 +398,17 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
 	s.conns.Unlock()
 
 	// Notify goroutines waiting for a direct connection
-
-	// Go routines interested in waiting for direct connection first acquire this lock and then
-	// acquire conns.RLock. Do not acquire this lock before conns.Unlock to prevent deadlock.
-	s.directConnNotifs.Lock()
 	if !c.Stat().Transient {
+		// Go routines interested in waiting for direct connection first acquire this lock
+		// and then acquire s.conns.RLock. Do not acquire this lock before conns.Unlock to
+		// prevent deadlock.
+		s.directConnNotifs.Lock()
 		for _, ch := range s.directConnNotifs.m[p] {
 			close(ch)
 		}
 		delete(s.directConnNotifs.m, p)
+		s.directConnNotifs.Unlock()
 	}
-	s.directConnNotifs.Unlock()
 
 	// Emit event after releasing `s.conns` lock so that a consumer can still
 	// use swarm methods that need the `s.conns` lock.
@@ -466,15 +466,15 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
 	//
 	// TODO: Try all connections even if we get an error opening a stream on
 	// a non-closed connection.
-	dialed := false
-	for i := 0; i < 1; i++ {
+	numDials := 0
+	for {
 		c := s.bestConnToPeer(p)
 		if c == nil {
 			if nodial, _ := network.GetNoDial(ctx); !nodial {
-				if dialed {
+				numDials++
+				if numDials > DialAttempts {
 					return nil, errors.New("max dial attempts exceeded")
 				}
-				dialed = true
 				var err error
 				c, err = s.dialPeer(ctx, p)
 				if err != nil {
@@ -503,7 +503,6 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
 		}
 		return str, nil
 	}
-	return nil, network.ErrNoConn
 }
 
 // waitForDirectConn waits for a direct connection established through hole punching or connection reversal.
@@ -524,14 +523,17 @@ func (s *Swarm) waitForDirectConn(ctx context.Context, p peer.ID) (*Conn, error)
 	s.directConnNotifs.m[p] = append(s.directConnNotifs.m[p], ch)
 	s.directConnNotifs.Unlock()
 
-	// Wait for notification.
-	// There's no point waiting for more than a minute here.
-	ctx, cancel := context.WithTimeout(ctx, time.Minute)
+	// apply the DialPeer timeout
+	ctx, cancel := context.WithTimeout(ctx, network.GetDialPeerTimeout(ctx))
 	defer cancel()
+
+	// Wait for notification.
 	select {
 	case <-ctx.Done():
 		// Remove ourselves from the notification list
 		s.directConnNotifs.Lock()
+		defer s.directConnNotifs.Unlock()
+
 		s.directConnNotifs.m[p] = slices.DeleteFunc(
 			s.directConnNotifs.m[p],
 			func(c chan struct{}) bool { return c == ch },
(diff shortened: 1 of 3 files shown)


p2p/net/swarm/swarm.go
```diff
@@ -390,6 +397,19 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
 	c.notifyLk.Lock()
 	s.conns.Unlock()
 
+	// Notify goroutines waiting for a direct connection
+
+	// Go routines interested in waiting for direct connection first acquire this lock and then
+	// acquire conns.RLock. Do not acquire this lock before conns.Unlock to prevent deadlock.
+	s.directConnNotifs.Lock()
+	if !c.Stat().Transient {
```
marten-seemann, line 405: Any reason not to move the locking of the mutex into the if? That way we avoid locking altogether for non-transient conns.

p2p/net/swarm/swarm.go
```diff
@@ -435,46 +455,101 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
...
-		c := s.bestAcceptableConnToPeer(ctx, p)
-
+	dialed := false
+	for i := 0; i < 1; i++ {
+		c := s.bestConnToPeer(p)
 		if c == nil {
-			if nodial, _ := network.GetNoDial(ctx); nodial {
+			if nodial, _ := network.GetNoDial(ctx); !nodial {
+				if dialed {
+					return nil, errors.New("max dial attempts exceeded")
```
marten-seemann, line 474: I'm probably missing something here, but why don't we need to check against `DialAttempts` here?

p2p/net/swarm/swarm.go
```diff
@@ -435,46 +455,101 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
...
+func (s *Swarm) waitForDirectConn(ctx context.Context, p peer.ID) (*Conn, error) {
+	s.directConnNotifs.Lock()
+	c := s.bestConnToPeer(p)
+	if c == nil {
+		s.directConnNotifs.Unlock()
+		return nil, network.ErrNoConn
+	} else if !c.Stat().Transient {
+		s.directConnNotifs.Unlock()
+		return c, nil
+	}
```
marten-seemann, line 518: Why do we need to check again here?

p2p/net/swarm/swarm.go
```diff
@@ -435,46 +455,101 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
...
+
+	// Wait for transient connection to upgrade to a direct connection either by
+	// connection reversal or hole punching.
+	ch := make(chan struct{})
+	s.directConnNotifs.m[p] = append(s.directConnNotifs.m[p], ch)
+	s.directConnNotifs.Unlock()
+
+	// Wait for notification.
+	// There's no point waiting for more than a minute here.
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
```
marten-seemann, line 528: This is a magical number. Why is this not covered by our normal dial timeout?

p2p/net/swarm/swarm.go
```diff
@@ -435,46 +455,101 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
...
+		}
+		s.directConnNotifs.Unlock()
+		return nil, ctx.Err()
+	case <-ch:
+		// We do not need to remove ourselves from the list here as the notifier
+		// clears the map
+		c := s.bestConnToPeer(p)
+		if c == nil {
+			return nil, network.ErrNoConn
+		} else if c.Stat().Transient {
```
marten-seemann, line 549: Nit: don't need an else here.

p2p/net/swarm/swarm.go
```diff
@@ -435,46 +455,101 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
...
+	s.directConnNotifs.Unlock()
+
+	// Wait for notification.
+	// There's no point waiting for more than a minute here.
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
+	defer cancel()
+	select {
+	case <-ctx.Done():
+		// Remove ourselves from the notification list
+		s.directConnNotifs.Lock()
```
marten-seemann, line 533: Nit: defer unlocking.

p2p/net/swarm/swarm.go
sukunrt: To ensure that a direct connection wasn't established in the interval this function started executing. If a direct connection was established in the interval, this goroutine will keep waiting for a notification.

p2p/net/swarm/swarm.go
sukunrt: I made use of the fact that `DialAttempts` was set to 1. But using `DialAttempts` is better, and there was a bug in the previous version. I've modified this to use `DialAttempts` and a counter `numDials`.

p2p/net/swarm/swarm.go
sukunrt: Fixed this. Moved the mutex lock in to the if statement. We will not lock for transient connections. We need to lock for direct connections since we need to notify waiters.

p2p/net/swarm/swarm.go
sukunrt: fixed this. Using network.GetDialPeerTimeout

p2p/test/basichost/basic_host_test.go
```diff
@@ -62,11 +65,81 @@ func TestNoStreamOverTransientConnection(t *testing.T) {
...
+	defer cancel()
+	s, err := h1.NewStream(ctx, h2.ID(), "/testprotocol")
+	require.ErrorIs(t, err, context.DeadlineExceeded)
+	require.Nil(t, s)
+
+	// NewStream should return a stream if a direct connection is established
+	// while waiting
+	ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	time.AfterFunc(time.Second, func() {
```
marten-seemann, line 133: Any way we can test this without sleeping for a second?

p2p/test/swarm/swarm_test.go
```diff
@@ -68,3 +70,64 @@ func TestDialPeerTransientConnection(t *testing.T) {
...
+	defer cancel()
+	s, err := h1.Network().NewStream(ctx, h2.ID())
+	require.ErrorIs(t, err, context.DeadlineExceeded)
+	require.Nil(t, s)
+
+	// NewStream should return a stream if a direct connection is established
+	// while waiting
+	ctx, cancel = context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	time.AfterFunc(time.Second, func() {
```
marten-seemann, line 121: Here too.

marten-seemann (approved):
Nice!

Would be nice if there was a way to test this without sleeping for a second. Other than that, this lgtm. Feel free to merge any time.

PR #2544, opened 2023-08-31
chore: update examples to v0.31

p-shahi:


examples/go.mod
```diff
@@ -4,13 +4,13 @@ go 1.20
...
+	github.com/google/uuid v1.3.1
 	github.com/ipfs/go-datastore v0.6.0
 	github.com/ipfs/go-log/v2 v2.5.1
-	github.com/libp2p/go-libp2p v0.30.0
-	github.com/libp2p/go-libp2p-kad-dht v0.20.1-0.20230209220319-6c2045abad23
+	github.com/libp2p/go-libp2p v0.31.0
+	github.com/libp2p/go-libp2p-kad-dht v0.25.0
 	github.com/multiformats/go-multiaddr v0.11.0
-	github.com/prometheus/client_golang v1.14.0
+	github.com/prometheus/client_golang v1.16.0
```
sukunrt, line 13: Can we not upgrade this prometheus dependency? We can do it when we upgrade the main repo dependency. I don't see this harming us, but it's better to keep both of them in sync.

sukunrt (approved):

PR #2547, opened 2023-08-31
swarm: fix DialPeer behaviour for transient connections

### LINKED ISSUE #1603: Connect should wait for an acceptable connection

sukunrt:
This fixes a bug where the first call to `swarm.DialPeer` succeeds and returns a transient connection with no error while the second call to `swarm.DialPeer` returns `(nil, network.ErrTransientConn)`.

For dialing, we now only rely on `network.WithForceDirectDial` to force a direct connection.
For new stream, we open a stream on a transient connection only if `network.WithUseTransient` is used.

This feels correct to me since once we fix #1603 we can have `NewStream` called without `network.WithUseTransient` or `network.WithForceDirectDial` and it'll wait for the transient connection to be upgraded to a direct connection before returning.  

I'm not fixing it the other way round, i.e. Failing if `network.WithUseTransient` is not used, because there are users of this API who are probably relying on the first call to succeed. https://github.com/search?q=Network%28%29.DialPeer&type=code

### DIFF
--- a/p2p/net/swarm/swarm.go
+++ b/p2p/net/swarm/swarm.go
@@ -428,7 +428,8 @@ func (s *Swarm) StreamHandler() network.StreamHandler {
 }
 
 // NewStream creates a new stream on any available connection to peer, dialing
-// if necessary.
+// if necessary. Use network.WithUseTransient to open a stream over a transient(relayed)
+// connection.
 func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error) {
 	log.Debugf("[%s] opening stream to peer [%s]", s.local, p)
 
@@ -447,10 +448,7 @@ func (s *Swarm) NewStream(ctx context.Context, p peer.ID) (network.Stream, error
 	dials := 0
 	for {
 		// will prefer direct connections over relayed connections for opening streams
-		c, err := s.bestAcceptableConnToPeer(ctx, p)
-		if err != nil {
-			return nil, err
-		}
+		c := s.bestAcceptableConnToPeer(ctx, p)
 
 		if c == nil {
 			if nodial, _ := network.GetNoDial(ctx); nodial {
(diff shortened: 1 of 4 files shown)


p2p/test/swarm/swarm_test.go
```diff
@@ -0,0 +1,77 @@
...
+	h1.Peerstore().AddAddr(h2.ID(), relayaddr, peerstore.TempAddrTTL)
+
+	// swarm.DialPeer should connect over transient connections
+	conn1, err := h1.Network().DialPeer(context.Background(), h2.ID())
+	require.NoError(t, err)
+	require.NotNil(t, conn1)
+
+	conn2, err := h1.Network().DialPeer(context.Background(), h2.ID())
+	require.NoError(t, err)
+	require.NotNil(t, conn2)
```
sukunrt, line 68: On current master, the first call succeeds while the second one fails.

p2p/net/swarm/swarm.go
```diff
@@ -548,26 +546,17 @@ func (s *Swarm) bestConnToPeer(p peer.ID) *Conn {
 	return best
 }
 
-// - Returns the best "acceptable" connection, if available.
-// - Returns nothing if no such connection exists, but if we should try dialing anyways.
-// - Returns an error if no such connection exists, but we should not try dialing.
-func (s *Swarm) bestAcceptableConnToPeer(ctx context.Context, p peer.ID) (*Conn, error) {
+// bestAcceptableConnToPeer returns the best acceptable connection in the ctx passed. If
+// network.WithForceDirectDial is used, it'll only returns a direct connection ignoring
+// any transient(relayed) connections to the peer.
```
marten-seemann, line 551:
```suggestion
// bestAcceptableConnToPeer returns the best acceptable connection, considering the passed in ctx.
// If network.WithForceDirectDial is used, it only returns a direct connections, ignoring
// any transient (relayed) connections to the peer.
```

p2p/test/swarm/swarm_test.go
```diff
@@ -0,0 +1,69 @@
...
+	relayaddr := ma.StringCast("/p2p/" + relay1info.ID.String() + "/p2p-circuit/p2p/" + h2.ID().String())
+
+	h1.Peerstore().AddAddr(h2.ID(), relayaddr, peerstore.TempAddrTTL)
+
+	// swarm.DialPeer should connect over transient connections
+	conn1, err := h1.Network().DialPeer(context.Background(), h2.ID())
+	require.NoError(t, err)
+	require.NotNil(t, conn1)
+
+	conn2, err := h1.Network().DialPeer(context.Background(), h2.ID())
```
marten-seemann, line 58:
```suggestion
	// Test that repeated calls return the same connection.
	conn2, err := h1.Network().DialPeer(context.Background(), h2.ID())
```

p2p/net/swarm/swarm.go
```diff
@@ -428,7 +428,8 @@ func (s *Swarm) StreamHandler() network.StreamHandler {
 }
 
 // NewStream creates a new stream on any available connection to peer, dialing
-// if necessary.
+// if necessary. Use network.WithUseTransient to open a stream over a transient(relayed)
+// connection.
```
marten-seemann, line 432:
```suggestion
// if necessary.
// Use network.WithUseTransient to open a stream over a transient (relayed) connection.
```

marten-seemann (approved):

p2p/net/swarm/swarm.go
sukunrt: thanks. This reads much better.

marten-seemann (approved):

PR #2549, opened 2023-08-31
fix typos

howjmay:


sukunrt (commented):
Thanks for your PR @howjmay. Two things:
1. Can we separate the variable renaming into a different PR from the typo fixes for comments? 
2. Let's not update examples/multipro/pb/p2p.pb.go. We should fix the typo in the proto file and use a `go:generate` directive to generate the protobuf file.

marten-seemann:
Closing due to inactivity.

PR #2551, opened 2023-09-03
basichost: don't wait for Identify

marten-seemann:
By not waiting for Identify to finish we save 1 RTT during connection establishment. Establishing a QUIC connection now only takes a single roundtrip.

Multistream is (again) giving us a hard time here, mostly due to https://github.com/multiformats/go-multistream/issues/20, and due to the lack of stream reset error codes. The situation is expected to improve once we introduce and roll out error codes (https://github.com/libp2p/specs/issues/479).

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -736,20 +736,8 @@ func (h *BasicHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
 // the connection once it has been opened.
 func (h *BasicHost) dialPeer(ctx context.Context, p peer.ID) error {
 	log.Debugf("host %s dialing %s", h.ID(), p)
-	c, err := h.Network().DialPeer(ctx, p)
-	if err != nil {
+	if _, err := h.Network().DialPeer(ctx, p); err != nil {
 		return fmt.Errorf("failed to dial: %w", err)
-	}
-
-	// TODO: Consider removing this? On one hand, it's nice because we can
-	// assume that things like the agent version are usually set when this
-	// returns. On the other hand, we don't _really_ need to wait for this.
-	//
-	// This is mostly here to preserve existing behavior.
-	select {
-	case <-h.ids.IdentifyWait(c):
-	case <-ctx.Done():
-		return fmt.Errorf("identify failed to complete: %w", ctx.Err())
 	}
 
 	log.Debugf("host %s finished dialing %s", h.ID(), p)
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/libp2p/go-libp2p/core/connmgr"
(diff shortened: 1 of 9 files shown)


marten-seemann:
I tested this manually by setting the RTT to 100ms (`sudo tc qdisc add dev lo root netem delay 50ms`) and instrumenting `TestPing` (in p2p/test/transport) to output the duration. It came down from 3 RTTs to 2 RTTs. Unfortunately, we don't have rigorous tests for this (https://github.com/libp2p/test-plans/issues/52).

Stebalien:
Error codes aren't relevant here, the issue is that multistream supports multiple rounds of negotiation and doesn't have a clear "success" message.

I'm not going to block this, but:

1. This is unsound and will always be unsound unless multistream is entirely replaced.
2. We need to do very careful testing of upstream applications as we've been assuming this behavior for quite a while. For example, in the past, successfully opening a stream to a peer on some protocol _generally_ meant that that peer supported said protocol. Now, opening a stream with exactly one protocol will always succeed instantly regardless of what that peer supports.

marten-seemann:
> 1\. This is unsound and will always be unsound unless multistream is entirely replaced.

Agreed.

> Error codes aren't relevant here, the issue is that multistream supports multiple rounds of negotiation and doesn't have a clear "success" message.

They would help us return the correct error here, without having to guess what a stream reset means, as this PR currently does. That logic feels very brittle. With error codes, we'd have a dedicated error code for "protocol not supported", so you could immediately tell a stream reset initiated by multistream and a stream reset initiated by the application protocol itself apart.

> 2\. We need to do very careful testing of upstream applications as we've been assuming this behavior for quite a while.

Any suggestion how to do that, other than running a Kubo / Lotus node for a while and seeing if anything breaks?

Stebalien:
> Any suggestion how to do that, other than running a Kubo / Lotus node for a while and seeing if anything breaks?

Unfortunately, that's your best bet. Run kubo, check the tests, make a PR to lotus with the change and I'm happy to run it on my node for a while.

sukunrt (commented):
I will review closely later but I think this case has not been handled: 

>so the logic would be: if we receive a stream reset while multistream is running
if Identify claimed that the application protocol is supported, we return the stream reset error
otherwise: we replace that error with the multistream ErrNotSupported

from: https://filecoinproject.slack.com/archives/C03K82MU486/p1693817331703339?thread_ts=1693737018.663209&cid=C03K82MU486

If identify has completed we need to return the Reset error and not Proto not supported.

marten-seemann:
Do we really need this complexity? I’d prefer just implementing the actual solution (error codes).

p2p/host/basic/basic_host.go
```diff
@@ -1050,14 +1047,26 @@ func (h *BasicHost) Close() error {
...
+		}
+	}
+	return n, err
 }
 
 func (s *streamWrapper) Write(b []byte) (int, error) {
-	return s.rw.Write(b)
+	n, err := s.rw.Write(b)
+	if s.calledRead.Load() && errors.Is(err, network.ErrReset) {
+		return n, msmux.ErrNotSupported[protocol.ID]{Protos: []protocol.ID{s.Protocol()}}
```
sukunrt, line 1067: I think this should be `!s.calledRead.Load()`

sukunrt: There is a race condition here, but I think it doesn't matter becuase I'm not sure if you can use streams in the way that is required to trigger this condition.

The race is:
goroutine1:
does successful read and is going to do CompareAndSwap

then:
goroutine2:
does write and receives StreamReset for some reason(can it?).

now goroutine2 does the if s.calledRead.Load() && errors.Is(err, network.ErrReset) before goroutine1 could do CompareAndSwap.

sukunrt (commented):
I think we will see the client side(the one using optimistic select) to be resetting the stream on protocols which exchange length prefixed messages. 
  The server side of multistream select will be waiting to receive the next protocol. The client will optimistically write `<length><msg>`. The server side reads this and replies with `<na>` provided the message is less than 1kb in size. Receiving this will cause the client to reset the stream.

p2p/host/basic/basic_host.go
```diff
@@ -1050,14 +1047,26 @@ func (h *BasicHost) Close() error {
 type streamWrapper struct {
 	network.Stream
 	rw io.ReadWriteCloser
+
+	calledRead atomic.Bool
```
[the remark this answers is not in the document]
sukunrt: Why do we need this `streamWrapper`? 
All it does is wrap CloseWrite, which can be concurrent with Read and Write.
@MarcoPolo @Stebalien

MarcoPolo (commented):
I think this PR is missing a couple things:

1. A test to assert the expected number of round trips.
2. I think it would be great if this new behavior was an option. Then we can default it off, and allow users to opt-in. In time we can discuss flipping the default, ideally after major users opt-in to it.

I do think this is a good addition and will improve certain use cases (DHT off the top of my head). So it would be good to keep advancing this.

Stebalien:
Oh, I see we have removed the "old" DHT protocol. Yeah, this could make a significant difference.

p2p/host/basic/basic_host.go
sukunrt: I now think this whole logic should be a part of `multistream.LazyClientConn`

MarcoPolo: yeah, I agree.

MarcoPolo:
I'm closing this issue since it's been close to a year without activity. This is useful a change, but it's tricky to implement and it needs to be rolled out carefully. It effectively changes the core semantics of go-libp2p.

A future version should of this change should:

- Find a user that wants this and would use it.
- Include round trip tests
- provide this as opt-in (per connection/newstream?)
- Document this and add examples.

Issue #2550, opened 2023-09-01
identify: dns addresses shouldn't be filtered out


richard-ramos:
I'm dialing a node I have at `/dns4/node-01.do-ams3.status.prod.statusim.net/tcp/30303/p2p/16Uiu2HAm6HZZr7aToTvEBPpiys4UxajCTU97zj5v7RNR2gbniy1D` and after Identify is executed, when I tried to retrieve the peer addresses from the peer store, the peer was not returning any address at all.

Is this expected behavior? Notice in line 1001 the filter function  manet.IsPublicAddr is being used which will only take into account IP addresses and not DNS4 addresses.

https://github.com/libp2p/go-libp2p/blob/9f14eb71580e358c485b700d576014015e86cd15/p2p/protocol/identify/id.go#L994-L1002

richard-ramos:
For more context on why this could be a problem:
I'm using `basicHost.Connect` to connect to that node with dns4 multiaddress. `Connect` will add the dns4 address to the peerstore with a `TempAddressTTL` as it can be seen [here](https://github.com/libp2p/go-libp2p/blob/9f14eb71580e358c485b700d576014015e86cd15/p2p/host/basic/basic_host.go#L722). However, later Identify protocol will [filter out all the addresses](https://github.com/libp2p/go-libp2p/blob/9f14eb71580e358c485b700d576014015e86cd15/p2p/protocol/identify/id.go#L772) returned by Identify, ignoring the `dns4` multiaddresses, and [expire all `TempAddressTTL`](https://github.com/libp2p/go-libp2p/blob/9f14eb71580e358c485b700d576014015e86cd15/p2p/protocol/identify/id.go#L775) effectively removing from the peer store the dns4 address used to dial the node

sukunrt:
Nice catch! This is a bug. We should keep the dns address if the peer provides us with one. 

The fix loop simple to me, check if it's an IP address before doing the filtering, do you want to submit a fix?

sukunrt:
Thanks @richard-ramos for reporting. I'd like to get this fix in the next patch release so I'm taking this up. If you're interested in contributing, see issues with [help-wanted label](https://github.com/libp2p/go-libp2p/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22help+wanted%22)

richard-ramos:
Thank you, @sukunrt! :)

PR #2553, opened 2023-09-03
identify: don't filter dns addresses based on remote addr type

sukunrt:
closes: #2550

### DIFF
--- a/p2p/protocol/identify/id_glass_test.go
+++ b/p2p/protocol/identify/id_glass_test.go
@@ -12,6 +12,7 @@ import (
 	recordPb "github.com/libp2p/go-libp2p/core/record/pb"
 	blhost "github.com/libp2p/go-libp2p/p2p/host/blank"
 	swarmt "github.com/libp2p/go-libp2p/p2p/net/swarm/testing"
+	ma "github.com/multiformats/go-multiaddr"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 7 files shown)


marten-seemann (approved):
Do we need to make the same change in Kademlia?

Would it make sense to move all of this logic to go-multiaddr? It seems like every protocol that is accepting address advertisements from peers needs exactly the same filtering logic. (If so, we can do that in a follow-up PR, this code looks correct).

marten-seemann:
> I think it makes sense to modify `multiaddr.IsPublicAddr` to return true for `/dns,/dns4/dns6` addresses, but only in cases where the name is not one of the special use dns names defined in [RFC 6761](https://datatracker.ietf.org/doc/html/rfc6761)

Makes sense.

p2p/protocol/identify/id_glass_test.go
```diff
@@ -173,3 +174,34 @@ func TestInvalidSignedPeerRecord(t *testing.T) {
 	require.True(t, ok)
 	require.Nil(t, cab.GetPeerRecord(h2.ID()))
 }
+
+func TestIncomingAddrFilter(t *testing.T) {
+	lhAddr := ma.StringCast("/ip4/127.0.0.1/udp/123/quic-v1")
+	privAddr := ma.StringCast("/ip4/192.168.1.101/tcp/123")
+	pubAddr := ma.StringCast("/ip6/2::1/udp/123/quic-v1")
+	dnsAddr := ma.StringCast("/dns/example.com/udp/123/quic-v1")
```
marten-seemann, line 182: Should we add a `/dns/localhost` address here?

marten-seemann (approved):

p2p/protocol/identify/id_glass_test.go
sukunrt: I'd prefer adding this when we start using the updated functions in [go-multiaddr](https://github.com/multiformats/go-multiaddr/pull/221) package.

marten-seemann: Here's the v0.12 release PR: https://github.com/multiformats/go-multiaddr/pull/223

marten-seemann: v0.12 was released: https://github.com/multiformats/go-multiaddr/releases/tag/v0.12.0

@sukunrt Can you update this PR?

sukunrt:
sorry about the delay. I was procrastinating this thinking autorelay tests would fail, but turns out those tests don't actually use /dns4/localhost anywhere.

marten-seemann (approved):

PR #2557, opened 2023-09-05
upgrader: drop support for multistream simultaneous open

### LINKED ISSUE #2330: Consider dropping multistream-select sim open
By default go-libp2p supports the multistream-select simultaneous open extension:

> In order to support direct connections through NATs with hole punching, we need to account for simultaneous open. In such cases, there is no single initiator and responder, but instead both peers act as initiators. This breaks protocol negotiation in multistream-select, which assumes a single initator.
>
> This draft proposes a simple extension to the multistream protocol negotiation in order to select a single initator when both peers are acting as such.

https://github.com/libp2p/specs/blob/master/connections/simopen.md

See also [`go-multistream/client.go`](https://github.com/multiformats/go-multistream/blob/5b2f48e2db22aa1db90bb5a21aa4e6702d4ee33c/client.go#L115-L117) as the entrypoint.

The simultaneous open extension is needed for uncoordinated simultaneous connection attempts on TCP. Note that coordinated simultaneous connection attempts in libp2p's hole punching does not require the extension. In that case the role (initiator, responder) is determined through the DCUtR protocol. See [DCUtR](https://github.com/libp2p/specs/blob/master/relay/DCUtR.md#the-protocol) specification.

There is a downside to supporting the multistream-select simultaneous open extension, namely when negotiating a single security protocol on a new normal TCP connection. "Normal" as in not a simultaneous connect, i.e. initiated only from one side.

Theoretically, when negotiating a single protocol only, one can use [multistream-select's optimistic protocol negotiation](https://github.com/multiformats/go-multistream/blob/5b2f48e2db22aa1db90bb5a21aa4e6702d4ee33c/client.go#L43-L47). More specifically the dialer can propose the single protocol and then directly proceed, without waiting for confirmation from the remote, saving one round-trip in the happy path. 

Though unfortunately, one can not combine the simultaneous open extension with optimistic protocol negotiation, as the former needs to wait for the response from the remote to determine whether to enter the _initiator selection phase_.

multistream-select with simultaneous open extension on "normal" TCP connection:

``` mermaid
sequenceDiagram
    A ->> B: /libp2p/simultaneous-connect
    Note left of A: A can not proceed optimistically and negotiate the security protocol.<br/>B might enter _initiator selection phase_.<br/>Need to wait for B's response.
    B ->> A: `/libp2p/simultaneous-connect` or `na`
```

Dropping multistream-select sim open would allow go-libp2p to compete with tcp+tls+http in connection-establishment latency **when supporting a single security protocol only**. More specifically it would allow go-libp2p to establish a TCP+TLS connection in two round trips. At the expense of failed connection establishment on uncoordinated TCP simultaneous open.

Worth noting:

- **Most go-libp2p deployments support two security protocols on top of TCP, namely Noise and TLS. The proposed optimization does not apply here, as the multistream-select optimistic protocol negotiation can only be used when negotiating a single protocol.**
- There are measurements on the frequency of uncoordinated simultaneous connect in the wild. See https://github.com/libp2p/test-plans/pull/163/#discussion_r1206303871.
- See https://github.com/libp2p/test-plans/pull/163/#discussion_r1203549587 for a recent discussion on this topic.
- See https://github.com/libp2p/specs/issues/389 suggesting to not always do TCP port reuse and thus prevent uncoordinated simultaneous connects in the first place.
- One could argue that go-libp2p's QUIC is the better TCP+XXX in the first place. Thus there is not much value in investing time here.

marten-seemann:
Fixes #2330. For https://github.com/libp2p/specs/issues/573.

Turns out there wasn't even a test for this...

We can also merge https://github.com/multiformats/go-multistream/pull/107 first, cut a new release, and then update go-multistream in this PR.

### DIFF
--- a/p2p/net/upgrader/upgrader.go
+++ b/p2p/net/upgrader/upgrader.go
@@ -152,7 +152,8 @@ func (u *upgrader) upgrade(ctx context.Context, t transport.Transport, maconn ma
 		return nil, ipnet.ErrNotInPrivateNetwork
 	}
 
-	sconn, security, server, err := u.setupSecurity(ctx, conn, p, dir)
+	isServer := dir == network.DirInbound
+	sconn, security, err := u.setupSecurity(ctx, conn, p, isServer)
 	if err != nil {
 		conn.Close()
 		return nil, fmt.Errorf("failed to negotiate security protocol: %w", err)
@@ -179,7 +180,7 @@ func (u *upgrader) upgrade(ctx context.Context, t transport.Transport, maconn ma
 		}
 	}
 
-	muxer, smconn, err := u.setupMuxer(ctx, sconn, server, connScope.PeerScope())
+	muxer, smconn, err := u.setupMuxer(ctx, sconn, isServer, connScope.PeerScope())
 	if err != nil {
 		sconn.Close()
 		return nil, fmt.Errorf("failed to negotiate stream multiplexer: %w", err)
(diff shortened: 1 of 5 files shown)


p2p/net/upgrader/upgrader.go
```diff
@@ -201,9 +201,7 @@ func (u *upgrader) upgrade(ctx context.Context, t transport.Transport, maconn ma
 
 func (u *upgrader) setupSecurity(ctx context.Context, conn net.Conn, p peer.ID, dir network.Direction) (sec.SecureConn, protocol.ID, bool, error) {
```
sukunrt, line 202: You don't need the third return, bool value right? 

AFAICT it is used to indicate whether we are the server which the caller now has complete context of.

marten-seemann: Good catch. This allowed a small refactor to simplify things a tiny bit.

sukunrt (approved):

marten-seemann:
Updated the go-multistream dependency to v0.5.0.

PR #2560, opened 2023-09-07
swarm: don't dial unspecified addresses

sukunrt:


marten-seemann (approved):

marten-seemann:
Should we have a test case for this?

sukunrt:
added a test.

marten-seemann (approved):

PR #2561, opened 2023-09-07
swarm: use shared black hole filters for autonat

sukunrt:
I'm sharing the black hole filters and added a ReadOnly mode to black hole detector where the detector doesn't update the state of the individual filters. This ReadOnly mode is used in the AutoNAT(both v1 and v2) swarms.  This looks like the nicest option. 

- No metrics tracking from autonat swarms.
- No updates from autonat swarm.
- Rejecting autonat requests when our black hole state is unknown.


sukunrt:
Merging this in to the main PR #2422

PR #2564, opened 2023-09-10
swarm: fix recursive resolving of DNS multiaddrs

marten-seemann:

PR #2571, opened 2023-09-14
websocket: don't resolve /dnsaddr addresses

marten-seemann:
/dnsaddr allows forwarding to arbitrary multiaddresses. We can't assume that the host name stays the same.

Thanks to @sukunrt for pointing this out in https://github.com/libp2p/go-libp2p/pull/2564#pullrequestreview-1621750665.


sukunrt (approved):
Thank you!

PR #2573, opened 2023-09-15
swarm: use happy eyeballs ranking for TCP dials

sukunrt:
The required change is much simpler than my previous tries. Closing other open PRs in favour of this.

### DIFF
--- a/core/transport/transport.go
+++ b/core/transport/transport.go
@@ -126,6 +126,12 @@ type Upgrader interface {
 	Upgrade(ctx context.Context, t Transport, maconn manet.Conn, dir network.Direction, p peer.ID, scope network.ConnManagementScope) (CapableConn, error)
 }
 
+// DialUpdater provides updates on in progress dials.
+type DialUpdater interface {
+	// DialWithUpdates dials a remote peer and provides updates on the passed channel.
+	DialWithUpdates(context.Context, ma.Multiaddr, peer.ID, chan<- DialUpdate) (CapableConn, error)
+}
+
 // DialUpdateKind indicates the type of DialUpdate event.
 type DialUpdateKind int
 
@@ -134,6 +140,8 @@ const (
 	DialFailed DialUpdateKind = iota
 	// DialSuccessful indicates dial succeeded.
 	DialSuccessful
+	// TCPConnectionEstablished indicates successful completion of the TCP 3-way handshake
+	TCPConnectionEstablished
 )
 
 func (k DialUpdateKind) String() string {
@@ -142,6 +150,8 @@ func (k DialUpdateKind) String() string {
 		return "DialFailed"
 	case DialSuccessful:
 		return "DialSuccessful"
+	case TCPConnectionEstablished:
+		return "TCPConnectionEstablished"
 	default:
 		return fmt.Sprintf("DialUpdateKind<Unknown-%d>", k)
 	}
--- a/core/transport/transport.go
+++ b/core/transport/transport.go
@@ -5,6 +5,7 @@ package transport
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net"
 
 	"github.com/libp2p/go-libp2p/core/network"
@@ -123,4 +124,37 @@ type Upgrader interface {
 	UpgradeListener(Transport, manet.Listener) Listener
 	// Upgrade upgrades the multiaddr/net connection into a full libp2p-transport connection.
 	Upgrade(ctx context.Context, t Transport, maconn manet.Conn, dir network.Direction, p peer.ID, scope network.ConnManagementScope) (CapableConn, error)
+}
+
+// DialUpdateKind indicates the type of DialUpdate event.
+type DialUpdateKind int
+
+const (
+	// DialFailed indicates dial failed.
+	DialFailed DialUpdateKind = iota
+	// DialSuccessful indicates dial succeeded.
+	DialSuccessful
+)
+
+func (k DialUpdateKind) String() string {
+	switch k {
+	case DialFailed:
+		return "DialFailed"
+	case DialSuccessful:
+		return "DialSuccessful"
+	default:
+		return fmt.Sprintf("DialUpdateKind<Unknown-%d>", k)
+	}
+}
+
+// DialUpdate is used by DialUpdater to provide dial updates.
+type DialUpdate struct {
+	// Kind is the kind of update event.
+	Kind DialUpdateKind
+	// Addr is the peer's address.
+	Addr ma.Multiaddr
+	// Conn is the resulting connection on success.
+	Conn CapableConn
+	// Err is the reason for dial failure.
+	Err error
 }
--- a/core/transport/transport.go
+++ b/core/transport/transport.go
@@ -136,21 +136,22 @@ type DialUpdater interface {
 type DialUpdateKind int
 
 const (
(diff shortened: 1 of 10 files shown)


p2p/net/swarm/dial_ranker.go
```diff
@@ -103,7 +104,29 @@ func getAddrDelay(addrs []ma.Multiaddr, tcpDelay time.Duration, quicDelay time.D
...
+		idx := 0
+		for i, a := range addrs {
+			if isProtocolAddr(a, ma.P_TCP) {
+				idx = i
+				break
+			}
+		}
+		if isProtocolAddr(addrs[idx], ma.P_TCP) && isProtocolAddr(addrs[idx], ma.P_IP6) {
+			for i := idx + 1; i < len(addrs); i++ {
+				if isProtocolAddr(addrs[i], ma.P_TCP) && isProtocolAddr(addrs[i], ma.P_IP4) {
```
marten-seemann, line 122: This could be done in a single loop over addrs. It would also be helpful to have a comment here to explain what this logic is doing.

p2p/net/swarm/dial_ranker.go
```diff
@@ -112,24 +135,32 @@ func getAddrDelay(addrs []ma.Multiaddr, tcpDelay time.Duration, quicDelay time.D
...
 	res := make([]network.AddrDelay, 0, len(addrs))
 
+	var tcpIdx int
 	var totalTCPDelay time.Duration
 	for i, addr := range addrs {
 		var delay time.Duration
 		switch {
 		case isQUICAddr(addr):
-			// For QUIC addresses we dial an IPv6 address, then after quicDelay an IPv4
-			// address, then after quicDelay we dial rest of the addresses.
```
marten-seemann, line 121: Why was this comment deleted? Does it now apply to both QUIC and TCP? Would be worth keeping it.

marten-seemann (commented):
Do we also need to use this for WebTransport? WebTransport also has a two-step handshake: The QUIC handshake finishes first, then we do the WebTransport session setup.

p2p/net/swarm/dial_worker.go
```diff
@@ -62,6 +64,8 @@ type addrDial struct {
 	createdAt time.Time
 	// dialRankingDelay is the delay in dialing this address introduced by the ranking logic
 	dialRankingDelay time.Duration
+	// expectedTCPUpgTime is the expected time by which security upgrade will complete
+	expectedTCPUpgTime time.Time
```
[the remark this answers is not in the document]
sukunrt: This is the field that provides how much to wait on for next dials. Used here for ex: https://github.com/libp2p/go-libp2p/blob/sukun/happy-eyeballs-tcp-4/p2p/net/swarm/dial_worker.go#L328-L340

p2p/net/swarm/dial_ranker.go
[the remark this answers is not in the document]
sukunrt: Yes. Updated.

core/transport/transport.go
```diff
@@ -124,3 +125,47 @@ type Upgrader interface {
...
+type DialUpdateKind int
+
+const (
+	// UpdateKindDialFailed indicates dial failed.
+	UpdateKindDialFailed DialUpdateKind = iota
+	// UpdateKindDialSuccessful indicates dial succeeded.
+	UpdateKindDialSuccessful
+	// UpdateKindTCPConnectionEstablished indicates successful completion of the TCP 3-way
+	// handshake
+	UpdateKindTCPConnectionEstablished
```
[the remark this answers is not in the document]
sukunrt: fixed.

p2p/net/swarm/dial_worker.go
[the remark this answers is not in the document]
sukunrt: I have renamed the field.

p2p/transport/tcp/tcp.go
```diff
@@ -176,21 +177,25 @@ func (t *TcpTransport) maDial(ctx context.Context, raddr ma.Multiaddr) (manet.Co
...
+	c, err := t.dialWithScope(ctx, raddr, p, connScope, updCh)
 	if err != nil {
 		connScope.Done()
 		return nil, err
 	}
 	return c, nil
 }
 
-func (t *TcpTransport) dialWithScope(ctx context.Context, raddr ma.Multiaddr, p peer.ID, connScope network.ConnManagementScope) (transport.CapableConn, error) {
+func (t *TcpTransport) dialWithScope(ctx context.Context, raddr ma.Multiaddr, p peer.ID, connScope network.ConnManagementScope, updCh chan<- transport.DialUpdate) (transport.CapableConn, error) {
```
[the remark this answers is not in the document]
sukunrt: agreed.

p2p/net/swarm/dial_ranker.go
sukunrt: I've added comments and made a single loop. Are the comments helpful?

p2p/net/swarm/dial_ranker.go
sukunrt: You're right. I don't remember why I chose to remove this instead of adding a similar comment to the TCP block.

sukunrt:
@marten-seemann Addressed review comments. I'll take webtransport up in a later PR.

p2p/net/swarm/dial_ranker.go
```diff
@@ -88,48 +99,100 @@ func DefaultDialRanker(addrs []ma.Multiaddr) []network.AddrDelay {
...
+	if len(addrs) == 0 {
+		return nil
+	}
 
 	sort.Slice(addrs, func(i, j int) bool { return score(addrs[i]) < score(addrs[j]) })
 
-	// If the first address is (QUIC, IPv6), make the second address (QUIC, IPv4).
-	happyEyeballs := false
-	if len(addrs) > 0 {
+	// addrs is now sorted by (Transport, IPVersion). Reorder addrs for happy eyeballs dialing.
```
marten-seemann, line 108: Should this sorting logic be part of `score`?

p2p/net/swarm/dial_ranker.go
```diff
@@ -88,48 +99,100 @@ func DefaultDialRanker(addrs []ma.Multiaddr) []network.AddrDelay {
...
-	var totalTCPDelay time.Duration
+	var tcpFirstDialDelay time.Duration
 	for i, addr := range addrs {
 		var delay time.Duration
 		switch {
 		case isQUICAddr(addr):
-			// For QUIC addresses we dial an IPv6 address, then after quicDelay an IPv4
-			// address, then after quicDelay we dial rest of the addresses.
+			// We dial an IPv6 address, then after quicDelay an IPv4
+			// address, then after a further quicDelay we dial rest of the addresses.
```
marten-seemann, line 166:
```suggestion
			// address, then after a further quicDelay we dial the rest of the addresses.
```

p2p/net/swarm/dial_ranker.go
sukunrt: I'm not sure if that'd be more readable. For this ranking we need to look at other addresses, so an IPv4 address gets higher score than other IPv6 addresses only if there are some IPv6 addresses present. I think the present implementation makes the logic explicit.

marten-seemann (approved):

PR #2576, opened 2023-09-20
webrtcprivate: add transport

sukunrt:


marten-seemann:
#2586 was merged. @sukunrt Can you rebase please?

marten-seemann (commented):
Review of the non-transport parts of this PR.

config/config.go
```diff
@@ -65,6 +67,8 @@ type Security struct {
 	Constructor interface{}
 }
 
+type ICEServer = webrtc.ICEServer
```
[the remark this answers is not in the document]
sukunrt: Nope. Just so that we redefine it a different way later if we want. But I think it's fine to remove this and just take the pion specific object. We can make a breaking change if this needs to be changed.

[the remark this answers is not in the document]
marten-seemann: It’s a breaking change anyway if we change the underlying type. That should be fine.

p2p/net/swarm/swarm_transport.go
```diff
@@ -27,7 +27,10 @@ func (s *Swarm) TransportForDialing(a ma.Multiaddr) transport.Transport {
 		}
 		return nil
 	}
-	if isRelayAddr(a) {
+	if isProtocolAddr(a, ma.P_WEBRTC) {
+		return s.transports.m[ma.P_WEBRTC]
```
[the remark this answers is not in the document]
sukunrt: The peer address is a circuitv2 address. So we need to check it before the circuitv2 check but the check is the same as the circuit v2 check. 

The address is of the form
`/ip4/1.2.3.4/udp/1/quic-v1/p2p/<relay-id>/p2p-circuit/webrtc`
So we need to ensure that we don't use the quic-v1 or the p2p-circuit transport for dialing.

p2p/protocol/holepunch/svc.go
```diff
@@ -283,3 +285,39 @@ func (s *Service) DirectConnect(p peer.ID) error {
...
+			defer s.refCount.Done()
+			select {
+			// waiting for Identify here will allow us to access the peer's public and observed addresses
+			// that we can dial to for a hole punch.
+			case <-s.ids.IdentifyWait(conn):
+			case <-s.ctx.Done():
+				return
+			}
+			p := conn.RemotePeer()
+			// Peer supports DCUtR, let it trigger holepunch
```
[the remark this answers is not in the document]
sukunrt: That case is handled here: https://github.com/libp2p/go-libp2p/blob/webrtcprivate/transport/p2p/protocol/holepunch/holepuncher.go#L142

In this case the peer(DCUtR initiator) is best suited to handle the call so I've kept it on that side.

p2p/protocol/holepunch/svc.go
```diff
@@ -283,3 +285,39 @@ func (s *Service) DirectConnect(p peer.ID) error {
...
+			if protos, err := s.host.Peerstore().SupportsProtocols(p, Protocol); err == nil && len(protos) > 0 {
+				return
+			}
+			// No DCUtR support, connect with peer over /webrtc
+			for _, addr := range s.host.Peerstore().Addrs(p) {
+				if _, err := addr.ValueForProtocol(ma.P_WEBRTC); err == nil {
+					ctx := network.WithForceDirectDial(s.ctx, "webrtc holepunch")
+					s.host.Connect(ctx, peer.AddrInfo{ID: p}) // address is already in peerstore
```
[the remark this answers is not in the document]
sukunrt: It returns immediately after this line, so this should be fine, no?

p2p/host/autorelay/relay_finder.go
```diff
@@ -737,9 +738,15 @@ func (rf *relayFinder) relayAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
 		addrs := cleanupAddressSet(rf.host.Peerstore().Addrs(p))
 		relayAddrCnt += len(addrs)
 		circuit := ma.StringCast(fmt.Sprintf("/p2p/%s/p2p-circuit", p))
+		webrtc := libp2pwebrtcprivate.WebRTCAddr
 		for _, addr := range addrs {
 			pub := addr.Encapsulate(circuit)
 			raddrs = append(raddrs, pub)
+			if rf.conf.webRTCSupport && isBrowserDialableAddr(addr) {
```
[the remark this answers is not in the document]
sukunrt: Yeah. This was a poor choice. I've moved this to host.Addrs. Even this is not great but I'm not sure what else to do without an address pipeline. The flow now is:

```
BasicHost.AllAddrs => RelayFinder.Addrs => BasicHost.AddrFactory => BasicHost.Addrs
```

p2p/host/autorelay/relay_finder.go
```diff
@@ -808,3 +815,14 @@ func (rf *relayFinder) resetMetrics() {
 	rf.metricsTracer.RelayAddressCount(0)
 	rf.metricsTracer.ScheduledWorkUpdated(&scheduledWorkTimes{})
 }
+
+var browserProtocols []int = []int{ma.P_WEBTRANSPORT, ma.P_WEBRTC_DIRECT, ma.P_WS, ma.P_WSS}
```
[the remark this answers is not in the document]
sukunrt: I've removed ma.P_WS. That shouldn't be here.

config/config.go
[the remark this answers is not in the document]
marten-seemann: Can you remove it?

p2p/transport/webrtcprivate/listener.go
```diff
@@ -0,0 +1,319 @@
...
+	// de-register candidate callback
+	defer pc.OnICECandidate(func(_ *webrtc.ICECandidate) {})
+
+	// read an incoming offer
+	var msg pb.Message
+	if err := r.ReadMsg(&msg); err != nil {
+		err = fmt.Errorf("failed to read offer: %w", err)
+		return nil, err
+	}
+	if msg.Type == nil || *msg.Type != pb.Message_SDP_OFFER {
```
marten-seemann, line 206:
```suggestion
	if msg.GetType() != pb.Message_SDP_OFFER {
```

p2p/transport/webrtcprivate/listener.go
```diff
@@ -0,0 +1,319 @@
...
+	})
+
+	// de-register candidate callback
+	defer pc.OnICECandidate(func(_ *webrtc.ICECandidate) {})
+
+	// read an incoming offer
+	var msg pb.Message
+	if err := r.ReadMsg(&msg); err != nil {
+		err = fmt.Errorf("failed to read offer: %w", err)
+		return nil, err
```
marten-seemann, line 204: Nit: here and below
```suggestion
		return nil, fmt.Errorf("failed to read offer: %w", err)
```

p2p/transport/webrtcprivate/listener.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+	if msg.Type == nil || *msg.Type != pb.Message_SDP_OFFER {
+		err = fmt.Errorf("invalid message: msg.Type expected %s got %s", pb.Message_SDP_OFFER, msg.Type)
+		return nil, err
+	}
+	if msg.Data == nil || *msg.Data == "" {
```
marten-seemann, line 210:
```suggestion
	if len(msg.GetData()) == 0 {
```

p2p/transport/webrtcprivate/listener.go
```diff
@@ -0,0 +1,319 @@
...
+	readErr := make(chan error, 1)
+	// start a goroutine to read candidates
+	go func() {
+		for {
+			if ctx.Err() != nil {
+				return
+			}
+			err := r.ReadMsg(&msg)
+			if err == io.EOF {
+				// remote has done writing
```
marten-seemann, line 245:
```suggestion
				// remote has finished writing
```

p2p/transport/webrtcprivate/listener.go
```diff
@@ -0,0 +1,319 @@
...
+				return
+			}
+			if err != nil {
+				readErr <- fmt.Errorf("failed to read candidate: %w", err)
+				return
+			}
+
+			if msg.Type == nil || *msg.Type != pb.Message_ICE_CANDIDATE {
```
marten-seemann, line 253:
```suggestion
			if msg.GetType() != pb.Message_ICE_CANDIDATE {
```

p2p/transport/webrtcprivate/listener.go
```diff
@@ -0,0 +1,319 @@
...
+				readErr <- fmt.Errorf("invalid message: msg.Type expected %s got %s", pb.Message_ICE_CANDIDATE, msg.Type)
+				return
+			}
+			// Ignore without Debuging on empty message.
+			// Pion has a case where OnCandidate callback may be called with a nil
+			// candidate
+			if msg.Data == nil || *msg.Data == "" {
```
marten-seemann, line 260:
```suggestion
			if len(msg.GetData()) == 0 {
```

p2p/transport/webrtcprivate/listener.go
```diff
@@ -0,0 +1,319 @@
...
+				log.Debugf("received empty candidate from %s", s.Conn().RemotePeer())
+				continue
+			}
+
+			var init webrtc.ICECandidateInit
+			if err := json.Unmarshal([]byte(*msg.Data), &init); err != nil {
```
marten-seemann, line 266:
```suggestion
			if err := json.Unmarshal([]byte(msg.GetData()), &init); err != nil {
```

p2p/transport/webrtcprivate/transport.go
```diff
@@ -0,0 +1,501 @@
...
+	gater                  connmgr.ConnectionGater
+	maxInFlightConnections int
+
+	mu       sync.Mutex
+	listener *listener
+}
+
+var _ tpt.Transport = &transport{}
+
+func AddTransport(h host.Host, gater connmgr.ConnectionGater, stunServers []webrtc.ICEServer) (*transport, error) {
```
marten-seemann, line 63: Exported function returning unexported type.

Why is this method needed? Why don't we have such a method for other transports?

p2p/transport/webrtc/listener.go
```diff
@@ -25,14 +25,14 @@ import (
 	"go.uber.org/zap/zapcore"
 )
 
-type connMultiaddrs struct {
-	local, remote ma.Multiaddr
+type ConnMultiaddrs struct {
```
marten-seemann, line 28: Nit: Don't make this public. Redefine if necessary.

marten-seemann (commented):
Should we move the shared parts to a separate package that's imported by both webrtc and webrtcprivate?

p2p/transport/webrtcprivate/transport.go
sukunrt: We have this method for circuit-v2 transport. We need this method for the same reason for webrtc private transport. 

The peer address is `<relay-addr>/p2p-circuit/webrtc`. This transport will first make a relayed connection with the peer, which is why it needs access to the host. After this point it establishes the connection over the signaling stream.

p2p/net/swarm/swarm_transport.go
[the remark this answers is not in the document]
sukunrt: Yes but we don't want the circuit-v2 transport to handle dialing to the webrtc address. 
The previous code was 
if isRelayAddr => Use circuitv2 transport. 

But now we need
if isWebRTCAddr => Use webrtcprivate transport.
else if isRelayAddr => Use circuitv2 transport. 
This is because the webrtcprivate address has a p2p-circuit component but we want to use webrtcprivate transport. I could have written this as:
```
if isRelayAddr(addr):
	if isWebRTCAddr(addr):
		use webrtcprivate transport
	else
		use circuitv2 transport
```

The flow I've implemented is: 
peerAddr: `<relay-addr>/p2p-circuit/webrtc`

1. webrtcprivate.Transport.Dial is called with peerAddr
2. webrtcprivate.Transport.Dial adds the relay address inferred from the webrtcprivate address inferred from the peerstore. Then makes a host.Connect call to establish a relay connection with the peer. 
3. Once relay connection is established it opens a signaling stream and establishes a webrtc connection.

I've chosen this because it is similar to circuitv2 transport which first establishes a direct connection with the relay node and then a circuitv2 connection to the peer over the relayed connection. This also has the nice property that Transport.Dial can connect to the peer over the address no matter the state of the host when Transport.Dial is called. In this case, it means if no circuit-v2 connection exists, it'll make a circuit-v2 connection first.

cpeliciari:
@sukunrt What do we need for webrtcprivate to work?

cpeliciari:
@sukunrt I would like to be able to create private meetings. Then I would share the video between the participants.

simsman2695:
@sukunrt currently webrtc-direct works only for public networks, my understanding now is that there is no means to peer browser's directly in private networks over the internet. Direct peering for web applications would reduce the overhead for relay requirements for private networks for pubsub. An example would be building a video meeting room with chat and file sharing.

MarcoPolo:

PR #2579, opened 2023-09-21
webrtc: fix flaky TestStreamWriteDeadlineAsync

sukunrt:


marten-seemann (changes requested):
I don't understand why this is necessary. We already have a 5x timeout on CI. Why is that not sufficient?

It seems like we have two ways here that are trying to achieve the same thing.

sukunrt:
I'll rethink the whole test. 

I'm not sure why we have the first extension of timeout in CI especially to 500ms. 

The test checks that when we set the deadline to `timeout` duration it does eventually give the `deadline exceeded` error. Then in the last check it checks that this error was provided after a sufficient duration. The last check defines sufficient currently as `1.5 * timeout`. I changed it to `3 * timeout` since it was flaking on windows as the test took more than 750 ms(500 * 1.5) and it generally took less than 900ms(as per my memory).
I am not convinced that the first extension of timeout makes sense. Why should we have to set deadline for 500ms when we are in CI?

p-shahi:
What's the status of this PR - are we targeting it for the release or is it a nice to have?

marten-seemann:
@sukun says this test hasn't failed in a while. Closing.

PR #2580, opened 2023-09-22
webrtc: fix deadlock on connection close

sukunrt:
also fixes a persistent failure in CI: https://github.com/libp2p/go-libp2p/actions/runs/6261705440/job/17007197102

### DIFF
--- a/p2p/transport/webrtc/stream.go
+++ b/p2p/transport/webrtc/stream.go
@@ -123,7 +123,23 @@ func newStream(
 			}
 		}
 
-		s.maybeDeclareStreamDone()
+		if s.isDone() {
+			// onDone removes the stream from the connection and requires the connection lock.
+			// This callback(onBufferedAmountLow) is executing in the sctp readLoop goroutine.
+			// If the connection is closed concurrently, the closing goroutine will acquire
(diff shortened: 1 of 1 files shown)


p2p/transport/webrtc/stream.go
```diff
@@ -200,6 +214,12 @@ func (s *stream) maybeDeclareStreamDone() {
 	}
 }
 
+func (s *stream) isDone() bool {
```
marten-seemann, line 217: Add a comment here that it requires locking?

p2p/transport/webrtc/stream.go
```diff
@@ -123,7 +123,23 @@ func newStream(
 			}
 		}
 
-		s.maybeDeclareStreamDone()
+		if s.isDone() {
+			// onDone removes the stream from the connection and requires the connection lock.
+			// This callback(onBufferedAmountLow) is executing in the sctp readLoop goroutine.
+			// If the connection is closed concurrently, the closing goroutine will acquire
```
marten-seemann, line 129:
```suggestion
			// If Connection.Close is called concurrently, the closing goroutine will acquire
```

marten-seemann (approved):

Issue #2536, opened 2023-08-29
WebRTC UDP Mux: more tests


MarcoPolo:
I'm surprised we only need two tests here. Should we add a test for:

- A muxed connection will be created *before* calling the unknown ufrag callback
- If unknown ufrag callback returns false, then we remove the conn from the muxer map

_Originally posted by @MarcoPolo in https://github.com/libp2p/go-libp2p/pull/2337#discussion_r1300773350_

PR #2586, opened 2023-09-26
webrtc: correctly report incoming packet address on muxed connection

sukunrt:
The address in ReadFrom should be the address of the received packet and not the address of the first packet received from the peer. If the interface used to send the first packet disallows incoming packets, this will lead to a connection not being established. 

The simplest way to test this is to use tailscale. Packets originated from tailscale address can be sent to localhost addresses, packets from localhost can't be sent to tailscale listeners. So if the first packet from the dialer is from the tailscale interface, the connection will never be established because muxedConnection.ReadFrom will always return the address of the packet as the tailscale address and we will ignore all reachable localhost packets.

My theory is that this is what is causing flakiness like the one observed here: https://github.com/libp2p/go-libp2p/actions/runs/6312484104/job/17138640142?pr=2584
The difference being the interface is a docker interface on the machine. 

The change is inline with the pion provided udp_mux implementation: https://github.com/pion/ice/blob/master/udp_muxed_conn.go#L140. They maintain a list of addresses on the connection, we keep this list on the udpmux. 

Also fixes: #2536

### DIFF
--- a/p2p/transport/webrtc/udpmux/mux.go
+++ b/p2p/transport/webrtc/udpmux/mux.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strings"
 	"sync"
 
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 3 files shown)


p2p/transport/webrtc/udpmux/mux.go
```diff
@@ -42,9 +43,15 @@ type UDPMux struct {
...
-	mx       sync.Mutex
+	mx sync.Mutex
+	// ufragMap allows us to multiplex incoming STUN packets based on ufrag
 	ufragMap map[ufragConnKey]*muxedConnection
-	addrMap  map[string]*muxedConnection
+	// addrMap allows us to correctly direct incoming packets after the connection
+	// is established and ufrag isn't available on all packets
+	addrMap map[string]*muxedConnection
+	// ufragAddrMap allows us to clean up all addresses from the addrMap once
+	// connection is closed
```
marten-seemann, line 53:
```suggestion
	// ufragAddrMap allows cleaning up all addresses from the addrMap once the connection is closed
	// During the ICE connectivity checks, the same ufrag might be used on multiple addresses.
```

marten-seemann (approved):

PR #2590, opened 2023-10-01
tcp: fix build on riscv64

marten-seemann:


Jorropo (changes requested):
Can you add comments linking to https://github.com/marten-seemann/tcp/pull/1 there is no reason this can't work on riscv, just we need theses constants defined.

p2p/transport/tcp/metrics.go
```diff
@@ -1,4 +1,4 @@
-//go:build !windows
+//go:build !windows && !riscv64
```
Jorropo, line 1:
```suggestion
// riscv64 see: https://github.com/marten-seemann/tcp/pull/1

//go:build !windows && !riscv64
```

p2p/transport/tcp/metrics_general.go
```diff
@@ -1,4 +1,4 @@
-//go:build !linux && !darwin && !windows
+//go:build !linux && !darwin && !windows && !riscv64
```
Jorropo, line 1:
```suggestion
// riscv64 see: https://github.com/marten-seemann/tcp/pull/1

//go:build !linux && !darwin && !windows && !riscv64
```

p2p/transport/tcp/metrics_none.go
```diff
@@ -1,4 +1,4 @@
-//go:build windows
+//go:build windows || riscv64
```
Jorropo, line 1:
```suggestion
// riscv64 see: https://github.com/marten-seemann/tcp/pull/1

//go:build windows || riscv64
```

Jorropo (approved):

p2p/transport/tcp/metrics.go
marten-seemann: Are you sure this works? Doesn't the `go:build` need to come first?

Jorropo: I don't think so, `go:build` line has to be come before `package`:
```console
> cat a.go 
// test

//go:build doNotBuild

package aaa

const "test" + 42
> cat b.go 
package aaa

func A() {}
> go build .
> echo $status
0
```

Jorropo:
Could this be included in `v0.31.1` ?

marten-seemann:
> Could this be included in `v0.31.1` ?

@Jorropo Can you elaborate on the sudden urgency please? The Kubo issue has been open for years.

Jorropo:
I'm now running Kubo on some riscv64 hardware and I'm lazy, so I would rather not have to repatch it every time but that is subjective.
We now have a patch that you are not in-comfortable merging due to the extra maintenance it would bring and it is free given it's stubbing statistics values to be 0.

Imo the best time to fix this would have been fixed in december 2021 or january 2022 when https://github.com/marten-seemann/tcp/pull/1 was sent but it didn't happend for reasons unknown to me and I don't have a time machine to do anything about it, so the next best time to do it is now.

marten-seemann:
This doesn't seem like a kind of patch that justifies cutting a patch release. This will be included in v0.32.

Jorropo:
I miss-understood there would be a v0.31.1 for the resource manager desync issue.

marten-seemann:
I added one of the comments for the build flags. I don't think we need to repeat in 3 files.

Issue #2592, opened 2023-10-03
webrtc: Figure out a way to discover webrtc-direct addresses


sukunrt:
Currently node doesn't dial out of the UDP packet conn that it is listening on for `/webrtc-direct`. Because of this it cannot discover its publicly dialable address. 

We find our listen address by dialing out using reuseport and then checking the observed address on identify. In webrtc-direct we never do a reuseport and so there's no way to find our listening address unless we manage to do port mapping on the router. So address discovery won't work on EC2 machines which don't have access to public IP from within the instance or UPnP like port mapping system. 

@mxinden how does rust-libp2p do this address discovery?

mxinden:
rust-libp2p's `/webrtc-direct` implementation reuses the listen port for outgoing dials.

Related: currently rust-libp2p's `/webrtc-direct` implementation has a bug, where one can not dial without a previous listen. https://github.com/libp2p/rust-libp2p/issues/4071

sukunrt:
Fixed by #2446

PR #2593, opened 2023-10-03
webrtc: reuse port when dialing

sukunrt:
depends on #2586 

fixes: #2592 

The fix turned out much simpler that I'd expected. Reusing the udp mux for dialing is straightforward. 

Here we only reuse the listening udpmux. If not listening, a new mux is created everytime we dial.

Adds certhashes to discovered /webrtc-direct addresses. Unfortunately this is impossible to unit test currently. I'll manually verify this for now. The situation is the same for webtransport.


sukunrt:
This is not as useful without further work. Even if we reuse port when dialing, we will not dial anyone with a /webrtc-direct address as we will always prefer QUIC to webrtc-direct. This means, it'll not help us much with address discovery.

Issue #2589, opened 2023-09-29
race: simulatenous `NewStream` blocks on one side despite successful connection


dennis-tra:
Simultaneously calling `NewStream` blocks on one side until the dial timeout was reached. Minimum repro example:

```go
package main

import (
	"context"
	"fmt"
	"sync"
	"testing"

	"github.com/libp2p/go-libp2p"
	"github.com/libp2p/go-libp2p/core/host"
	"github.com/libp2p/go-libp2p/core/peerstore"
	"github.com/stretchr/testify/require"
)

func newHost(t *testing.T) host.Host {
	listenAddr := libp2p.ListenAddrStrings("/ip4/127.0.0.1/tcp/0")

	h, err := libp2p.New(listenAddr)
	require.NoError(t, err)

	t.Cleanup(func() {
		if err = h.Close(); err != nil {
			t.Logf("unexpected error when closing host: %s", err)
		}
	})
	return h
}

func TestBlock(t *testing.T) {
	for i := 0; i < 100; i++ {
		t.Run(fmt.Sprintf("[%d]", i), func(t *testing.T) {
			h1 := newHost(t)
			h2 := newHost(t)

			h1.Peerstore().AddAddrs(h2.ID(), h2.Addrs(), peerstore.PermanentAddrTTL)
			h2.Peerstore().AddAddrs(h1.ID(), h1.Addrs(), peerstore.PermanentAddrTTL)

			var wg sync.WaitGroup
			wg.Add(2)
			go func() {
				defer wg.Done()
				_, _ = h1.NewStream(context.Background(), h2.ID(), "/any/protocol")
			}()

			go func() {
				defer wg.Done()
				_, _ = h2.NewStream(context.Background(), h1.ID(), "/any/protocol")
			}()

			wg.Wait()
		})
	}
}
```

I cannot reproduce it on my Mac, but on my Linux server. I need to run the test a few times because it's only sometimes happening. This is an example output:

```
$ go test -v -run ./...
=== RUN   TestBlock
=== RUN   TestBlock/[0]
=== RUN   TestBlock/[1]
...
=== RUN   TestBlock/[98]
=== RUN   TestBlock/[99]
--- PASS: TestBlock (11.99s)
    --- PASS: TestBlock/[0] (0.03s)
    --- PASS: TestBlock/[1] (0.02s)
    ...
    --- PASS: TestBlock/[11] (0.02s)
    --- PASS: TestBlock/[12] (5.02s) <--------------------
[...]

PR #2595, opened 2023-10-06
swarm: add loopback to low timeout filter

dennis-tra:
@sukunrt suggested lowering `swarm.WithDialTimeoutLocal` to prevent our [tests from timing out](https://github.com/plprobelab/zikade/issues/48) because of [blocking dials](https://github.com/libp2p/go-libp2p/issues/2589). In our tests, we connect via loopback addresses, and I noticed that the `swarm.WithDialTimeoutLocal` wasn't applied.

In this PR, I added the loopback CIDR to the list of CIDRs to which the local timeout applies. 

There are a few more ranges from http://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml that aren't in that list. I'm not sure if there's a reason or if it was just forgotten. I'm happy to extend this PR to add the missing ones as well.

These are missing:

```csv
"/ip4/192.0.0.9/ipcidr/32"
"/ip4/192.0.0.10/ipcidr/32"
"/ip4/192.31.196.0/ipcidr/24"
"/ip4/192.52.193.0/ipcidr/24"
"/ip4/192.88.99.0/ipcidr/24"
"/ip4/192.175.48.0/ipcidr/24"
"/ip4/198.18.0.0/ipcidr/15"
```


sukunrt (changes requested):
I've changed my mind here considering this: 

> There are a few more ranges from http://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml that aren't in that list. I'm not sure if there's a reason or if it was just forgotten. I'm happy to extend this PR to add the missing ones as well.


Let's remove this lowTimeoutFilters construct and use `manet.IsPublicAddr` or `manet.IsPrivateAddr`. That will keep this definition in one place. 
https://pkg.go.dev/github.com/multiformats/go-multiaddr/net#IsPrivateAddr

dennis-tra:
Changed to use `manet.IsPrivateAddr` 👍

marten-seemann (approved):
Nice job deleting addrs.go!

PR #2598, opened 2023-10-08
examples: remove unused 'SetStreamHandler'

joohhnnn:
in this echo case, sender is no need to SetStreamHandler, the sender node only used to interact with listener node.


sukunrt:
Thanks @joohhnnn

PR #2601, opened 2023-10-10
webrtcprivate: add transport integration tests

sukunrt:
Depends on: #2576 

The code duplication is unfortunate, but this is reasonable for now. When we have circuit v2 transport integration tests, we can refactor this.


MarcoPolo:

PR #2603, opened 2023-10-12
chore: testing https://github.com/pl-strflt/uci/pull/53 [DO NOT MERGE]

galargh:
TODO:
- [x] test case when PR changes protoc versions in comments
- [x] test case when the new workflow param is not provided
- [x] test case when the new workflow param is false


marten-seemann:
We have this is on master now. Thank you @galargh for resolving yet another pain point!

PR #2611, opened 2023-10-19
circuitv2: don't check ASN for private addrs

sukunrt:
This is responsible for some of the flakiness of the autorelay testsuite. Loading the ASN table takes a lot of time. I think we can optimise the underlying library but this change is harmless since localhost / private addresses wouldn't be associated with an ASN.


marten-seemann (approved):
LGTM. Feel free to merge once you've added the comment.

PR #2613, opened 2023-10-20
test/basichost: fix flaky test due to rcmgr

marten-seemann:


sukunrt (approved):
were these tests running into rcmgr limits?

PR #2615, opened 2023-10-22
webrtc: wait for FIN_ACK before closing data channels

sukunrt:
part of #2656 

This removes the control message queue because that only ever adds 500 bytes to the queue. This will not cause too much HOL Blocking for other streams. 

There is no problem with backpressure since sctp writes never actually block. There is no blocking on sctp Write.

To verify:
```
func TestStreamBackpressure(t *testing.T) {
	client, _ := getDetachedDataChannels(t)

	clientStr := newStream(client.dc, client.rwc, nil)

	b := make([]byte, 1024)
	for i := 0; i < 1e7; i++ {
		n, err := clientStr.dataChannel.Write(b)
		if n != 1024 {
			t.Fatalf("failed to write to datachannel")
		}
		if err != nil {
			t.Fatalf("failed to write to datachannel")
		}
	}
}

```


p2p/test/transport/transport_test.go
```diff
@@ -314,9 +314,6 @@ func TestManyStreams(t *testing.T) {
 	const streamCount = 128
 	for _, tc := range transportsToTest {
 		t.Run(tc.Name, func(t *testing.T) {
-			if strings.Contains(tc.Name, "WebRTC") {
```
marten-seemann, line 317: Exciting!

p2p/transport/webrtc/stream.go
```diff
@@ -52,34 +54,42 @@ type sendState uint8
...
 	writeDeadlineUpdated chan struct{}
 	writeAvailable       chan struct{}
 
-	readLoopOnce sync.Once
+	controlMessageReaderOnce sync.Once
+	// controlMessageReaderEndTime is the end time for reading FIN_ACK from the control
+	// message reader. We cannot rely on SetReadDeadline to do this since that is prone to
+	// race condition where a previous deadline timer fires after the latest call to
+	// SetReadDeadline
+	// See: https://github.com/pion/sctp/pull/290
```
marten-seemann, line 89: Is there anything we can change / simplify here after 290 was merged?

p2p/transport/webrtc/stream.go
```diff
@@ -52,34 +54,42 @@ type sendState uint8
...
 type stream struct {
 	mx sync.Mutex
-	// pbio.Reader is not thread safe,
-	// and while our Read is not promised to be thread safe,
-	// we ourselves internally read from multiple routines...
-	reader pbio.Reader
+
+	// readerOnce ensures that only a single goroutine reads from the reader. Read is not threadsafe
+	// But we may need to read from reader for control messages from a different goroutine.
+	readerOnce chan struct{}
```
marten-seemann, line 68: Nit: Don't call it once, since that's usually used for a `sync.Once`. Alternative: `readerSem`, to make clear that it's a semaphore.

marten-seemann (commented):
Partial review.

Discussed offline: If we move closing of the stream scope to the transports, we don't need to introduce the `AsyncCloser` interface. We need to decide if that structure makes sense first.

p2p/transport/webrtc/connection.go
```diff
@@ -144,16 +126,41 @@ func (c *connection) ConnState() network.ConnectionState {
...
+		// cancel must be called after closeErr is set. This ensures interested goroutines waiting on
+		// ctx.Done can read closeErr without holding the conn lock.
+		c.cancel()
+		c.m.Lock()
+		streams := c.streams
+		c.streams = nil
+		c.m.Unlock()
+		for _, str := range streams {
+			str.Reset()
+		}
```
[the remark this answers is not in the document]
sukunrt: Good point. We don't need to, pion does handle this correctly. I'll remove this.

sukunrt:
I don't have a good idea on how to ensure that we collect the stream handler goroutine and then close the stream rcmgr scope if the transport stream object is responsible for closing the scope and not the swarm stream object. 

However a public interface that exposes AsyncClose doesn't seem required either. I've made the assertion private to swarm package. This is subtle action at a distance but at least it makes it clear that we don't want this AsyncClose system to be copied in other transports.

p2p/transport/webrtc/stream.go
sukunrt: This hasn't been merged yet. 


When it's done we can remove the `controlMsgReaderEndTime` and just set the deadline in AsyncClose & Close to be now + 10 seconds. We currently cannot do that because there's no way of knowing whether a deadlineexceeded error occurred because of a past canceled deadline or because of the most recently set deadline.

p2p/net/swarm/swarm_stream.go
```diff
@@ -75,9 +75,19 @@ func (s *Stream) Write(p []byte) (int, error) {
 	return n, err
 }
 
+type asyncCloser interface {
+	AsyncClose(onDone func()) error
```
Jorropo, line 79: Add docs on the method please, I had to read the implementation to understand this actually also started the close process.

p2p/net/swarm/swarm_stream_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+type asyncStreamWrapper struct {
+	network.MuxedStream
+	beforeClose func()
+}
+
+func (s *asyncStreamWrapper) AsyncClose(onDone func()) error {
+	s.beforeClose()
+	err := s.Close()
+	onDone()
```
Jorropo, line 21: Could we throw a `go` statement somewhere to exercise potential deadlocks and races ?
I think it would be doable by adding `sync.WaitGroup` inside `asyncStreamWrapper`.

We could also add a second `sync.Mutex` which would allow the test the behavior of listing streams while an `AsyncClose` is still in flight (does that deadlocks ? Can we still find the open stream ?).

p2p/net/swarm/swarm_stream.go
```diff
@@ -75,9 +75,19 @@ func (s *Stream) Write(p []byte) (int, error) {
...
+
 // Close closes the stream, closing both ends and freeing all associated
 // resources.
 func (s *Stream) Close() error {
+	if as, ok := s.stream.(asyncCloser); ok {
+		err := as.AsyncClose(func() {
+			s.closeAndRemoveStream()
+		})
+		return err
+	}
```
Jorropo, line 90: This means consumers can't acutally know if the remote ACKed received data (since `.Write` does not block for acks).
But I checked and we don't promise this in the first place:
> ```go
> // Close may be asynchronous and _does not_ guarantee receipt of the data.
> ```

I guess it's fine then.

p2p/transport/webrtc/stream.go
```diff
@@ -150,15 +136,50 @@ func newStream(
...
+		return errors.Join(closeWriteErr, closeReadErr)
 	}
-	return closeReadErr
+
+	s.mx.Lock()
+	s.controlMessageReaderEndTime = time.Now().Add(maxFINACKWait)
+	s.mx.Unlock()
+	s.SetReadDeadline(time.Now().Add(-1 * time.Hour))
+	<-s.controlMessageReaderDone
+	return nil
```
Jorropo, line 152: Shouldn't this error if reading the ack fails ?
Like you can't actually fail to close a stream (unless maybe it's already closed) because even if the remote refuse to close you would still stop responding on this stream, so I always thought the `error` field was used to indicate already written data has been lost.

p2p/transport/webrtc/transport.go
```diff
@@ -76,6 +77,7 @@ const (
 // timeout values for the peerconnection
 // https://github.com/pion/webrtc/blob/v3.1.50/settingengine.go#L102-L109
 const (
+	receiveMTUBytes            = 1500
```
Jorropo, line 80: Pion defaults to 1460 which sounds correct:

https://github.com/pion/webrtc/blob/c259e89eb53b6d599c9dc7ec3e259fdb26e15f95/constants.go#L9-L10

Can you add a comment please (or remove it entirely if it's not needed) ?

Jorropo (commented):
I havn't reviewed everything yet, I have a few questions.

p2p/transport/webrtc/transport.go
sukunrt: This failed once with ErrShortBuffer when the MTU was only 1460. But I think that's because this has a higher MTU and we should remove both the settings(https://github.com/libp2p/go-libp2p/blob/master/p2p/transport/webrtc/udpmux/mux.go#L20)

p2p/net/swarm/swarm_stream.go
sukunrt: yeah, there is no guarantee that the data will be delivered after Close so the semantics seem okay.

p2p/transport/webrtc/transport.go
sukunrt: I've removed this for now. I'll investigate this when this fails.

p2p/net/swarm/swarm_stream_test.go
sukunrt: I updated it with test for 100 concurrent stream Close with GetStreams check. Is this what you had in mind?

p2p/net/swarm/swarm_stream.go
Jorropo: Thx for explaining why it's private.
Just so you know this doesn't actually prevent anyone from implementing it.
You can use a badge using an internal package if you want to do.

p2p/net/swarm/swarm_stream_test.go
Jorropo: I didn't thought of having more than one stream running at once but I think it's good idea.

My point was just about calling onDone out-of-order.

p2p/transport/webrtc/transport.go
Jorropo: Dumb question but shouldn't udpmux also use 1460 MTU ?

sukunrt: I think so. Will take that in a separate PR.

p2p/transport/webrtc/stream.go
```diff
@@ -150,15 +136,50 @@ func newStream(
...
+	if closeWriteErr != nil || closeReadErr != nil {
+		s.Reset()
+		return errors.Join(closeWriteErr, closeReadErr)
 	}
-	return closeReadErr
+
+	s.mx.Lock()
+	s.controlMessageReaderEndTime = time.Now().Add(maxFINACKWait)
+	s.mx.Unlock()
+	s.SetReadDeadline(time.Now().Add(-1 * time.Hour))
```
Jorropo, line 150: That weird.

Jorropo (dismissed):

p2p/transport/webrtc/stream_write.go
```diff
@@ -193,9 +151,12 @@ func (s *stream) CloseWrite() error {
 		return nil
 	}
 	s.sendState = sendStateDataSent
-	if err := s.sendControlMessage(&pb.Message{Flag: pb.Message_FIN.Enum()}); err != nil {
+	select {
+	case s.sendStateChanged <- struct{}{}:
```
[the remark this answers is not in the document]
sukunrt: The channel will keep the message, no? when the Write goroutine comes around to select block? 
There's a problem with concurrent Write calls but that's a race anyway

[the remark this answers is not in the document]
Jorropo: I overlooked the fact that it is buffered.
It looks correct then, I would rather use already existing primitives than reimplementing what is well tested in the `sync` package but if you think this is fine then keep it.

p2p/transport/webrtc/stream.go
```diff
@@ -52,34 +60,41 @@ type sendState uint8
...
 type stream struct {
 	mx sync.Mutex
-	// pbio.Reader is not thread safe,
-	// and while our Read is not promised to be thread safe,
-	// we ourselves internally read from multiple routines...
-	reader pbio.Reader
+
+	// readerSem ensures that only a single goroutine reads from the reader. Read is not threadsafe
+	// But we may need to read from reader for control messages from a different goroutine.
+	readerSem chan struct{}
```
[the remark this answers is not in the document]
sukunrt: Yeah, I don't remember why I did it this way.

p2p/transport/webrtc/stream.go
```diff
@@ -173,59 +194,113 @@ func (s *stream) SetDeadline(t time.Time) error {
...
+		if s.nextMessage != nil {
+			s.processIncomingFlag(s.nextMessage.Flag)
+			s.nextMessage = nil
+		}
+		s.mx.Unlock()
+
+		for !isSendCompleted() {
+			var msg pb.Message
+			if !setDeadline() {
+				return
```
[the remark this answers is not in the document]
sukunrt: Not sure if there's a race here but there's no point risking it.

p2p/transport/webrtc/stream_write.go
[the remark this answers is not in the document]
sukunrt: I don't mind changing it if the performance is better, but since this PR didn't introduce this particular structure, let's take this in a separate PR.

p2p/transport/webrtc/transport.go
sukunrt: This doesn't matter. It's just used for receiving and not for deciding the packet size on the write path. Renamed the variable and added a comment.

p-shahi:
@Jorropo can you help review this PR?

p2p/transport/webrtc/stream.go
```diff
@@ -19,6 +21,9 @@ const (
 	// We can change this value in the SettingEngine before creating the peerconnection.
 	// https://github.com/pion/webrtc/blob/v3.1.49/sctptransport.go#L341
 	maxBufferedAmount = 2 * maxMessageSize
+	// maxTotalControlMessagesSize is the maximum total size of all control messages we will
+	// write on this stream
+	maxTotalControlMessagesSize = 500
```
guillaumemichel, line 26: where does this number come from? 

giving the rationale or link to reference would be great!

p2p/transport/webrtc/stream.go
```diff
@@ -37,6 +42,9 @@ const (
 	// is less than or equal to 2 ^ 14, the varint will not be more than
 	// 2 bytes in length.
 	varintOverhead = 2
+	// maxFINACKWait is the maximum amount of time a stream will wait to read
+	// FIN_ACK before closing the data channel
+	maxFINACKWait = 10 * time.Second
```
guillaumemichel, line 47: why 10 seconds?

guillaumemichel (approved):
LGTM!

It is my first time looking at the WebRTC implementation, so other reviews are also welcome.

p-shahi:
go-libp2p open maintainers call: We have @guillaumemichel 's approval and pending other review comments we will merge this PR. There are other Pion fixes that are necessary to take WebRTC out of experimental as well but merging this will help get the next v0.33 release out the door

p2p/transport/webrtc/transport_test.go
```diff
@@ -665,9 +739,11 @@ func TestConnectionTimeoutOnListener(t *testing.T) {
 	// TODO: return timeout errors here
 	for {
 		if _, err := str.Write([]byte("test")); err != nil {
-			require.True(t, os.IsTimeout(err))
+			// TODO (sukunrt): Decide whether we want to keep this behaviour
+			//require.True(t, os.IsTimeout(err))
```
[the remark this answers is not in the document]
sukunrt: I wanted to confirm whether quic and yamux also do this. They do. I have fixed it to return the correct error and removed the TODOs

p2p/transport/webrtc/stream.go
sukunrt: This should be 50. Added an explanation. Please comment if it needs elaborating.

p2p/net/swarm/swarm_stream.go
Stebalien: This seems totally unnecessary.  Why not just call `closeAndRemoveStream` immediately? From the perspective of libp2p, the stream is closed. If calling `closeAndRemoveStream` causes issues, that's a separate bug that needs to be fixed.


Really, _all_ stream closure is async, this isn't unique to webrtc.

Stebalien (changes requested):

sukunrt:
@Stebalien I have removed `AsyncClose`. 

The difference between webrtc and yamux/quic is that you're relying on the transport stack to discard all the buffers when you do a Close. This happens with yamux. With quic, the receive side of the stream is not immediately discarded but the peer doesn't get credits to open a new stream till it cleanly finishes the stream we have sent `STOP_SENDING` on. 

No such mechanism exists in webrtc. There is no limit to streams you can open(At least not in pion) or unilateral stream closing since streams are reusable. In webrtc `datachannel.Close` is a 1RTT sync procedure over the FIN / FIN_ACK closing that we do. So an effective `AsyncClose` procedure needs to do it after the datachannel is Closed and not after a FIN_ACK is received.  Moreover pion never drops a reference to the datachannel(including all the transport layer allocated buffers) till the connection is open. So a system like this only makes sense if we fix: https://github.com/pion/webrtc/issues/2672 

There is one way to handle this other than doing an explicit async close without waiting for 1RTT which would require us to check a misbehaving peer which is forcing us to keep all the datachannels open and just close the connection to such a peer. We can discuss that strategy in a separate issue. That'd also require https://github.com/pion/webrtc/issues/2672

Stebalien:
> With quic, the receive side of the stream is not immediately discarded but the peer doesn't get credits to open a new stream till it cleanly finishes the stream after we have sent STOP_SENDING on.

Can't we do the same thing here? My objection is adding this hack to libp2p itself as this is a transport specific detail.

p2p/transport/webrtc/stream.go
guillaumemichel: It is still `500` (not `50`) and I don't see the explanation

sukunrt:
>My objection is adding this hack to libp2p itself as this is a transport specific detail.

That's a fair point. I'll open an issue to discuss.

guillaumemichel (approved):

sukunrt:
@Stebalien 
>Why not just call closeAndRemoveStream immediately? 

I have addressed this and removed the `AsyncClose` logic. Is it fine to merge?

p2p/transport/webrtc/stream_read.go
```diff
@@ -35,18 +37,28 @@ func (s *stream) Read(b []byte) (int, error) {
 			var msg pb.Message
 			if err := s.reader.ReadMsg(&msg); err != nil {
 				s.mx.Lock()
```
Stebalien, line 39: nit: not related, but this locking is strange. We could write this as:

```go
s.mx.Unlock()
var msg pb.Message
err := s.reader.ReadMsg(&msg)
s.mx.Lock()
if err != nil {
    // ...
}
// no second "lock" call needed.

File: p2p/transport/webrtc/stream.go [hunk 18]
```diff
@@ -102,45 +120,16 @@ func newStream(
 		writeDeadlineUpdated: make(chan struct{}, 1),
 		writeAvailable:       make(chan struct{}, 1),
 
+		controlMessageReaderDone: sync.WaitGroup{},
```
Stebalien, line 123: nit: no need to initialize this.

p2p/transport/webrtc/stream.go
```diff
@@ -150,15 +139,40 @@ func newStream(
 }
 
 func (s *stream) Close() error {
+	s.mx.Lock()
+	if s.closeForShutdownErr != nil {
```
Stebalien, line 143: nit: same as the other case, I generally prefer:

```go
s.mx.Lock()
closed := s.closeForShutdownErr != nil
s.mx.Unlock()
...
```

p2p/transport/webrtc/stream.go
```diff
@@ -167,65 +181,134 @@ func (s *stream) Reset() error {
...
 }
 
+func (s *stream) closeForShutdown(closeErr error) {
+	defer s.cleanup()
+
+	s.mx.Lock()
+	defer s.mx.Unlock()
+
+	s.closeForShutdownErr = closeErr
+	s.SetReadDeadline(time.Now().Add(-1 * time.Hour))
```
Stebalien, line 191: This seems like a strange way to do this. It'll also be racy (e.g., if the user calls the same function to change the read deadline).

Do we need to change the read deadline? If this is the only way to do this, let's add something to `SetReadDeadline` so it can't be changed again by the user.

p2p/transport/webrtc/stream.go
```diff
@@ -167,65 +181,134 @@ func (s *stream) Reset() error {
...
 		}
 		select {
 		case s.sendStateChanged <- struct{}{}:
 		default:
 		}
+	case pb.Message_FIN_ACK:
+		s.sendState = sendStateDataReceived
+		select {
+		case s.sendStateChanged <- struct{}{}:
+		default:
```
Stebalien, line 225: Future cleanup: make some form of `notifyStateChanged` function.

p2p/transport/webrtc/stream.go
```diff
@@ -167,65 +181,134 @@ func (s *stream) Reset() error {
...
+					return true
+				}
+				return false
+			}
 
-func (s *stream) setCloseError(e error) {
-	s.mx.Lock()
-	defer s.mx.Unlock()
+			// Unblock any Read call waiting on reader.ReadMsg
+			s.SetReadDeadline(time.Now().Add(-1 * time.Hour))
```
Stebalien, line 263: same race comment as above.

p2p/transport/webrtc/stream_write.go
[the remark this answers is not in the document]
Stebalien: Using a channel as a "is this changed" flag is a common idiom, assuming there's only _one_ reader. `sync.Cond` is mostly useful when there are many readers.

p2p/transport/webrtc/transport_test.go
```diff
@@ -17,6 +18,7 @@ import (
 	quicproxy "github.com/quic-go/quic-go/integrationtests/tools/proxy"
 
 	"golang.org/x/crypto/sha3"
+	"golang.org/x/exp/rand"
```
Stebalien, line 21: Is this the rand you want to use?

Stebalien (commented):
Some comments, mostly questions, otherwise LGTM.

p2p/transport/webrtc/stream.go
sukunrt: I've replaced the three channels with 1 `writeStateChanged` and using a method to notify on the channel removing all the `select` blocks everywhere.

p2p/transport/webrtc/transport_test.go
sukunrt: I did mostly for this: https://github.com/golang/go/issues/21835

But looks like some of the changes have made it to math/rand and go1.23 will have a migrate tool to move to math/rand/v2. So the stdlib rand is fine here.

p2p/transport/webrtc/stream.go
sukunrt: This is the only way to unblock a waiting reader. Even closing the datachannel won't unblock it till 1RTT later when the peer replies with a close datachannel on its part. 

I've changed this to make SetReadDeadline a noop after the read half is closed so the user can't change this again.

Stebalien:
> I have addressed your concerns here. I want to do a release today. Happy to take any comments in a follow up PR and if something major is missed we can do a patch release.


Nope, LGTM!

p2p/transport/webrtc/transport.go
sukunrt: I think I found the bug:
This was causing pion to send larger packets some times. 
https://github.com/pion/sctp/pull/301/files

PR #2618, opened 2023-10-25
Refactor the proxy example.

kaichaosun:
When go through the http-proxy example, start client service and server service confuse me, here I refactor the code to make it less confusion for new readers.


marten-seemann:
Can you elaborate why you think it's less confusing?

kaichaosun:
There are two roles (client/server) in this example, I have to look deep into the shared `NewProxyService` to tell which logic for client and which for server. This is not necessary and consume more time reading the code.

kaichaosun:
@marten-seemann I don't get any new feedback about the PR, if the change is not helpful, feel free to close, but I would be happy if this is helpful and get merged.

PR #2631, opened 2023-11-01
ci: ignore protoc version comment on go generate

galargh:
Go generate leaves protoc version comments in some generated files. This PR ensures that we ignore it in the go check workflow.


.github/workflows/go-check.yml
```diff
@@ -15,4 +15,6 @@ concurrency:
 
 jobs:
   go-check:
-    uses: pl-strflt/uci/.github/workflows/go-check.yml@v0.0
+    uses: pl-strflt/uci/.github/workflows/go-check.yml@release-v0.0.13
```
galargh, line 18: This is just testing before release. We should restore v0.0 once v0.0.13 is out.

```suggestion
    uses: pl-strflt/uci/.github/workflows/go-check.yml@v0.0
```

sukunrt (approved):

sukunrt:
Thanks @galargh

.github/workflows/go-check.yml
sukunrt: So should we merge this as is or wait for the release?

galargh: We can go back to this now:
```suggestion
    uses: pl-strflt/uci/.github/workflows/go-check.yml@v0.0
```

marten-seemann (approved):

PR #2635, opened 2023-11-06
Remove unused stream handler for sender in example/echo

kaichaosun:
It seems to me such stream handler is used by the listener side or used by chat use cases, for echo, it don't need to use this steam handler.


sukunrt:
Can you rebase this?

kaichaosun:
Another fix already merged [here](https://github.com/libp2p/go-libp2p/pull/2598).

PR #2636, opened 2023-11-06
swarm: fix timer Leak in the dial loop

Jorropo:
Ensure we never leak timers out of the dial loop.
See https://discuss.ipfs.tech/t/constant-100-cpu-utilization-on-aws-ec2-ipfs-node/17172


Jorropo:
Screenshot from 2023-11-06 13-55-34
Screenshot from 2023-11-06 14-04-59

sukunrt (approved):

PR #2639, opened 2023-11-06
release v0.31.1 (#2636)

sukunrt:
We should do a patch release here as v0.32 removes support for `peer.ID.Pretty` and it might not be trivial for users to upgrade.


marten-seemann:
Was this bug recently introduced?

sukunrt:
No. It's been there from the first smart dialing implementation(v0.28)

PR #2640, opened 2023-11-12
Fix Swarm Grafana Dashboard

burdiyan:
When using the existing Grafana setup, the newer versions of Grafana fail to provision the Swarm dashboard. It works when importing it manually, but it fails during the automatic provisioning, complaining about not being able to transform legacy templates.

Apparently the issues is in the format of the `templating` field of the dashboard JSON. I adapted it from the other dashboards that seem to work just fine with provisioning.

### DIFF
--- a/dashboards/swarm/swarm.json
+++ b/dashboards/swarm/swarm.json
@@ -232,9 +232,7 @@
       "options": {
         "orientation": "auto",
         "reduceOptions": {
-          "calcs": [
-            "lastNotNull"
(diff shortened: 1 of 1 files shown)


sukunrt:
Thanks @burdiyan Can you tell me the grafana version it is failing with?

sukunrt (approved):

sukunrt:
Thanks @burdiyan!

Issue #2643, opened 2023-11-13
Peerstore reports empty protocolID list


juligasa:
In our network, we identify peers talking `hypermedia` protocol by checking if that protocol is present in the list of protocols when connecting to a peer. We get the list of protocols by doing `<caller>.Peerstore().GetProtocols(<called_pid>)` (we could be using `FirstSupportedProtocol` but for the sake of this issue, getting all supported protocols is best) after connecting with a peer on the network. The usual response is like:
`[/ipfs/lan/kad/1.0.0 /hypermedia/0.2.0 /ipfs/bitswap /ipfs/bitswap/1.0.0 /ipfs/bitswap/1.1.0 /ipfs/bitswap/1.2.0 /ipfs/id/push/1.0.0 /ipfs/id/1.0.0 /ipfs/ping/1.0.0 /libp2p/circuit/relay/0.2.0/stop]`
However, some peers report no protocols at all `[]` (as if we were issuing `RemoveProtocols`) and the only way to make them work properly again is either restarting the called node or restarting the caller node. Note that while they are in this _protocolless_ state, any subsequent calls to `NewStream()` will fail (with the error `failed to negotiate protocol: context deadline exceeded`), which is the real blocker in our case.
After restarting the caller, a new random set of peers will face the same issue.
In unitary tests, this seems to work. Any idea on what's going on?

<details>
<summary>Version Information</summary>
<pre>

mintter
cloud.google.com/go v0.105.0
cloud.google.com/go/accessapproval v1.5.0
cloud.google.com/go/accesscontextmanager v1.4.0
cloud.google.com/go/aiplatform v1.27.0
cloud.google.com/go/analytics v0.12.0
cloud.google.com/go/apigateway v1.4.0
cloud.google.com/go/apigeeconnect v1.4.0
cloud.google.com/go/appengine v1.5.0
cloud.google.com/go/area120 v0.6.0
cloud.google.com/go/artifactregistry v1.9.0
cloud.google.com/go/asset v1.10.0
cloud.google.com/go/assuredworkloads v1.9.0
cloud.google.com/go/automl v1.8.0
cloud.google.com/go/baremetalsolution v0.4.0
cloud.google.com/go/batch v0.4.0
cloud.google.com/go/beyondcorp v0.3.0
cloud.google.com/go/bigquery v1.44.0
cloud.google.com/go/billing v1.7.0
cloud.google.com/go/binaryauthorization v1.4.0
[...]

PR #2644, opened 2023-11-14
pstoremanager: fix race condition when removing peers from peer store

marten-seemann:
For https://github.com/libp2p/go-libp2p/issues/2643.

@juligasa could you try out this PR?


juligasa:
Unfortunately, after doing `go get github.com/libp2p/go-libp2p@5a419b2a0327bb9e9f97c25410ed42f7a312fb0e` in local app and re-running, I still see peers (who run the previous libp2p version) reporting zero protocols -according to peerstore-

sukunrt (approved):

Issue #2628, opened 2023-10-31
Question: rcmgr limit conflicts with connmgr limit


richard-ramos:
While running this example code with `GOLOG_LOG_LEVEL=WARN go run main.go` , I always see this warning being printed:
```
2023-10-31T12:36:25.909-0400    WARN    p2p-config      config/config.go:304    rcmgr limit conflicts with connmgr limit: conn manager high watermark limit: 192, exceeds the system connection limit of: 1
```
What does this warning mean? I tried to follow go-libp2p code for the resource manager and connection manager to find out but had some trouble understanding it. Is there something I should change to avoid having this warning?

```go
package main

import (
	"fmt"
	"os"
	"os/signal"
	"syscall"

	"github.com/libp2p/go-libp2p"
)

func main() {
	host, err := libp2p.New()
	if err != nil {
		panic(err.Error())
	}

	fmt.Println(host.ID())

	// Wait for a SIGINT or SIGTERM signal
	ch := make(chan os.Signal, 1)
	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
	<-ch
	fmt.Println("\n\n\nReceived signal, shutting down...")
}
```

Thank you in advance.

burdiyan:
Facing the same issue.

sukunrt:
This means that you have configured the connection manager in such a way that it will never be used. 

Libp2p uses two modules for connection managements. Resource manager limit(rcmgr) is a hard limit. If the rcmgr limit is set to 1 you will only have 1 open connection. 

The ConnMgr limit is a soft limit of sort. It allows for a burst of connections between ConnMgr.LowWaterMark and ConnMgr.HighWaterMark. This allows for a burst of connections between Low and high watermark. But when you exceed the high watermark, the system trims connections down to low watermark. 

In your case HighWatermark trimming never happens because it's higher than the rcmgr limit. Ideally your high watermark limit should be a bit less than the rcmgr connections limit.

juligasa:
I have the same issue, even if I set unlimited resources, it still complains about the system limit being 1 `WARN    p2p-config      config/config.go:304    rcmgr limit conflicts with connmgr limit: conn manager high watermark limit: 768, exceeds the system connection limit of: 1`
My configuration so far:
```
cm := mustConnMgr(connmgr.NewConnManager(
	512,
	768,
	connmgr.WithGracePeriod(r.cfg.ConnMgr.ConnMgrGrace),
))

opts = append(opts,
[...]

PR #2648, opened 2023-11-18
rcmgr: fix connmgr connection limit conflict warning

sukunrt:
closes: #2628 

The ConnLimit is associated with Connection scope and always has connection limit 1 because within a connection scope you only need 1 connection. 

The correct system connection limit is provided by the system scope. 

Thanks! @juligasa


marten-seemann (approved):

marten-seemann:
Good catch!

PR #2650, opened 2023-11-22
defaults: do TLS by default for encryption

Jorropo:
Tls has much better throughput, the handshake benchmark is fairly noisy, there is no significant performance difference, however it does allocate more.
```
goos: linux
goarch: amd64
cpu: AMD Ryzen 5 3600 6-Core Processor
BenchmarkNoise/throughput/32KiB-12 24984             46605 ns/op         703.10 MB/s          37 B/op          2 allocs/op
BenchmarkNoise/throughput/1MiB-12   1134           1459483 ns/op         718.46 MB/s         663 B/op         34 allocs/op
BenchmarkNoise/handshakes-12        1302           1054533 ns/op           32691 B/op        348 allocs/op
BenchmarkTls/throughput/32KiB-12   49006             24309 ns/op        1347.99 MB/s          50 B/op          2 allocs/op
BenchmarkTls/throughput/1MiB-12     1747            778498 ns/op        1346.92 MB/s        1603 B/op         64 allocs/op
BenchmarkTls/handshakes-12          1116           1045475 ns/op          105257 B/op       1478 allocs/op
```

### DIFF
--- a/p2p/security/noise/transport_test.go
+++ b/p2p/security/noise/transport_test.go
@@ -17,7 +17,6 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/protocol"
 	"github.com/libp2p/go-libp2p/core/sec"
-	"github.com/libp2p/go-libp2p/p2p/security/internal/benchmark"
 	"github.com/libp2p/go-libp2p/p2p/security/noise/pb"
 
 	"github.com/stretchr/testify/assert"
@@ -714,12 +713,4 @@ func TestHandshakeWithTransportEarlyData(t *testing.T) {
 			noiseHandshake(t, test.clientProtos, test.serverProtos, test.expectedResult)
 		})
 	}
-}
-
-func BenchmarkNoise(b *testing.B) {
-	benchmark.Bench(b, func(b *testing.B, priv crypto.PrivKey) sec.SecureTransport {
-		tpt, err := New("", priv, nil)
-		assert.NoError(b, err)
-		return tpt
-	})
 }
--- a/p2p/security/noise/transport_test.go
+++ b/p2p/security/noise/transport_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/protocol"
 	"github.com/libp2p/go-libp2p/core/sec"
+	"github.com/libp2p/go-libp2p/p2p/security/internal/benchmark"
 	"github.com/libp2p/go-libp2p/p2p/security/noise/pb"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


Jorropo:
Throughput is basically comparing go's hardware accelerated AES impl vs the software chacha one.

marten-seemann:
Any idea why TLS is allocating like crazy during the transfer?

Jorropo:
So is `noise`, I think it could be `net.Pipe`.
AFAIK tls allocates once per refresh:
- https://go-review.googlesource.com/c/go/+/464835
- https://github.com/golang/go/issues/53142

p2p/security/noise/transport_test.go
```diff
@@ -714,3 +715,11 @@ func TestHandshakeWithTransportEarlyData(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkNoise(b *testing.B) {
```
marten-seemann, line 719: Can we have them all in one place? `p2p/test/transport` or `p2p/test/security` would probably be a good place.

p2p/security/tls/transport_test.go
```diff
@@ -709,3 +710,11 @@ func TestInvalidCerts(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkTls(b *testing.B) {
```
marten-seemann, line 714:
```suggestion
func BenchmarkTLS(b *testing.B) {
```

p2p/security/noise/transport_test.go
Jorropo: I moved it to `p2p/test/security`

sukunrt (approved):
LGTM Thanks @Jorropo!
Should we remove the asserts from the benchmarks?

Jorropo:
The assert is extremely cheap, it ensure the benchmark is actually benchmarking as intended.

aschmahmann:
Note: If merging this I would make sure to call out this change in the release notes so people can decide if/when they want to make the change. IIUC this will add one roundtrip to all TCP/WS/WSS connections in between go-libp2p nodes in the short term (while nodes update) and may also add one roundtrip to all of those connection types to libp2p implementations that don't support TLS (is that all the others, or is my info out of date?).

If I recall the reason Noise was made the default over TLS was that TLS was only really supported in go-libp2p and the potential for extra roundtrips weren't deemed worthwhile.

sukunrt:
>this will add one roundtrip to all TCP/WS/WSS connections in between go-libp2p nodes in the short term (while nodes update)

This shouldn't happen, I'll confirm though. 
The dialer(new go-libp2p version) should eagerly write /tls now and the server should select this preference, so this should only take 1 Round Trip. In the other case the dialer(old go-libp2p version) should eagerly write /noise and the server(new go-libp2p) version should select the client's preference. 

>may also add one roundtrip to all of those connection types to libp2p implementations that don't support TLS

Jorropo:
Libp2p's noise require ChaChaPoly sha256, that would require a new spec and break compat which is bad.
We also don't have a way to negociate multiple noise suites without round trips, TLS sends all the supported suites at once and figure it out.
The go tls implementation is smart, it will dynamically prefer AES over ChaCha (or not) depending on what the CPU supports.

The cost of encapsulation seems trivial over the cost of the cipher, your `Noise_XX_DH25519_AESGCM_BLAKE2b` benchmark is within benchmark of TLS.

Even if TLS was too slow, `crypto/tls` isn't very well optimized and could be replaced.

PR #2660, opened 2023-12-06
feat: support ipv6 link-local address

### LINKED ISSUE #2659: Connect swarm with ipv6 link-local address.
Make available p2p in before ip configure.

The problems I found were:

1. IPv6 advertising incorrectly
`/ip6zone/br0/ip6/fe80::7369:44d4:4573:9c7f/tcp/12345` When listen to an IP like this, ip6zone is included when advertising the address. This is just for own interface, so don't need to advertise it.

the code below roughly works.

```
		func(cfg *libp2p.Config) error {
			cfg.AddrsFactory = func(addrs []ma.Multiaddr) []ma.Multiaddr {
				var filtered []ma.Multiaddr
				for _, addr := range addrs {
					_, err := addr.ValueForProtocol(ma.P_IP6ZONE)
					if err == nil {
						tokens := multiaddr.Split(addr)
						addr = multiaddr.Join(tokens[1:]...)
					}
					filtered = append(filtered, addr)
				}

				return filtered
			}
			return nil
		},
```

2. Unable to connect via ip6zone.

```
var dialMatcher = mafmt.And(mafmt.IP, mafmt.Base(ma.P_TCP))

// CanDial returns true if this transport believes it can dial the given
//multiaddr.
func (t *TcpTransport) CanDial(addr ma.Multiaddr) bool {
return dialMatcher.Matches(addr)
}
```

link-local address cannot start with /ip6/ because an interface must be bound for communication. But it fails with CanDial.

3. Note
It may be a good idea to listen on link-local addresses but not advertise them.
You can discover and connect with mdns.

jclab-joseph:
Close #2659


marten-seemann (commented):
Is there any reason why a well-behaved peer would ever advertise an ip6zone address? I can’t see any, but I might be missing something.

If that’s the case, we don’t need to make any changes to the TCP transport.

jclab-joseph:
@marten-seemann 

As an example, an IoT node before IP is set can use it to communicate via p2p.
IPv6 link-local address can be used even if a static IP is not set.
In this case, you can communicate by searching for each other using mdns, etc. within the local network.

Since link-local address cannot be used on the Internet, it was intentionally excluded from announce.

marten-seemann:
What do you mean by „before the IP address is set“? Are you referring to DHCP?

How would and ip6zone address passed to the TCP transport look like?

marten-seemann:
Can you just advertise /ip6/fe80::3811:2aff:fe94:adbe/tcp/1234? For the peer, it’s irrelevant that it’s on your eth0 interface, right?

jclab-joseph:
@marten-seemann Is there a reason to announce it? A link-local address is only useful if the receiving party can determine the listening interface, such as mdns.
I thought it would be a good idea not to advertise unnecessary addresses. Still, is it better to advertise?

marten-seemann:
> A link-local address is only useful if the receiving party can determine the listening interface, such as mdns.

I don't understand how this would work. What is configured to be eth0 is only relevant to me, not to the remote node. Therefore, we should never advertise it.

MarcoPolo:

PR #2662, opened 2023-12-06
swarm: remove /ip6zone component from listen addresses

### LINKED ISSUE #2659: Connect swarm with ipv6 link-local address.

marten-seemann:
Fixes #2659.

/ip6zone addresses shouldn't be advertised. This PR adds logic to remove the ip6zone component from addresses returned by `Swarm.ListenAddresses` and `Swarm.InterfaceListenAddresses`.

@jclab-joseph Does this solve your problem?

**TODO: add tests**

### DIFF
--- a/p2p/net/swarm/swarm_addr.go
+++ b/p2p/net/swarm/swarm_addr.go
@@ -18,7 +18,17 @@ func (s *Swarm) ListenAddresses() []ma.Multiaddr {
 func (s *Swarm) listenAddressesNoLock() []ma.Multiaddr {
 	addrs := make([]ma.Multiaddr, 0, len(s.listeners.m)+10) // A bit extra so we may avoid an extra allocation in the for loop below.
 	for l := range s.listeners.m {
-		addrs = append(addrs, l.Multiaddr())
+		a := l.Multiaddr()
+		// remove ip6zone from the addresses
+		ma.ForEach(a, func(c ma.Component) bool {
+			if c.Protocol().Code == ma.P_IP6ZONE {
+				_, a = ma.SplitFirst(a)
+			}
+			return false
+		})
+		if a != nil {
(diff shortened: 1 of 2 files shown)


jclab-joseph:
No this is the exact opposite of my intention.
My PR(#2660) was to enable Listen for ip6zone to enable p2p communication with link-local address.
However, this PR excludes ip6zone.

MarcoPolo:

Issue #193, opened 2017-05-14
KCP transport (reliable UDP)


ghost:
Let's check it out:

https://github.com/xtaci/kcp-go

> **kcp-go** is a **Production-Grade Reliable-UDP** library for [golang](https://golang.org/). 
>
> It provides **fast, ordered and error-checked** delivery of streams over **UDP** packets, has been well tested with opensource project [kcptun](https://github.com/xtaci/kcptun). Millions of devices(from low-end MIPS routers to high-end servers) are running with **kcp-go** at present, including applications like **online games, live broadcasting, file synchronization and network acceleration**.

> 1. Optimized for **Realtime Multiplayer Games, Audio/Video Streaming**.
> 1. Compatible with [skywind3000's](https://github.com/skywind3000) C version with language specific optimizations.
> 1. **Cache friendly** and **Memory optimized** design, offers extremely **High Performance** core.
> 1. Compatible with [net.Conn](https://golang.org/pkg/net/#Conn) and [net.Listener](https://golang.org/pkg/net/#Listener), easy to use.
> 1. [FEC(Forward Error Correction)](https://en.wikipedia.org/wiki/Forward_error_correction) Support with [Reed-Solomon Codes](https://en.wikipedia.org/wiki/Reed%E2%80%93Solomon_error_correction)
> 1. Packet level encryption support with [AES](https://en.wikipedia.org/wiki/Advanced_Encryption_Standard), [TEA](https://en.wikipedia.org/wiki/Tiny_Encryption_Algorithm), [3DES](https://en.wikipedia.org/wiki/Triple_DES), [Blowfish](https://en.wikipedia.org/wiki/Blowfish_(cipher)), [Cast5](https://en.wikipedia.org/wiki/CAST-128), [Salsa20]( https://en.wikipedia.org/wiki/Salsa20), etc. in [CFB](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation#Cipher_Feedback_.28CFB.29) mode.
> 1. **O(1) goroutines** created for the entire server application, minimized goroutine context switch.

It's used by Lantern and Syncthing among others

whyrusleeping:
Yeah, i'd love to try this out. Should be fairly simple to plug it in

whyrusleeping:
They also have a multiplexer: https://github.com/xtaci/smux

whyrusleeping:
[...]

PR #2672, opened 2023-12-14
feat: add kcp transport with listener implement 

derrandz:
### Description

This PR aims to introduce KCP as a supported protocol in the libp2p library

### Missing

* Metrics
* Reuseport


marten-seemann:
Which issue does this resolve? I’m not convinced KCP is a good idea, but we should discuss this in an issue and not on a PR.

marten-seemann:
There already is a 3rd party KCP transport implementation: https://github.com/paralin/go-libp2p-kcp.

marten-seemann:
#193 was the tracking issue, and we closed it once QUIC was merged. There really shouldn’t be any reason to use KCP if you can use QUIC instead.

derrandz:
Hi @marten-seemann thanks for the context. This was old and wanted to do a performance test against quic through libp2p to provide a rational for using kcp alongside quic or instead of quic. I would have to come with rationale and benchmark stats but closing this for now makes sense.

Cheers,

PR #2673, opened 2023-12-19
chore: update go-libp2p-asn-util

### LINKED ISSUE #2664: go-libp2p-asn-util adds ~4MiB to builds and increases build times, even in configurations where it will never be used
I've found that go-libp2p-asn-util is linked to production binaries that import libp2p, even when they do not use any of the relay functionality that needs this util. This is due to interfaces being linked from `protocol/circuitv2/relay`. This util seems to be responsible for as much as 4MiB in the final build and it seems to increase our production build time by 19%. This is because it has a rather large array which our static checks and pgo have to grind through. 

Please consider refactoring use of this dependency such that users of go-libp2p do not have to import it unless they are actually using it.

Jorropo:
Fixes #2664


p2p/protocol/circuitv2/relay/constraints.go
```diff
@@ -73,13 +73,10 @@ func (c *constraints) AddReservation(p peer.ID, a ma.Multiaddr) error {
...
-	// Only public addresses have an ASN. Skip checking ASN for private addresses as
-	// initialising the ASN store is a costly operation. Skipping this check reduces a lot of
-	// flakiness in tests
-	if ip.To4() == nil && manet.IsPublicAddr(a) {
-		asn, _ = asnutil.Store.AsnForIPv6(ip)
-		if asn != "" {
+	var asn uint32
+	if ip.To4() == nil {
+		asn = asnutil.AsnForIPv6(ip)
+		if asn != 0 {
```
sukunrt, line 79: Is it fine to check this for Private addresses now? Previously this was a slow operation causing flakiness in CI

Jorropo: I removed all the initialization, it operate straight on binary data mmaped from in `.rodata` part of the executable, one lookup is now ~50ns on my computer.
Would be more if this part of the executable is cold (not in L2 and or the pages are still tripped), but never something significant (there are ~17 steps of binary search to do).

sukunrt (approved):
This is great! Thanks @Jorropo

Issue #2681, opened 2024-01-09
Auto Relay Doesn't Work in libp2p Network


amirvalhalla:
Hi,

In the [Pactus Blockchain Network](https://github.com/pactus-project/pactus), we're using `go-libp2p` as our `p2p` network. However, we are currently facing an issue with auto relay, even though we have configured everything as mentioned in the `libp2p` documents and examples. Despite setting up metrics, etc., we don't observe any relay traffic.

You can review the relevant code [here](https://github.com/pactus-project/pactus/blob/main/network/network.go) in the `newNetwork` method where we initialize `libp2p`.

sukunrt:
Difficult to debug without more info. 
What does this log say? https://github.com/pactus-project/pactus/blob/main/network/network.go#L312

Can you run it with logs for autorelay enabled?
GOLOG_LOG_LEVEL="autorelay=debug"

amirvalhalla:
> Difficult to debug without more info. What does this log say? https://github.com/pactus-project/pactus/blob/main/network/network.go#L312
> 
> Can you run it with logs for autorelay enabled? GOLOG_LOG_LEVEL="autorelay=debug"

We made a mistake regarding the relay, and the issue has been identified. Thanks.

PR #2690, opened 2024-01-25
config: error when relay service is enabled without relay transport

sukunrt:
fixes: #2681


sukunrt:
This isn't needed. Relay service can work without the relay transport.

PR #2695, opened 2024-01-28
fix:Add HTTPS to documentation link

keienWang:
This PR updates a specific link in the documentation to use `https`, enhancing link security. It's a minor improvement to the quality of our project documentation but significantly important for protecting end-users and elevating our project's security standards.


Jorropo (approved):
This doesn't do what you think it does, `href="libp2p.io"` is understood as a relative link (which is wrong).

That is still a fix. Thx

PR #2696, opened 2024-01-29
Add a "transient" network connectivity state

### LINKED ISSUE #2692: Add Transient as a "connectedness" state
This will let users:

1. Quickly detect if the connection to a peer is transient.
2. Detect transitions from transient to "full" connections (and back) via events.

Stebalien:
Previously, we'd consider "transiently" connected peers to be connected. This meant:

1. We wouldn't fire a second event when transitioning to "really connected". The only option for users was to listen on the old-style per-connection notifications.
2. "Connectedness" checks would be a little too eager to treat a peer as connected.

For 99% of users, "transient" peers should be treated as disconnected. So while it's technically a breaking change to split-out "transient" connectivity into a separate state, I expect it's more likely to fix bugs than anything.

Unfortunately, this change _did_ require several changes to go-libp2p itself because go-libp2p _does_ care about transient connections:

1. We want to keep peerstore information for transient peers.
2. We may sometimes want to treat peers as "connected" in the host.
3. Identify still needs to run over transient connections.

fixes #2692


core/network/network.go
```diff
@@ -61,10 +61,13 @@ const (
 	// CannotConnect means recently attempted connecting but failed to connect.
 	// (should signal "made effort, failed")
 	CannotConnect
+
+	// Transient means we have a transient connection to the peer, but aren't fully connected.
+	Transient
```
Stebalien, line 66: Call this something else?

core/network/network.go
```diff
@@ -61,10 +61,13 @@ const (
 	// CannotConnect means recently attempted connecting but failed to connect.
 	// (should signal "made effort, failed")
 	CannotConnect
```
Stebalien, line 63: Maybe we can just get rid of these? I don't think _anyone_ is referencing them...

Stebalien: That would also simplify checks to `!= disconnected` (transient ok) `== connected` (transient not ok).

p2p/host/basic/basic_host.go
```diff
@@ -723,8 +723,10 @@ func (h *BasicHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
 	h.Peerstore().AddAddrs(pi.ID, pi.Addrs, peerstore.TempAddrTTL)
 
 	forceDirect, _ := network.GetForceDirectDial(ctx)
+	canUseTransient, _ := network.GetUseTransient(ctx)
 	if !forceDirect {
-		if h.Network().Connectedness(pi.ID) == network.Connected {
+		connectedness := rh.Network().Connectedness(pi.ID)
```
Stebalien, line 728: I'm not sure if this is now correct, but it was definitely wrong before. We need to _block_ until we have a direct connection. Does that now work?

p2p/host/routed/routed.go
```diff
@@ -48,8 +48,10 @@ func Wrap(h host.Host, r Routing) *RoutedHost {
 func (rh *RoutedHost) Connect(ctx context.Context, pi peer.AddrInfo) error {
 	// first, check if we're already connected unless force direct dial.
 	forceDirect, _ := network.GetForceDirectDial(ctx)
+	canUseTransient, _ := network.GetUseTransient(ctx)
 	if !forceDirect {
-		if rh.Network().Connectedness(pi.ID) == network.Connected {
+		connectedness := rh.Network().Connectedness(pi.ID)
```
Stebalien, line 53: Ideally we wouldn't try to perform routing if we're _just_ going to wait on an ongoing connection. But that's not a new issue.

core/network/network.go
sukunrt: I prefer `limited`. 
This is fine too. At least it's consistent with the network.UseTransient.

p2p/host/basic/basic_host.go
sukunrt: There is inconsistency between Connect and NewStream. 

Connect will just connect and doesn't care about whether the connection is transient or not. Using ForceDirect tells connect to *only* dial direct addresses and not relayed addresses. This is used by the DCUTR code to ensure that relayed addresses are not dialled. 

NewStream cares about whether the connection is transient or not and will block for hole punching. So NewStream uses Connect to first connect to the peer over a relay and then waits for the hole punch to succeed if WithUseTransient option is not selected. 

The fix here is to have NewStream call Connect with network.WithUseTransient and then wait for the hole punch to happen. 

I suggest we keep it as it is in this PR and take that change in a separate PR.

core/network/network.go
sukunrt: It'd be nice to fix these too. Can you remove them?

core/network/network.go
marten-seemann: relayed?

Stebalien: Limited/Temporary make the most sense to me. It may be relayed, it may not be.

core/network/network.go
Stebalien: Removing these will be a breaking change (it'll break actual users. IMO, we should deprecate first.

p2p/host/basic/basic_host.go
Stebalien: I disagree. The fix is to make `Connect` wait until you have a connection with the desired properties, not to mess with `NewStream`. Wasn't that fixed in https://github.com/libp2p/go-libp2p/issues/1603?

sukunrt: That fixed it for NewStream which now blocks for a direct connection and kept Connect behaviour the same as before. Part of the reason was because Connect was broken previously. Connect without network.WithUseTransient would return with no error and transient connectivity without blocking.  

Thinking more about this my suggested solution is incorrect and I'm not against fixing Connect behaviour. Just to take it in a different PR.

Stebalien: > Thinking more about this my suggested solution is incorrect and I'm not against fixing Connect behaviour. Just to change the behaviour in a different PR.

SGTM. I can make that PR first.

sukunrt: That would be amazing. Otherwise I can also take that up.

Stebalien: It's probably going to be several patches. "transient" and "indirect" (through a proxy) are _not_ the same but the code seems to use these concepts interchangeably.

core/network/network.go
sukunrt: SGTM.

p2p/net/swarm/swarm.go
```diff
@@ -397,8 +399,12 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
 	c.notifyLk.Lock()
 	s.conns.Unlock()
 
-	// Notify goroutines waiting for a direct connection
-	if !c.Stat().Transient {
+	newState := network.Transient
```
sukunrt, line 402: We should calculate newState just after we add it to conns map with connectednessUnlocked. We already have the lock it's a small set to check anyway.
There's a problem here that if someone makes a relayed connection after we have a direct connection this will send a limited connectivity event.

p2p/net/swarm/swarm.go
```diff
@@ -412,10 +418,10 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
 
 	// Emit event after releasing `s.conns` lock so that a consumer can still
 	// use swarm methods that need the `s.conns` lock.
-	if isFirstConnection {
+	if oldState != newState {
 		s.emitter.Emit(event.EvtPeerConnectednessChanged{
 			Peer:          p,
-			Connectedness: network.Connected,
+			Connectedness: newState,
 		})
```
sukunrt, line 425: You need the same changes here on the disconnection side: https://github.com/libp2p/go-libp2p/blob/57736f1ec68de8471650a65ff383cb0ada0a0c4d/p2p/net/swarm/swarm.go#L777

sukunrt (changes requested):
I'm fine with these changes. We just need to 

1. Fix this for the close path.
2. Keep existing Connect behavior. I'll raise a PR to make Connect consistent with NewStream. 
3. Deprecate the items in network.Connectedness that we don't care about.

core/network/network.go
sukunrt: I prefer `Limited` because the opposite `Unlimited` also makes sense. I don't think `Temporary` / `Permanent` conveys that correctly. Because whether a connection is transient or not depends on how much data we can transfer over it. Relayed connections with no data limit count as Nontransient.

p2p/host/basic/basic_host.go
Stebalien: > That would be amazing. Otherwise I can also take that up.

Ok, I'm going to be realistic... I'm not likely going to get to this soon (possibly within a month or two, but not "soon"). Go ahead and take it if you find time/motivation.

core/network/network.go
MarcoPolo: I would prefer we rename this while we are here. `Limited` sounds good to me.

After a couple of months of not working on go-libp2p, it took me a while to remember what a "transient" connection really way. Transient implies that the connection may disappear soon. Which may not be the case

Stebalien: Limited sounds like the way to go then.

guillaumemichel: Renamed to `Limited`

core/network/network.go
guillaumemichel: `CanConnect` and `CannotConnect` deprecated

p2p/net/swarm/swarm.go
guillaumemichel: I am not sure I addressed it correctly, can you please double check @sukunrt ?

p2p/net/swarm/swarm.go
guillaumemichel: What checks exactly are required?

IIUC it isn't necessary to delete the conn from `s.conns.m` ?

sukunrt: maintaining the conns map is okay, my comment is regarding sending the `PeerConnectednessEvent` when a connection is disconnected. 
https://github.com/libp2p/go-libp2p/blob/57736f1ec68de8471650a65ff383cb0ada0a0c4d/p2p/net/swarm/swarm.go#L777-L787
These lines should now check whether the connection we have to the peer is a Limited connection or an Unlimited connection and send a connectedness change event appropriately.

p2p/net/swarm/swarm.go
sukunrt: We need. 
```
	oldState := s.connectednessUnlocked(p)
	s.conns.m[p] = append(s.conns.m[p], c)
	newState := s.connectednessUnlocked(p)
```

This prevents us from sending a PeerConnectedness Limited event when we make a relayed connection *after* a direct connection.

p2p/net/swarm/swarm.go
```diff
@@ -748,6 +772,8 @@ func (s *Swarm) removeConn(c *Conn) {
 
 	cs := s.conns.m[p]
 
+	oldState := s.connectednessUnlocked(p)
+
 	if len(cs) == 1 {
 		delete(s.conns.m, p)
 		s.conns.Unlock()
```
guillaumemichel, line 779: Autorelay tests are failing if no event `NotConnected` is emitted when `oldState` is `NotConnected` and `newState` is `NotConnected` (no state change).

lidel:
@sukunrt is there any remaining work here, or awaiting your final review?

MarcoPolo:
I'll review this as well

p2p/net/swarm/swarm.go
```diff
@@ -646,10 +650,34 @@ func isDirectConn(c *Conn) bool {
...
+	defer s.conns.RUnlock()
+
+	return s.connectednessUnlocked(p, false)
+}
+
+// connectednessUnlocked returns the connectedness of a peer. For sending peer connectedness
+// changed notifications consider closed connections. When remote closes a connection
+// the underlying transport connection is closed first, so tracking changes to the connectedness
+// state requires considering this recently closed connections impact on Connectedness.
+func (s *Swarm) connectednessUnlocked(p peer.ID, considerClosed bool) network.Connectedness {
```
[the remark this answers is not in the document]
guillaumemichel: We may also have limited connections to the peer. So once a connection is closed, we check:
* If there is at least 1 unlimited connection, return `Connected`
* Else, if there is at least 1 limited connection, return `Limited`
* Else, return `NotConnected`

> Why are we potentially returning `Connected` in the case that every conn in the map returns `c.IsClosed()=true` with `considerClosed=true`

It is true. However, `considerClosed` is `true` only when called from `addConn` and `removeConn`, which hold the lock. `removeConn` makes sure to remove the connection from `s.conns.m`, so the function will not return `Connected` if we removed the last unlimited connection.

Though, I don't really understand why `considerClosed` is needed at all. @sukunrt could you provide some context?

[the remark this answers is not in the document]
sukunrt: I agree this is confusing. 

There are two different states this function is called from: 

When closing the connection from our side by calling Close, this function is called when the underlying transport connection is not closed. 

When closing the connection because the remote Closed the connection, this function is called when the underlying transport connection has already been closed.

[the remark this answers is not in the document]
sukunrt: Consider that we are connected to the peer over a QUIC and a relayed connection. 

Now the peer closes the QUIC connection. So we have moved from a `Connected` state to `Limited` connectivity state. 

To make this work, when calculating oldState in removeConn we need to consider connectivity assuming QUIC connection was still connected.

Alternatively we can store the last sent connectedness event for a peer. That would make this calculation cleaner at the expense of more state for an edge case.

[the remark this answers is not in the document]
MarcoPolo: > Alternatively we can store the last sent connectedness event for a peer. That would make this calculation cleaner at the expense of more state for an edge case.

I think we should do this.

MarcoPolo (approved):

sukunrt:
We have a deadlock in our event sending mechanism when the consumer closes the connection in the subscription goroutine. 

Closing the connection causes the subscription goroutine to emit an event and assuming the subscription queue is full this will deadlock. Check the test in https://github.com/libp2p/go-libp2p/pull/2696/commits/57c688ecdc6b2f2b3c5e8188bdfda24a0bd8647b

There are two options here: 
1. We disallow Close in subscription. 
2. We enqueue the events in a slice and we send the events in a separate goroutine. This slice will only take 2*NumConns space since for a given connection we will only add a connectedness event at most twice. This is also the current memory usage since the goroutines block on sending the connectedness event.  

I have implemented 2 in https://github.com/libp2p/go-libp2p/pull/2696/commits/0dcc73b81ea6ea36ad16f980736a313e47ebeaaa
I'll refactor a bit if we do decide to go ahead with 2.

p2p/net/swarm/swarm.go
```diff
@@ -745,33 +770,51 @@ func (s *Swarm) removeConn(c *Conn) {
...
+				break
+			}
+		}
+	}
+	s.conns.Unlock()
+	// Do this in a separate go routine to not block the caller.
+	// This ensures that if a event subscriber closes the connection from the subscription goroutine
+	// this doesn't deadlock
+	s.refs.Add(1)
+	go func() {
```
MarcoPolo, line 794: I would really like to avoid spawning a go routine on close

Stebalien: Yeah, we shouldn't do this. I haven't read the rest of the patch so I might be missing an obvious solution, but it _looks_ like we're just trying to "raise" a flag for this peer.

I'd just avoid channels entirely and use the pattern I used in https://github.com/ipfs/boxo/blob/2559ec2c7b93ac6905fe83d7a4b38b8e9a703b0a/bitswap/network/connecteventmanager.go#L69-L81, https://github.com/ipfs/boxo/blob/2559ec2c7b93ac6905fe83d7a4b38b8e9a703b0a/bitswap/network/connecteventmanager.go#L97-L122. It's not simple, but it works and avoids stacking goroutines.

MarcoPolo: I haven’t looked closely, but it seems like changeQueue is unbounded. Is that right?

Stebalien: Technically, but it shouldn't be an issue in practice. We only add a peer to the queue the first time the connectivity state changes. If we end up with a lot of "dead" peers in this queue, we should burn through them very quickly (transition from disconnected -> disconnected doesn't require any events).

We _can_ do better by removing peers that have transitioned back to the same state from the queue, but it's likely not worth the complexity.

sukunrt: I think a queue with this will work 

> We can do better by removing peers that have transitioned back to the same state from the queue, but it's likely not worth the complexity.


We do need the complexity because a malicious peer can connect, disconnect very quickly causing this queue to grow quickly if a consumer is slow. It can also do this on different peer ids.

Stebalien: > We do need the complexity because a malicious peer can connect, disconnect very quickly causing this queue to grow quickly if a consumer is slow. It can also do this with different peer ids.

It's _really_ unlikely to matter.

The situation is: peers quickly connect/disconnect, faster than we can process events.

1. With the unoptimized bitswap implementation, we'll end up with entries in the queue for "dead peers".
2. With an optimization, we can get rid of these entries the second the peer disconnects.

_However_, this doesn't matter in practice unless the goroutine processing this queue is _completely_ 100% blocked. If it's able to make any progress (even if it's slow), processing a string of "no-op" events is effectively free so it shouldn't fall behind.

Stebalien: Hm.... Actually, I'm not sure we can safely drop disconnect events, even if we never emitted a connect event. Services may already have observed the peer (we don't guarantee that the connect event will be delivered _before_ any inbound streams).

I think we need backpressure all the way up the stack to prevent new connections if we're backed-up with notifications.

Stebalien: Yeah, I think that, on connectivity change, we always have to emit an event for the final status _even if_ it's the same as the last status we emitted because libp2p services may have observed a connect/disconnect through some other means.

MarcoPolo: As a general rule we should have no unbounded queues in the core logic. In this case I can think of a way that a reasonable application can fall victim to a dos attack.

Example: An application opens a new stream and does an exchange on a newly connected peer. If this is done by the same go routine that reads events then the attacker controls the rate events are popped off, and thus exploit this.

Stebalien: Ideally, I'd agree. But in cases like this we can argue that the queue will hit a steady-state max size and stop growing (assuming the read side isn't deadlocked forever). It's not great, but if someone can DoS us due to this queue, they can _also_ DoS us by running us out of memory before GC can kick in.

But ideally we'd just have some form of backpressure somewhere. And I think that's what we'll need given https://github.com/libp2p/go-libp2p/pull/2696#discussion_r1588742137

sukunrt: I am trying an approach where we have backpressure on the add side and to have an unbounded queue on the remove side(close connection) because the remove side is effectively bounded by the total number of connections since on the add connection side we do have backpressure.

sukunrt: How does https://github.com/libp2p/go-libp2p/pull/2696/commits/5271dd85e9dd0a0b5f1caa4f2fbe7917ab435aed look? 

We apply backpressure when adding connections to the swarm. This allows us to enqueue remove conn events in a slice without spawning a goroutine. Back pressure on adding connections limits the slice length. We also ensure that we send *at least 1* disconnection event for every peer we connect to.

Stebalien: I don't really have time for a full review but:

1. I'd be careful to avoid _too_ many channels/goroutines. Go tends to push you in that direction, but it can quickly get harder to understand than simple locks & conditions.
2. Make sure backpressure is applied _all the way_ to accepting connections in the first place. It sounds like this might be a gap in libp2p and may be something we want to fix in a separate patch.

(I defer to @MarcoPolo on all of this, I'm just kibitzing from the sidelines)

p2p/net/swarm/swarm.go
```diff
@@ -337,7 +338,7 @@ func (s *Swarm) close() {
 	wg.Wait()
 }
 
-func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn, error) {
+func (s *Swarm) addConn(ctx context.Context, tc transport.CapableConn, dir network.Direction) (*Conn, error) {
```
MarcoPolo, line 341: Unused?

p2p/net/swarm/swarm_event_test.go
```diff
@@ -113,3 +116,197 @@ func TestNoDeadlockWhenConsumingConnectednessEvents(t *testing.T) {
...
+			cancel()
+			wg.Done()
+		}(i)
+	}
+	wg.Wait()
+	s1.Close()
+
+	close(done)
+	subWG.Wait()
+	fmt.Println("swarm closed")
```
MarcoPolo, line 311: Vestige?

p2p/net/swarm/connectedness_event_emitter.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (c *connectednessEventEmitter) AddConn(p peer.ID) {
+	c.mx.RLock()
+	defer c.mx.RUnlock()
+	if c.ctx.Err() != nil {
+		return
+	}
+
+	c.newConns <- p
```
MarcoPolo, line 58: Did you mean to take a ctx here and exit if ctx is cancelled?

MarcoPolo (commented):
Just one question, but I think this looks a lot cleaner than what was there before. Thanks!

p2p/net/swarm/connectedness_event_emitter.go
sukunrt: I tried, but this turned out very difficult. 

There isn't a nice context to use here. The dial worker loop context doesn't work too well because it expires after a minute(dial timeout) and only blocking for a minute seemed too low. 
Say we have a nice enough context: 
1. we skip sending the connectedness event on expiry, we will still would use the swarm call back notification system to send notifications. I wanted to maintain parity, if we send a callback we will send an event. 
2. we skip the notification callback, this can be done with some extra state on the connection. This extra state is needed because at this point we've added the connection to the conns map. Swarm close at this point will trigger the disconnection callback unless we store the state that we never send the connected callback for this connection. 

We can do 2 and should do 2, by timing out after a few minutes of waiting. I decided not to fix that in this PR because this is the current behavior and this doesn't cause any deadlock since the dial worker loop(which calls this method) is independent of the `DialPeer` call so consumers making dials will only wait for a minute and then will continue consuming and not deadlock.

MarcoPolo: got it. That makes sense 👍

MarcoPolo (approved):

sukunrt (approved):

PR #2698, opened 2024-02-01
basichost: don't error if identify doesn't complete within deadline

sukunrt:


Stebalien:
Why? This only happens if the context gets canceled, right?

Stebalien:
Or is the idea: "we have a connection, whatever"?

sukunrt:
>Or is the idea: "we have a connection, whatever"?

Yeah, I imagine in most cases this happens when the context times out. In which case it seems better to return the connection. 
The connection is also returned when identify fails, so it seems more consistent to return the connection when identify is just slow.

Stebalien:
Well, it's the _caller's_ context and this can also happen if:

1. The caller cancels the request.
2. The caller sets a really short timeout.

IMO, we should (and, IIRC, do) have a timeout for identify itself. If we hit that timeout, sure, we should return the connection. But not if the caller's timeout is canceled.

sukunrt:
My problem with the API returning error is that the host is now Connected to the peer despite the fact that we return an error which is very confusing. 

I'm closing this because the impact doesn't seem much especially considering we should take #2551 over the finish line.

Stebalien:
Yeah, I agree that that's annoying. But that can happen in any case so I'm not _too_ concerned.

PR #2708, opened 2024-02-16
chore: drop support for go1.20 

sukunrt:


Stebalien (approved):
From what I can tell, go 1.20 still "works", it'll just use a slower sha implementation. Which is fine, IMO.

core/crypto/ecdsa.go
```diff
@@ -10,9 +10,10 @@ import (
 	"io"
 	"math/big"
```
Jorropo, line 12:
```suggestion
```

core/crypto/rsa_go.go
```diff
@@ -8,9 +8,10 @@ import (
 	"errors"
 	"io"
```
Jorropo, line 10:
```suggestion
```

p2p/net/swarm/swarm_addr_test.go
```diff
@@ -5,11 +5,12 @@ import (
 	"fmt"
 	"testing"
```
Jorropo, line 7:
```suggestion
```

p2p/security/noise/handshake.go
```diff
@@ -10,10 +10,11 @@ import (
 	"runtime/debug"
 	"time"
```
Jorropo, line 12:
```suggestion
```

Jorropo (approved):

PR #2710, opened 2024-02-18
quic: upgrade to quic-go v0.41.0

sukunrt:


p2p/transport/webtransport/listener.go
```diff
@@ -212,5 +212,14 @@ func (l *listener) Close() error {
...
 	<-l.serverClosed
+loop:
+	for {
+		select {
+		case conn := <-l.queue:
+			conn.Close()
+		default:
+			break loop
+		}
+	}
```
guillaumemichel, line 223: why is this change required to upgrade to quic-go 0.41?

sukunrt: In v0.40 quic-go changed the behaviour of Listener.Close to not close all the existing connections that were enqueued. 
See: https://github.com/quic-go/quic-go/pull/4072

guillaumemichel (approved):

PR #2716, opened 2024-02-22
webrtc: setup datachannel handlers before connecting to a peer

sukunrt:
If done after connecting to a peer, there's a small window of time when datachannels created by the peer may cause a memory leak.


sukunrt:
Please let me know if I elaborate on anything.

PR #2717, opened 2024-02-23
webrtc: close mux when closing listener

sukunrt:
There is currently a leak in the webrtc listener. When the listener is closed the udp mux readloop just keeps running.

depends on #2716


guillaumemichel (approved):
Good catch!

PR #2719, opened 2024-02-24
Add reuseport functionality to websocket protocol (v2)

Maelkum:
This PR adds the reuseport functionality for websocket protocol.

The `TestReusePortOnDial` tests this implementation by starting to listen on a random port, then dialing another endpoint and verifying that the dial used the same port we were previously listening on.

`TestReusePortOnListen` starts two listeners and twenty dialers and (on Linux) verifies that requests were distributed among both listeners; on Windows and MacOS we can't make such guarantees, so we just verify that everything worked, and all requests were indeed handled.

This PR offers an alternative implementation and operates in the same space as https://github.com/libp2p/go-libp2p/pull/2261.
Kudos to @chaitanyaprem for his work, I have implemented his documentation references and was inspired by his test for multiple listeners and connection distribution.

The tests for websocket implementation are somewhat noisy (on master too), I don't think warnings like those below stem from these changes:
```
2024/02/24 17:11:16 http: TLS handshake error from 127.0.0.1:51918: remote error: tls: bad certificate
2024/02/24 17:11:16 http: TLS handshake error from 127.0.0.1:39826: remote error: tls: bad certificate
2024/02/24 17:11:17 http: TLS handshake error from 127.0.0.1:52002: read tcp4 127.0.0.1:38471->127.0.0.1:52002: use of closed network connection
2024/02/24 17:11:17 http: TLS handshake error from 127.0.0.1:34080: read tcp4 127.0.0.1:46259->127.0.0.1:34080: use of closed network connection
2024/02/24 17:11:17 http: TLS handshake error from 127.0.0.1:38720: remote error: tls: bad certificate
2024/02/24 17:11:18 http: TLS handshake error from 127.0.0.1:60538: remote error: tls: bad certificate
```

### DIFF
--- a/p2p/transport/websocket/reuseport.go
+++ b/p2p/transport/websocket/reuseport.go
@@ -1,0 +1,9 @@
+package websocket
+
+import (
+	"github.com/libp2p/go-reuseport"
+)
+
+func reuseportIsAvailable() bool {
+	return reuseport.Available()
+}
--- a/p2p/transport/websocket/websocket_test.go
+++ b/p2p/transport/websocket/websocket_test.go
@@ -15,6 +15,7 @@ import (
 	"math/big"
 	"net"
 	"net/http"
+	"runtime"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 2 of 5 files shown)


p2p/transport/websocket/reuseport.go
```diff
@@ -0,0 +1,9 @@
+package websocket
+
+import (
+	"github.com/libp2p/go-reuseport"
+)
+
+func reuseportIsAvailable() bool {
```
Maelkum, line 7: Right now this is just overhead - I added it for parity with the TCP implementation that includes checking of a `LIBP2P_TCP_REUSEPORT` environment variable to change the setting.

p2p/transport/websocket/websocket_test.go
```diff
@@ -549,3 +552,151 @@ func TestResolveMultiaddr(t *testing.T) {
...
+	for i, listener := range listeners {
+		for j := 0; j < connectionCount; j++ {
+			go func(index int, listener manet.Listener) {
+
+				conn, err := listener.Accept()
+				if err != nil {
+					// Stop condition - this happens when the listener is closed.
+					require.ErrorIs(t, err, transport.ErrListenerClosed)
+					return
+				}
```
Maelkum, line 643: It's possible to just start listeners above and defer `Close()` calls to it. Then when the test function completes and execution "falls off the end" the `listener.Close()` call will unblock this and `Accept()` call will fail with an error.

Still, I think I prefer that one can see what will happen during the test just by reading it. To me it's better than "something will happen after the test is done too". This is a bit ugly but explicit. I'm open to any criticism :)

MarcoPolo:

PR #2722, opened 2024-03-02
Small code improvements

AnomalRoil:
Here are a few small changes.
Happy to do 3 PRs if you want a single one per change.

The only actual code change that is reachable is that I have now interlocked the Locks in the stream `spawnControlMessageReader` function. 
This could be an issue if another goroutine were to lock the stream lock and wait on being able to lock on the reader lock before releasing the stream lock, but if that were the case, it would also mean that the previous code didn't get the guarantee no other function was able to "do reader.ReadMsg" as the comment says.
This is the reason why we should Lock the stream mutex before unlocking the reader one, IMO.

### DIFF
--- a/p2p/transport/webrtc/stream.go
+++ b/p2p/transport/webrtc/stream.go
@@ -243,16 +243,14 @@ func (s *stream) spawnControlMessageReader() {
 			s.setDataChannelReadDeadline(time.Now().Add(-1 * time.Hour))
 
 			s.readerMx.Lock()
-			// We have the lock any readers blocked on reader.ReadMsg have exited.
+			// We have the lock: any readers blocked on reader.ReadMsg have exited.
+			s.mx.Lock()
+			defer s.mx.Unlock()
 			// From this point onwards only this goroutine will do reader.ReadMsg.
-
-			//lint:ignore SA2001 we just want to ensure any exising readers have exited.
+			// We just wanted to ensure any exising readers have exited.
 			// Read calls from this point onwards will exit immediately on checking
 			// s.readState
 			s.readerMx.Unlock()
-
-			s.mx.Lock()
-			defer s.mx.Unlock()
 
 			if s.nextMessage != nil {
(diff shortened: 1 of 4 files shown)


p2p/transport/webrtc/stream.go
```diff
@@ -243,17 +243,15 @@ func (s *stream) spawnControlMessageReader() {
...
-			// We have the lock any readers blocked on reader.ReadMsg have exited.
+			// We have the lock: any readers blocked on reader.ReadMsg have exited.
+			s.mx.Lock()
+			defer s.mx.Unlock()
 			// From this point onwards only this goroutine will do reader.ReadMsg.
-
-			//lint:ignore SA2001 we just want to ensure any exising readers have exited.
+			// We just wanted to ensure any exising readers have exited.
 			// Read calls from this point onwards will exit immediately on checking
 			// s.readState
```
sukunrt, line 252: This change is not required since the rule is:

If you need both reader and stream lock. You must acquire them in order reader first then mx second. 

Any other lock acquisition order will deadlock anyway since Read calls do reader and then mx. 

In this case, we do not need both locks. However if you think this is the right thing to do for consistency with Read behaviour, I would prefer that we release the ReaderMx once the goroutine has exited. So it becomes:

```
s.readerMx.Lock()
defer s.readerMx.Unlock()

s.mx.Lock()
defer s.mx.Unlock()
```

sukunrt: My suggestion won't work. we need to release the readerMx so that future calls to Read() fail with error and not hang on s.readerMx.Lock. 

The first point stands though: the lock order *MUST* be readerMx before mx.

MarcoPolo: Does the suggested change not work? It seems reasonable to me but I haven't been in this part of the codebase for a bit.

sukunrt (approved):

sukunrt:
Thanks @AnomalRoil!

PR #2723, opened 2024-03-05
ping: use context.Afterfunc to avoid a lingering goroutine

Jorropo:


MarcoPolo:
That's a useful new function. Thanks

MarcoPolo (approved):

PR #2726, opened 2024-03-08
feat: add connection selection logic

wlynxg:
Add connection comparison logic and set more preferences to choose better connections:
1. LAN > WAN > PROXY;
2. Prefer udp based connection over tcp based;


sukunrt:
Won't both LAN and WAN have the same performance? 
I am not sure if the WAN connection between two nodes that have a LAN connection would be routed over the public internet except in some edge cases.

wlynxg:
> Won't both LAN and WAN have the same performance? I am not sure if the WAN connection between two nodes that have a LAN connection would be routed over the public internet except in some edge cases.

If the connection between two nodes with a LAN connection is via a WAN address, two situations can occur:
1. The router does not have hairpin capability and the connection cannot be successful;
2. The router has hairpin capability. The router will modify the dst ip and then send the data packet out.
All WAN IP packets need to be modified through the hairpin mode of the router, which will put a certain amount of pressure on the router.

About hairpin mode: https://datatracker.ietf.org/doc/html/rfc5128#autoid-12

p2p/net/swarm/swarm.go
```diff
@@ -602,6 +595,38 @@ func isBetterConn(a, b *Conn) bool {
...
+	switch c.ConnState().Transport {
+	case "quic", "quic-v1", "webtransport", "webrtc":
+		priority += 5
+	case "tcp", "websocket":
+		priority += 3
+	}
+
+	//  If one is transient and not the other, prefer the non-transient connection.
+	if !c.Stat().Transient {
+		priority += 2
```
Jorropo, line 624: I think that should be in `isBetterConn` and not here.
Transient connections are fundamentally different than non transient one as applications can't and don't want to use theses for transferring data.

In other words, there is no score difference that will make us want to use a transient conn over a non transient one.
Even if the transient conn is quic-v1 over localhost and the non transient one is tcp over wan.
Because consumers (`go-libp2p-kad-dht`, `boxo/bitswap`, ...) wont use the transient connection over localhost to send and receive data.

p2p/net/swarm/swarm.go
```diff
@@ -602,6 +595,38 @@ func isBetterConn(a, b *Conn) bool {
...
+		priority += 15
+	}
+
+	// We prefer udp protocols
+	switch c.ConnState().Transport {
+	case "quic", "quic-v1", "webtransport", "webrtc":
+		priority += 5
+	case "tcp", "websocket":
+		priority += 3
+	}
```
Jorropo, line 620: - `quic` is now unsupported since `quic-go` `v0.37.0`.

---

That looks somewhat fragile to me, what if I run a transport that not on there, altho this can be fixed later.
It should be more granular:
- `quic-v1` > `webtransport` (is a direct downgrade as webtransport is `quic-v1` plus other stuff) > `webrtc` (relies on hacks to function properly due to lack of proper stream close)

Jorropo (commented):
I don't really understand the point of this, when is this code called ?

As far I can tell this code is only used when switching from a transient / relayed connection to a direct one after doing hole punching. Which was broken in this PR.
As the dialer already implements similar logic but with inclusion of RTT in the mix.

p2p/net/swarm/swarm.go
wlynxg: You make a good point, transient connections should be excluded from the beginning. Thanks for the advice❤, I will make changes later.

p2p/net/swarm/swarm.go
wlynxg: Thank you for your correction. It is indeed necessary to make a more fine-grained distinction between protocols.

wlynxg:
> I don't really understand the point of this, when is this code called ?
> 
> As far I can tell this code is only used when switching from a transient / relayed connection to a direct one after doing hole punching. Which was broken in this PR. As the dialer already implements similar logic but with inclusion of RTT in the mix.

Each time NewStream is executed, swarm will find the best connection and then create the stream:
https://github.com/libp2p/go-libp2p/blob/91e1025f04519a5560361b09dfccd4b5239e36e6/p2p/net/swarm/swarm.go#L471

Jorropo:
@wlynxg we are trying to not get into a situation where we have more than one connection in the first place.

For example the dialler cancel dialling once we got a success. I've meant to link the code that does this but that harder than it needs to be because the dialler use the worker goroutine pattern.

wlynxg:
> we are trying to not get into a situation where we have more than one connection in the first place.
> 
> For example the dialler cancel dialling once we got a success. I've meant to link the code that does this but that harder than it needs to be because the dialler use the worker goroutine pattern.

The strategy I adopted when implementing p2p myself was to first sort the target addresses according to priority, and then perform traversal connections until one succeeds.

To some extent, maintaining multiple connections is not a bad thing. It can quickly switch to a connection with another address when one connection is unavailable. In the implementation of tailscale, it also adopts a multi-link strategy, but it also sets up a dedicated goroutine to regularly perform RTT speed testing on each connection, and then selects the best connection.

p2p/net/swarm/swarm.go
```diff
@@ -602,6 +603,38 @@ func isBetterConn(a, b *Conn) bool {
...
+
+	//  If one is transient and not the other, prefer the non-transient connection.
+	if !c.Stat().Transient {
+		priority += 2
+	}
+
```
Jorropo, line 634:
```suggestion
```

This now do nothing because both connections have identical transience.

wlynxg: Sorry, I must be an idiot, I forgot to delete this part of the code...

Jorropo: <3 you're not

wlynxg:
> @wlynxg I assumed too much mb. Your idea is good, multiple connections have cases where they are great indeed as you explained.
> 
> The problem I have with this code is that it's not how go-libp2p operates, your current code looks correct now (as non Transient connections have priority). But it also doesn't do anything in 99% of the cases, you are fixing a bug that does not yet exists because we don't even try to have multiple connections right now. Do you have more plans to change other things in future pull requests ?
> 
> On a personal note, I do want multiple connections but not at the libp2p layer because they leak to applications, generate stream resets and require applications. Instead I think this makes more sense to use QUIC Connection Migrations and MPQUIC to do that on a per packet level in the QUIC transport.
> 
> There is [libp2p/specs#328](https://github.com/libp2p/specs/issues/328) on the subject but I believe work there is halted.

I think the best performance of multi connections is like this:
Let's perform another upper-level abstraction on the stream. The stream is not bound to a fixed connection. It can freely choose what conn is used at the bottom for transmission.
However, this requires fragmentation of data packets for reliable transmission, which will increase a lot of workload😥...

Jorropo:
> However, this requires fragmentation of data packets for reliable transmission, which will increase a lot of workload

***https://github.com/libp2p/specs/issues/328 https://github.com/libp2p/specs/pull/406*** exists for theses purposes

PR #2729, opened 2024-03-08
webrtc: run onDone callback immediately on close

sukunrt:


guillaumemichel (approved):

sukunrt:
We modified the specs to do a synchronous datachannel close after finding out that chrome implementation of datachannels drop enqueued messages if the datachannel is closed. see https://github.com/libp2p/specs/issues/575#issuecomment-1717438840

When implementing this spec change, I implemented this such that the stream resource manager scope is closed only after reading the FIN_ACK. This required a rather ugly hack in swarm_stream by introducing a asyncCloser interface. See this comment: https://github.com/libp2p/go-libp2p/pull/2615#discussion_r1484693365

After implementing this, I realized that the datachannel close procedure itself was a synchronous close where SCTP layer doesn't drop references to the stream till it get a data channel message from the other end. So calling onDone in a separate go routine immediately after data channel close didn't help anything since the reference to the data channel would be kept for 1RTT. I removed the `AsyncClose` bit for the same reason and decided to close the stream scope early. I should have done this when I changed #2615. See this comment: https://github.com/libp2p/go-libp2p/pull/2615#issuecomment-1941655375

MarcoPolo:
Does this mean that the resource manager is off from the true resource usage? For example, if a client closes a stream and the on close is called immediately, the resource manager frees that stream, but that memory is actually still reserved by pion until we finish the close. Is that correct?

sukunrt:
Yes, it'll be freed only on datachannel close which happens 1RTT after we call `datachannel.Close`. If we want to keep the resource manager usage in sync we have to do it after datachannel close and not after we receive FIN_ACK from peer.

MarcoPolo:
Is it significantly harder to call the onDone callback after we close the datachannel? (and thus free the resource token in the resource manager) That seems to me the most correct solution because it protects against a hypothetical memory exhaustion attack where a server can have a client think it has freed its resource usage to a stream but actually still have that memory reserved. Unless I'm misunderstanding something and we get this protection some other way?

MarcoPolo (approved):

Issue #2614, opened 2023-10-20
webrtc: potential problem with the Pion state machine


marten-seemann:
We might be using the Pion state machine incorrectly:
```
=== RUN   TestTransportWebRTC_Deadline/SetWriteDeadline
      transport_test.go:401: 
          	Error Trace:	/home/runner/work/go-libp2p/go-libp2p/p2p/transport/webrtc/transport_test.go:401
          	Error:      	Received unexpected error:
          	            	failed to secure outbound connection: error reading handshake message: sending reset packet in non-established state: state=Closed
          	Test:       	TestTransportWebRTC_Deadline/SetWriteDeadline
  --- FAIL: TestTransportWebRTC_Deadline (20.14s)
      --- PASS: TestTransportWebRTC_Deadline/SetReadDeadline (0.13s)
      --- FAIL: TestTransportWebRTC_Deadline/SetWriteDeadline (20.00s)
```

sukunrt:
There's one other bug which causes failure in connection establishment:
```
2023-10-22T06:47:19.9957753Z === RUN   TestTransportWebRTC_DialerCanCreateStreamsMultiple
2023-10-22T06:47:19.9959034Z panic: test timed out after 10m0s
2023-10-22T06:47:19.9959911Z running tests:
2023-10-22T06:47:19.9960992Z 	TestTransportWebRTC_DialerCanCreateStreamsMultiple (9m58s)
```
https://github.com/libp2p/go-libp2p/actions/runs/6602211220/job/17934010538

sukunrt:
Both these errors haven't happened since merging #2717 

The first errors was fixed by fixing the tests. When one side closes the connection the other side sees this line when it does a stream Close.
>sending reset packet in non-established state: state=Closed

MarcoPolo:
> Both these errors haven't happened since merging #2717
> 
> The first errors was fixed by fixing the tests. When one side closes the connection the other side sees this line when it does a stream Close.
> 
> > sending reset packet in non-established state: state=Closed

How was the other error fixed?

sukunrt:
I am not convinced it is fixed completely. We just find it flaking less for some reason. 

One of the reasons the other error occurs is this https://github.com/libp2p/go-libp2p/actions/runs/8125238982/job/22214622459?pr=2722

This will be fixed when pion/sctp does a new release. This was fixed by: https://github.com/pion/sctp/pull/301

[...]

PR #2732, opened 2024-03-12
webrtc: test pion fixes for state change callbacks ordering

sukunrt:
Fixes: #2614 
This is a problem with pion/webrtc's OnConnectionStateChange handler. It should wait for the handler to finish. Right now connection notifications might get reordered leading to listener thinking the connection has established and the dialer thinks connection hasn't established yet.

The fix in pion/webrtc is here:
https://github.com/pion/webrtc/pull/2702/files

The same error in pion/ice
https://github.com/pion/ice/pull/656/files
 

Update: the text below helped me explains how I went about debugging this
~I think this fixes the problem~ This might help debug the problem of tests taking 10m. My theory is that the ICE connection succeeds on dialer and times out on listener(20s). So the listener removes the ufrag from its udp mux. After keepalive time, the dialer sends a STUN Request and initiates a new connection on the listener. This of course will fail because the dialer has exited the ice connection establishment mode and is simply sending STUN packets to keep the connection alive. 

If we look at the failure here: https://github.com/libp2p/go-libp2p/actions/runs/8244362080/job/22546464639

listener is stuck on 
2024-03-12T06:24:47.2310488Z 	/home/runner/work/go-libp2p/go-libp2p/p2p/transport/webrtc/listener.go:234 +0x886
```
	select {
	case <-ctx.Done():
		return nil, ctx.Err()
	case err := <-errC:
		if err != nil {
			return nil, fmt.Errorf("peer connection failed for ufrag: %s", candidate.Ufrag)
		}
```
This context expires after 20 seconds. 
And all the goroutines on the dialer side are stuck for 9 minutes. And all the goroutines on the listener side are stuck for < minute(which is why that's not shown in the logs)
The logs also show that the dialer believes that ice connection on its side has completed.

Unfortunately this is very difficult to test. This has not flaked in the 16 tries I have done after that. 

We still have to explain why the dialer doesn't exit on DTLS failure. DTLS failure should reflect in the OnConnectionStateChange callback.


p2p/transport/webrtc/transport.go
```diff
@@ -78,7 +78,7 @@ const (
 const (
 	DefaultDisconnectedTimeout = 20 * time.Second
 	DefaultFailedTimeout       = 30 * time.Second
-	DefaultKeepaliveTimeout    = 15 * time.Second
+	DefaultKeepaliveTimeout    = 5 * time.Second
 )
```
sukunrt, line 82: I think it's better to have 4 keep alives before timing out than just 1 as before.

sukunrt:
the webrtc fixes have landed and will be included in v0.34

PR #2733, opened 2024-03-13
p2p/http: disable transfer encoding on stream listener

Jorropo:
cc @MarcoPolo you could always do this

Note: this doesn't prevent stream reusing when `Content-Length` is set, because `Content-Length` is useful outside of message framing (giving download progress bars, bounding range requests, preallocating a destination file, ...).

### DIFF
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -304,7 +304,7 @@ func (h *Host) Serve() error {
 		h.httpTransport.listenAddrs = append(h.httpTransport.listenAddrs, h.StreamHost.Addrs()...)
 
 		go func() {
-			errCh <- http.Serve(listener, h.ServeMux)
+			errCh <- http.Serve(listener, doNotDefaultToTransferEncodingHandler{h.ServeMux})
 		}()
 	}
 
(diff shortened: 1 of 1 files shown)


MarcoPolo:
I don't understand what this is for. Can you please explain why this is desired? Isn't Transfer-Encoding already nil by default?

Jorropo:
By default things are not in the map, this adds a nil value inside the map.
This means the two form map access will return `nil, true` instead of `nil, false`.

Here is the relevant docs:
https://pkg.go.dev/net/http#ResponseWriter.Header
> ```go
>  // To suppress automatic response headers (such as "Date"), set
> // their value to nil.
> ```

This PR proves you don't gain anything by disallowing pipelining, as this is already an optional part of HTTP1.1 for servers.

Assuming the client is never gonna reuse streams, this is more efficient.
For message framing go will try in order:
- `Content-Length`
- `Transfer-Encoding`
- Closing the underlying stream

`Content-Length` is only usable when the length is known, either because the response is short enough it is sent within one buffer action, or because the application set `Content-Length` itself in the first place (for example `http.ServeFile` will use `.Seek` to `SeekEnd, 0` to get the last index of the file).
Then when you are transmitting a stream of data of unknown size (example streaming cars over trustless gateway), go will use `Transfer-Encoding: chunked` which require extra buffering steps as the data is written in chunks. This disable that.
Instead when transferring an unknown sized stream the http server will write the stream as-is and close the underlying libp2p stream to indicate the end of the response.

MarcoPolo:
I don't think this does what you think it does. Here's an example:

Server:
```go
package main

import (
	"net/http"
)

func main() {
	http.ListenAndServe(":8080", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header()["Transfer-Encoding"] = nil

		a := [1_000_000]byte{}
		w.Write(a[:])
	}))

}
```

Client:
```
curl -vvv localhost:8080 --output /dev/null                        
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.6.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Thu, 14 Mar 2024 21:47:08 GMT
< Content-Type: application/octet-stream
< Transfer-Encoding: chunked
< 
```

I think you'd have to set transfer encoding to "identity".

server:
```go
-		w.Header()["Transfer-Encoding"] = nil
+ 		w.Header()["Transfer-Encoding"] = []string{"identity"}
```

client:
```
❯ curl -vvv localhost:8080 --output /dev/null
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET / HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.6.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Thu, 14 Mar 2024 21:44:14 GMT
< Connection: close
< 
```

Is the goal with this PR to have the server not chunk responses when sending them over libp2p streams? and to have the server close the stream after sending the response? If so, I'm in favor of that idea.

Looking at server code in Go std lib: https://github.com/golang/go/blob/412623c53fbb1b3ff29035022ef70374db35e826/src/net/http/server.go#L1479

We can do this by setting the Transfer-Encoding = "identity". But now that I'm here, I'm now thinking we should set the Connection = Close header as well (https://datatracker.ietf.org/doc/html/rfc2616#section-14.10).

Jorropo:
@MarcoPolo yes mb, I miss-remembered what I did last time I disabled it.

MarcoPolo:
Closing in favor of #2756

PR #2736, opened 2024-03-15
chore: Added missing explanation for 'basicEquals' function

JonyBepary:
Maybe not a big of a deal but again calling this will show a nice little message for developer.


MarcoPolo:
Thanks for the effort, but this is a non-exported function and straightforward. I think in this case it's not a worthwhile change.

PR #2749, opened 2024-03-21
autonat: Clean up after close

### LINKED ISSUE #2743: AutoNAT fails to close its peerstore
The AutoNAT service creates a peerstore but fails to close it, leaking goroutines.

MarcoPolo:
Autonat would leave a couple of dangling go routines because it wouldn't close the swarm or the peerstore. Fixes #2743.

### DIFF
--- a/p2p/host/autonat/svc.go
+++ b/p2p/host/autonat/svc.go
@@ -273,6 +273,11 @@ func (as *autoNATService) Disable() {
 	}
 }
 
+func (as *autoNATService) Close() error {
+	as.Disable()
+	return as.config.dialer.Close()
+}
+
 func (as *autoNATService) background(ctx context.Context) {
 	defer close(as.backgroundRunning)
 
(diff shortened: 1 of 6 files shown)


p2p/host/autonat/svc.go
```diff
@@ -273,6 +273,11 @@ func (as *autoNATService) Disable() {
 	}
 }
 
+func (as *autoNATService) Close() error {
+	as.Disable()
+	return as.config.dialer.Close()
+}
+
```
sukunrt, line 280: Should we just close the dialer from the autonat instance? That seems better, we ensure that the dialer is always closed even if for some reason there's an issue with service being set to nil for some reason. 
I don't have strong opinions here.

sukunrt (approved):

p2p/host/autonat/svc.go
MarcoPolo: My read here is that the dialer is owned by the service, so it's the service's responsibility to close it.

PR #2750, opened 2024-03-22
feat: add tls `WithKeyLogWriter` option

wlynxg:
`KeyLogWriter` can help us use Wireshark to analyze TLS decrypted network packets.
Details can be found at: https://wiki.wireshark.org/TLS#using-the-pre-master-secret


MarcoPolo:
Nice! I usually hack this in when I need it. I think it makes sense to expose this for easier debugging.

MarcoPolo (approved):

PR #2751, opened 2024-03-24
quic: make server cmd use RFC 9000 instead of draft-29

leonrinkel:
Support for QUIC draft-29 has been dropped in v0.30.0, however the transport server cmd has still been using draft-29 and thus fails with error `unknown QUIC version` since that release.

This PR makes `p2p/transport/quic/cmd/server/main.go` use RFC 9000, i.e. `/quic-v1` instead of `/quic`, so that the standalone client/server test commands work again.

### DIFF
--- a/p2p/transport/quic/cmd/server/main.go
+++ b/p2p/transport/quic/cmd/server/main.go
@@ -28,7 +28,7 @@ func main() {
 }
 
 func run(port string) error {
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Thanks, I didn't even know we had this. We should have an [Example test](https://go.dev/blog/examples) for this. I'll add one soon unless you beat me to it.

MarcoPolo:
Closing this in favor of #2753 so that I can push the change. You will still have the original commit associated with you @leonrinkel

PR #2753, opened 2024-03-25
quic: make server cmd use RFC 9000 instead of draft-29 

MarcoPolo:
Continues from https://github.com/libp2p/go-libp2p/pull/2751. I was not able to push to that branch.

- Does the same as #2751 
- Moves run code to a lib package and runs a test on it.

Closes #2751

### DIFF
--- a/p2p/http/example_test.go
+++ b/p2p/http/example_test.go
@@ -57,7 +57,7 @@ func ExampleHost_withAStockGoHTTPClient() {
 }
 
 func ExampleHost_listenOnHTTPTransportAndStreams() {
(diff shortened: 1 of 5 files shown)


MarcoPolo:
I still need to fix the Windows test which does not let you dial on `0.0.0.0`:
```
2024-03-28T03:35:37.9745956Z     lib_test.go:16: INTERNAL_ERROR (local): write udp4 0.0.0.0:63731->0.0.0.0:63730: wsasendto: The requested address is not valid in its context.
```

sukunrt (approved):

PR #2757, opened 2024-04-03
libp2phttp: Rename well-known resource

MarcoPolo:
Renaming the well-known libp2p http resource. Context here: https://github.com/libp2p/specs/pull/508#discussion_r1550043195

### DIFF
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -30,6 +30,7 @@ import (
 var log = logging.Logger("libp2phttp")
 
 const ProtocolIDForMultistreamSelect = "/http/1.1"
+const WellKnownProtocols = "/.well-known/libp2p/protocols"
 const peerMetadataLimit = 8 << 10 // 8KB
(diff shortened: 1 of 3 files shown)


sukunrt (approved):

p2p/http/libp2phttp.go
```diff
@@ -30,6 +30,7 @@ import (
 var log = logging.Logger("libp2phttp")
 
 const ProtocolIDForMultistreamSelect = "/http/1.1"
+const WellKnownProtocols = "/.well-known/libp2p/protocols"
```
gammazero, line 33: @MarcoPolo Changing this from `"/.well-known/libp2p"` to `"/.well-known/libp2p/protocols"` will break the ability to communicate with existing servers deployed without upgrading them first. Since these servers are running in places not under our control, the will no be immediately upgradable.

This means that libp2phttp needs to support the both the old and the new value, or revert to using the old value.

Suggest:
```go
const WellKnownProtocols = "/.well-known/libp2p"
```

Issue #2355, opened 2023-06-12
Proposal: a radically simpler Peerstore


marten-seemann:
Observation: The current `Peerstore` does a lot of things, but it doesn’t any of them well. 

Problems:

- It’s a giant interface (32 methods in total (!!!), plus 2 for the `CertifiedAddressBook`).
- Entries are GC’ed when the peer disconnects.
- It’s trying to offer applications a range of options to store peer-related data, without offering options to persist this data beyond the time a peer disconnects.
- It doesn’t keep track of any metrics (e.g. regarding peer addresses stored).
- `SupportsProtocol` doesn’t distinguish between Identify having completed and not, leading to complexity in applications ([example](https://github.com/libp2p/go-libp2p/blob/1b0df37769a6c5ffa0e2db0d1cd3d59cb443e65b/p2p/host/autorelay/relay_finder.go#L475-L488)).

For example, this shows up:

- In QUIC / TLS: it can’t be used to store session tickets for session resumption / 0-RTT.
- In Kademlia: it can’t be used to store peer / provider records.

This is a proposal to radically shrink the peerstore, such that it offers *only* what is needed to dial connections. Applications running on top of QUIC, and even transports like QUIC that want to use 0-RTT, will need to come up with their own data structures to store peer-related data. This is inevitable, since different applications will have different needs regarding persisting data when a peer disconnects, and different heuristics for limiting the size of these data structures.

In particular, we’ll get rid of the following components:

1. `PeerMetadata`: Only used to save the agent / protocol version sent in Identify. There’s no real reason to save that information in the first place.
2. `Metrics`: Only used to store RTT statistics. Arguably, that’s not a per-peer, but a per-connection (or rather, per path, if dealing with multipath transports) property. We should consider exposing the RTT on the `Connection` interface.
3. `ProtoBook`: Used to store the protocols supported by this peer (as learned via Identify). Takes up huge amounts of storage, for little benefit. Protocols that need a list of peers supporting their protocol can easily subscribe to the respective `EventBus` notification. [^1]

The `KeyBook` will be radically simplified to only allow storing of the public key.

As a result, this leaves us with a peer store that only stores:

1. addresses
2. public keys
3. certified peer records (in their serialized form)

It basically becomes an address book, and maybe we should honor that fact by renaming it accordingly.

The following changes will make it a lot more smarter and more useful:

[...]

Issue #2693, opened 2024-01-26
Per-connection protocol list


Stebalien:
Currently, identify operates at a _peer_ level. However, this doesn't match up with reality. In practice, there are some protocols we only want to speak over connections with specific properties. Mostly with respect to transient connections, but we may end up having, e.g., network, interface, or even IP-based restrictions.

To solve this, we should:

1. Run identify on all connections (we already do this).
2. Associate the protocol list with the connection itself (see https://github.com/libp2p/go-libp2p/issues/2355).
3. When creating a new stream, find the best connection that supports the given protocol.
4. Whenever a connection is created/destroyed, emit a "protocol change" event describing the connection-specific protocols lost/gained.
5. Provide a way to set a protocol handler with some set of constraints. Ideally an arbitrary function that's called on each new connection (during identify) and on a stream's connection before the handler is invoked. That shouldn't be that slow and it's _really_ flexible.

Why? Bitswap, for example only wants non-transient connections and it doesn't even want to hear about peers with transient-only connections. Honestly, this applies to pretty much all of our protocols except the DHT and hole-punching protocols.

MarcoPolo:
Is there any other property besides limited (née transient) connections that we care about today? 

We currently do a [happy eyeballs type](https://github.com/libp2p/go-libp2p/pull/2260) thing for dialing a peer given a list of addresses, so for most cases we'll end up with a single connection.

Perhaps unrelated, but it might make sense to have peers be able to exchange which protocols they support over a limited connection.

Stebalien:
Nothing else that I can think of today. The only other reason to do this right now is to remove the protocol list from the peer store. See number two above. But this probably isn't a priority.

PR #2759, opened 2024-04-05
Identify: emit useful events after identification

MarcoPolo:
Identify has an event that it emits, but it's not very useful in that it will only tell you a peer has been identified, but not tell you anything about that peer. This design is intended to have the user then lookup peer information somewhere else after they get the event. However, that approach is brittle because it relies on identify itself probing in and updating all the "somewhere elses". This issue is highlight in #2754 and #2355.

Instead, Identify should only concern itself with emitting an event about what it has learned. Other services can subscribe to those events and do their appropriate logic. This helps decouple these services from each other.

Relates to #2754 and #2355 

This is a partial fix for #2754. The next step after this is merged is to update GossipSub to make use of these events.

### DIFF
--- a/core/event/identify.go
+++ b/core/event/identify.go
@@ -1,11 +1,29 @@
 package event
 
-import "github.com/libp2p/go-libp2p/core/peer"
+import (
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/protocol"
+	"github.com/libp2p/go-libp2p/core/record"
+	"github.com/multiformats/go-multiaddr"
+)
 
 // EvtPeerIdentificationCompleted is emitted when the initial identification round for a peer is completed.
 type EvtPeerIdentificationCompleted struct {
 	// Peer is the ID of the peer whose identification succeeded.
 	Peer peer.ID
+
+	// Conn is the connection we identified.
+	Conn network.Conn
+
+	// ListenAddrs is the list of addresses the peer is listening on.
+	ListenAddrs []multiaddr.Multiaddr
+
+	// Protocols is the list of protocols the peer advertised on this connection.
+	Protocols []protocol.ID
+
+	// SignedPeerRecord is the provided signed peer record of the peer. May be nil.
+	SignedPeerRecord *record.Envelope
 }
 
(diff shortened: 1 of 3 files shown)


core/event/identify.go
```diff
@@ -1,11 +1,29 @@
...
+	"github.com/multiformats/go-multiaddr"
+)
 
 // EvtPeerIdentificationCompleted is emitted when the initial identification round for a peer is completed.
 type EvtPeerIdentificationCompleted struct {
 	// Peer is the ID of the peer whose identification succeeded.
 	Peer peer.ID
+
+	// Conn is the connection we identified.
+	Conn network.Conn
```
Stebalien, line 17: It makes sense to have this right now, but I'm concerned about the future if/when we have some sort of connection-independent identify. But I guess we have no plans there.

core/event/identify.go
```diff
@@ -1,11 +1,29 @@
...
+
+	// ListenAddrs is the list of addresses the peer is listening on.
+	ListenAddrs []multiaddr.Multiaddr
+
+	// Protocols is the list of protocols the peer advertised on this connection.
+	Protocols []protocol.ID
```
Stebalien, line 23: All addrs/protocols, or only the ones the peer told us about?

core/event/identify.go
sukunrt: Don't we want to double down on this connection dependent identify approach? https://github.com/libp2p/go-libp2p/issues/2693

core/event/identify.go
```diff
@@ -1,11 +1,29 @@
...
+
+	// SignedPeerRecord is the provided signed peer record of the peer. May be nil.
+	SignedPeerRecord *record.Envelope
```
sukunrt, line 26: Should we include the other fields?
AgentVersion, ProtocolVersion and as we have the connection, observed addrs might be useful too?

core/event/identify.go
Stebalien: In general, yes. I'm just calling this out as a decision.

core/event/identify.go
MarcoPolo: Only the one they told us about on this connection

MarcoPolo: (what it says in the comment)

core/event/identify.go
MarcoPolo: agreed. Thanks

sukunrt (approved):

PR #2760, opened 2024-04-07
Update and fix examples

apenzk:
This PR fixes some readmes and examples in the example section 

- libp2p readme example does not work in its current form, which is fixed in this PR
- routed echo example missing in the main readme
- inconsistent style in the examples. Sometimes the main function is in the beginning sometimes at the end. This PR moves the main function in chat.go and proxy.go at the start of the file in line with the other examples.
- in chat.go the handleStream function is not applied consistently. This is fixed.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -9,6 +9,7 @@ Let us know if you find any issue or if you want to contribute and add a new tut
 - [The libp2p 'host'](./libp2p-host)
 - [Building an http proxy with libp2p](./http-proxy)
 - [An echo host](./echo)
+- [Routed echo host](./routed-echo/)
 - [Multicodecs with protobufs](./multipro)
 - [P2P chat application](./chat)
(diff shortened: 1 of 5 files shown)


sukunrt:
Thanks for the PR @apenzk. There are too many different changes here, can you break this in to smaller PRs?
 1. libp2p readme example fix
 2. fix docs to include routed echo example
 3. fix consistency with handleStream(not sure if this is needed)
 
There's no reason to move the main function around. It'd be a bit nicer if there were consistency in where the main function is in files, it's infeasible to maintain this consistency without automated tooling.

apenzk:
@sukunrt done as requested. 
- 1: fix libp2p-host example: https://github.com/libp2p/go-libp2p/pull/2766
- 2: update readmes: https://github.com/libp2p/go-libp2p/pull/2767

- 3: has now been abondened as the current state allows for better testing and i would need to apply more changes to improve the consistency

PR #2766, opened 2024-04-12
fix code snippets to achieve working libp2p-host example

apenzk:
This PR updates the readme of `examples/libp2p-host/`.

In the current form the code snippets of the readme do not work and lead to errors. The code snippets now align with the example code.


MarcoPolo:
Is there value in having this code duplicated in multiple places? README.md and host.go in the example? Did you find it helpful when you read through this?

apenzk:
For consistency I  would argue against duplication. As for exactly this reason errors were present. Also the readme code is not tested. 

In contrast, personally I learnt something from it that I would otherwise not. So for me it was valuable.

PR #2774, opened 2024-04-18
basichost: append certhash for webrtc addresses provided via address factory

sukunrt:
This change is similar to how we append certhashes for webtransport.


MarcoPolo (approved):
One small fix, but looks good.

Happy to see that this change was relatively straightforward :)

PR #2788, opened 2024-05-13
rcmgr: Add conn_limiter to limit number of conns per ip cidr

MarcoPolo:
This change makes the resource manager aware of IPs. Adds some reasonable defaults too.

### DIFF
--- a/p2p/host/resource-manager/conn_limiter.go
+++ b/p2p/host/resource-manager/conn_limiter.go
@@ -1,0 +1,138 @@
+package rcmgr
+
+import (
+	"net/netip"
+	"sync"
+)
+
+type ConnLimitPerCIDR struct {
+	// How many leading 1 bits in the mask
+	BitMask   int
+	ConnCount int
+}
+
+// 8 for now so that it matches the number of concurrent dials we may do
+// in swarm_dial.go. With future smart dialing work we should bring this
+// down
+var defaultMaxConcurrentConns = 8
+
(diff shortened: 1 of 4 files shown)


p2p/host/resource-manager/conn_limiter.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil
+	}
+}
+
+type connLimiter struct {
+	mu                  sync.Mutex
+	connLimitPerCIDRIP4 []ConnLimitPerCIDR
+	connLimitPerCIDRIP6 []ConnLimitPerCIDR
+	ip4connsPerLimit    []map[string]int
+	ip6connsPerLimit    []map[string]int
```
marten-seemann, line 51: Why is this a slice? How are these maps ever garbage collected?

marten-seemann (commented):
Just had a quick glance so far.

p2p/host/resource-manager/conn_limiter.go
MarcoPolo: It’s a slice because the limits are a slice. Each map corresponds to each limit. 

You’re right about the gc of course, I need to delete after removing all conns from a subnet.

sukunrt (approved):
Should we add an exception for no limits on localhost so that tests don't break?

MarcoPolo:
> Should we add an exception for no limits on localhost so that tests don't break?

Are there any tests that currently break?

sukunrt:
I think they're fine.

PR #2790, opened 2024-05-14
upgrade checkout action to version with node20

cario-dev:
Node16 is deprecated and Actions using it will stop being used 3rd of June
https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/


MarcoPolo:
Thanks for this. I'll take a look once we cut the next release.

MarcoPolo (approved):

PR #2793, opened 2024-05-15
identify: refactor observed address manager to do address mapping at thin waist(IP+TCP/UDP) layer

sukunrt:
Provides an option to disable observed address manager for users who know their external IP addresses.

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"slices"
 	"sync"
 	"time"
 
@@ -52,6 +53,8 @@ var (
 	// DefaultAddrsFactory is the default value for HostOpts.AddrsFactory.
 	DefaultAddrsFactory = func(addrs []ma.Multiaddr) []ma.Multiaddr { return addrs }
 )
+
+const maxPeerRecordSize = 4096 // 4k to be compatible with rust-libp2p identify implementation
 
 // AddrsFactory functions can be passed to New in order to override
 // addresses returned by Addrs.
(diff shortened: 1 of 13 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -53,6 +54,8 @@ var (
 	DefaultAddrsFactory = func(addrs []ma.Multiaddr) []ma.Multiaddr { return addrs }
 )
 
+const maxPeerRecordSize = 4096 // 4k to be compatible with rust-libp2p identify implementation
```
MarcoPolo, line 57: I don't think we need this because the rust-libp2p identify implementation isn't even aware of signed peer records: https://github.com/libp2p/rust-libp2p/blob/master/protocols/identify/src/generated/structs.proto#L1

p2p/protocol/identify/id.go
```diff
@@ -34,24 +34,27 @@ import (
...
-
-const (
-	legacyIDSize = 2 * 1024 // 2k Bytes
-	signedIDSize = 8 * 1024 // 8K
-	maxMessages  = 10
+	ServiceName = "libp2p.identify"
+
+	legacyIDSize          = 2 * 1024
+	signedIDSize          = 8 * 1024
+	maxOwnIdentifyMsgSize = 4096 // smaller than what we accept. This is 4k to be compatible with rust-libp2p
```
MarcoPolo, line 50: This should be the max size of the first message for rust-libp2p compatibility. The signed peer records may go in the next message and can occupy the whole 8k.

p2p/protocol/identify/opts.go
```diff
@@ -38,3 +39,11 @@ func WithMetricsTracer(tr MetricsTracer) Option {
 		cfg.metricsTracer = tr
 	}
 }
+
+// DisableObservedAddrManager disables the observed address manager. It also
+// effectively disables the nat emitter and EvtNATDeviceTypeChanged
```
MarcoPolo, line 44: Whoops! thanks for the correction

p2p/host/basic/basic_host.go
MarcoPolo: I've changed it to 8k instead.

p2p/protocol/identify/id.go
MarcoPolo: It already is, nvm :)

MarcoPolo (approved):

ivan4th:
This PR broke disconnect handling in `identify`, fix: #2825

PR #2794, opened 2024-05-16
fix: Set missing deadlines

MarcoPolo:
A couple of places were missing deadlines so I added them.

### DIFF
--- a/p2p/net/gostream/listener.go
+++ b/p2p/net/gostream/listener.go
@@ -3,11 +3,14 @@ package gostream
 import (
 	"context"
 	"net"
+	"time"
 
 	"github.com/libp2p/go-libp2p/core/host"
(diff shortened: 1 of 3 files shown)


p2p/net/gostream/listener.go
```diff
@@ -60,6 +63,7 @@ func Listen(h host.Host, tag protocol.ID) (net.Listener, error) {
 	}
 
 	h.SetStreamHandler(tag, func(s network.Stream) {
+		s.SetDeadline(time.Now().Add(ServerTimeout))
```
marten-seemann, line 66: This seems like the wrong place to set the deadline. The consumer might want to use the stream for a long time.

MarcoPolo: You're right. This `Conn` this transforms to should have the deadline set by the user.

PR #2797, opened 2024-05-16
Use previous well-known resource

gammazero:
Changing from "/.well-known/libp2p" to "/.well-known/libp2p/protocols" will break the ability to communicate with existing servers deployed without upgrading them first. Since these servers are running in places not under our control, they will not be immediately upgradable. This means that libp2phttp needs to support both the old and the new values, or revert to using the old value.

This PR reverts to using the old resource name, which was changed in #2757.

### DIFF
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -30,7 +30,7 @@ import (
 var log = logging.Logger("libp2phttp")
 
 const ProtocolIDForMultistreamSelect = "/http/1.1"
-const WellKnownProtocols = "/.well-known/libp2p/protocols"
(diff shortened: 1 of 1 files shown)


MarcoPolo:
What do you think about supporting both versions for a release for a bit, and then removing the old version?

gammazero:
> What do you think about supporting both versions for a release for a bit, and then removing the old version?

I think it will be one of those things we will need to support for a very long time or forever. Hundreds of storage providers are using the old resource name, and it is very difficult to get SPs to upgrade.

MarcoPolo:
Closing in favor of #2798

PR #2798, opened 2024-05-17
Add backwards compatibility option to libp2phttp for the well-known resource

MarcoPolo:
We've bike shedded on the well-known resource so long that we have a number of users that are relying on the old well-known path being at `.well-known/libp2p`. This patch gives them a migration path. Enabling this option allows clients to request both the new and old path in parallel (so it doesn't cost one whole round trip). On the server side, it serves both well-known resources so that legacy clients don't break.

This change is meant to give our early adopters of libp2phttp a migration path. The goal is to get rid of this code after some time (2 releases? 3?) to allow users to upgrade.

### DIFF
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -4,6 +4,7 @@ package libp2phttp
 
 import (
 	"bufio"
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"errors"
@@ -15,6 +16,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	lru "github.com/hashicorp/golang-lru/v2"
 	logging "github.com/ipfs/go-log/v2"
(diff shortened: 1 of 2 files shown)


gammazero (approved):

p2p/http/libp2phttp.go
```diff
@@ -29,8 +31,16 @@ import (
...
 
+var WellKnownRequestTimeout = 30 * time.Second
+
 const ProtocolIDForMultistreamSelect = "/http/1.1"
 const WellKnownProtocols = "/.well-known/libp2p/protocols"
+
+// LegacyWellKnownProtocols refer to a the well-known resource used in an early
+// draft of the libp2p+http spec. Some users have deployed this, and need backwards compatibility.
+// Hopefully we can phase this out in the future. Context: https://github.com/libp2p/go-libp2p/pull/2797
+const LegacyWellKnownProtocols = "/.well-known/libp2p"
```
sukunrt, line 42: Do we need to export this?

sukunrt (approved):

p2p/http/libp2phttp.go
MarcoPolo: I think so. For similar reasons as we export the other one. Also when we remove compatibility support we’ll remove this, which could help folks realize this is gone.

PR #2801, opened 2024-05-17
webrtc: add a stress test for many connections

sukunrt:


p2p/transport/webrtc/listener.go
```diff
@@ -329,8 +329,8 @@ func (l *listener) Multiaddr() ma.Multiaddr {
 func addOnConnectionStateChangeCallback(pc *webrtc.PeerConnection) <-chan error {
 	errC := make(chan error, 1)
 	var once sync.Once
-	pc.OnConnectionStateChange(func(state webrtc.PeerConnectionState) {
-		switch state {
+	pc.OnConnectionStateChange(func(_ webrtc.PeerConnectionState) {
+		switch pc.ConnectionState() {
```
sukunrt, line 333: This works fine, but the change makes it better in case there are ordering issues for the callback. The callback does need to guarantee that we can access connection state.

p2p/transport/webrtc/transport_test.go
```diff
@@ -860,3 +861,108 @@ func TestMaxInFlightRequests(t *testing.T) {
...
+			return
+		}
+		var b [4]byte
+		if _, err := s.Read(b[:]); err != nil {
+			t.Errorf("read stream failed for listener: %s", err)
+			errCh <- err
+			return
+		}
+		s.Write(b[:])
+		s.Read(b[:]) // peer will close the connection after read
```
MarcoPolo, line 898: Should we expect an error here then?

p2p/transport/webrtc/listener.go
sukunrt: Apparently it doesn't work fine. https://github.com/pion/webrtc/pull/2702#issuecomment-2156006877

So we do need this change.

p2p/transport/webrtc/transport_test.go
sukunrt: added a check.

MarcoPolo (approved):

Issue #523, opened 2019-01-23
Select relays based on latency


vyzo:
With the planned deployment of relay infrastructure, it will make sense to select relays based on minimum latency.

Proposed action: 
- Modify the autorelay relay selection strategy to select based on ping latency.

vyzo:
cc @Stebalien

Stebalien:
That sounds like a decent initial approach (we can try other benchmarks later).

marten-seemann:
The problem here is that we don't want to flood relay nodes that happen to be very well connected to many other nodes with reservation requests.

reubenr0d:
@marten-seemann would fuzzing RTTs be sufficient here? I would like to have a crack at this.

Do you think we should also account for how many reservations a relay has received? Would that acts as a sufficient proxy for actual network utilisation?

cc @Stebalien

PR #2804, opened 2024-05-21
autorelay: implement candidates sorted by RTT

wlynxg:
Implement https://github.com/libp2p/go-libp2p/issues/523.

### DIFF
--- a/p2p/host/autorelay/relay_finder.go
+++ b/p2p/host/autorelay/relay_finder.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"sort"
 	"sync"
 	"time"
 
@@ -17,6 +18,7 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/host/eventbus"
 	circuitv2 "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/client"
 	circuitv2_proto "github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/proto"
+	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
 
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr/net"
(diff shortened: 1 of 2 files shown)


p2p/host/autorelay/relay_finder.go
```diff
@@ -699,11 +718,17 @@ func (rf *relayFinder) selectCandidates() []*candidate {
...
 
-	// TODO: better relay selection strategy; this just selects random relays,
-	// but we should probably use ping latency as the selection metric
-	rand.Shuffle(len(candidates), func(i, j int) {
-		candidates[i], candidates[j] = candidates[j], candidates[i]
-	})
+	// If the node speed test is enabled, the rtt will be sorted,
+	// otherwise the order will be randomly shuffled.
+	if rf.sortByRTT {
+		sort.Slice(candidates, func(i, j int) bool {
```
vyzo, line 724: this should be fuzzy, otherwise it can create hot spots in the network.

wlynxg: > this should be fuzzy, otherwise it can create hot spots in the network.

Sorry, I didn't quite get your point. Could you be a little more detailed?

vyzo: fuzz the RTTs, if they are exact then well positioned nodrs in the network might become hot spots.

wlynxg: it is true that those high quality nodes in the network are preferred, but the node has already shown that he can improve the relay service for us when he accepted to be able to be our candidate.

MarcoPolo: I agree with Vyzo. We don't want to introduce a change that could cause all well-behaved nodes to possibly DoS a well behaved relay.

wlynxg: We do not enable rtt speed measurement by default, which should alleviate this problem.

Regarding the DoS issue, I think we should protect the server side instead of the client side. The libp2p code itself is open source, and anyone can change the behavior of the libp2p relay client, so restricting it here is useless for hackers who want to perform DoS attacks on our public network nodes.

MarcoPolo: There's a difference between a malicious user doing something bad with nodes they control and us shipping code that causes *every* node to behave badly.

p-shahi:
@wlynxg I'm going to close this PR - there's been good feedback from Marco and Vyzo, you can address it and re-open if you decide to pick this up again and submit for a re-review

PR #2806, opened 2024-05-21
rcmgr: Breaking change. Add OpenConnectionNoIP to interface

MarcoPolo:
To merge before we make the next minor release.

This adds a `OpenConnectionNoIP` method to the interface for the resource manager. It's used by users who want a connection scope, but don't have an IP address to tie to the connection. I believe the only user is circuitv2 relays.

The reasons to prefer a new method over changing `OpenConnectionIP` to be circuitv2 aware is:
- The caller should be explicit about not having an IP address.
- The resource manager code should not be aware of circuitv2.


sukunrt:
Why make an exception for IP addresses vs Non IP addresses. Users who want to implement their custom behavior anyway need to wrap our implementation and do their custom stuff on top.

I think the better way is to remove the method `OpenConnectionNoIP` and let users who want custom limiting wrap OpenConnection and implement it themselves. 

Similar comment here: https://github.com/libp2p/go-libp2p/pull/2805#discussion_r1609969616

MarcoPolo:
Lets move discussion to https://github.com/libp2p/go-libp2p/pull/2805#discussion_r1609969616

PR #2807, opened 2024-05-21
rcmgr: Support specific network prefix in conn limiter

MarcoPolo:
Despite the field being called "ConnLimitPerCIDR" you weren't able to define limits for a specific CIDR address block. I've renamed this to ConnLimitPerSubnet to make it clearer that this connection limit applies to each subnet with the same N bit prefix.

I've also added an option to allow defining limits on specific address prefix blocks. If an IP address matches one of these blocks then that limit is used instead of the more generic `ConnLimitPerSubnet`. By default you are allowed unlimited connections from localhost. This should prevent test breakages we've seen.

This does break the existing API. I think that's okay because:
- We just released this API, so hopefully no one is using it yet. Or, if they are, it's fresh on their mind.
- Most users probably just stuck with the defaults. And the defaults here are better.
- The old names were confusing at best.

### DIFF
--- a/p2p/host/resource-manager/conn_limiter.go
+++ b/p2p/host/resource-manager/conn_limiter.go
@@ -16,8 +16,8 @@ type ConnLimitPerSubnet struct {
 	ConnCount int
 }
 
-type CIDRLimit struct {
-	// The CIDR prefix for which this limit applies.
+type NetworkPrefixLimit struct {
+	// The Network prefix for which this limit applies.
 	Network netip.Prefix
 
 	// The maximum number of connections allowed for this subnet.
@@ -44,14 +44,14 @@ var defaultIP6Limits = []ConnLimitPerSubnet{
 	},
 }
 
-var DefaultCIDRLimitV4 = sortCIDRLimits([]CIDRLimit{
+var DefaultNetworkPrefixLimitV4 = sortNetworkPrefixes([]NetworkPrefixLimit{
 	{
 		// Loopback address for v4 https://datatracker.ietf.org/doc/html/rfc6890#section-2.2.2
 		Network:   netip.MustParsePrefix("127.0.0.0/8"),
 		ConnCount: math.MaxInt, // Unlimited
 	},
 })
-var DefaultCIDRLimitV6 = sortCIDRLimits([]CIDRLimit{
+var DefaultNetworkPrefixLimitV6 = sortNetworkPrefixes([]NetworkPrefixLimit{
 	{
 		// Loopback address for v6 https://datatracker.ietf.org/doc/html/rfc6890#section-2.2.3
 		Network:   netip.MustParsePrefix("::1/128"),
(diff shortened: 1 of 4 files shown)


p2p/host/resource-manager/conn_limiter.go
```diff
@@ -1,13 +1,26 @@
...
+type ConnLimitPerSubnet struct {
+	// This defines how big the subnet is. For example, a /24 subnet has a
+	// BitMask of 24. All IPs that share the same 24 bit prefix are in the same
+	// subnet.  Are in the same subnet, and bound to the same limit.
+	BitMask int
+	// The maximum number of connections allowed for each subnet.
+	ConnCount int
+}
+
+type CIDRLimit struct {
```
MarcoPolo, line 19: Maybe we should call this NetworkPrefixLimit and get rid of the CIDR vocab here. Go std lib doesn’t use it either.

MarcoPolo: Would also help with the transition since the old thing had CIDR in the name, but this is different.

p2p/host/resource-manager/conn_limiter.go
```diff
@@ -1,13 +1,26 @@
...
 )
 
-type ConnLimitPerCIDR struct {
-	// How many leading 1 bits in the mask
-	BitMask   int
+type ConnLimitPerSubnet struct {
+	// This defines how big the subnet is. For example, a /24 subnet has a
+	// BitMask of 24. All IPs that share the same 24 bit prefix are in the same
+	// subnet.  Are in the same subnet, and bound to the same limit.
+	BitMask int
```
sukunrt, line 14: Is `PrefixLength` better? 
bitmask suggest's it is the network's bitmask that you can `AND` the ip address with.

p2p/host/resource-manager/conn_limiter.go
sukunrt: Prefix limit is better. CIDR refers to the routing method, right?

p2p/host/resource-manager/conn_limiter.go
```diff
@@ -31,51 +44,146 @@ var defaultIP6Limits = []ConnLimitPerCIDR{
...
+			if connsPerCidr[i]+1 > limit.ConnCount {
+				return false
+			}
+			connsPerCidr[i]++
+			if isIP6 {
+				cl.connsPerCIDRV6 = connsPerCidr
+			} else {
+				cl.connsPerCIDRV4 = connsPerCidr
+			}
+
```
sukunrt: should this be in the block above? Line 159

p2p/host/resource-manager/conn_limiter.go
```diff
@@ -31,51 +44,146 @@ var defaultIP6Limits = []ConnLimitPerCIDR{
...
+		limits = cl.connLimitPerSubnetV6
+		connsPerLimit = cl.ip6connsPerLimit
 	}
 
-	if len(countsPerLimit) == 0 && len(limits) > 0 {
-		countsPerLimit = make([]map[string]int, len(limits))
+	// Check CIDR limits first
+	if len(connsPerCidr) == 0 && len(cidrLimits) > 0 {
+		// Initialize the counts
+		connsPerCidr = make([]int, len(cidrLimits))
```
sukunrt: Should this be: 

```
                // Initialize the counts
		connsPerCidr = make([]int, len(cidrLimits))
                if isIP6 {
				cl.connsPerCIDRV6 = connsPerCidr
			} else {
				cl.connsPerCIDRV4 = connsPerCidr
			}
```

p2p/host/resource-manager/conn_limiter.go
```diff
@@ -110,12 +218,52 @@ func (cl *connLimiter) addConn(ip netip.Addr) bool {
...
+	if len(connsPerCidr) == 0 && len(cidrLimits) > 0 {
+		// Initialize just in case. We should have already initialized in
+		// addConn, but if the callers calls rmConn first we don't want to panic
+		connsPerCidr = make([]int, len(cidrLimits))
+	}
+	for i, limit := range cidrLimits {
+		if limit.Network.Contains(ip) {
+			count := connsPerCidr[i]
+			if count <= 0 {
+				log.Errorf("unexpected conn count for ip %s. Was this not added with addConn first?", ip)
```
sukunrt, line 243: Should there be a `return` here?

p2p/host/resource-manager/conn_limiter.go
```diff
@@ -31,51 +44,146 @@ var defaultIP6Limits = []ConnLimitPerCIDR{
...
+				return false
+			}
+			connsPerNetworkPrefix[i]++
+			if isIP6 {
+				cl.connsPerNetworkPrefixV6 = connsPerNetworkPrefix
+			} else {
+				cl.connsPerNetworkPrefixV4 = connsPerNetworkPrefix
+			}
+
+			return true
```
sukunrt, line 177: Should we add a comment explaining why we are done here?

sukunrt (approved):

p2p/host/resource-manager/conn_limiter.go
MarcoPolo: Yeah, I think that makes sense as an effort to reduce the unexpectedness of the unexpected case.

MarcoPolo:
On second thought, we shouldn't make a breaking change in a patch release. I'm going to add back the old name of `ConnLimitPerCIDR` and mark it as deprecated. That will avoid breaking users who update to a patch release. We can remove the poorly named one in the next major version. I'll make an issue for it.

p2p/host/resource-manager/conn_limiter.go
```diff
@@ -16,81 +29,177 @@ type ConnLimitPerCIDR struct {
...
 	},
 	{
-		ConnCount: 8 * defaultMaxConcurrentConns,
-		BitMask:   48,
+		ConnCount:    8 * defaultMaxConcurrentConns,
+		PrefixLength: 48,
 	},
 }
 
-func WithLimitPeersPerCIDR(ipv4 []ConnLimitPerCIDR, ipv6 []ConnLimitPerCIDR) Option {
```
sukunrt, line 34: We can't change this name for the patch release. I'm fine with just making a new release. v0.35.

sukunrt: On the same topic can we introduce a new Option in a patch release(`WithNetworkPrefixLimit`)? It's backwards compatible but definitely not a bug fix.

MarcoPolo: Okay, let's make a new release to v0.35.

> On the same topic can we introduce a new Option in a patch release(WithNetworkPrefixLimit)? It's backwards compatible but definitely not a bug fix.

Technically you're right. It's not a patch release per semver. We don't have a great way of distinguishing between major and minor releases right now though and I'd argue a minor release is closer to a patch release in spirit than a minor release is to a major relase.

MarcoPolo:
Plan is to merge this and release it in v0.35 with just this change.

PR #2810, opened 2024-05-22
config: fix "Insecure-security" constructor

sukunrt:


config/config.go
```diff
@@ -430,7 +432,12 @@ func (cfg *Config) NewNode() (host.Host, error) {
 	}
 
 	if err := cfg.addAutoNAT(bh); err != nil {
-		rh.Close()
+		app.Stop(context.Background())
+		if cfg.Routing != nil {
```
MarcoPolo, line 436: This should be part of the stop lifecycle, no? Maybe we defer the fix until later

MarcoPolo (approved):

MarcoPolo:
For folks skimming this PR. This is fixing the constructor for the "Insecure" security layer. It's not fixing an insecure construction of some other security layer.

Folks should not be using the Insecure module except outside of testing.

config/config.go
sukunrt: wdym? We are not closing the host with the Stop lifecycle hook.

PR #2820, opened 2024-06-03
identify: Don't filter addr if remote is neither public nor private

MarcoPolo:
Updates the `filterAddrs` logic to no-op if the address is neither public nor private.

This fixes an issue in mocknet that assigns each node an address in the [IPv6 discard prefix](https://datatracker.ietf.org/doc/html/rfc6666) space. That doesn't interact well with this logic in identify.

The issue mocknet hits is that it filters out all received listen addresses and then doesn't remember any address for the peer.


sukunrt (approved):
Should we classify blackhole prefix addresses as Private for manet.IsPrivate?

MarcoPolo:
I don't think so. I think this logic is correct. It's neither public nor private.

Also this might be what we want anyways if this code ever runs into non-IP multiaddrs.

PR #2823, opened 2024-06-06
quic: add support for quic-go metrics

sukunrt:


sukunrt:
Pending: quic-go patch release.

p2p/transport/quicreuse/options.go
```diff
@@ -10,9 +12,12 @@ func DisableReuseport() Option {
 }
 
 // EnableMetrics enables Prometheus metrics collection.
```
MarcoPolo, line 14: Let's add to the comment saying if nil is passed, it will use the default registerer

MarcoPolo (approved):
One suggestion to improve the comment for the exported Option func.

sukunrt:
@galargh The go-check is failing with a diff in the generated protobuf files. How do i find out what the diff in the generated protobuf file is? Locally I only see the version being changed which should be ignored as per https://github.com/libp2p/go-libp2p/pull/2631

galargh:
Here's a run with full diff - https://github.com/libp2p/go-libp2p/actions/runs/9974916892/job/27563846282

I'll make sure we have the diff available on failures in the next release of uci.

PR #2828, opened 2024-06-08
identify: add test for observed address handling

ivan4th:
This modifies TestObservedAddrManager to verify the fix in #2825


sukunrt (approved):
Thanks @ivan4th

PR #2829, opened 2024-06-10
fix: Close host on SIGINT

### LINKED ISSUE #2785: Process unresponsive to SIGINT
This bug was discovered on the master branch after the merge at [`9d149fa3cbba11f2985fbc8e42cd2cf6b9a679f7`](https://github.com/libp2p/go-libp2p/pull/2118)

Using [`./examples/echo`](https://github.com/libp2p/go-libp2p/tree/master/examples/echo) to reproduce, first set libp2p to the problematic version, the process can't be killed with ctrl-c

```
[root@railway echo]# go get github.com/libp2p/go-libp2p@9d149fa3cbba11f2985fbc8e42cd2cf6b9a679f7 
go: upgraded github.com/libp2p/go-libp2p v0.33.1-0.20240321170946-39242a499ce6 => v0.33.1-0.20240321171834-9d149fa3cbba
[root@railway echo]# go run . -l 1234
2024/05/06 21:56:06 I am /ip4/127.0.0.1/tcp/1234/p2p/QmRvs2Uj3rkx9RUbawDPSTDkM6og4C4A4RtB15xdXjwSB9
2024/05/06 21:56:06 listening for connections
2024/05/06 21:56:06 Now run "./echo -l 1235 -d /ip4/127.0.0.1/tcp/1234/p2p/QmRvs2Uj3rkx9RUbawDPSTDkM6og4C4A4RtB15xdXjwSB9" on a different terminal
^C^C^C
```

Now downgrade libp2p to its parent commit, the process gets interrupted as expected

```
[root@railway echo]# go get github.com/libp2p/go-libp2p@39242a499ce6fcd64dd414b504f538ab329b87fc
go: downgraded github.com/libp2p/go-libp2p v0.33.1-0.20240321171834-9d149fa3cbba => v0.33.1-0.20240321170946-39242a499ce6
[root@railway echo]# go run . -l 1234
2024/05/06 22:02:58 I am /ip4/127.0.0.1/tcp/1234/p2p/QmZELJtUuWc255ytkFrPHSqEfpGJyJxZVswLRa2aEPVcoW
2024/05/06 22:02:58 listening for connections
2024/05/06 22:02:58 Now run "./echo -l 1235 -d /ip4/127.0.0.1/tcp/1234/p2p/QmZELJtUuWc255ytkFrPHSqEfpGJyJxZVswLRa2aEPVcoW" on a different terminal
^Csignal: interrupt
```

MarcoPolo:
Closes #2785 

This is closer to the old behavior prior to V0.34.

The first SIGINT closes the libp2p host. The second SIGINT follows the default Go behavior, unless the user is handling it.

This means, unless the user does something different, the default behavior will be 2 ctrl-c's are necessary to exit the program. I don't think this is perfect, but it's best we can do since FX will handle the first ctrl-c, and I don't think we should try to replicate the default Go behavior there. For now, I've added a log.Info to provide some info.

I'm open to other suggestions here.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -417,9 +417,13 @@ func (cfg *Config) NewNode() (host.Host, error) {
 
 	var bh *bhost.BasicHost
 	fxopts = append(fxopts, fx.Invoke(func(bho *bhost.BasicHost) { bh = bho }))
+	hostStopped := make(chan struct{})
 	fxopts = append(fxopts, fx.Invoke(func(h *bhost.BasicHost, lifecycle fx.Lifecycle) {
 		lifecycle.Append(fx.StartHook(h.Start))
 		lifecycle.Append(fx.StopHook(h.Close))
+		lifecycle.Append(fx.StopHook(func() {
+			close(hostStopped)
+		}))
 	}))
 
 	var rh *routed.RoutedHost
@@ -443,9 +447,12 @@ func (cfg *Config) NewNode() (host.Host, error) {
 	}
 
 	go func() {
-		<-app.Done() // Listen for SIGINT or SIGTERM and close the host
-		log.Info("Shutting down libp2p host. Ctrl-c once more to exit...")
-		app.Stop(context.Background())
+		select {
+		case <-app.Done(): // Listen for SIGINT or SIGTERM and close the host
+			log.Info("Shutting down libp2p host. Ctrl-c once more to exit...")
+			app.Stop(context.Background())
+		case <-hostStopped: // host has stopped, we can return
+		}
 	}()
 
 	if cfg.Routing != nil {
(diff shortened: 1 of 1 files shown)


sukunrt (commented):
Alternatively, we can expose the `app.Done()` channel and let users decide what to do with the signal. Users with more context can just call `os.Exit` after stopping the host and exit the program.  
 
If I understand correctly previously, the default go signal handler terminated the program which users could override by registering their own signal handler. 
If we make this change, users cannot override the behavior of stopping on `ctrl+c`

config/config.go
```diff
@@ -441,6 +446,15 @@ func (cfg *Config) NewNode() (host.Host, error) {
 		return nil, err
 	}
 
+	go func() {
+		select {
+		case <-app.Done(): // Listen for SIGINT or SIGTERM and close the host
+			log.Info("Shutting down libp2p host. Ctrl-c once more to exit...")
+			app.Stop(context.Background())
```
sukunrt, line 453: can we use `return` to exit the goroutine?

richard-ramos:
I like the idea of exposing `app.Done()`, because then it would be possible to implement at the app level a behavior more natural of interrupting the app on a single SIGINT, compared to the proposal of having to press it twice

MarcoPolo:
I think the real solution here is if fx didn't register signal handlers at all. This would allow users to:

- Rely on the [default behavior of signals in Go programs](https://pkg.go.dev/os/signal#hdr-Default_behavior_of_signals_in_Go_programs).
- Handle signals in a standard Go way with `signal.Notify`.

MarcoPolo:
closing in favor of fixing fx instead.

PR #2840, opened 2024-06-13
libp2phttp.Host implements RoundTripper

MarcoPolo:
After working on the js-libp2p API for libp2p+HTTP, I realized it's quite natural to make an HTTP request given a multiaddr URI (who would've guessed?).

This change makes using the HTTP client with libp2p easier and more natural. Instead of having to create a specialized roundtripper for each server, you can now have a single `http.Client` that can make requests to multiple servers. The usage feels very familiar to the classic HTTP api. You can now do this:

```
client := http.Client{Transport: &libp2phttp.Host{ options... }}
client.Get("https://example.com") // Standard way of making an HTTP request
client.Get("multiaddr:/dns/example.com/tls/http") // Same as above, but with a multiaddr scheme
client.Get("multiaddr:/dnsaddr/example.com/p2p/QmFoo") // galaxy brain; make an HTTP request over a libp2p transport
```

### DIFF
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -674,11 +674,15 @@ func (h *Host) RoundTrip(r *http.Request) (*http.Response, error) {
 			// completeness, but I don't expect us to hit it often.
 			rt = rt.Clone()
 			rt.TLSClientConfig.ServerName = parsed.sni
-			defer rt.CloseIdleConnections()
 		}
 
-		r.URL.Scheme = scheme
-		r.URL.Host = parsed.host + ":" + parsed.port
+		// TODO add http-path support
+		url := url.URL{
+			Scheme: scheme,
+			Host:   parsed.host + ":" + parsed.port,
+		}
+
+		r.URL = &url
 		return rt.RoundTrip(r)
 	}
 
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -406,11 +406,14 @@ type PeerMetadataGetter interface {
 }
 
 type streamRoundTripper struct {
-	server      peer.ID
-	addrsAdded  sync.Once
-	serverAddrs []ma.Multiaddr
-	h           host.Host
-	httpHost    *Host
+	server peer.ID
+	// if true, we won't add the server's addresses to the peerstore. This
+	// should only be set when creating the struct.
+	skipAddAddrs bool
+	addrsAdded   sync.Once
+	serverAddrs  []ma.Multiaddr
+	h            host.Host
+	httpHost     *Host
 }
 
 // streamReadCloser wraps an io.ReadCloser and closes the underlying stream when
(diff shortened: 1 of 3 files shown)


p2p/http/libp2phttp_test.go
```diff
@@ -755,3 +755,71 @@ func TestResponseWriterShouldNotHaveCancelledContext(t *testing.T) {
...
+	// serverHttpHost.SetHTTPHandlerAtPath("/hello-again", "/hello", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	// 	w.Write([]byte("hello"))
+	// }))
+
+	go serverHttpHost.Serve()
+	defer serverHttpHost.Close()
+
+	testCases := []string{
+		// Version that has an http-path. Will uncomment when we get the http-path changes in go-multiaddr
+		// "multiaddr:" + serverHost.Addrs()[0].String() + "/http-path/hello",
```
sukunrt, line 786: Nice! :heart:

p2p/http/libp2phttp.go
```diff
@@ -406,11 +406,14 @@ type PeerMetadataGetter interface {
...
 type streamRoundTripper struct {
-	server      peer.ID
-	addrsAdded  sync.Once
-	serverAddrs []ma.Multiaddr
-	h           host.Host
-	httpHost    *Host
+	server peer.ID
+	// if true, we won't add the server's addresses to the peerstore. This
+	// should only be set when creating the struct.
+	skipAddAddrs bool
```
sukunrt, line 412: Can you elaborate on why we need this?

p2p/http/libp2phttp.go
```diff
@@ -620,6 +625,76 @@ func (h *Host) NamespacedClient(p protocol.ID, server peer.AddrInfo, opts ...Rou
...
+	if isHTTP {
+		parsed := parseMultiaddr(addr)
+		scheme := "http"
+		if parsed.useHTTPS {
+			scheme = "https"
+		}
+		h.initDefaultRT()
+		rt := h.DefaultClientRoundTripper
+		if parsed.sni != parsed.host {
+			// We have a different host and SNI (e.g. using an IP address but specifying a SNI)
```
sukunrt, line 664: Does this work with `/dns` addrs?

Do we need something like: https://github.com/libp2p/go-libp2p/blob/master/p2p/transport/websocket/websocket.go#L123 here?

sukunrt (approved):

p2p/http/libp2phttp.go
MarcoPolo: It really doesn't matter, but it saves us the synchronization overhead of the `addrsAdded sync.Once` call.

p2p/http/libp2phttp.go
MarcoPolo: Yes `/dns` works. Roundtripper handles this implicitly. Fails if the server doesn't support the SNI. Example: `/dns/example.com/tls/sni/example.net/http`works, but `/dns/example.net/tls/sni/notexample.com/http` fails.

PR #2843, opened 2024-06-20
add peerId compare

Einstellung:
add peerId compare, otherwise this program will occasionally cause TLS shake sequence error

### DIFF
--- a/examples/pubsub/chat/main.go
+++ b/examples/pubsub/chat/main.go
@@ -94,6 +94,9 @@ type discoveryNotifee struct {
 // support PubSub.
 func (n *discoveryNotifee) HandlePeerFound(pi peer.AddrInfo) {
 	fmt.Printf("discovered new peer %s\n", pi.ID)
+	if pi.ID > n.h.ID() {
+		return
+	} 
 	err := n.h.Connect(context.Background(), pi)
 	if err != nil {
 		fmt.Printf("error connecting to peer %s: %s\n", pi.ID, err)


MarcoPolo:
This doesn't make sense. This codepath happens after the TLS handshake. Also the peer ids are random. Closing, but if you want to add more context go ahead.

Einstellung:
like this file [code](https://github.com/libp2p/go-libp2p/blob/master/examples/chat-with-mdns/main.go) in line 119. In concurrency program, occasionally different channels want to establish connection at the same time, which will cause connection error: libp2p failed to negotiate security protocol: tls: received unexpected handshake message of type *tls.clientHelloMsg when waiting for *tls.serverHelloMsg

Einstellung:
> This doesn't make sense. This codepath happens after the TLS handshake. Also the peer ids are random. Closing, but if you want to add more context go ahead.

like this file [code](https://github.com/libp2p/go-libp2p/blob/master/examples/chat-with-mdns/main.go) in line 119. In concurrency program, occasionally different channels want to establish connection at the same time, which will cause connection error: libp2p failed to negotiate security protocol: tls: received unexpected handshake message of type *tls.clientHelloMsg when waiting for *tls.serverHelloMsg

MarcoPolo:
Are you seeing a tcp simultaneous open? Is this on a local network?

Einstellung:
Yes, I just test on my local network (localhost). I didn't test whether tcp simultaneous open.

MarcoPolo:
Can you switch the listen address to a quic one and see if you have the same issue?

Einstellung:
Sorry, based on my poor network knowledge I don't know how to switch the listen address to a quic one🥹

PR #2850, opened 2024-06-25
feat: libp2phttp `/http-path`

MarcoPolo:
Adds support for multiaddr URIs that contain `/http-path` components. In other words, this works now:

```
client := http.Client{Transport: &libp2phttp.Host{ options... }}
client.Get("multiaddr:/dns/example.com/tls/http/http-path/some%2fpath")
// equivalent to
client.Get("https://example.com/some/path")
// And of course works with any libp2p stream transport:
client.Get("multiaddr:/dnsaddr/example.com/p2p/QmFoo/http-path/some%2fpath")
```

### DIFF
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -499,6 +499,8 @@ func (rt *streamRoundTripper) RoundTrip(r *http.Request) (*http.Response, error)
 	return resp, nil
 }
 
+var errNotRelative = errors.New("not relative")
+
 // relativeMultiaddrURIToAbs takes a relative multiaddr URI and turns it into an
 // absolute one. Useful, for example, when a server gives us a relative URI for a redirect.
 // It allows the following request (the one after redirected) to reach the correct server.
@@ -510,7 +512,7 @@ func relativeMultiaddrURIToAbs(original *url.URL, relative *url.URL) (*url.URL, 
 	// multiaddr://here-instead.
 	if relative.OmitHost {
 		// Not relative (at least we can't tell). Nothing we can do here
-		return nil, errors.New("not relative")
+		return nil, errNotRelative
 	}
 	originalStr := original.RawPath
 	if originalStr == "" {
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -750,10 +750,10 @@ func (h *Host) RoundTrip(r *http.Request) (*http.Response, error) {
 	if parsed.peer == "" {
 		return nil, fmt.Errorf("no peer ID in multiaddr")
 	}
-	addr, _ = ma.SplitFunc(addr, func(c ma.Component) bool {
-		return c.Protocol().Code == ma.P_P2P
+	withoutHTTPPath, _ := ma.SplitFunc(addr, func(c ma.Component) bool {
+		return c.Protocol().Code == ma.P_HTTP_PATH
 	})
-	h.StreamHost.Peerstore().AddAddrs(parsed.peer, []ma.Multiaddr{addr}, peerstore.TempAddrTTL)
+	h.StreamHost.Peerstore().AddAddrs(parsed.peer, []ma.Multiaddr{withoutHTTPPath}, peerstore.TempAddrTTL)
 
 	// Set the Opaque field to the http-path so that the HTTP request only makes
 	// a reference to that path and not the whole multiaddr uri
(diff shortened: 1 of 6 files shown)


p2p/http/libp2phttp.go
```diff
@@ -657,12 +660,22 @@ func (h *Host) RoundTrip(r *http.Request) (*http.Response, error) {
 		return nil, err
 	}
 	addr, isHTTP := normalizeHTTPMultiaddr(addr)
```
sukunrt, line 662: This panics for input `multiaddr:/https/http-path/xyz`

p2p/http/libp2phttp.go
```diff
@@ -657,12 +660,22 @@ func (h *Host) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	parsed, err := parseMultiaddr(addr)
```
sukunrt, line 663: We need this parsing only if it's a http multiaddr, right? Otherwise we can just use the `addr`? 

I would prefer calling this only when it's an `/http` multiaddr because `parseMultiaddr` looks specific to parsing `/http` multiaddrs

p2p/http/libp2phttp.go
MarcoPolo: Do you also find it surprising that the multiaddr library returns a nil Multiaddr type, and panics when given the type it just returned? I think this variant should hold:

```
before, after := ma.Split(original, ...)
roundTrip := ma.Join(before, after)
roundTrip == original // sometimes true, sometimes panics
```

p2p/http/libp2phttp.go
```diff
@@ -676,28 +689,23 @@ func (h *Host) RoundTrip(r *http.Request) (*http.Response, error) {
...
 
-	addr, pid := peer.SplitAddr(addr)
-	if pid == "" {
+	if parsed.peer == "" {
 		return nil, fmt.Errorf("no peer ID in multiaddr")
 	}
-	h.StreamHost.Peerstore().AddAddrs(pid, []ma.Multiaddr{addr}, peerstore.TempAddrTTL)
+	addr, _ = ma.SplitFunc(addr, func(c ma.Component) bool {
+		return c.Protocol().Code == ma.P_P2P
+	})
```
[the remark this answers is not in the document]
MarcoPolo: My code is wrong. Doesn't handle circuit addrs. I just need to split on http-paths in case they are there.

p2p/http/libp2phttp.go
MarcoPolo: > We need this parsing only if it's a http multiaddr, right? 

Both. We use the http-path, host, and port in both code paths.

Would you prefer we name this something else?

p2p/http/libp2phttp.go
sukunrt: I completely agree. This should always work
```
before, after := ma.SplitFirst(addr)
addr.Equal(ma.Join(before, after))
before, after = ma.SplitLast(addr)
addr.Equal(ma.Join(before, after))
```

We should change `ma.Join` to just ignore nils. I don't think it would break anyone but would help with a lot of such cases.

p2p/http/libp2phttp.go
sukunrt: You only need `parsed.peer` in that code path so I was wondering if we should just get the peerID from the addr in that code path.  

This is fine too though.

p2p/http/libp2phttp.go
```diff
@@ -477,9 +480,59 @@ func (rt *streamRoundTripper) RoundTrip(r *http.Request) (*http.Response, error)
...
+// It allows the following request (the one after redirected) to reach the correct server.
+func relativeMultiaddrURIToAbs(original *url.URL, relative *url.URL) (*url.URL, error) {
+	// Is this a relative uri? We know if it is because non-relative URI's of the form:
+	// "multiaddr:/ip4/1.2.3.4/tcp/9899" when parsed by Go's url package will have url.OmitHost == true
+	// But if it is relative (just a path to an http resource e.g. /here-instead)
+	// a redirect will inherit the multiaddr scheme, but set url.OmitHost == false. It will also stringify as something like
+	// multiaddr://here-instead.
+	if relative.OmitHost {
+		// Not relative (at least we can't tell). Nothing we can do here
+		return nil, errors.New("not relative")
```
sukunrt, line 513: You can reuse this error and avoid allocation for all the calls. This looks like it'll be used often.

p2p/http/libp2phttp.go
MarcoPolo: We need http path as well

PR #2852, opened 2024-06-27
fix: Close conn in quic listener when wrapping fails

### LINKED ISSUE #2841: Memory Leak With Go-Libp2p
In Prysm, we have support for running Prysm nodes both with the Quic transport and the TCP transport.  Recently we updated our version to be v0.35.0. This has been working for the most part, however while profiling long-running nodes running we have come across what appears to be a memory leak. Over the course of a few weeks the heap occupied by a prysm node has grown rather than stayed constant size. If you look at the heap profiles below, you can see the large amount of memory allocated by the respective quic and tcp dialers.
  

image
image

The above 2 screenshots capture the heatmaps occupied by QUIC connections. If you take a look, it is sizeable amount occupied by the pre-setup step of quic connections:

```
github.com/quic-go/quic-go.(*connection).preSetup
external/com_github_quic_go_quic_go/connection.go

  Total:    698.35MB   734.85MB (flat, cum) 12.86%
    452            .          .           	} 
    453            .          .           	return s 
    454            .          .           } 
    455            .          .            
    456            .          .           func (s *connection) preSetup() { 
    457          5MB        9MB           	s.initialStream = newCryptoStream() 
    458          1MB        8MB           	s.handshakeStream = newCryptoStream() 
    459       9.50MB     9.50MB           	s.sendQueue = newSendQueue(s.conn) 
    460       3.50MB     3.50MB           	s.retransmissionQueue = newRetransmissionQueue() 
    461          1MB        1MB           	s.frameParser = *wire.NewFrameParser(s.config.EnableDatagrams) 
    462       1.50MB     1.50MB           	s.rttStats = &utils.RTTStats{} 
    463       2.50MB     2.50MB           	s.connFlowController = flowcontrol.NewConnectionFlowController( 
    464            .          .           		protocol.ByteCount(s.config.InitialConnectionReceiveWindow), 
    465            .          .           		protocol.ByteCount(s.config.MaxConnectionReceiveWindow), 
    466     512.01kB   512.01kB           		s.onHasConnectionWindowUpdate, 
    467     512.01kB   512.01kB           		func(size protocol.ByteCount) bool { 
    468            .          .           			if s.config.AllowConnectionWindowIncrease == nil { 
    469            .          .           				return true 
    470            .          .           			} 
    471            .          .           			return s.config.AllowConnectionWindowIncrease(s, uint64(size)) 
    472            .          .           		}, 
    473            .          .           		s.rttStats, 
    474            .          .           		s.logger, 
    475            .          .           	) 
    476       2.50MB     2.50MB           	s.earlyConnReadyChan = make(chan struct{}) 
    477          4MB       28MB           	s.streamsMap = newStreamsMap( 
    478            .          .           		s.ctx, 
    479            .          .           		s, 
    480     512.01kB   512.01kB           		s.newFlowController, 
    481            .          .           		uint64(s.config.MaxIncomingStreams), 
    482            .          .           		uint64(s.config.MaxIncomingUniStreams), 
    483            .          .           		s.perspective, 
    484            .          .           	) 
    485          2MB        2MB           	s.framer = newFramer(s.streamsMap) 
    486     647.84MB   647.84MB           	s.receivedPackets = make(chan receivedPacket, protocol.MaxConnUnprocessedPackets) 
    487       3.50MB     3.50MB           	s.closeChan = make(chan closeError, 1) 
    488       1.50MB     1.50MB           	s.sendingScheduled = make(chan struct{}, 1) 
    489            .     1.50MB           	s.handshakeCtx, s.handshakeCtxCancel = context.WithCancel(context.Background()) 
    490            .          .            
    491            .          .           	now := time.Now() 
    492            .          .           	s.lastPacketReceivedTime = now 
    493            .          .           	s.creationTime = now 
    494            .          .            
    495       2.50MB     2.50MB           	s.windowUpdateQueue = newWindowUpdateQueue(s.streamsMap, s.connFlowController, s.framer.QueueControlFrame) 
    496          9MB        9MB           	s.datagramQueue = newDatagramQueue(s.scheduleSending, s.logger) 
    497            .          .           	s.connState.Version = s.version 
    498            .          .           } 
    499            .          .            
    500            .          .           // run the connection main loop 
    501            .          .           func (s *connection) run() error { 
 ```
 
 I am not very familiar with the internals of quic-go but it appears these streams/connections are not being appropriately garbage collected by go-libp2p ? You have old connections somehow still living on the heap. 
 
 Finally we did a profile differential for the same node of its current state and its state a few days ago:
 
 
image

With this heatmap it becomes obvious where the memory leak is coming from, it does appear the new release has introduced either a regression or new bug for old connections. 

<details>
<summary>Version Information</summary>
<pre>
v0.35.0
</pre>
</details>

### PR DESCRIPTION
**Pull request by MarcoPolo**
When wrapping a quic connection into a transport.CapableConn we wouldn't always close the underlying connection if wrapping failed. This fixes that and adds some tests to ensure all transports do this properly.

I believe this closes #2841

### DIFF
--- a/p2p/test/transport/transport_test.go
+++ b/p2p/test/transport/transport_test.go
@@ -33,7 +33,6 @@ import (
 	libp2pwebrtc "github.com/libp2p/go-libp2p/p2p/transport/webrtc"
 	"go.uber.org/mock/gomock"
 
-	"github.com/multiformats/go-multiaddr"
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/stretchr/testify/require"
 )
@@ -673,7 +672,7 @@ func TestDiscoverPeerIDFromSecurityNegotiation(t *testing.T) {
 
 			ai := &peer.AddrInfo{
 				ID:    bogusPeerId,
-				Addrs: []multiaddr.Multiaddr{h1.Addrs()[0]},
+				Addrs: []ma.Multiaddr{h1.Addrs()[0]},
 			}
 
 			// Try connecting with the bogus peer ID
--- a/p2p/test/transport/transport_test.go
+++ b/p2p/test/transport/transport_test.go
@@ -20,7 +20,9 @@ import (
 	"github.com/libp2p/go-libp2p/core/connmgr"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
+	mocknetwork "github.com/libp2p/go-libp2p/core/network/mocks"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/core/sec"
 	rcmgr "github.com/libp2p/go-libp2p/p2p/host/resource-manager"
 	"github.com/libp2p/go-libp2p/p2p/muxer/yamux"
@@ -29,8 +31,10 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/security/noise"
 	tls "github.com/libp2p/go-libp2p/p2p/security/tls"
 	libp2pwebrtc "github.com/libp2p/go-libp2p/p2p/transport/webrtc"
+	"go.uber.org/mock/gomock"
 
 	"github.com/multiformats/go-multiaddr"
+	ma "github.com/multiformats/go-multiaddr"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: p2p/transport/quic/listener_test.go [hunk 1]
```diff
@@ -113,3 +118,51 @@ func TestCorrectNumberOfVirtualListeners(t *testing.T) {
...
+
+	// No error yet, let's continue using the conn
+	s, err := conn.OpenStream(ctx)
+	if err != nil && errors.As(err, &quicErr) {
+		// We hit our expected application error
+		return
+	}
+
+	// No error yet, let's continue using the conn
+	s.SetReadDeadline(time.Now().Add(1 * time.Second))
```
sukunrt, line 159: We can keep this 10 seconds. It will error before that any way.

p2p/test/transport/transport_test.go
```diff
@@ -688,3 +692,34 @@ func TestDiscoverPeerIDFromSecurityNegotiation(t *testing.T) {
...
+			client := tc.HostGenerator(t, TransportTestCaseOpts{NoListen: true})
+			defer server.Close()
+			defer client.Close()
+
+			client.Peerstore().AddAddrs(server.ID(), server.Addrs(), peerstore.PermanentAddrTTL)
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			_, err := client.NewStream(ctx, server.ID(), ping.ID)
+			require.Error(t, err)
+			require.False(t, errors.Is(err, context.DeadlineExceeded), "expected error to be not be context deadline exceeded")
```
sukunrt, line 722: This will succeed if the implementation is using lazy multistream negotiation, right?

p2p/test/transport/transport_test.go
```diff
@@ -688,3 +692,34 @@ func TestDiscoverPeerIDFromSecurityNegotiation(t *testing.T) {
 		})
 	}
 }
+
+// TestCloseConnWhenBlocked tests that the server closes the connection when the rcmgr blocks it.
+func TestCloseConnWhenBlocked(t *testing.T) {
+	for _, tc := range transportsToTest {
+		if tc.Name == "WebRTC" {
+			continue // WebRTC doesn't have a connection when we block so there's nothing to close
```
sukunrt, line 700: Should we test this, `NewStream` on client will still fail, which is what we want. It'll fail with `DeadlineExceeded` though

sukunrt (approved):

p2p/test/transport/transport_test.go
MarcoPolo: I don't think we should. It's not testing what we think. The only thing it tests is whether the server ignores us and we hit a deadline.

MarcoPolo: Maybe we could make another tests that just focuses on that behavior for WebRTC though?

p2p/test/transport/transport_test.go
MarcoPolo: In which case the require.Error will fail. I haven't seen that happen which is probably because we require the identify to happen before returing a new stream (a separate issue). For now I think let's leave this as is and when we change the Identify/NewStream interaction we can update this test.

p2p/test/transport/transport_test.go
sukunrt: It would test that if the respurce manager blocks no connection is established. But we can handle that separately. I am fine with keeping it as is.

MarcoPolo: #2856

sukunrt:
Did you merge this instead of squash and merge? :sweat_smile:

MarcoPolo:
Unfortunately yes. Not on purpose. I feel like the defaults might have changed.

nisdas:
When will this PR be included in a new release ?

PR #2870, opened 2024-07-12
Adding tau to `Notable Users` in README.md

samyfodil:
[tau](https://github.com/taubyte/tau) is a distributed PaaS that uses libp2p for networking, routing and discovery. As of now tau has over 1500 stars on github.


MarcoPolo (approved):
Cool!

PR #2875, opened 2024-07-17
basichost: reset stream if SetProtocol fails

sukunrt:


MarcoPolo (approved):

rvagg:
What kinds of conditions would cause `SetProtocol` to fail?

PR #2876, opened 2024-07-17
Update go-check.yml

galargh:


sukunrt:
@galargh can you rebase this to master? I think this should be default behavior.

galargh:
No need to rebase, I just released unified github workflows with a patch that adds `git diff` to the failure output of generation check - https://github.com/ipdxco/unified-github-workflows/releases/tag/v1.0.10

PR #2879, opened 2024-07-19
chore(p2p): use `errors.New` to replace `fmt.Errorf` with no parameters

yukionfire:


MarcoPolo:
I don't think this change is worth it. It saves a couple function calls and a comparison on the error path. https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/fmt/errors.go;drc=4a0a2b33dfa3c99250efa222439f2c27d6780e4a;l=30

PR #2881, opened 2024-07-22
chore: Update connmgr.go/NewConnManager()'s commnet

elecbug:
Although not very important, the 'low' used in the variable name and the 'lo' in the comment are different. This is the result of renaming the variable in the comment.


sukunrt:
I'm closing this since this has no impact as the comment as is, is not misleading. You can check the github issues for good first issues.

PR #2889, opened 2024-07-30
feat: WebRTC reuse QUIC conn

MarcoPolo:
This lets users listen on the same port for WebRTC and QUIC. Kubo folks asked for this as it would be very beneficial to their deployment.

One open question for me is:
- Do we need to make sure this works well with the Observed Address Manager? I think it should just work, but we should double check. Ideally we should know our public webrtc-direct address if we've discovered our public quic address when they use the same port.

### DIFF
--- a/p2p/net/swarm/swarm_listen.go
+++ b/p2p/net/swarm/swarm_listen.go
@@ -3,6 +3,7 @@ package swarm
 import (
 	"errors"
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/libp2p/go-libp2p/core/canonicallog"
@@ -12,13 +13,44 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+type OrderedListener interface {
+	// Transports optionally implement this interface to indicate the relative
+	// ordering that listeners should be setup. Some transports may optionally
+	// make use of other listeners if they are setup. e.g. WebRTC may reuse the
+	// same UDP port as QUIC, but only when QUIC is setup first.
+	// lower values are setup first.
+	ListenOrder() int
+}
+
 // Listen sets up listeners for all of the given addresses.
 // It returns as long as we successfully listen on at least *one* address.
 func (s *Swarm) Listen(addrs ...ma.Multiaddr) error {
 	errs := make([]error, len(addrs))
 	var succeeded int
-	for i, a := range addrs {
-		if err := s.AddListenAddr(a); err != nil {
+
+	type addrAndListener struct {
+		addr ma.Multiaddr
+		lTpt transport.Transport
+	}
+	sortedAddrsAndTpts := make([]addrAndListener, 0, len(addrs))
+	for _, a := range addrs {
+		t := s.TransportForListening(a)
+		sortedAddrsAndTpts = append(sortedAddrsAndTpts, addrAndListener{addr: a, lTpt: t})
+	}
+	slices.SortFunc(sortedAddrsAndTpts, func(a, b addrAndListener) int {
+		aOrder := 0
+		bOrder := 0
+		if l, ok := a.lTpt.(OrderedListener); ok {
+			aOrder = l.ListenOrder()
+		}
+		if l, ok := b.lTpt.(OrderedListener); ok {
+			bOrder = l.ListenOrder()
+		}
+		return aOrder - bOrder
+	})
+
+	for i, a := range sortedAddrsAndTpts {
+		if err := s.AddListenAddr(a.addr); err != nil {
 			errs[i] = err
 		} else {
 			succeeded++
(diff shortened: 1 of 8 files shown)


p2p/net/swarm/swarm_listen.go
```diff
@@ -12,13 +13,44 @@ import (
...
 
+type OrderedListener interface {
+	// Transports optionally implement this interface to indicate the relative
+	// ordering that listeners should be setup. Some transports may optionally
+	// make use of other listeners if they are setup. e.g. WebRTC may reuse the
+	// same UDP port as QUIC, but only when QUIC is setup first.
+	// lower values are setup first.
+	ListenOrder() int
+}
+
```
sukunrt, line 24: Does this matter for webrtc? 
There are two cases here
1. User specifies port: User listens on `/udp/4001/webrtc-direct` and `/udp/4001/quic-v1`
2. User doesn't specify port: User listens on `/udp/0/webrtc-direct` and `/udp/0/quic-v1`

In both the cases quic listener can start after webrtc listener without any impact.

p2p/transport/quicreuse/connmgr.go
```diff
@@ -131,6 +131,28 @@ func (c *ConnManager) onListenerClosed(key string) {
...
+	c.quicListenersMu.Lock()
+	defer c.quicListenersMu.Unlock()
+	key := laddr.String()
+	entry, ok := c.quicListeners[key]
+	if !ok {
+		return nil, errors.New("expected to be able to share with a QUIC listener, but no QUIC listener found. The QUIC listener should start first")
+	}
+	t := entry.ln.transport
+	if t, ok := t.(*refcountedTransport); ok {
+		t.IncreaseCount()
```
sukunrt, line 144: If we move this logic to `quicreuse`, we can do away with the ordering requirement between quic and webrtc listener. 
In quicreuse we'll have to introduce something like `TransportForNonQuicStuff`(hopefully a better name) which is then used for listening on quic.

p2p/transport/quicreuse/nonquic_packetconn.go
```diff
@@ -0,0 +1,74 @@
...
+
+// SetReadDeadline implements net.PacketConn.
+func (n *nonQUICPacketConn) SetReadDeadline(t time.Time) error {
+	if t.IsZero() && n.readCtx != nil {
+		n.readCancel()
+		n.readCtx = nil
+	}
+	n.readCtx, n.readCancel = context.WithDeadline(n.ctx, t)
+	return nil
+}
```
sukunrt, line 61: We don't need this but doesn't hurt to keep this here.

p2p/transport/webrtc/transport.go
```diff
@@ -105,7 +108,9 @@ type iceTimeouts struct {
 	Keepalive  time.Duration
 }
 
-func New(privKey ic.PrivKey, psk pnet.PSK, gater connmgr.ConnectionGater, rcmgr network.ResourceManager, opts ...Option) (*WebRTCTransport, error) {
+type ListenUDPFn func(network string, laddr *net.UDPAddr) (net.PacketConn, error)
+
+func New(privKey ic.PrivKey, psk pnet.PSK, gater connmgr.ConnectionGater, rcmgr network.ResourceManager, listenUDP ListenUDPFn, opts ...Option) (*WebRTCTransport, error) {
```
sukunrt, line 113: Should we make this depend on `quicreuse.Connmgr` like webtransport and quic constructors?
I don't have a strong opinion here.

MarcoPolo:
Discussed out of band about changes to make this order agnostic with the QUIC transport, since, as it is implemented in this PR, the QUIC listener starts first. Sukun and I agreed the existing logic is good enough. Which is to sort the listeners and start the QUIC listener first (if there is one). Summary of the points we discussed:

- To reuse the same port, WebRTC has to go through a QUIC transport.
- It doesn't make sense to always instantiate a QUIC transport when setting up a WebRTC listener.
  - It makes WebRTC unnecessarily depend on a QUIC transport being present.
  - Can lead to hard to debug issues. And cause surprising behavior.
- It's much easier to depend on a QUIC transport that has already been started from a prior QUIC listener.
- Sorting the listeners before starting is relatively easy and straightforward.

p2p/transport/webrtc/transport.go
MarcoPolo: I purposely did not since it's nicer to require the bare minimum interface we actually need. WebTransport and QUIC do need the connmgr, WebRTC only needs a function that returns a `net.PacketConn`.

lidel (changes requested):
Thank you for pushing this forward, reusing same port has big impact for DX/UX, allowing us faster rollout with less hiccups.

Took it for a spin in https://github.com/ipfs/kubo/pull/10463 and  `fx` seems to have issue with `libp2pwebrtc.ListenUDPFn`.

To reproduce, rebase this PR on top of https://github.com/libp2p/go-libp2p/pull/2887, to ensure WebRTC Direct is enabled by default, then `fx` errors with `missing type: libp2pwebrtc.ListenUDPFn` (same error as in https://github.com/ipfs/kubo/pull/10463/files#r1697532128).

MarcoPolo:
> We should infer the address from the quic address like we do for webtransport. It'll work seamlessly once we have the same logic as [inferWebtransportAddrsFromQuic](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/basic/basic_host.go#L974)

I don't think we actually need this anymore. I've made a PR to remove that function here https://github.com/libp2p/go-libp2p/pull/2892.

MarcoPolo:
> Took it for a spin in https://github.com/ipfs/kubo/pull/10463 and  fx seems to have issue with libp2pwebrtc.ListenUDPFn.

That was my mistake sorry. I think it was complaining that it couldn't create the WebRTC transport for the AutoNAT host because I didn't provide it to it (only for the basic host).

sukunrt (approved):

sukunrt (approved):
LGTM! 
Thanks @MarcoPolo :tada:

PR #2920, opened 2024-08-14
Fix proto import paths

joshklop:
A proto file's *import path* is relative to one of the `--proto-path`s.
    
    Previously, the proto files were compiled separately. Some invocations
    used different values for the `--proto_path`, which led to inconsistent
    import paths in proto file descriptors.
    
    Typically, this wouldn't be a problem. However, if a downstream project
    uses `protoregistry.GlobalFiles` to inspect proto dependencies, it will fail
    to find a dependency's file descriptor when the dependency was compiled
    with a different `--proto_path`.
    
    By using a single script to generate all protobuf files, we can ensure
    the `--proto_path` is always set to the same sane value (the root of the
    project, as suggested in the [official documentation]).
    
    [official documentation]: https://protobuf.dev/programming-guides/proto2/#importing

Cosmos SDK modules may inspect `protoregistry.GlobalFiles` in SDK v0.50.8 and later. As a result, this appears to be necessary to use libp2p in the same binary as those modules.


joshklop:
The commits start with a failing test and build up to a fully working solution. They should probably be squashed into a single commit before merging to `master`.

MarcoPolo (approved):
Seems like a reasonable change. Thanks for doing it.

MarcoPolo:
Interop test is failing, but it's not this PR's fault. It's just running out of space on the runner.

PR #2931, opened 2024-08-21
fix: enable dctur when interface address is public 

### LINKED ISSUE #2913: QUIC/WebTransport Transport sometimes uses the wrong socket to dial out.
### Context

Given the following scenario:
- Two peers: 
  - Peer A: with a public IP and a firewall blocking arbitrary remote connections. Running latest [Vole](https://github.com/ipfs-shipyard/vole)
  - Peer B (Kubo Node) behind NAT with a circuit relay reservation
- Peer A is running go-libp2p 0.36.1 as part of Vole
- Peer A is constructed with `h, err := libp2p.New(libp2p.DefaultMuxers, libp2p.EnableHolePunching())`
- Peer A is bound to a public IP
- The machine with Peer A has a firewall configured that blocks incoming connections on public interfaces, so even though `h.Addrs()` contains public IPs: 
  ```
  /ip4/127.0.0.1/tcp/35689 
  /ip4/127.0.0.1/udp/39191/webrtc-direct/certhash/uEiB5brq0KqYrJLT8O1Iq2Eo8E3gx8kThAJLs_pyncP74aA 
  /ip4/127.0.0.1/udp/39779/quic-v1/webtransport/certhash/uEiAvszmD8ETmRbFNmM1ZNPe1NDqJEGUSQku9xBhVR-n_MQ/certhash/uEiC9XsVxwVbJabPvDfAeooM4rrurb94qShBeDqmiS30Nfw 
  /ip4/127.0.0.1/udp/42886/quic-v1 
  /ip4/139.178.82.19/tcp/35689 
  /ip4/139.178.82.19/udp/39191/webrtc-direct/certhash/uEiB5brq0KqYrJLT8O1Iq2Eo8E3gx8kThAJLs_pyncP74aA 
  /ip4/139.178.82.19/udp/39779/quic-v1/webtransport/certhash/uEiAvszmD8ETmRbFNmM1ZNPe1NDqJEGUSQku9xBhVR-n_MQ/certhash/uEiC9XsVxwVbJabPvDfAeooM4rrurb94qShBeDqmiS30Nfw 
  /ip4/139.178.82.19/udp/42886/quic-v1 
  /ip6/::1/tcp/39833 
  /ip6/::1/udp/42162/quic-v1 
  /ip6/::1/udp/46915/quic-v1/webtransport/certhash/uEiAvszmD8ETmRbFNmM1ZNPe1NDqJEGUSQku9xBhVR-n_MQ/certhash/uEiC9XsVxwVbJabPvDfAeooM4rrurb94qShBeDqmiS30Nfw 
  /ip6/::1/udp/59420/webrtc-direct/certhash/uEiB5brq0KqYrJLT8O1Iq2Eo8E3gx8kThAJLs_pyncP74aA 
  /ip6/2604:1380:4642:6600::1/tcp/39833 
  /ip6/2604:1380:4642:6600::1/udp/42162/quic-v1 
  /ip6/2604:1380:4642:6600::1/udp/46915/quic-v1/webtransport/certhash/uEiAvszmD8ETmRbFNmM1ZNPe1NDqJEGUSQku9xBhVR-n_MQ/certhash/uEiC9XsVxwVbJabPvDfAeooM4rrurb94qShBeDqmiS30Nfw 
  /ip6/2604:1380:4642:6600::1/udp/59420/webrtc-direct/certhash/uEiB5brq0KqYrJLT8O1Iq2Eo8E3gx8kThAJLs_pyncP74aA]
  ```
  they are not dialable

## Problem

- Hole punching between the will fail consistently, because Peer A will not have the `/libp2p/dcutr` protocol enabled. 
- This causes a stream reset (logged by Peer A) and the `"error": "failed to open hole-punching stream: failed to negotiate protocol: protocols not supported: [/libp2p/dcutr]"}` error (logged by Peer B).

## Source of the problem

Based on debugging with @sukunrt, it appears to be related to the following lines:
- https://github.com/libp2p/go-libp2p/blob/16a27ff08cfda4ec780e6dd69e178a2031adbeab/p2p/protocol/holepunch/svc.go#L109
- https://github.com/libp2p/go-libp2p/blob/16a27ff08cfda4ec780e6dd69e178a2031adbeab/p2p/protocol/holepunch/svc.go#L174

We were able to work around this problem and successfully hole punch by replacing `s.ids.OwnObservedAddrs()` with `s.host.Addrs())`. This was a temporary measure, as we may just want to **add** the listen addrs to the observed addrs. 

## Related issue

- https://github.com/ipfs-shipyard/ipfs-check/issues/40

2color:
## What's in this PR

This change attempts to fix #2913.

In essence, it enables the `dctur` protocol, when a peer has a public IP (if it's bound to an interface with a public IP) without having to wait for an identify response (which would then make the public address visible via `OwnObservedAddrs()`).

## Open questions

- The `OwnObservedAddrs()` function in the identify service (which we rely on in the holepunch service) seems to **not** return observed addresses if they aren't dialable. Why is that? 
  - Should be fixed by https://github.com/libp2p/go-libp2p/pull/2936

### DIFF
--- a/p2p/protocol/holepunch/svc.go
+++ b/p2p/protocol/holepunch/svc.go
@@ -19,6 +19,7 @@ import (
 	"github.com/libp2p/go-msgio/pbio"
 
 	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
 )
 
 // Protocol is the libp2p protocol for Hole Punching.
@@ -106,13 +107,9 @@ func (s *Service) watchForPublicAddr() {
 	t := time.NewTimer(duration)
 	defer t.Stop()
 	for {
-		interfaceListenAddrs, err := s.host.Network().InterfaceListenAddresses()
 
-		if err != nil {
-			log.Debugf("failed to get to get InterfaceListenAddresses: %s", err)
-		}
 		// Use both host and observed to enable hole punching for undialable public ips that might not be observed
-		if containsPublicAddr(s.ids.OwnObservedAddrs()) || containsPublicAddr(interfaceListenAddrs) {
+		if len(s.getPublicAddrs()) > 0 {
 			log.Debug("Host now has a public address. Starting holepunch protocol.")
 			s.host.SetStreamHandler(Protocol, s.handleNewStream)
 			break
@@ -177,13 +174,8 @@ func (s *Service) incomingHolePunch(str network.Stream) (rtt time.Duration, remo
 	if !isRelayAddress(str.Conn().RemoteMultiaddr()) {
 		return 0, nil, nil, fmt.Errorf("received hole punch stream: %s", str.Conn().RemoteMultiaddr())
 	}
-	ownAddrs = removeRelayAddrs(s.ids.OwnObservedAddrs())
-	interfaceListenAddrs, err := s.host.Network().InterfaceListenAddresses()
 
-	if err != nil {
-		log.Debugf("failed to get to get InterfaceListenAddresses: %s", err)
-	}
-	ownAddrs = ma.Unique(append(ownAddrs, interfaceListenAddrs...))
+	ownAddrs = s.getPublicAddrs()
 
 	if s.filter != nil {
 		ownAddrs = s.filter.FilterLocal(str.Conn().RemotePeer(), ownAddrs)
@@ -285,6 +277,29 @@ func (s *Service) handleNewStream(str network.Stream) {
 	dt := time.Since(start)
 	s.tracer.EndHolePunch(rp, dt, err)
 	s.tracer.HolePunchFinished("receiver", 1, addrs, ownAddrs, getDirectConnection(s.host, rp))
+}
+
+// getPublicAddrs returns public observed and interface addresses
+func (s *Service) getPublicAddrs() []ma.Multiaddr {
+	addrs := removeRelayAddrs(s.ids.OwnObservedAddrs())
+
+	interfaceListenAddrs, err := s.host.Network().InterfaceListenAddresses()
+	if err != nil {
+		log.Debugf("failed to get to get InterfaceListenAddresses: %s", err)
+	} else {
+		addrs = append(addrs, interfaceListenAddrs...)
+	}
+
+	addrs = ma.Unique(addrs)
+
+	publicAddrs := make([]ma.Multiaddr, 0, len(addrs))
+
+	for _, addr := range addrs {
+		if manet.IsPublicAddr(addr) {
+			publicAddrs = append(publicAddrs, addr)
+		}
+	}
+	return publicAddrs
 }
 
 // DirectConnect is only exposed for testing purposes.
--- a/p2p/protocol/holepunch/svc.go
+++ b/p2p/protocol/holepunch/svc.go
@@ -106,7 +106,8 @@ func (s *Service) watchForPublicAddr() {
 	t := time.NewTimer(duration)
 	defer t.Stop()
 	for {
-		if containsPublicAddr(s.ids.OwnObservedAddrs()) {
+		// Use both host and observed to enable hole punching for undialable public ips that might not be observed
+		if containsPublicAddr(s.ids.OwnObservedAddrs()) || containsPublicAddr(s.host.Addrs()) {
 			log.Debug("Host now has a public address. Starting holepunch protocol.")
 			s.host.SetStreamHandler(Protocol, s.handleNewStream)
 			break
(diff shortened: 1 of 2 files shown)


p2p/protocol/holepunch/svc.go
```diff
@@ -106,7 +106,8 @@ func (s *Service) watchForPublicAddr() {
 	t := time.NewTimer(duration)
 	defer t.Stop()
 	for {
-		if containsPublicAddr(s.ids.OwnObservedAddrs()) {
+		// Use both host and observed to enable hole punching for undialable public ips that might not be observed
+		if containsPublicAddr(s.ids.OwnObservedAddrs()) || containsPublicAddr(s.host.Addrs()) {
```
sukunrt, line 110: This won't work with hosts that have autorelay enabled. https://github.com/libp2p/go-libp2p/blob/master/p2p/host/autorelay/relay_finder.go#L728
Autorelay wraps host.Addrs and filters out all public host addresses and adds relayed addresses. 

Instead, we can do, `host.Network().ListenAddresses()` and `host.Network().InterfaceListenAddresses()`

p2p/protocol/holepunch/svc.go
```diff
@@ -172,6 +173,7 @@ func (s *Service) incomingHolePunch(str network.Stream) (rtt time.Duration, remo
 		return 0, nil, nil, fmt.Errorf("received hole punch stream: %s", str.Conn().RemoteMultiaddr())
 	}
 	ownAddrs = removeRelayAddrs(s.ids.OwnObservedAddrs())
+	ownAddrs = append(ownAddrs, s.host.Addrs()...)
```
sukunrt, line 176: 1. remove relay addresses here too. After changing `host.Addrs()` to `host.Network().ListenAddrs...`
2. Do a unique after adding all the addresses.

sukunrt (changes requested):
Thanks @2color

p2p/protocol/holepunch/svc.go
sukunrt: If using `Network().ListenAddress` etc  you'll also have to filter for public addresses.

MarcoPolo:
I have some questions about the issue. Let's continue the discussion there. But one quick thing:

> The OwnObservedAddrs() function in the identify service (which we rely on in the holepunch service) seems to not return observed addresses if they aren't dialable. Why is that?

What do you mean by dialable? By not dialable do you mean that the host will drop unsolicited incoming packets?

p2p/protocol/holepunch/svc.go
2color: What's the difference between `host.Network().ListenAddresses()` and `host.Network().InterfaceListenAddresses()`?

Jorropo: The transport listen on `/ip4/0.0.0.0/tcp/12345` then `InterfaceListenAddresses()` use https://pkg.go.dev/net#Interfaces to explode this to:
- `/ip4/127.0.0.1/tcp/12345`
- `/ip4/192.168.1.123/tcp/12345`
- `/ip4/10.13.37.42/tcp/12345`

(numbers made up for example sake, varies based on the machine config)

2color: Given that, we probably only want to rely on `host.Network().InterfaceListenAddresses()` here to get the actual IPs of the interfaces. 

Or is there a case where we want to also check `host.Network().ListenAddresses()`?

2color:
> By not dialable do you mean that the host will drop unsolicited incoming packets?

Exactly. It binds to a port on an interface with a public IP, however that port is blocked by the firewall.

p2p/protocol/holepunch/svc.go
2color: Here too, we should just rely on `host.Network().InterfaceListenAddresses()`

2color:
I've addressed your feedback @sukunrt.

p2p/protocol/holepunch/svc.go
```diff
@@ -172,6 +178,13 @@ func (s *Service) incomingHolePunch(str network.Stream) (rtt time.Duration, remo
 		return 0, nil, nil, fmt.Errorf("received hole punch stream: %s", str.Conn().RemoteMultiaddr())
 	}
 	ownAddrs = removeRelayAddrs(s.ids.OwnObservedAddrs())
+	interfaceListenAddrs, err := s.host.Network().InterfaceListenAddresses()
+
+	if err != nil {
+		log.Debugf("failed to get to get InterfaceListenAddresses: %s", err)
+	}
+	ownAddrs = ma.Unique(append(ownAddrs, interfaceListenAddrs...))
```
sukunrt, line 186: We need to filter for public addresses in the list. 

I suggest making a method
```
 func (s *Service) getPublicAddrs() []ma.Multiaddr 
``` 

and using it in both the places.

sukunrt (changes requested):

sukunrt:
@MarcoPolo The change looks reasonable to me. If the node does have access to its own public address there's no reason to depend on identify for address discovery for hole punching.

MarcoPolo (commented):
This change makes sense. After addressing @sukunrt's comment this should be good to merge

p2p/protocol/holepunch/svc.go
2color: Addressed

MarcoPolo (approved):

PR #2934, opened 2024-08-22
fix: deadlock on panic in transport testsuite

Jorropo:
I had a panic in `list.Multiaddr()` which would deadlock the test instead of properly showing up the stack trace.

This happens because in the defer block it waits for <-done however because the test never proceed the other goroutine is stuck on `.Accept()`, never sending `close(done)`.


MarcoPolo:

PR #2935, opened 2024-08-22
fix: reuse the quic port in webtransport

2color:
This sets the ListenOrder for WebTransport so that it starts listening after QUIC, thereby allowing port reuse when not listening on an explicit port. 

## Related bugs

Related to #2913 and fixes a bug whereby a libp2p peer doesn't register an observed public IP, because the outgoing connection uses the WebTransport port which is different to the QUIC port.

### DIFF
--- a/p2p/transport/webtransport/transport.go
+++ b/p2p/transport/webtransport/transport.go
@@ -20,6 +20,7 @@ import (
 	tpt "github.com/libp2p/go-libp2p/core/transport"
 	"github.com/libp2p/go-libp2p/p2p/security/noise"
 	"github.com/libp2p/go-libp2p/p2p/security/noise/pb"
+	libp2pquic "github.com/libp2p/go-libp2p/p2p/transport/quic"
 	"github.com/libp2p/go-libp2p/p2p/transport/quicreuse"
 
 	"github.com/benbjohnson/clock"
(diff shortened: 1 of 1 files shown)


sukunrt:
We don't need this change. The current behavior of using different ports when we listen on 
```
/ip4/0.0.0.0/udp/0/quic-v1
/ip4/0.0.0.0/udp/0/quic-v1/webtransport
```
is the expected behavior and also one that on changing might break existing users. 

To reuse the same port, either pass a port explicitly or listen on port 0 for quic first and then listen on webtransport with the port allocated to the quic listener.

2color:
Fair enough.  

If we allow WebTransport and QUIC to be allocated different ports, we still need to fix the bug whereby go-libp2p fails to [correctly verify an observed ip as observed in this comment](https://github.com/libp2p/go-libp2p/issues/2913#issuecomment-2304541410), because the request goes out from the WebTransport port which then doesn't match the QUIC one (as we also noticed in our testing @sukunrt)

Any thoughts on that? Should we open a separate issue for that?

MarcoPolo:
> we still need to fix the bug whereby go-libp2p fails to [correctly verify an observed ip as observed in this comment](https://github.com/libp2p/go-libp2p/issues/2913#issuecomment-2304541410), because the request goes out from the WebTransport port which then doesn't match the QUIC one

Fixed in #2936

2color:
> To reuse the same port, either pass a port explicitly or listen on port 0 for quic first and then listen on webtransport with the port allocated to the quic listener.

Naive question, if this change is not necessary, why do we define `ListenOrder` for the WebRTC transport?

https://github.com/libp2p/go-libp2p/blob/412daa471815271b8b15bf25c3c755e3cbc17a2e/p2p/transport/webrtc/transport.go#L166-L168

2color:
Got it. So this can be closed?

MarcoPolo:
> is the expected behavior and also one that on changing might break existing users.

One more thought on this. I think we _can_ change this, but we should just do it deliberately through an option this is default off, and at some point default on. (like all breaking changes).

PR #2936, opened 2024-08-22
fix: QUIC/Webtransport Transports now will prefer their owned listeners for dialing out

### LINKED ISSUE #2913: QUIC/WebTransport Transport sometimes uses the wrong socket to dial out.

MarcoPolo:
Fixes #2913 

The issue is presented in https://github.com/libp2p/go-libp2p/issues/2913#issuecomment-2305818394.

The fix here introduces "associations" to refcountedTransports. a libp2p transport can associate itself with the transport used for listening, and then prefer to use that for dialing by passing itself as the association.

### DIFF
--- a/libp2p_test.go
+++ b/libp2p_test.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/netip"
 	"regexp"
@@ -586,4 +587,71 @@ func TestWebRTCReuseAddrWithQUIC(t *testing.T) {
 		require.Equal(t, 1, len(h1.Addrs()))
 		require.Contains(t, h1.Addrs()[0].String(), "quic-v1")
 	})
+}
+
+func TestUseCorrectTransportForDialOut(t *testing.T) {
+	listAddrOrder := [][]string{
+		{"/ip4/127.0.0.1/udp/0/quic-v1", "/ip4/127.0.0.1/udp/0/quic-v1/webtransport"},
+		{"/ip4/127.0.0.1/udp/0/quic-v1/webtransport", "/ip4/127.0.0.1/udp/0/quic-v1"},
+		{"/ip4/0.0.0.0/udp/0/quic-v1", "/ip4/0.0.0.0/udp/0/quic-v1/webtransport"},
+		{"/ip4/0.0.0.0/udp/0/quic-v1/webtransport", "/ip4/0.0.0.0/udp/0/quic-v1"},
+	}
+	for _, order := range listAddrOrder {
+		h1, err := New(ListenAddrStrings(order...), Transport(quic.NewTransport), Transport(webtransport.New))
+		require.NoError(t, err)
+		t.Cleanup(func() {
+			h1.Close()
+		})
+
+		go func() {
+			h1.SetStreamHandler("/echo-port", func(s network.Stream) {
+				m := s.Conn().RemoteMultiaddr()
+				v, err := m.ValueForProtocol(ma.P_UDP)
+				if err != nil {
+					s.Reset()
+					return
+				}
+				s.Write([]byte(v))
+				s.Close()
+			})
+		}()
+
+		for _, addr := range h1.Addrs() {
+			t.Run("order "+strings.Join(order, ",")+" Dial to "+addr.String(), func(t *testing.T) {
+				h2, err := New(ListenAddrStrings(
+					"/ip4/0.0.0.0/udp/0/quic-v1",
+					"/ip4/0.0.0.0/udp/0/quic-v1/webtransport",
+				), Transport(quic.NewTransport), Transport(webtransport.New))
+				require.NoError(t, err)
+				defer h2.Close()
+				t.Log("H2 Addrs", h2.Addrs())
+				var myExpectedDialOutAddr ma.Multiaddr
+				addrIsWT, _ := webtransport.IsWebtransportMultiaddr(addr)
+				isLocal := func(a ma.Multiaddr) bool {
+					return strings.Contains(a.String(), "127.0.0.1")
+				}
+				addrIsLocal := isLocal(addr)
+				for _, a := range h2.Addrs() {
+					aIsWT, _ := webtransport.IsWebtransportMultiaddr(a)
+					if addrIsWT == aIsWT && isLocal(a) == addrIsLocal {
+						myExpectedDialOutAddr = a
+						break
+					}
+				}
+
+				err = h2.Connect(context.Background(), peer.AddrInfo{ID: h1.ID(), Addrs: []ma.Multiaddr{addr}})
+				require.NoError(t, err)
+
+				s, err := h2.NewStream(context.Background(), h1.ID(), "/echo-port")
+				require.NoError(t, err)
+
+				port, err := io.ReadAll(s)
+				require.NoError(t, err)
+
+				myExpectedPort, err := myExpectedDialOutAddr.ValueForProtocol(ma.P_UDP)
+				require.NoError(t, err)
+				require.Equal(t, myExpectedPort, string(port))
+			})
+		}
+	}
 }
--- a/p2p/transport/quic/transport.go
+++ b/p2p/transport/quic/transport.go
@@ -136,6 +136,7 @@ func (t *transport) dialWithScope(ctx context.Context, raddr ma.Multiaddr, p pee
 	}
 
 	tlsConf, keyCh := t.identity.ConfigForPeer(p)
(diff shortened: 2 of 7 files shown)


p2p/transport/webtransport/listener.go
```diff
@@ -295,5 +299,15 @@ loop:
 			break loop
 		}
 	}
+
+	l.transport.listenersMu.Lock()
+	for i, ln := range l.transport.listeners {
+		if ln == l {
+			l.transport.listeners = append(l.transport.listeners[:i], l.transport.listeners[i+1:]...)
```
sukunrt, line 306: you can use `slices.DeleteFunc`

p2p/transport/webtransport/transport.go
```diff
@@ -207,10 +231,28 @@ func (t *transport) dial(ctx context.Context, addr ma.Multiaddr, url, sni string
...
+		clientConf.AllowConnectionWindowIncrease = t.allowWindowIncrease
+		naddr, _, err := quicreuse.FromQuicMultiaddr(addr)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		conn, err = listener.reuseListener.Dial(ctx, naddr, tlsConf, clientConf)
+		if err != nil {
+			return nil, nil, err
+		}
```
sukunrt, line 248: Should we expose this as a method in `ConnManager` as `DialWithListener` or `DialWithTransport` or something. I am specifically interested in keeping this logic in one place. 

```
clientConf := t.connManager.ClientConfig().Clone()
clientConf.AllowConnectionWindowIncrease = t.allowWindowIncrease
naddr, _, err := quicreuse.FromQuicMultiaddr(addr)
if err != nil {
	return nil, nil, err
}
```

p2p/transport/quic/transport.go
```diff
@@ -129,16 +129,56 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (_c
...
+		return nil, err
+	}
+	t.listenersMu.Lock()
+	defer t.listenersMu.Unlock()
+	for _, ls := range t.listeners {
+		for _, l := range ls {
+			lnetw, _, err := manet.DialArgs(l.Multiaddr())
+			if err != nil {
+				continue
+			}
```
sukunrt, line 146: should we log this?

libp2p_test.go
```diff
@@ -587,3 +588,61 @@ func TestWebRTCReuseAddrWithQUIC(t *testing.T) {
 		require.Contains(t, h1.Addrs()[0].String(), "quic-v1")
 	})
 }
+
+func TestUseCorrectTransportForDialOut(t *testing.T) {
+	listAddrOrder := [][]string{
+		{"/ip4/127.0.0.1/udp/0/quic-v1", "/ip4/127.0.0.1/udp/0/quic-v1/webtransport"},
+		{"/ip4/127.0.0.1/udp/0/quic-v1/webtransport", "/ip4/127.0.0.1/udp/0/quic-v1"},
+	}
```
sukunrt, line 596: Can you add a test case for Listening on `0.0.0.0` too?

sukunrt (approved):

p2p/transport/webtransport/transport.go
MarcoPolo: It would be a small benefit for reducing reuse here, but at the expense of introducing a new public method in ConnManager. If ConnManager was internal, I think that would be okay. But I don't think the reuse here is worth the tradeoff.

MarcoPolo:
I think there's an issue here where we don't have any logic to pick the interface that can actually route the packet.

MarcoPolo:
I've changed this PR completely. I think it's a little cleaner now. Minimal changes to the transports, and leveraging existing logic in the quicreuse package.

Another review please @sukunrt ?

sukunrt (approved):
This is definitely nicer. Thanks!

ethan-gallant:
@MarcoPolo I think this may have broken hole punching as it seems to prefer the port which the relay signaled on opposed to the higher available hole punch ports.

ethan-gallant:
Notably removing this line resolves the issue where it tries to establish the hole punch from the origin port `443` even though that port is filtered out in the local and remote available addresses for the punch.

https://github.com/libp2p/go-libp2p/blob/0fde4dcb01d83117495f0c920c34c8f540467d4c/p2p/transport/webtransport/transport.go#L210

MarcoPolo:
@ethan-gallant, can you please open an issue here and outline what your setup is? I don't want this to get lost. Thanks!

PR #2948, opened 2024-09-03
chore: parameterise s3 build cache setup

galargh:
As we're setting up a new cache bucket, we'd like to be able to control its' configuration via GitHub vars/secrets fully.


sukunrt (approved):

sukunrt:
Thanks @galargh

Issue #2958, opened 2024-09-11
Provide metrics for in progress dials


sukunrt:
We have metrics in swarm for connections that are opened, closed, handshake latency etc. However, these don't provide much insight in to ongoing dials. We should add metrics for:
1. In progress dials by transport
2. Queued dials by transport

Thanks @gammazero for pointing this out.

sukunrt:
fixed by #2959

PR #2959, opened 2024-09-12
swarm: add a peer dial latency metric

sukunrt:
Part of #2958

### DIFF
--- a/p2p/net/swarm/swarm_metrics.go
+++ b/p2p/net/swarm/swarm_metrics.go
@@ -77,6 +77,15 @@ var (
 		},
 		[]string{"outcome", "num_dials"},
 	)
+	dialLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metricNamespace,
+			Name:      "dial_latency_seconds",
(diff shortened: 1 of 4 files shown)


p2p/net/swarm/swarm_metrics.go
```diff
@@ -77,6 +77,15 @@ var (
 		},
 		[]string{"outcome", "num_dials"},
 	)
+	dialLatency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metricNamespace,
+			Name:      "dial_latency_seconds",
+			Help:      "time taken to establish connection with the peer",
+			Buckets:   []float64{0.001, 1.3, 35},
```
MarcoPolo, line 85: Where are these numbers coming from?

MarcoPolo (approved):

p2p/net/swarm/swarm_metrics.go
sukunrt: I copied it from somewhere :sweat_smile: 
The range is 1ms to 10s

PR #2960, opened 2024-09-13
fix: Better GC in membacked peerstore

### LINKED ISSUE #1698: peerstore: memoryAddrBook GC interval is hard-coded
I'm writing an application that uses very small TTLs in order to quickly detect unavailable peers.  The TTL values are on the order of seconds.

Currently, `memoryAddrBook` waits 1h between GC cycles, which is unacceptably long.  A naive solution would be make the [hard-coded value](https://github.com/libp2p/go-libp2p-peerstore/blob/b1f58c0365c0a8af3928cb1937d95e2554cb89a6/pstoremem/addr_book.go#L89) configurable, but this is likely to be very costly at low interval values.  I think we should use a slightly more sophisticated method.

I've evaluated a few different approaches, which fall into two rough categories:

1. Remove the need for periodic polling (Approach 1)
2. Make periodic polling very cheap (Approaches 2 -- 4)

I'm wondering whether any of these are good fits for `memoryAddrBook`, and I'm happy to contribute a patch if there's any interest.

### Approach 1:  Let the Go runtime handle timeouts

The Go runtime is surprisingly smart.  The runtime maintains a heap of [`runtime.timer`](https://github.com/golang/go/blob/release-branch.go1.7/src/runtime/time.go#L15) instances, min-ordered by expiration time, and sleeps until the first timer fires.  I'm curious to know whether this approach was ever attempted, and if so, whether it was ever benchmarked.  In my experience, deferring to the Go runtime in such matters is almost always the best solution.

Here's a rough sketch of what this would look like.  Note that my example doesn't make use of sharding (as with `addrSegments`), though there is nothing preventing us from doing so.  Note also how timers are reused.

```go
type addrSet struct {
    sync.Mutex
    as map[peer.ID]*entry
}

type entry struct{
    set *addrSet

    timerLock sync.Mutex
    timer *time.Timer

    key peer.ID
    addrs map[string]multiaddr.Multiaddr
}
```

Any time a peer's addresses/TTL is updated, you need to run `entry.setExpire`:

```go
func (e *entry) setExpire(d time.Duration) {
    e.timerLock.Lock()
    defer e.timerLock.Unlock()

    // if this entry was just created, there won't be a TTL job yet
    if e.timer != nil {
        // cancel the existing TTL expiration job
        if !e.timer.Stop() {
            // The timer fired right before this function was called, so we need to re-add it.
            // Assumes setExpire's caller holds the addrSet lock.
            e.set.as[e.key] = e
        }
    }

    e.ttlTimer = time.AfterFunc(TTL, e.expire)
}

func (e *entry) expire() {
    e.set.Lock()
    defer e.set.Unlock()
    delete(e.set.as, e.key)
}
```
There are so low-hanging optimizations we can do here.  The above is just trying to paint a "big picture".

The potential drawback I see is that the runtime's timer heap is protected by a mutex.  At _some_ scale, I'd expect contention to become an issue, and there's little we can do about it.  That said, I'm not sure what that inflection point actually is, and in practice, I've never encountered it.

### Approach 2:  Heap-based expiration

As mentioned above, we can continue to run a periodic GC task, and focus on making it cheap enough to run often.  A 1h interval probably makes sense for most applications, but it would ideally be _feasible_ to lower the interval to the order of seconds (or even tens of milliseconds).

This approach would involve maintaining a heap alongside the map of expiring addrs.  The GC loop would then proceed to do the following on each iteration:

```go
func runGC(now time.Time) {
    for heap.Peek().Deadline.Before(now) {
        deleteAddrsFromAddrBook(heap.Pop().ID)
    }
}
```

Performance improvements here are going to come from avoiding full map iterations on each GC run.  Mutex Un/Locks are on the order of tens of nanoseconds on modern hardware , so are unlikely to be problematic.  This solution is also amenable to sharding.

### Approach 3:  Treap-based expiration

We can use a similar approach as above, substituting the map+heap for a persistent treap + CAS.  The trade-off here is twofold:

- persistent treaps need to allocate each time they are modified
- although they perform well for uniformly-distributed weights (O log n), they fall to O(n) in the worst-case for non-random weights.

In practice, this is likely to be quite slow.  I mention it only for completeness, and in case someone thinks of a clever optimization.

### Approach 4:  Hierarchical-Timing-Wheel-based expiration

Instead of a heap, we can maintain one or more timing wheels.  Inserting/Updating entries involves incrementing a counter and scheduling it's decrementation some time in the future.  When the counter reaches zero, we delete the entry from the map.

HTWs have an amortized time-complexity of O(1).

MarcoPolo:
Closes #1698 

Makes GC ~60x faster (on my synthetic benchmark) under certain circumstances. So by that logic, it should be fine to run this more often.

Adds a min heap list to track the next address to expire. This lets us short circuit gc if we see the next address to expire isn't expired yet.

Also bumps the gc interval from 60 minutes to 1 minute. This should be fine since we don't traverse every address anymore.

This relies on existing tests to assert nothing has broken.

Of course there are many more improvements that could be made. See the linked issue for inspiration. The next lowest hanging fruit is to do something similar to this across peers.

### DIFF
--- a/p2p/host/peerstore/pstoremem/addr_book.go
+++ b/p2p/host/peerstore/pstoremem/addr_book.go
@@ -21,16 +21,16 @@ var SignedPeerRecordBound = 100_000
 var log = logging.Logger("peerstore")
 
 type expiringAddr struct {
-	Addr    ma.Multiaddr
-	TTL     time.Duration
-	Expires time.Time
-	Peer    peer.ID
+	Addr   ma.Multiaddr
+	TTL    time.Duration
+	Expiry time.Time
+	Peer   peer.ID
 	// to sort by expiry time
 	heapIndex int
 }
 
 func (e *expiringAddr) ExpiredBy(t time.Time) bool {
-	return !t.Before(e.Expires)
+	return !t.Before(e.Expiry)
 }
 
 type peerRecordState struct {
@@ -54,7 +54,7 @@ func newPeerAddrs() peerAddrs {
 
 func (pa *peerAddrs) Len() int { return len(pa.expiringHeap) }
 func (pa *peerAddrs) Less(i, j int) bool {
-	return pa.expiringHeap[i].Expires.Before(pa.expiringHeap[j].Expires)
+	return pa.expiringHeap[i].Expiry.Before(pa.expiringHeap[j].Expiry)
 }
 func (pa *peerAddrs) Swap(i, j int) {
 	pa.expiringHeap[i], pa.expiringHeap[j] = pa.expiringHeap[j], pa.expiringHeap[i]
@@ -81,7 +81,6 @@ func (pa *peerAddrs) Pop() any {
 			delete(pa.Addrs, a.Peer)
 		}
 	}
-
 	return a
 }
 
@@ -112,11 +111,11 @@ func (pa *peerAddrs) NextExpiry() time.Time {
 	if len(pa.expiringHeap) == 0 {
 		return time.Time{}
 	}
-	return pa.expiringHeap[0].Expires
+	return pa.expiringHeap[0].Expiry
 }
 
 func (pa *peerAddrs) PopIfExpired(now time.Time) (*expiringAddr, bool) {
-	// Use !Before and not After to ensure that we expire *at* now and not just after now.
+	// Use `!Before` instead of `After` to ensure that we expire *at* now, and not *just after now*.
 	if len(pa.expiringHeap) > 0 && !now.Before(pa.NextExpiry()) {
 		a := heap.Pop(pa)
 		return a.(*expiringAddr), true
@@ -304,7 +303,7 @@ func (mab *memoryAddrBook) addAddrsUnlocked(p peer.ID, addrs []ma.Multiaddr, ttl
 		a, found := mab.addrs.FindAddr(p, addr)
 		if !found {
 			// not found, announce it.
-			entry := &expiringAddr{Addr: addr, Expires: exp, TTL: ttl, Peer: p}
+			entry := &expiringAddr{Addr: addr, Expiry: exp, TTL: ttl, Peer: p}
 			heap.Push(&mab.addrs, entry)
 			mab.subManager.BroadcastAddr(p, addr)
 		} else {
@@ -314,9 +313,9 @@ func (mab *memoryAddrBook) addAddrsUnlocked(p peer.ID, addrs []ma.Multiaddr, ttl
 				changed = true
 				a.TTL = ttl
 			}
-			if exp.After(a.Expires) {
+			if exp.After(a.Expiry) {
 				changed = true
-				a.Expires = exp
+				a.Expiry = exp
 			}
 			if changed {
 				mab.addrs.Fix(a)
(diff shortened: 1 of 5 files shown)


p2p/host/peerstore/pstoremem/addr_book.go
```diff
@@ -109,7 +163,7 @@ func WithClock(clock clock) AddrBookOption {
 // background periodically schedules a gc
 func (mab *memoryAddrBook) background(ctx context.Context) {
 	defer mab.refCount.Done()
-	ticker := time.NewTicker(1 * time.Hour)
+	ticker := time.NewTicker(1 * time.Minute)
 	defer ticker.Stop()
```
sukunrt, line 167: Can you replace this with a timer and a signal channel? That would mean we don't have to run this every minute and simply run it when the first element of the heap expires.

MarcoPolo: I like the idea. I've implemented it on the branch `marco/better-peerstore-gc-and-timer`. Testing it though requires a bit more changes to use our mockclock instead of the stdlib timer and then updating the existing tests to use it. Doable, but a bit of a hassle.

Let's do it in a follow up PR.

p2p/host/peerstore/pstoremem/addr_book.go
```diff
@@ -32,24 +38,89 @@ type peerRecordState struct {
...
 }
 
-type addrSegments [256]*addrSegment
+// Essentially Go stdlib's Priority Queue example
+var _ heap.Interface = &peerAddrs{}
 
-type addrSegment struct {
-	sync.RWMutex
+type peerAddrs struct {
+	addrs        map[peer.ID]map[string]*expiringAddr // peer.ID -> addr.Bytes() -> *expiringAddr
```
sukunrt, line 45: Should be public, as it's accessed from outside.

p2p/host/peerstore/pstoremem/addr_book.go
```diff
@@ -114,11 +114,12 @@ func (pa *peerAddrs) NextExpiry() time.Time {
 	if len(pa.expiringHeap) == 0 {
 		return time.Time{}
 	}
-	return pa.expiringHeap[len(pa.expiringHeap)-1].Expires
+	return pa.expiringHeap[0].Expires
```
MarcoPolo, line 117: Are you sure this is right? The min value should be at the end, no?

sukunrt: Yes, I added a couple of tests. See: `TestPeerAddrsNextExpiry`

If you're confused, as I was, by the fact that `heap.Interface.Pop` requires removal from the end, that's because `heap.Pop` swaps the first and last element before calling `heap.Interface.Pop`. 

from container/heap/heap.go
```
// Pop removes and returns the minimum element (according to Less) from the heap.
// The complexity is O(log n) where n = h.Len().
// Pop is equivalent to [Remove](h, 0).
func Pop(h Interface) any {
	n := h.Len() - 1
	h.Swap(0, n)
	down(h, 0, n)
	return h.Pop()
}
```

sukunrt:
I've removed the change that limited the number of signed peer records we could store in the peer store. I removed the limit because:
1. There's no way to opt out of it. 
2. A rate limiter for new connections from the same IP / subnet is a better solution to the underlying problem. 
3. If we don't limit addresses in the peer store, the benefits are small. 

I suggest we introduce the limit in a separate PR, and provide an option to opt out of the limit.

sukunrt (approved):

p2p/host/peerstore/pstoremem/addr_book.go
MarcoPolo: ah, thanks. That makes sense.

MarcoPolo:
Changes look good. I appreciate the extra tests. Thank you.

PR #2961, opened 2024-09-13
connmgr: reduce log level for untagging untracked peers

burdiyan:
Method UntagPeer() logs an "info" log message whenever it's called for a peer that is untracked. It generates a lot of noise in the logs for something that is basically harmless.

This commit reduces the log level to "debug" for this message.


vyzo (approved):
thank you!

PR #2962, opened 2024-09-13
Feat: Expose Fx options directly

MarcoPolo:
This allows many things:

- Easily build libp2p in a modular way, relying on Fx to fill in the dependency graph.
- Allow user services to get access to shared resources like the Peerstore, or other services like Idenetify without having to expose a method in the Host interface for all these things. Or without having the user have to cast the `host.Host` into a basic host.
- Allow users who don't need resources like a Peerstore to avoid allocating one.
- In the case that the user is already using Fx, it plugs neatly into their constructor. In case it isn't, it should still be just as easy to construct the libp2p host as it is today.
- Better support for "setup packs". Preconfigured options for certain use-cases (e.g. everything please; I'm a public server; absolute bare minimum)

In draft until we get close to feature parity with the existing constructor. Please let me know if you'd be interested in trying this when we release it. (In particular I think this would significantly clean up Kubo's libp2p constructor).

### DIFF
--- a/fx/options.go
+++ b/fx/options.go
@@ -30,6 +30,7 @@ import (
 	libp2pquic "github.com/libp2p/go-libp2p/p2p/transport/quic"
 	"github.com/libp2p/go-libp2p/p2p/transport/quicreuse"
 	"github.com/libp2p/go-libp2p/p2p/transport/tcp"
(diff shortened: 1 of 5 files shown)


MarcoPolo:

Issue #2964, opened 2024-09-17
Bug: autorelay: enabling autorelay overrides custom AddrFactory


burdiyan:
In this line, autorelay overwrites the AddrFactory on the basic host: https://github.com/libp2p/go-libp2p/blob/921cc71072229344a78c7d794e3571c3cbf058e3/p2p/host/autorelay/autorelay.go#L54

It supposedly "caches" the original AddrFactory here: https://github.com/libp2p/go-libp2p/blob/921cc71072229344a78c7d794e3571c3cbf058e3/p2p/host/autorelay/autorelay.go#L41

but it seems like this happens before the custom `libp2p.AddrFactory` option is being applied on the libp2p host. I presume that `NewAutorelay` function is being passed a partially built instance of the basic host, which is probably due to some of the `fx` dependency injection magic.

In my custom AddrFactory option I add a public IP address that I discovery via STUN manually, and it's not being returned by calling `host.Addrs()` method unless I disable AutoRelay option.

PR #2967, opened 2024-09-18
autonat: fix interaction with autorelay

sukunrt:
Fixes: #2964

### DIFF
--- a/p2p/host/autonat/autonat.go
+++ b/p2p/host/autonat/autonat.go
@@ -261,7 +261,7 @@ func (as *AmbientAutoNAT) checkAddrs() (hasNewAddr bool) {
 		}
 		as.ourAddrs[string(a.Bytes())] = struct{}{}
 	}
-	return false
+	return hasNewAddr
 }
 
 // scheduleProbe calculates when the next probe should be scheduled for.
--- a/p2p/host/autonat/autonat.go
+++ b/p2p/host/autonat/autonat.go
@@ -33,6 +33,8 @@ type AmbientAutoNAT struct {
 
 	inboundConn   chan network.Conn
 	dialResponses chan error
+	// Used when testing the autonat service
+	observations chan network.Reachability
 	// status is an autoNATResult reflecting current status.
 	status atomic.Pointer[network.Reachability]
 	// Reflects the confidence on of the NATStatus being private, as a single
@@ -113,6 +115,7 @@ func New(h host.Host, options ...Option) (AutoNAT, error) {
 		config:            conf,
 		inboundConn:       make(chan network.Conn, 5),
 		dialResponses:     make(chan error, 1),
+		observations:      make(chan network.Reachability, 1),
 
 		emitReachabilityChanged: emitReachabilityChanged,
 		service:                 service,
@@ -201,6 +204,9 @@ func (as *AmbientAutoNAT) background() {
 			default:
 				log.Errorf("unknown event type: %T", e)
 			}
+		case obs := <-as.observations:
+			as.recordObservation(obs)
+			continue
 		case err, ok := <-as.dialResponses:
 			if !ok {
 				return
--- a/p2p/host/autonat/autonat.go
+++ b/p2p/host/autonat/autonat.go
@@ -174,10 +174,9 @@ func (as *AmbientAutoNAT) background() {
 	defer as.subscriber.Close()
 	defer as.emitReachabilityChanged.Close()
 
-	// We want an update when our public non-relay listen addresses have changed.
-	// EvtLocalAddressesUpdated is a poor proxy for that. It works when the host is Public,
-	// but fails when the host is private and used with AutoRelay.
-	addrChangeTicker := time.NewTicker(1 * time.Minute)
+	// Fallback timer to update address in case EvtLocalAddressesUpdated is not emitted.
+	// TODO: The event not emitting properly is a bug. This is a workaround.
+	addrChangeTicker := time.NewTicker(30 * time.Minute)
 	defer addrChangeTicker.Stop()
 
 	timer := time.NewTimer(delay)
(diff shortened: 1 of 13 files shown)


p2p/host/autonat/autonat.go
```diff
@@ -165,117 +170,126 @@ func (as *AmbientAutoNAT) background() {
...
-	for k, v := range as.recentProbes {
-		if fixedNow.Sub(v) > as.throttlePeerPeriod {
-			delete(as.recentProbes, k)
+func (as *AmbientAutoNAT) checkAddrs() (hasNewAddr bool) {
+	currentAddrs := as.addressFunc()
+	for _, a := range currentAddrs {
+		if !manet.IsPublicAddr(a) {
+			continue
+		}
+		if _, ok := as.ourAddrs[string(a.Bytes())]; !ok {
```
MarcoPolo, line 246: We only need to check if `a` is in the `ourAddrs` set. We don't need the IsPublicAddr function call. Also we can use slices.IndexFunc for a bit of semantic legibility.

p2p/host/autonat/autonat.go
```diff
@@ -165,117 +170,126 @@ func (as *AmbientAutoNAT) background() {
...
-			nextProbe = backoff
-		}
+	nextProbeAfter := as.config.refreshInterval
+	if forceProbe && currentStatus == network.ReachabilityUnknown {
+		// retry very quicky if forceProbe is true *and* we don't know our reachability
+		// limit all peers fetch from peerstore to 1 per second.
+		nextProbeAfter = 2 * time.Second
+	} else if currentStatus == network.ReachabilityUnknown ||
+		as.confidence < maxConfidence ||
+		(currentStatus != network.ReachabilityPublic && as.lastInbound.After(as.lastProbe)) {
```
MarcoPolo, line 273: Minor nit (optional). This might be cleaner with a switch

p2p/host/autonat/autonat.go
```diff
@@ -165,117 +170,126 @@ func (as *AmbientAutoNAT) background() {
...
+		// Retry quickly in case:
+		// 1. Our reachability is Unknown
+		// 2. We don't have enough confidence in our reachability.
+		// 3. We're private but we received an inbound connection.
+		nextProbeAfter = as.config.retryInterval
+	} else if currentStatus == network.ReachabilityPublic && as.lastInbound.After(as.lastProbe) {
+		// We are public and we received an inbound connection recently,
+		// wait a little longer
+		nextProbeAfter *= 2
```
MarcoPolo, line 282: We should set some max limit here

p2p/host/autonat/autonat.go
```diff
@@ -399,33 +399,30 @@ func (as *AmbientAutoNAT) getPeerToProbe() peer.ID {
...
+	fixedNow := time.Now()
+	for k, v := range as.recentProbes {
+		if fixedNow.Sub(v) > as.throttlePeerPeriod {
+			delete(as.recentProbes, k)
+		}
+	}
 
-	for _, p := range peers {
+	n := len(peers)
+	for i, j := rand.Intn(n), 0; j < n; i, j = (i+1)%n, j+1 {
```
MarcoPolo, line 411: I don't understand the point of starting at a random offset. First, the code is slightly confusing. But, more importantly, you probably want to actually shuffle the list, rather than starting at a random offset. Shuffling correctly is surprisingly [subtle](https://blog.codinghorror.com/the-danger-of-naivete/).

p2p/host/autonat/autonat.go
```diff
@@ -399,33 +399,30 @@ func (as *AmbientAutoNAT) getPeerToProbe() peer.ID {
 		return ""
 	}
 
-	candidates := make([]peer.ID, 0, len(peers))
+	// clean old probes
```
MarcoPolo, line 402: ~Do we need to decrement pending probes?~

Nevermind, this is about our throttling, not about outstanding requests. May help to wordsmith the comment.

MarcoPolo (changes requested):
I don't love this service. It's very messy, interacts with the basic host in deep ways, and seems hard to test.

It'll be nice when we drop it in favor of AutoNAT V2.

MarcoPolo (approved):

p2p/host/autonat/autonat.go
sukunrt: We are only interested in the first element that satisfies our condition, so we I think it was fine to not shuffle the whole thing. 

Separately, theres `rand.Shuffle` if you do want to shuffle.

p2p/host/autonat/autonat.go
```diff
@@ -165,117 +170,126 @@ func (as *AmbientAutoNAT) background() {
 	defer as.subscriber.Close()
 	defer as.emitReachabilityChanged.Close()
 
+	// We want an update when our public non-relay listen addresses have changed.
+	// EvtLocalAddressesUpdated is a poor proxy for that. It works when the host is Public,
+	// but fails when the host is private and used with AutoRelay.
+	addrChangeTicker := time.NewTicker(1 * time.Minute)
```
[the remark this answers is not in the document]
sukunrt: There's no bug. The public address is not reachable so we don't advertise it. 

Once we obtain relay reservations with AutoRelay, we will only advertise those public circuit v2 addresses and not our unreachable public addresses. 
For AutoNAT we need our public, reachable or unreachable, non circuitv2, addresses.

sukunrt:
```
I don't love this service. It's very messy, interacts with the basic host in deep ways, and seems hard to test.
```

I agree.

p2p/host/autonat/autonat.go
sukunrt: We start at a random offset to not rely on the peerstore's ordering of peers.

MarcoPolo: TIL about rand.Shuffle. We should make a PR that updates our usages of shuffling throughout the codebase.

I think we do want to shuffle here, although I don't think I could prove why.

PR #2969, opened 2024-09-19
RoutedHost: Embed the underlying host instead of hiding it

burdiyan:
When using a libp2p host with routing, the underlying concrete type is RoutedHost. This type wraps the original Host type and hides it into a private field, which makes it hard to make interface type assertions on the host.

E.g. sometimes it's useful to have access to the underlying instance of the IDService, which is exposed by BasicHost as a method `IDService() identityIDService`, but is inaccessible for type assertions because RoutedHost hides the BasicHost in its private field.

This commit uses type embedding, instead of a private field, to fix it.

### DIFF
--- a/p2p/host/routed/routed.go
+++ b/p2p/host/routed/routed.go
@@ -5,13 +5,12 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/libp2p/go-libp2p/core/connmgr"
-	"github.com/libp2p/go-libp2p/core/event"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/core/protocol"
+	"github.com/libp2p/go-libp2p/core/routing"
 
 	logging "github.com/ipfs/go-log/v2"
 
@@ -28,10 +27,12 @@ const AddressTTL = time.Second * 10
 // This allows the Host to find the addresses for peers when
 // it does not have them.
 type RoutedHost struct {
-	host  host.Host // embedded other host.
-	route Routing
+	host.Host // Embedding the underlying host.
+	route     routing.PeerRouting
 }
 
+// Routing is an interface for peer routing.
+// Deprecated: use [routing.PeerRouting] instead.
 type Routing interface {
 	FindPeer(context.Context, peer.ID) (peer.AddrInfo, error)
 }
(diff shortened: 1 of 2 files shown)


p2p/host/routed/routed.go
```diff
@@ -28,15 +27,12 @@ const AddressTTL = time.Second * 10
 // This allows the Host to find the addresses for peers when
 // it does not have them.
 type RoutedHost struct {
-	host  host.Host // embedded other host.
-	route Routing
+	host.Host
+	route routing.PeerRouting
 }
```
sukunrt, line 32: Do we need to change route from Routing to routing.PeerRouting? 

The rest of the change looks good to me.

sukunrt: SGTM.

p2p/host/routed/routed.go
```diff
@@ -28,15 +27,12 @@ const AddressTTL = time.Second * 10
...
 
-type Routing interface {
-	FindPeer(context.Context, peer.ID) (peer.AddrInfo, error)
-}
-
```
sukunrt, line 38: Can you deprecate this with a comment saying use `routing.PeerRouting`

sukunrt (approved):

MarcoPolo:
This change seems fine, but I think a better solution to the problem of trying to get the ID service is with Fx options. You could provide them in the constructor with Fx's [Populate](https://pkg.go.dev/go.uber.org/fx#Populate) (https://github.com/libp2p/go-libp2p/pull/2956) or when you build libp2p with Fx directly (https://github.com/libp2p/go-libp2p/pull/2962).

burdiyan:
> This change seems fine, but I think a better solution to the problem of trying to get the ID service is with Fx options. You could provide them in the constructor with Fx's [Populate](https://pkg.go.dev/go.uber.org/fx#Populate) (#2956) or when you build libp2p with Fx directly (#2962).

@MarcoPolo Not sure if I agree it would be better. Requiring people to learn and use Fx seems like an overkill for something that can be done with existing language features. I personally stopped using Fx and now prefer doing things explicitly. But of course exposing more Fx-related flexibilities in Libp2p could be helpful too.

p2p/host/routed/routed.go
burdiyan: Done.

MarcoPolo:
The problem is that we don't want the `Host` interfaced type to have to expose all the sub-services a user might want access to (e.g. the identify service). We also don't really want users to assume that the result of creating a libp2p Host will be a `BasicHost` or any other concrete type. If we did, we would return that type directly. Therefore type assertions here are brittle.

Fx solves this problem in a fairly way. It lets you get a reference to a service without having to expose that service in the Host. It also won't create the service if nothing needs it.

burdiyan:
Are the Fx affordances for doing that in libp2p exposed to users? I haven't seen how could I do that. Anyway, I'm not feeling strongly about this change, so feel free to close it :)

MarcoPolo:
You'll be able to get the ID service in the next release. Refer to this test for an example: https://github.com/libp2p/go-libp2p/blob/c4c3a34545aab54b5b825a7adbd6a3db2c680afa/fx_options_test.go#L50-L60

Issue #2963, opened 2024-09-17
relay: recycle reservations for the same peer


burdiyan:
I have a libp2p node configured with `EnableRelay()`, `EnableHolePunching()`, `ForceReachabilityPrivate()`, and `EnableAutoRelayWithStaticRelays()` options. So my node starts up and connects to a preconfigured relay server right away.

I have a few reservation slots per peer configured on the relay server.

If I restart my libp2p node a few times, after some time it will fail to obtain a reservation on the relay because all slots are exhausted.

I'm not sure the motivation for this design of slots, e.g. why there need to be more than 1 slot per peer at all? But I'd like to suggest an enhancement to recycle the slots, so when a relay server notices the peer has disconnected, it should free up those slots, and allocate new ones when the peer reconnects.

With the current design, seems like even temporary network flickering could cause all the reservation slots to finish, although in fact there're no resources busy on the relay server.

PR #2974, opened 2024-09-23
relay: make only 1 reservation per peer

sukunrt:
We only need 1 reservation per peer.

This PR also proactively removes relay reservations on disconnecting from the peer. This helps when the relay client is restarting frequently.

Fixes: https://github.com/libp2p/go-libp2p/issues/2963


MarcoPolo (approved):
Looks good mod some nits

Issue #2966, opened 2024-09-18
Hole Punching: Use node's addresses coherently


burdiyan:
This issue is a more scoped subset of the problems described in https://github.com/libp2p/go-libp2p/issues/2965.

The hole punching code doesn't use the addresses from the Libp2p host. Instead it only uses the observed addresses from Identify, and the addresses of the network interfaces.

Actually the "server" and the "client" parts of hole punching don't use the same set of addresses, as seen in the following code fragments:

https://github.com/libp2p/go-libp2p/blob/921cc71072229344a78c7d794e3571c3cbf058e3/p2p/protocol/holepunch/svc.go#L279-L307

https://github.com/libp2p/go-libp2p/blob/921cc71072229344a78c7d794e3571c3cbf058e3/p2p/protocol/holepunch/holepuncher.go#L209

I suggest to use `host.Addrs()` directly in both places, as those includes all the possible addresses, including the ones provided by the custom AddrsFactory.

sukunrt:
@burdiyan Can you check if https://github.com/libp2p/go-libp2p/pull/2979 fixes your problem?

p-shahi:
@burdiyan please re-open if that didn't fix the problem

PR #2979, opened 2024-09-25
holepuncher: pass address function in constructor

sukunrt:
fixes: #2966


MarcoPolo (approved):
A couple of nits. I would like to see a test for the issue here

Issue #2972, opened 2024-09-21
`BasicHost` submits `EvtLocalAddressesUpdated`  but addresses were not changed


Wondertan:
### Context
I am running a test network in the k8s cluster, and while debugging an orthogonal issue, I observed unexpected behavior.

### Bug (?)
To my knowledge, the addresses of running pods do not change. Also, the logs below show that the addresses on every event are constant, however  `BasicHost` still submits the  `EvtLocalAddressesUpdated` .

### Trace
The trace was from PubSub node spamming `go-libp2p-pubsub@v0.12.0/pubsub.go:698 already have connection to peer`, triggered by `EvtPeerIdentificationCompleted`, initiated by identify push from a peer, prompted by `BasicHost` background loop on the peer detecting address changes. 

However, as per identify logs on the PubSub node, the addresses didn't change, and the exact same log with the same address repeats over and over, potentially suggesting that diffing logic in [makeUpdatedAddrsEvent](https://github.com/libp2p/go-libp2p/blob/9038a72c4863778e6a26814aab201cee6d93d640/p2p/host/basic/basic_host.go#L498) has a bug.
> DEBUG	net/identify	identify/id.go:832 received listen addrs for 12D3KooWQHiNsPQoiRMnixa1SzEMUukJo3r1UQ7tZaxH8kjMSm7L: [/ip4/151.115.12.199/udp/2121/quic-v1 /ip4/151.115.12.199/tcp/2121 /ip4/100.64.42.71/tcp/2121 /ip4/100.64.42.71/udp/2121/quic-v1 /ip4/100.64.42.71/udp/2121/quic-v1/webtransport/certhash/uEiCRTETNTiCP0LDvirFICa3zihf3qB14vOa6F7N-GoFntQ/certhash/uEiAr9jX-WLX19pVfR5MY1vL2y_ytxxcRQNPtCfE6oVlH_Q /ip4/100.64.42.71/udp/2121/quic-v1/webtransport/certhash/uEiCRTETNTiCP0LDvirFICa3zihf3qB14vOa6F7N-GoFntQ/certhash/uEiAr9jX-WLX19pVfR5MY1vL2y_ytxxcRQNPtCfE6oVlH_Q]

I couldn't find a way to inspect the state of the peer submitting the identify push with logs, and the only data point is the log from the peer above.

sukunrt:
I'm unable to reproduce this. Is it possible that the id pushes were triggered by Local Protocols Updated event?

[...]

PR #2980, opened 2024-09-26
basichost: ensure no duplicates in Addrs output

sukunrt:
fixes: https://github.com/libp2p/go-libp2p/issues/2972

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -500,6 +500,7 @@ func (h *BasicHost) makeUpdatedAddrEvent(prev, current []ma.Multiaddr) *event.Ev
 		return nil
 	}
 	prevmap := make(map[string]ma.Multiaddr, len(prev))
+	currmap := make(map[string]ma.Multiaddr, len(current))
 	evt := &event.EvtLocalAddressesUpdated{Diffs: true}
 	addrsAdded := false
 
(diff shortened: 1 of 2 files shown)


Wondertan:
I can help testing this on my cluster once the fix is final

sukunrt:
@Wondertan, The fix is final. Can you check this?

Wondertan:
Looks good!
image

Wondertan:
Might be worth adding a unit test

p2p/host/basic/basic_host.go
```diff
@@ -500,13 +500,17 @@ func (h *BasicHost) makeUpdatedAddrEvent(prev, current []ma.Multiaddr) *event.Ev
 		return nil
 	}
 	prevmap := make(map[string]ma.Multiaddr, len(prev))
+	currmap := make(map[string]ma.Multiaddr, len(current))
```
MarcoPolo, line 503: Do we even use `currmap`?

MarcoPolo (approved):
Looks good minus the superfluous `currmap`

p2p/host/basic/basic_host.go
sukunrt: I changed the iteration below to be over this map. This ensures that this method works even if there are duplicates in `Addrs` output.

MarcoPolo:
@dimartiro Yes. Removing duplicates without allocating means that you have to change the order of the returned Addrs (https://github.com/multiformats/go-multiaddr/blob/d19cf5da7dc590c829b9a4b329341ea8847d4d2a/multiaddr.go#L253).

If you don't have duplicates, it shouldn't affect the order.

PR #2982, opened 2024-09-27
identify: fix faulty handling of conn entries in IdentifyWait

### LINKED ISSUE #2981: Bug in IdentifyWait
<details>
<summary>Version Information</summary>
<pre>
	github.com/libp2p/go-libp2p v0.36.4
</pre>
</details>

The code in `IdentifyWait`
```
	e, found := ids.conns[c]
	if !found {
		// No entry found. We may have gotten an out of order notification. Check it we should have this conn (because we're still connected)
		// We hold the ids.connsMu lock so this is safe since a disconnect event will be processed later if we are connected.
		if c.IsClosed() {
			log.Debugw("connection not found in identify service", "peer", c.RemotePeer())
			ch := make(chan struct{})
			close(ch)
			return ch
		} else {
			ids.addConnWithLock(c)
		}
	}
```

and even if `!found` we are working with `e` below, which leads to an identify deadlock. Patch forthcoming

vyzo:
Closes #2981

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -155,7 +155,7 @@ type idService struct {
 	// The conns map contains all connections we're currently handling.
 	// Connections are inserted as soon as they're available in the swarm
 	// Connections are removed from the map when the connection disconnects.
-	conns map[network.Conn]entry
+	conns map[network.Conn]*entry
 
 	addrMu sync.Mutex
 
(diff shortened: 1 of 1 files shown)


vyzo:
This was completely broken and it manifests itself in very nasty ways under load.

vyzo:
ok, so the bug is not there, sorry -- we are copying the entry.

PR #2737, opened 2024-03-17
Magiselect

Jorropo:
Implements https://github.com/libp2p/specs/issues/608 for `/tls` and `/noise`.

Todo:
- [ ] Fix existing tests.
- [ ] Add new tests.
- [ ] Improve maddr reachability extension computation (generalize reachable quic into reachable quic + webtransport into reachable a into reachable a + b).
- [ ] Spec
- [ ] Figure out 8962b2ae336d94627f2f4361f96799ee3a5bd9e4's situation
- [ ] Favor `/tls` over `/noise` over multistream-select in happy eyeballs.

PR #2984, opened 2024-09-27
feat(tcpreuse): add options for sharing TCP listeners amongst TCP, WS and WSS transports

### LINKED ISSUE #2684: TCP Bind issue
I am trying to listen on `/tcp/{port}, /tcp/{port}/ws, and /tcp/{port}/tls/ws` and getting error  `bind: Only one usage of each socket address (protocol/network address/port) is normally permitted."}`

I thought LibP2P was capable of this?

aschmahmann:
closes #2684 

@MarcoPolo @sukunrt I took a stab at this as a result of looking at https://github.com/ipfs/kubo/pull/10521. I also pulled from @Jorropo's work on #2737.

Initial testing seems to indicate things work ok, but this definitely needs more tests, API cleanup + eyes before merging.

Side note: Given that we're sharing the underlying TCP listener now I wonder if we basically have to tackle #1435 as well.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -38,6 +38,7 @@ import (
 	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
 	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
 	"github.com/libp2p/go-libp2p/p2p/transport/quicreuse"
+	"github.com/libp2p/go-libp2p/p2p/transport/tcpreuse"
 	libp2pwebrtc "github.com/libp2p/go-libp2p/p2p/transport/webrtc"
 	"github.com/prometheus/client_golang/prometheus"
 
@@ -145,6 +146,8 @@ type Config struct {
 	CustomIPv6BlackHoleSuccessCounter bool
 
 	UserFxOptions []fx.Option
+
+	ShareTCPListener bool
 }
 
 func (cfg *Config) makeSwarm(eventBus event.Bus, enableMetrics bool) (*swarm.Swarm, error) {
@@ -289,6 +292,12 @@ func (cfg *Config) addTransports() ([]fx.Option, error) {
 		fx.Provide(func() connmgr.ConnectionGater { return cfg.ConnectionGater }),
 		fx.Provide(func() pnet.PSK { return cfg.PSK }),
 		fx.Provide(func() network.ResourceManager { return cfg.ResourceManager }),
+		fx.Provide(func(gater connmgr.ConnectionGater, rcmgr network.ResourceManager) *tcpreuse.ConnMgr {
+			if !cfg.ShareTCPListener {
+				return nil
+			}
+			return tcpreuse.NewConnMgr(tcpreuse.EnvReuseportVal, gater, rcmgr)
+		}),
 		fx.Provide(func(cm *quicreuse.ConnManager, sw *swarm.Swarm) libp2pwebrtc.ListenUDPFn {
 			hasQuicAddrPortFor := func(network string, laddr *net.UDPAddr) bool {
 				quicAddrPorts := map[string]struct{}{}
--- a/options.go
+++ b/options.go
@@ -644,6 +644,7 @@ func WithFxOption(opts ...fx.Option) Option {
 	}
 }
 
+// ShareTCPListener shares the same listen address between TCP and Websocket transports.
 func ShareTCPListener() Option {
 	return func(cfg *Config) error {
 		cfg.ShareTCPListener = true
--- a/options.go
+++ b/options.go
@@ -644,7 +644,11 @@ func WithFxOption(opts ...fx.Option) Option {
 	}
 }
 
-// ShareTCPListener shares the same listen address between TCP and Websocket transports.
+// ShareTCPListener shares the same listen address between TCP and Websocket
+// transports. This lets both transports use the same TCP port.
+//
+// Currently this behavior is Opt-in. In a future release this will be the
+// default, and this option will be removed.
 func ShareTCPListener() Option {
 	return func(cfg *Config) error {
 		cfg.ShareTCPListener = true
--- a/options.go
+++ b/options.go
@@ -643,3 +643,10 @@ func WithFxOption(opts ...fx.Option) Option {
 		return nil
 	}
 }
+
+func ShareTCPListener() Option {
+	return func(cfg *Config) error {
+		cfg.ShareTCPListener = true
+		return nil
+	}
+}
--- a/p2p/transport/tcp/tcp.go
+++ b/p2p/transport/tcp/tcp.go
@@ -142,8 +142,6 @@ type TcpTransport struct {
 	rcmgr network.ResourceManager
 
 	reuse reuseport.Transport
-
-	metricsCollector *aggregatingCollector
 }
 
 var _ transport.Transport = &TcpTransport{}
@@ -233,7 +231,7 @@ func (t *TcpTransport) dialWithScope(ctx context.Context, raddr ma.Multiaddr, p 
 	c := conn
 	if t.enableMetrics {
 		var err error
-		c, err = newTracingConn(conn, t.metricsCollector, true)
+		c, err = newTracingConn(conn, true)
 		if err != nil {
 			return nil, err
 		}
@@ -279,7 +277,7 @@ func (t *TcpTransport) Listen(laddr ma.Multiaddr) (transport.Listener, error) {
 	}
 
 	if t.enableMetrics {
-		list = newTracingListener(&tcpListener{list, 0}, t.metricsCollector)
+		list = newTracingListener(&tcpListener{list, 0})
 	}
 	return t.upgrader.UpgradeListener(t, list), nil
 }
--- a/p2p/transport/tcp/tcp.go
+++ b/p2p/transport/tcp/tcp.go
@@ -148,7 +148,7 @@ var _ transport.Transport = &TcpTransport{}
 var _ transport.DialUpdater = &TcpTransport{}
 
 // NewTCPTransport creates a tcp transport object that tracks dialers and listeners
-// created. It represents an entire TCP stack (though it might not necessarily be).
+// created.
 func NewTCPTransport(upgrader transport.Upgrader, rcmgr network.ResourceManager, opts ...Option) (*TcpTransport, error) {
 	if rcmgr == nil {
 		rcmgr = &network.NullResourceManager{}
@@ -269,7 +269,7 @@ func (t *TcpTransport) Listen(laddr ma.Multiaddr) (transport.Listener, error) {
 	if t.sharedTcp == nil {
 		list, err = t.unsharedMAListen(laddr)
 	} else {
-		list, err = t.sharedTcp.DemultiplexedListen(laddr, tcpreuse.MultistreamSelect)
+		list, err = t.sharedTcp.DemultiplexedListen(laddr, tcpreuse.DemultiplexedConnType_MultistreamSelect)
 	}
 	if err != nil {
 		return nil, err
--- a/p2p/transport/tcp/tcp.go
+++ b/p2p/transport/tcp/tcp.go
@@ -117,13 +117,6 @@ func WithMetrics() Option {
 	}
 }
 
-func WithSharedTCP(mgr *tcpreuse.ConnMgr) Option {
-	return func(tr *TcpTransport) error {
-		tr.sharedTcp = mgr
-		return nil
-	}
-}
-
 // TcpTransport is the TCP transport.
 type TcpTransport struct {
 	// Connection upgrader for upgrading insecure stream connections to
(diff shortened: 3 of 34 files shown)


p2p/transport/tcp/tcp.go
```diff
@@ -113,6 +117,13 @@ func WithMetrics() Option {
 	}
 }
 
+func WithSharedTCP(mgr *tcpreuse.ConnMgr) Option {
```
aschmahmann, line 120: There's a few API questions on how we want to expose this functionality. For example:
1. Is it an option passed into the transport constructors for TCP + WebSockets, or is it one of the parameters controlled by fx and if nil is passed we don't allow shared listeners (this would lend itself better to the SharedTCP option being enabled at the host constructor layer)?
2. Should shared listeners be enabled by default in the host constructor (whether now or in the future)?

p2p/transport/tcpreuse/reuseport.go
```diff
@@ -11,13 +11,13 @@ import (
 // It default to true.
 const envReuseport = "LIBP2P_TCP_REUSEPORT"
 
-// envReuseportVal stores the value of envReuseport. defaults to true.
-var envReuseportVal = true
+// EnvReuseportVal stores the value of envReuseport. defaults to true.
+var EnvReuseportVal = true
```
aschmahmann, line 15: This doesn't feel good to export as it's only used for tests and is the result of me moving things to avoid dependency cycles (and separate concerns).

Any suggestions on where this reuseport logic lives? Could potentially move the tests or have enough options to thread the testing logic through

sukunrt:
Overall it looks quite good. 

The only thing major thing that we must fix before merging is this:
```
// TODO: if/how do we want to handle stalled connections and stop them from clogging up the pipeline?
// Drop connection because the buffer is full
```
We need a way to limit the number of connections being upgraded per IP/subnet. Otherwise, a single node can make the target node unresponsive. 

We don't have the endpoint multiaddr at this point. This makes it difficult to use the resource manager as is. 
1. We can just use the rcmgr as is. We have the IP and Port of the remote, that's all that's needed by the resource manager any way.
2. We can implement the ConnLimiter separately for this tcp reuse block.  

I think 1 is fine. AllowList anyway works with a IP / Subnet. We can make this clear in the documentation for Resource Manager.

p2p/transport/tcpreuse/listener.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (m *demultiplexedListener) Close() error {
+	return m.closeFn()
+}
+
+func (m *demultiplexedListener) Multiaddr() ma.Multiaddr {
+	// TODO: do we need to add a suffix for the rest of the transport?
+	return m.inner.Multiaddr()
+}
```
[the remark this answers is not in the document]
sukunrt: I think this is okay. The transport listeners can handle that.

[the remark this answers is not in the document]
sukunrt: Note: This is separate from the address being incorrect on a connection: #2988

p2p/transport/tcp/tcp.go
```diff
@@ -134,15 +148,16 @@ var _ transport.Transport = &TcpTransport{}
 var _ transport.DialUpdater = &TcpTransport{}
 
 // NewTCPTransport creates a tcp transport object that tracks dialers and listeners
-// created. It represents an entire TCP stack (though it might not necessarily be).
-func NewTCPTransport(upgrader transport.Upgrader, rcmgr network.ResourceManager, opts ...Option) (*TcpTransport, error) {
+// created.
+func NewTCPTransport(upgrader transport.Upgrader, rcmgr network.ResourceManager, sharedTCP *tcpreuse.ConnMgr, opts ...Option) (*TcpTransport, error) {
 	if rcmgr == nil {
```
sukunrt, line 153: This is a breaking change. 
It simplifies the top level basic host api a lot. You can do:

```
libp2p.New(Transport(tcp.NewTransport), Transport(websocket.New), ShareTCPListener())
```
And not need to specify the rcmgr, or the connection gater. 

I'm confused how to provide this API, enabling sharing by just specifying `ShareTCPListener`, and pass in the sharedTCP objects as one of the optional arguments TCP Transport. If we can pass in the sharedTCP object as an optional argument that will prevent breaking this.

p2p/transport/tcpreuse/listener_test.go
```diff
@@ -0,0 +1,465 @@
...
+				}()
+			}
+			wg.Wait()
+		})
+	}
+}
+
+func TestListenerMultiplexed(t *testing.T) {
+	listenAddr := ma.StringCast("/ip4/0.0.0.0/tcp/0")
+	const N = 20
```
sukunrt, line 229: For some reason this breaks with N > 60 on macOS. Just making more than 150 concurrent connections breaks with a raw tcp listener and dialer. So this is probably not a bug in the code somewhere.

p2p/transport/websocket/listener.go
```diff
@@ -126,13 +150,7 @@ func (l *listener) Accept() (manet.Conn, error) {
...
 			return nil, transport.ErrListenerClosed
 		}
-
-		mnc, err := manet.WrapNetConn(c)
-		if err != nil {
-			c.Close()
-			return nil, err
-		}
-		return mnc, nil
+		return c, nil
```
sukunrt, line 153: Wrapping it by manet shadows the `Scope` method which we need for the upgrader.

options.go
```diff
@@ -635,3 +635,11 @@ func IPv6BlackHoleSuccessCounter(f *swarm.BlackHoleSuccessCounter) Option {
...
 	}
 }
+
+// ShareTCPListener shares the same listen address between TCP and Websocket transports.
+func ShareTCPListener() Option {
+	return func(cfg *Config) error {
+		cfg.ShareTCPListener = true
+		return nil
+	}
+}
```
sukunrt, line 645: This API has the nice property of keeping all current code same and enable sharing by just specifying 
```
libp2p.New(
...
ShareTCPListener()
)
```

p2p/transport/tcpreuse/demultiplex.go
```diff
@@ -0,0 +1,107 @@
...
+	// Peek returns the next n bytes without advancing the reader. The bytes stop
+	// being valid at the next read call. If Peek returns fewer than n bytes, it
+	// also returns an error explaining why the read is short. The error is
+	// [ErrBufferFull] if n is larger than b's buffer size.
+	Peek(n int) ([]byte, error)
+}
+
+var _ peekAble = (*bufio.Reader)(nil)
+
+// TODO: We can unexport this type and rely completely on the multiaddr passed in to
```
MarcoPolo, line 23: I kind of like being explicit with what the conn type is.

p2p/transport/websocket/conn.go
```diff
@@ -99,9 +132,20 @@ func (c *Conn) Write(b []byte) (n int, err error) {
...
+	}); ok {
+		return sc.Scope()
+	}
+	return nil
+}
+
 // Close closes the connection. Only the first call to Close will receive the
 // close error, subsequent and concurrent calls will return nil.
 // This method is thread-safe.
+// TODO: Fix this ^
```
MarcoPolo, line 148: This should be easy with OnceValue

MarcoPolo (commented):
I think this mostly looks good. Just a couple things I'll clean up in the next pass

config/config.go
```diff
@@ -289,6 +292,12 @@ func (cfg *Config) addTransports() ([]fx.Option, error) {
 		fx.Provide(func() connmgr.ConnectionGater { return cfg.ConnectionGater }),
 		fx.Provide(func() pnet.PSK { return cfg.PSK }),
 		fx.Provide(func() network.ResourceManager { return cfg.ResourceManager }),
+		fx.Provide(func(gater connmgr.ConnectionGater, rcmgr network.ResourceManager) *tcpreuse.ConnMgr {
+			if !cfg.ShareTCPListener {
+				return nil
+			}
+			return tcpreuse.NewConnMgr(tcpreuse.EnvReuseportVal, gater, rcmgr)
```
aschmahmann, line 299: There seems to be a negative missing somewhere. `tcpreuse.EnvReuseportVal` is true if we should use reuseport and false otherwise, but we're passing it into a field labeled "disablereuseport". Either we should pass `!tcpreuse.EnvReuseportVal` or negate how it's used within `NewConnMgr`.

No strong feelings on which we do, although IMO passing enableX generally feels better than disableX if you are being forced to pass in a boolean.

p2p/transport/tcpreuse/listener.go
```diff
@@ -0,0 +1,313 @@
...
+func (m *multiplexedListener) DemultiplexedListen(connType DemultiplexedConnType) (manet.Listener, error) {
+	if !connType.IsKnown() {
+		return nil, fmt.Errorf("unknown connection type: %s", connType)
+	}
+
+	m.mx.Lock()
+	defer m.mx.Unlock()
+	l, ok := m.listeners[connType]
+	if ok {
+		return l, nil
```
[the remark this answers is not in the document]
aschmahmann: You're saying we should just return an error here if there's already an active listener? That seems reasonable.

options.go
lidel: As `libp2p.New` user I will wonder why I need to explicitly pass this Option  to enable sharing for TCP ports, but don't have to do anything extra for UDP ones.

No chance of making this implicit? Or at least explain in godoc comment why this needs to be explicit?

p2p/transport/tcpreuse/listener.go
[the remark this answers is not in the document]
MarcoPolo: fixed in c314a2ab80fa

p2p/transport/tcp/tcp.go
```diff
@@ -122,6 +133,9 @@ type TcpTransport struct {
 	disableReuseport bool // Explicitly disable reuseport.
 	enableMetrics    bool
 
+	// share and demultiplex TCP listeners across multiple transports
+	sharedTcp *tcpreuse.ConnMgr
```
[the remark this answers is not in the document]
MarcoPolo: Not sure. I also don't like the connmgr name in quicreuse. But I don't have a better name that isn't also confusing. At least this is consistently confusing (?)

p2p/transport/tcpreuse/listener.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	m.listeners[connType] = l
+	m.listenerCounter++
+
+	return l, nil
+}
+
+func (m *multiplexedListener) Run() error {
+	const numWorkers = 16
```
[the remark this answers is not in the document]
MarcoPolo: I think this was refactored to use a channel semaphore with N=64. That should be good enough for most (all?) use cases.

This thing should be very fast we are reading the first 3 bytes of the first packet.

p2p/transport/tcpreuse/reuseport.go
MarcoPolo: There's a refactor to be done here, but maybe we do that in a separate PR? This one is already quite big

p2p/transport/tcpreuse/listener.go
```diff
@@ -0,0 +1,313 @@
...
+		return nil, fmt.Errorf("invalid listen addr %s, need tcp address", listenAddr)
+	}
+	return addr, nil
+}
+
+// DemultiplexedListen returns a listener for laddr listening for `connType` connections. The connections
+// accepted from returned listeners need to be upgraded with a `transport.Upgrader`.
+// NOTE: All listeners for port 0 share the same underlying socket, so they have the same specific port.
+func (t *ConnMgr) DemultiplexedListen(laddr ma.Multiaddr, connType DemultiplexedConnType) (manet.Listener, error) {
+	if !connType.IsKnown() {
```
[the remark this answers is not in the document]
MarcoPolo: They both seem pretty equivalent to me. I'd opt to keep what we have for lack of a reason to change.

options.go
MarcoPolo: The main reason is that I want to get some confidence here before defaulting this behavior to on. In a future release it'll be the default and you wouldn't need this.

Aside: UDP transports don't currently share the port if you use port `0` across them (this does). That's also something that I'd like to change in a future release.

I'll update the comment.

p2p/transport/tcpreuse/demultiplex.go
```diff
@@ -0,0 +1,240 @@
...
+	case "GET", "HEA", "POS", "PUT", "DEL", "CON", "OPT", "TRA", "PAT":
+		return true
+	default:
+		return false
+	}
+}
+
+func IsTLS(s Sample) bool {
+	switch string(s[:]) {
+	case "\x16\x03\x01", "\x16\x03\x02", "\x16\x03\x03", "\x16\x03\x04":
```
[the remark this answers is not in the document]
MarcoPolo: Do you mean so should _not_ be here?

MarcoPolo (approved):

p2p/transport/tcpreuse/demultiplex.go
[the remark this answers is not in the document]
aschmahmann: Ya, thanks

PR #2985, opened 2024-09-27
feat(websocket): switch the underlying http server logger to use ipfs/go-log

aschmahmann:
Does what it says on the box 😅.

The motivation behind this change is that while implementing https://github.com/ipfs/kubo/pull/10521 I noticed a large number of logs of the form `"http: TLS handshake error from ip:port: EOF"` and I couldn't control them in the same way I can logs for any other transport.

This error in particular is one I don't expect many folks have run into since my understanding is that the WSS listener within go-libp2p is not highly utilized with many applications dependent on WSS choosing to use WS + do TLS termination in front of the libp2p node instead. However, in the linked issue the idea is to do the TLS termination within go-libp2p and so the issue becomes apparent.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,6 +60,7 @@ require (
 	go.uber.org/fx v1.22.1
 	go.uber.org/goleak v1.3.0
 	go.uber.org/mock v0.4.0
+	go.uber.org/zap v1.27.0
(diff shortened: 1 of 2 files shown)


sukunrt (approved):
Thanks @aschmahmann

PR #2988, opened 2024-09-30
Connection multiaddress should match the address actually dialed (WSS, WebTransport, WebRTC)

aschmahmann:
It seems like a bug that `wssConn.RemoteMultiaddr()` doesn't contain the TLS or SNI components (and may also have an over-resolved domain name, but haven't tested that yet) and shows a much simpler multiaddr.

Context: I noticed the domain names of wss addresses were not showing up in the https://check.ipfs.network tool due to https://github.com/ipfs/ipfs-check/blob/b35461c0ac6dda2f32efceec52f2559f4c105bc4/daemon.go#L256

If this is buggy behavior it seems like this can be fixed in the websocket conn type, should a generic test for this be added to the transport suite?

### DIFF
--- a/p2p/test/transport/gating_test.go
+++ b/p2p/test/transport/gating_test.go
@@ -101,7 +101,7 @@ func TestInterceptSecuredOutgoing(t *testing.T) {
 				connGater.EXPECT().InterceptAddrDial(h2.ID(), gomock.Any()).Return(true),
 				connGater.EXPECT().InterceptSecured(network.DirOutbound, h2.ID(), gomock.Any()).Do(func(_ network.Direction, _ peer.ID, addrs network.ConnMultiaddrs) {
 					// remove the certhash component from WebTransport and WebRTC addresses
-					require.Equal(t, stripCertHash(h2.Addrs()[0]).String(), addrs.RemoteMultiaddr().String())
+					require.Equal(t, h2.Addrs()[0].String(), addrs.RemoteMultiaddr().String())
 				}),
 			)
 			err := h1.Connect(ctx, peer.AddrInfo{ID: h2.ID(), Addrs: h2.Addrs()})
@@ -135,8 +135,7 @@ func TestInterceptUpgradedOutgoing(t *testing.T) {
 				connGater.EXPECT().InterceptAddrDial(h2.ID(), gomock.Any()).Return(true),
 				connGater.EXPECT().InterceptSecured(network.DirOutbound, h2.ID(), gomock.Any()).Return(true),
 				connGater.EXPECT().InterceptUpgraded(gomock.Any()).Do(func(c network.Conn) {
-					// remove the certhash component from WebTransport addresses
-					require.Equal(t, stripCertHash(h2.Addrs()[0]), c.RemoteMultiaddr())
+					require.Equal(t, h2.Addrs()[0], c.RemoteMultiaddr())
 					require.Equal(t, h1.ID(), c.LocalPeer())
 					require.Equal(t, h2.ID(), c.RemotePeer())
 				}))
@@ -170,12 +169,12 @@ func TestInterceptAccept(t *testing.T) {
 				// In WebRTC, retransmissions of the STUN packet might cause us to create multiple connections,
 				// if the first connection attempt is rejected.
 				connGater.EXPECT().InterceptAccept(gomock.Any()).Do(func(addrs network.ConnMultiaddrs) {
-					// remove the certhash component from WebTransport addresses
+					// remove the certhash component from WebRTC and WebTransport addresses
 					require.Equal(t, stripCertHash(h2.Addrs()[0]), addrs.LocalMultiaddr())
 				}).AnyTimes()
 			} else {
 				connGater.EXPECT().InterceptAccept(gomock.Any()).Do(func(addrs network.ConnMultiaddrs) {
-					// remove the certhash component from WebTransport addresses
+					// remove the certhash component from WebRTC and WebTransport addresses
 					require.Equal(t, stripCertHash(h2.Addrs()[0]), addrs.LocalMultiaddr())
 				})
 			}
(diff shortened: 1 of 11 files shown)


p2p/test/transport/transport_test.go
```diff
@@ -113,6 +122,26 @@ var transportsToTest = []TransportTestCase{
...
+			if opts.NoListen {
+				libp2pOpts = append(libp2pOpts, libp2p.NoListenAddrs)
+			} else {
+				dnsName := fmt.Sprintf("example%s.com", opts.HostSeed)
+				cert, err := generateSelfSignedCert(dnsName)
+				require.NoError(t, err)
+				libp2pOpts = append(libp2pOpts, libp2p.ListenAddrStrings(fmt.Sprintf("/ip4/127.0.0.1/tcp/0/tls/sni/%s/ws", dnsName)))
+				wsOpts = append(wsOpts, websocket.WithTLSConfig(&tls.Config{Certificates: []tls.Certificate{cert}}))
+			}
+			libp2pOpts = append(libp2pOpts, libp2p.Transport(websocket.New, wsOpts...))
```
aschmahmann, line 139: This makes WSS a standout compared to the others in that the others all support multiple transports, but this host only supports WSS. Should the other host generators also be narrowed down to support a single transport?

p2p/test/transport/transport_test.go
```diff
@@ -113,6 +122,26 @@ var transportsToTest = []TransportTestCase{
 			return h
 		},
 	},
+	{
+		Name: "Secure WebSocket with CA Certificate",
```
aschmahmann, line 126: As with the WebSockets example the second security layer negotiation + multiplexer are left undefined (i.e. use the defaults), is that a mistake?

p2p/test/transport/gating_test.go
```diff
@@ -170,12 +169,12 @@ func TestInterceptAccept(t *testing.T) {
...
 				connGater.EXPECT().InterceptAccept(gomock.Any()).Do(func(addrs network.ConnMultiaddrs) {
-					// remove the certhash component from WebTransport addresses
+					// remove the certhash component from WebRTC and WebTransport addresses
 					require.Equal(t, stripCertHash(h2.Addrs()[0]), addrs.LocalMultiaddr())
 				}).AnyTimes()
 			} else {
 				connGater.EXPECT().InterceptAccept(gomock.Any()).Do(func(addrs network.ConnMultiaddrs) {
-					// remove the certhash component from WebTransport addresses
+					// remove the certhash component from WebRTC and WebTransport addresses
 					require.Equal(t, stripCertHash(h2.Addrs()[0]), addrs.LocalMultiaddr())
```
aschmahmann, line 178: Wasn't sure what the correct move was for the connection gaters (both in terms of what should be expected and how people are using this and how it might break).

At the moment I made a decision here to strip out the certhashes for `InterceptAccept` but not `InterceptSecured` based on the idea that Accept shouldn't really be thinking about security-related, but if it's available maybe it's still worth exposing and perhaps the complexity of being different is too confusing.

Either way, once we decide what's going on here I think adding better documentation to the interfaces seems needed

p2p/test/transport/transport_test.go
```diff
@@ -798,3 +867,29 @@ func TestConnClosedWhenRemoteCloses(t *testing.T) {
 		})
 	}
 }
+
+func TestConnMatchingAddress(t *testing.T) {
```
aschmahmann, line 871: Here's the new test with the behavior that IIUC is that's expected and was driving most of the changes (although the connection gater ones might be different)

p2p/transport/webtransport/transport_test.go
```diff
@@ -133,7 +133,8 @@ func TestTransport(t *testing.T) {
 		require.NoError(t, err)
 		_, port, err := net.SplitHostPort(addr)
 		require.NoError(t, err)
-		require.Equal(t, ma.StringCast(fmt.Sprintf("/ip4/127.0.0.1/udp/%s/quic-v1/webtransport", port)), conn.RemoteMultiaddr())
+		require.Equal(t, ma.StringCast(fmt.Sprintf("/ip4/127.0.0.1/udp/%s/quic-v1/webtransport", port)), stripCertHashes(conn.RemoteMultiaddr()))
+		require.Equal(t, 2, len(extractCertHashes(conn.RemoteMultiaddr())))
```
aschmahmann, line 137: Changes like the ones in this file make me wonder if the fixes here are going to result in anyone's code breaking.

p2p/transport/websocket/conn.go
```diff
@@ -25,17 +28,72 @@ type Conn struct {
...
+	return newConn(raw, secure, sni, false)
+}
+
+// NewInboundConn creates an inbound Conn given a regular gorilla/websocket Conn.
+func NewInboundConn(raw *ws.Conn, secure bool, sni string) (*Conn, error) {
+	return newConn(raw, secure, sni, true)
+}
+
+// newConn creates a Conn given a regular gorilla/websocket Conn.
+func newConn(raw *ws.Conn, secure bool, sni string, inbound bool) (*Conn, error) {
```
aschmahmann, line 49: This function could use some review. Additionally, I don't feel like we've tested the `/dns/example.com/tcp/80/wss` case very well in the associated tests. It seems like maybe we could use with a normalization function within go-libp2p for working with WSS addresses since IIUC we have a number of forms:
- Ending in `/wss` or `/tls/ws` for when no SNI is needed
- The `/dns` (and dns4/6) which implies the domain to use for verification if there's no `/sni` component (can end in `/tls/ws` or `/wss`)
- Ending with `/tls/sni/<domain>/ws`

MarcoPolo: I haven't looked at this PR yet, but the connection's RemoteMultiaddr should not have a dns component. It should have the IP address that is used for the current connection (along with tls/sni info).

Apologies if that's already the case, and I haven't looked at this yet.

MarcoPolo:

Issue #1597, opened 2022-06-11
DNS resolution should be moved to the transports


marten-seemann:
We currently resolve multiaddrs in the basichost: https://github.com/libp2p/go-libp2p/blob/d7ba37217c6f1a0d739666f78ac19a87728a7280/p2p/host/basic/basic_host.go#L712

This leads to problems with transports that, for example, validate TLS certificates, as they're kept ignorant about the domain name, and only learn the resolved address. This applies to WebSockets (thanks to @aschmahmann for bringing it up in #1592) and will apply to WebTransport as well (although we will _mostly_ use certificate hashes there).

Architecturally, address resolution should probably live within the transport. This has a few advantages:
* it gives full control what to do with a multiaddr to that transport
* it makes address resolution part of the dial timeout
* it speeds up things, since we don't need to resolve _all_ the peer's addresses before we start dialing
* it parallelises address resolution (as dials are run concurrently)
* it moves logic out of the basic host

However, there are caching implications. Currently, we implement some rudimentary caching for a const 2 minutes, not based on the TTL of the DNS record:
https://github.com/libp2p/go-libp2p/blob/d7ba37217c6f1a0d739666f78ac19a87728a7280/p2p/host/basic/basic_host.go#L707
Arguably, caching should live inside the DNS resolver, so one could argue that this is safe to remove.

@Stebalien @vyzo @aschmahmann @MarcoPolo Thoughts?

vyzo:
I am not convinced its the transport's responsibility to do dns resolution; and we can pass the domain in a contrext option for websockets.

For caching, we should probably rely on the resolver, might make sense to make a caching layer for the vanilla/OS resolver as the DoH resolver already does it.

aschmahmann:
> This leads to problems with transports that, for example, validate TLS certificates, as they're kept ignorant about the domain name, and only learn the resolved address

IIUC this isn't technically what happens instead we get both `/dns4/foo.tld/tcp/443/wss/p2p/QmFoo` and `/ip4/1.2.3.4/tcp/443/wss/p2p/QmFoo` in the peerstore and then are stuck with the decision of dialing both (whether sequentially or in parallel) or choosing one.

At the moment WSS chooses the worst option by only choosing the IP one due to the filtering here https://github.com/libp2p/go-libp2p/blob/d7ba37217c6f1a0d739666f78ac19a87728a7280/p2p/transport/websocket/websocket.go#L26

[...]

PR #2989, opened 2024-09-30
feat: core/transport: Add SkipResolver interface

MarcoPolo:
Fixes an issue where the implied SNI is dropped when connecting over a p2p-circuit address. For example if you connected to a peer over `/dns/example.com/wss/p2p/qmFoo/p2p-circuit/p2p/qmTarget` the swarm would resolve example.com without preserving the SNI information needed for `WSS`. The WebSocket transport would only get the IP address of example.com without knowning the host name it needed to use for the SNI.

We had the same problem previously when dialing a wss address. See https://github.com/libp2p/go-libp2p/issues/1597. This was fixed by letting transports resolve the multiaddr instead of the swarm. This case is similar, except that we want to tell the swarm to just skip resolving altogether since the inner transport will resolve the multiaddr later.

### DIFF
--- a/core/transport/transport.go
+++ b/core/transport/transport.go
@@ -85,6 +85,16 @@ type Resolver interface {
 	Resolve(ctx context.Context, maddr ma.Multiaddr) ([]ma.Multiaddr, error)
 }
 
+// SkipResolver can be optionally implemented by transports that don't want to
+// resolve or transform the multiaddr. Useful for transports that wrap other
+// transports. This lets the inner transport specify how a multiaddr is
+// resolved later.
+// Also useful in cases where the transport doesn't need a resolved address to
+// dial.
+type SkipResolver interface {
+	SkipResolve(ctx context.Context, maddr ma.Multiaddr) bool
+}
+
 // Listener is an interface closely resembling the net.Listener interface. The
 // only real difference is that Accept() returns Conn's of the type in this
 // package, and also exposes a Multiaddr method as opposed to a regular Addr
--- a/core/transport/transport.go
+++ b/core/transport/transport.go
@@ -50,6 +50,10 @@ type CapableConn interface {
 // shutdown. NOTE: `Dial` and `Listen` may be called after or concurrently with
 // `Close`.
 //
+// In addition to the Transport interface, transports may implement
+// Resolver or SkipResolver interface. When wrapping/embedding a transport, you should
+// ensure that the Resolver/SkipResolver interface is handled correctly.
+//
 // For a conceptual overview, see https://docs.libp2p.io/concepts/transport/
 type Transport interface {
 	// Dial dials a remote peer. It should try to reuse local listener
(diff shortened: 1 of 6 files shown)


core/transport/transport.go
```diff
@@ -85,6 +85,16 @@ type Resolver interface {
...
 
+// SkipResolver can be optionally implemented by transports that don't want to
+// resolve or transform the multiaddr. Useful for transports that wrap other
+// transports. This lets the inner transport specify how a multiaddr is
+// resolved later.
+// Also useful in cases where the transport doesn't need a resolved address to
+// dial.
+type SkipResolver interface {
+	SkipResolve(ctx context.Context, maddr ma.Multiaddr) bool
+}
```
sukunrt, line 96: Can you add some text to the transport interface documentation that says something like: 

```
In addition to the Transport interface, transports may implement SkipResolver
interface. When wrapping/embedding a transport, you should ensure that 
the SkipResolver interface is handled correctly. 
```

It's very rare to wrap a transport. In case someone does, this documentation would be useful.

core/transport/transport.go
```diff
@@ -85,6 +85,16 @@ type Resolver interface {
...
 }
 
+// SkipResolver can be optionally implemented by transports that don't want to
+// resolve or transform the multiaddr. Useful for transports that wrap other
+// transports. This lets the inner transport specify how a multiaddr is
+// resolved later.
+// Also useful in cases where the transport doesn't need a resolved address to
+// dial.
+type SkipResolver interface {
+	SkipResolve(ctx context.Context, maddr ma.Multiaddr) bool
```
sukunrt, line 95: Do we need the context?

core/transport/transport.go
sukunrt: I am also happy to just make `SkipResolve` a part of the transport interface.

p2p/protocol/circuitv2/client/transport.go
```diff
@@ -46,8 +46,16 @@ func AddTransport(h host.Host, upgrader transport.Upgrader) error {
 
 // Transport interface
 var _ transport.Transport = (*Client)(nil)
+var _ transport.SkipResolver = (*Client)(nil)
```
sukunrt, line 49: Also similar documentation here, so that in case some one does actually embed this, they remember to wrap the `SkipResolver` interface too.

sukunrt (approved):

core/transport/transport.go
MarcoPolo: Not right now, but I think it's nice to keep the same interface as Resolve. It also enables this request to be use IO in the future.

core/transport/transport.go
MarcoPolo: > I am also happy to just make SkipResolve a part of the transport interface.

I don't want to change the existing transport interface. That would be a breaking change for something that almost no transports care about.

I'll add the text to the Transport documentation however.

p2p/net/swarm/swarm_dial.go
```diff
@@ -322,84 +328,142 @@ func (s *Swarm) addrsForDial(ctx context.Context, p peer.ID) (goodAddrs []ma.Mul
...
+			tpt := s.TransportForDialing(addr)
+			resolver, ok := tpt.(transport.SkipResolver)
+			if !ok {
+				return []ma.Multiaddr{addr}, nil
+			}
+			if resolver.SkipResolve(ctx, addr) {
+				skipped = append(skipped, addr)
+				return nil, nil
+			}
+			return []ma.Multiaddr{addr}, nil
```
sukunrt, line 426: As we are not recursively resolving addresses any more, can we implement a null resolver which just returns the address as is?  Do we need this new `SkipResolve` method?

sukunrt (approved):

p2p/net/swarm/swarm_dial.go
MarcoPolo: We do. It stores it in the skipped list which is a closured var.

PR #2990, opened 2024-10-01
fix: swarm: refactor address resolution

MarcoPolo:
Fixes an issue around p2p-circuit multiaddrs and DNS resolution.


p2p/net/swarm/swarm_dial.go
```diff
@@ -22,9 +22,17 @@ import (
 )
 
 // The maximum number of address resolution steps we'll perform for a single
-// peer (for all addresses).
+// peer (for all addresses). In other words, the maximum number of times we'll
+// call `.Resolve` for all of a peer's addresses
 const maxAddressResolution = 32
```
sukunrt, line 27: I prefer `maxDNSResolutions`? We only use it for DNS, right?

p2p/net/swarm/swarm_dial.go
```diff
@@ -372,6 +380,9 @@ func (s *Swarm) resolveAddrs(ctx context.Context, pi peer.AddrInfo) ([]ma.Multia
 			}
 			var added bool
 			for _, a := range resolvedAddrs {
+				if len(toResolve) >= maximumToResolveAddresses {
+					break
+				}
 				if !addr.Equal(a) {
 					toResolve = append(toResolve, a)
 					added = true
```
sukunrt, line 388: Do we need to maintain this variable `added`? When this block ends, we want to continue with the outer loop, as we have resolved the variable with the transport, right?

The current code says, if the resolver didn't modify the address, we should try to resolve it using our madns resolver. Do we want that? If so, can you add a comment for it?

p2p/net/swarm/swarm_dial.go
```diff
@@ -382,6 +393,11 @@ func (s *Swarm) resolveAddrs(ctx context.Context, pi peer.AddrInfo) ([]ma.Multia
 			}
 		}
 
+		if len(toResolve) >= maximumToResolveAddresses {
+			// No point in resolving any more addresses, we don't have space.
+			continue
+		}
```
sukunrt, line 399: Should this check be before the previous block that does resolution with Transport? 

If we don't have space in resolve list, there's not much the transport resolution can do either.

p2p/net/swarm/swarm_dial.go
```diff
@@ -391,6 +407,9 @@ func (s *Swarm) resolveAddrs(ctx context.Context, pi peer.AddrInfo) ([]ma.Multia
 
 		// add the results to the toResolve list.
 		for _, res := range resaddrs {
+			if len(toResolve) >= maximumToResolveAddresses {
+				break
+			}
 			pi, err := peer.AddrInfoFromP2pAddr(res)
 			if err != nil {
 				log.Infof("error parsing %s: %s", res, err)
```
sukunrt, line 415: Do we need to make a peerinfo object here?

p2p/net/swarm/swarm_dial.go
```diff
@@ -337,7 +345,7 @@ func (s *Swarm) resolveAddrs(ctx context.Context, pi peer.AddrInfo) ([]ma.Multia
 	// * Otherwise, resolve it and add the results to the "to resolve" list.
 	toResolve := append([]ma.Multiaddr{}, pi.Addrs...)
 	resolved := make([]ma.Multiaddr, 0, len(pi.Addrs))
```
sukunrt, line 347: I think this code would be nicer if we change the invariants to:

1. Addresses in the resolved list are all fully resolved.
2. Addresses in the toResolve list are DNS addresses that aren't fully resolved.

p2p/net/swarm/swarm_dial.go
```diff
@@ -382,6 +393,11 @@ func (s *Swarm) resolveAddrs(ctx context.Context, pi peer.AddrInfo) ([]ma.Multia
...
 		}
 
+		if len(toResolve) >= maximumToResolveAddresses {
+			// No point in resolving any more addresses, we don't have space.
+			continue
+		}
+
 		// otherwise, resolve it
 		reqaddr := addr.Encapsulate(p2paddr)
 		resaddrs, err := s.maResolver.Resolve(ctx, reqaddr)
```
sukunrt, line 403: Do you know why we do this? Do we need the peer ID to resolve DNS addresses?

MarcoPolo: I believe it's to assert that the dnsaddr we get has the matching peer id suffix (`/p2p/foo`). I don't love that it's so subtle

MarcoPolo:
I'm going to refactor this function and clean up things around here. As a side effect, I think we won't need #2989 anymore.

MarcoPolo:
@sukunrt this is ready now. Hopefully it's a lot clearer

p2p/net/swarm/swarm.go
```diff
@@ -846,3 +855,95 @@ func (c connWithMetrics) Stat() network.ConnStats {
...
+			toResolve = append(toResolve, addr)
+		} else {
+			resolved = append(resolved, addr)
+		}
+	}
+
+	for _, addr := range toResolve {
+		resolvedAddrs, err := r.ResolveDNSAddr(ctx, expectedPeerID, addr, recursionLimit-1, outputLimit-len(resolved))
+		if err != nil {
+			log.Warnf("failed to resolve dnsaddr %v %s: ", addr, err)
```
sukunrt, line 907: Is there a potential exponential case here? 

Consider a dnsaddr that resolves to 10 dnsaddrs. Each of these 10 dnsaddrs resolve to 10 dnsaddrs and so on. 

`outputLimit-len(resolved)` will be 10 because all of these are dnsaddrs and len(resolved) will be 0. So we call `ResolveDNSAddrs` on these 10 addrs and then call it again with outputLimit 10 for each of these and so on.

p2p/net/swarm/swarm.go
```diff
@@ -846,3 +855,95 @@ func (c connWithMetrics) Stat() network.ConnStats {
...
+			// Dropping this address
+			continue
+		}
+		resolved = append(resolved, resolvedAddrs...)
+	}
+
+	if len(resolved) > outputLimit {
+		resolved = resolved[:outputLimit]
+	}
```
sukunrt, line 916: Limit the size after every append.

p2p/net/swarm/swarm.go
```diff
@@ -145,6 +145,14 @@ func WithReadOnlyBlackHoleDetector() Option {
 	}
 }
 
+type MultiaddrDNSResolver interface {
+	// ResolveDNSAddr resolves the first /dnsaddr component in a multiaddr.
+	// Recurisvely resolves DNSADDRs up to the recursion limit
+	ResolveDNSAddr(ctx context.Context, expectedPeerID peer.ID, maddr ma.Multiaddr, recursionLimit, outputLimit int) ([]ma.Multiaddr, error)
+	// ResolveDNSComponent resolves the first /{dns,dns4,dns6} component in a multiaddr.
+	ResolveDNSComponent(ctx context.Context, maddr ma.Multiaddr, outputLimit int) ([]ma.Multiaddr, error)
+}
```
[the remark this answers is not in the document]
MarcoPolo: Discussed sync. Exporting this to prefer this interface over the concrete pointer type in madns.

options.go
```diff
@@ -495,7 +494,7 @@ func UserAgent(userAgent string) Option {
 }
 
 // MultiaddrResolver sets the libp2p dns resolver
-func MultiaddrResolver(rslv *madns.Resolver) Option {
+func MultiaddrResolver(rslv swarm.MultiaddrDNSResolver) Option {
```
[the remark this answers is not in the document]
MarcoPolo: Right now only the swarm uses this. No strong opinion on whether it should be a swarm opt or libp2p opt. It feels general enough to be a libp2p opt.

p2p/net/swarm/swarm.go
MarcoPolo: We shouldn't need to since the function that gives us resolvedAddrs is only returning up to the limit. Even this section is probably not necessary.

sukunrt: you're right. I missed the part where nextOutputLimit was reduced for successive calls based on len(resolved)

PR #2991, opened 2024-10-02
autonatv2: don't dial dns addresses

sukunrt:


sukunrt:
I'm not completely convinced about this change. Testing addresses end-to-end is the while point of autonat v2

PR #2992, opened 2024-10-02
autonatv2: recover from panics

sukunrt:


p2p/protocol/autonatv2/client.go
```diff
@@ -248,6 +250,12 @@ func newDialRequest(reqs []Request, nonce uint64) pb.Message {
 
 // handleDialBack receives the nonce on the dial-back stream
 func (ac *client) handleDialBack(s network.Stream) {
+	defer func() {
```
MarcoPolo, line 253: Do we need to ensure that the stream has been closed?

MarcoPolo (commented):
One question around cleaning up the stream

p2p/protocol/autonatv2/client.go
sukunrt: Yes. We should reset the stream on panic.

MarcoPolo (approved):

PR #2996, opened 2024-10-08
fix: Troubleshooting too many local IP allocation failure logs

elecbug:
Currently log systems periodically generate too many local IPv6 route failure messages.
It repeats the same content and only grows the size of the log.
Unless there are changes, the local IP binding failure message has been modified to occur only once.

PR #2999, opened 2024-10-08
fix: basichost: Add backoff for updating local IP addresses on error

MarcoPolo:
closes #2996


sukunrt:
Do we need this interface selection logic at all? Shouldn't we advertise all interface addresses when we are listening on `0.0.0.0` or `::`?

MarcoPolo:
Maybe. 

1. This isn't related to listen address.
1. Really the whole basichost's Addrs logic needs a overhaul.
1. This is semantically the same, but nicer in the case of errors.

sukunrt:
Isn't the whole point of this method to get our interface addresses so that we can accurately report our _listen addresses_ and not `0.0.0.0`?

sukunrt (approved):

PR #3022, opened 2024-10-24
WIP: Feat/memory transport

### LINKED ISSUE #2293: transport: add in memory transport
Add an in memory transport (similar to rust-libp2p) which can be used for testing https://docs.rs/libp2p/latest/libp2p/core/transport/struct.MemoryTransport.html

pyropy:
Closes #2293


MarcoPolo:
Friendly reminder to test your transport by adding a case here: https://github.com/libp2p/go-libp2p/blob/581571b349ef023d232c893118309bb95895ea99/p2p/test/transport/transport_test.go#L69

pyropy:
> Friendly reminder to test your transport by adding a case here:
> 
> https://github.com/libp2p/go-libp2p/blob/581571b349ef023d232c893118309bb95895ea99/p2p/test/transport/transport_test.go#L69

Thanks, will add it. 

Could you possibly take a look at this PR https://github.com/multiformats/go-multiaddr/pull/256 ?

pyropy:
I have few questions:

- When `OpenStream` is called on the `Connection`, how is the returned stream used? Is the `AcceptStream` called after `OpenStream`? My concern is about how the channels would be exchanged between the opened streams for the connection. I guess if `OpenStream` call is followed by `AcceptStream` we can send stream to "remote" connection.
- Should we keep using `io.Pipe` or switch it up with channels? I've seen that `io.Pipe` is using channels in the background.

sukunrt:
>When OpenStream is called on the Connection, how is the returned stream used? Is the AcceptStream called after OpenStream? My concern is about how the channels would be exchanged between the opened streams for the connection. I guess if OpenStream call is followed by AcceptStream we can send stream to "remote" connection.

`OpenStream` and `AcceptStream` happen on different libp2p hosts. Consider two hosts _A_ and _B_. When _A_ calls OpenStream, it creates a stream it can write to and read from. _B_ will get this stream when it calls `AcceptStream` on the connection. 
In the swarm, `AcceptStream` happens automatically because every open connection is waiting on `AcceptStream` here: https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/swarm_conn.go#L118
For implementing a transport, the transport should buffer streams till they are accepted by calling `AcceptStream`. 


>Should we keep using io.Pipe or switch it up with channels? I've seen that io.Pipe is using channels in the background.

I think io.Pipe is fine. This transport will only be used for tests, so it's nice to get as much of the work done by the stdlib as we can.

p2p/transport/memory/conn_test.go
```diff
@@ -0,0 +1 @@
+package memory
```
sukunrt, line 1: Write these before transport integration tests.

sukunrt (changes requested):
I think the tests are specifically failing because you don't support deadlines and basic_host expects deadline support. https://github.com/libp2p/go-libp2p/blob/master/p2p/host/basic/basic_host.go#L448

sukunrt:
Closing this. Memory transport for libp2p will be #3262

Issue #2791, opened 2024-05-15
`GenerateSecp256k1` doesn't use the provided random reader source.


jsimnz:
Based on the latest version (33.2).

Currently the [`GenerateSecp256k1Key(src io.Reader)`](https://github.com/libp2p/go-libp2p/blob/a86d94e4084861dafb35ce85af0bd7262324a42d/core/crypto/secp256k1.go#L23) accepts a `io.Reader` parameter to use as its source of random ness to generate the new key, however it doesn't actually use the provided reader, and instead uses the default random reader within the `secp256k1` package.

However, the underlying cryptographic package does support using a provided source of randomness as seen [here](https://pkg.go.dev/github.com/decred/dcrd/dcrec/secp256k1/v4#GeneratePrivateKeyFromRand).

Stebalien:
This should probably be documented, but likely isn't going to be fixed.

jsimnz:
Why do you feel it shouldn't be fixed? The function takes in an argument that doesn't do anything. 

Additionally, the equivalent function in the crypto package for generating Ed25519 keys which has the same `io.Reader` argument which *is* actually used if supplied.

So it's inconsistent behavior between equivalent functions?

MarcoPolo:
Seems like a reasonable request. feel free to open a PR and I can review.

@Stebalien am I missing some reason for why we shouldn't fix this?

Stebalien:
Ah, I didn't realize decred actually provided a way to do this.

My response comes from the fact that:

1. While the go standard library takes a random source when generating private keys, it _intentionally_ (and randomly) reads a few bytes from the source before generating the key to prevent users from relying on this for deterministic key generation (turned out to be a compatibility footgun).
2. The library we just switched to in Filecoin doesn't even provide a way to specify a random source.

I agree that taking an argument that doesn't do anything isn't great, but I expect it's that way for historical reasons (i.e., from before the switch to decred). But... we probably shouldn't have taken a random source in the first place.

- If you don't trust `crypto.Rand`, any crypto you do will likely be broken.
- As noted above, it's a compatibility footgun.

jsimnz:
I can certainly open a PR as I have a branch I'm using for this. 

@Stebalien at least from my POV it's not about trusting crypto.Rand, but intentionally wanting deterministic keys for certain testing scenarios.

It should be also noted that some of the standard lib crypto libs (eg ecdsa) has specific behavior for supplied `io.Reader` for Generate functions.

Specifically
```
[...]

PR #3029, opened 2024-11-06
fix: GenerateSecp256k1Key use src as seed

tvi:
I think `GeneratePrivateKeyFromRand` was not available before, so it was not passed.


tvi:
Hi @MarcoPolo, I was wondering if you could take a look at this PR.

MarcoPolo:
Could you comment on https://github.com/libp2p/go-libp2p/issues/2791 and see if any of the suggestions there serve your use case?

p-shahi:
@tvi can you comment on the issue Marco pointed out? Please reopen PR if necessary

Issue #2983, opened 2024-09-27
Issues with identify: identify failed to complete


vyzo:
The problem: when I disconnect and reconnect to a peer, the latter finds it impossible to open a stream, with the error indication that identified failed to complete. This happens consistently in our node, and I can trigger it reliably; so there is some bug related to identify. 

Relevant logs:
```
2024-09-27T12:48:21.519Z	DEBUG	net/identify	identify/id.go:543	/ipfs/id/push/1.0.0 received message from 12D3KooWEuVEZ26PwP7f8mYxuQ97ChiqLvXL5JeDfmXmqN4ds3YT /ip4/45.139.212.29/udp/17846/quic-v1
2024-09-27T12:48:24.657Z	WARN	libp2p	libp2p/libp2p.go:380	send: failed to open stream to peer 12D3KooWEuVEZ26PwP7f8mYxuQ97ChiqLvXL5JeDfmXmqN4ds3YT: identify failed to complete: context deadline exceeded
2024-09-27T12:48:24.680Z	DEBUG	basichost	basic/basic_host.go:483	negotiated: /ipfs/id/push/1.0.0 (took 27.52µs)
2024-09-27T12:48:24.680Z	DEBUG	net/identify	identify/id.go:543	/ipfs/id/push/1.0.0 received message from 12D3KooWEuVEZ26PwP7f8mYxuQ97ChiqLvXL5JeDfmXmqN4ds3YT /ip4/45.139.212.29/udp/17846/quic-v1
2024-09-27T12:48:30.475Z	DEBUG	basichost	basic/basic_host.go:483	negotiated: actor/root/messages/0.0.1 (took 48.66µs)
2024-09-27T12:48:30.479Z	DEBUG	actor	actor/dispatch.go:205	dispatching message from {BAAREICLTJJ63SCP5MUFVGX7GUM42QL2I5E46I355CASAEDIL55VX4SHTA====== did:key:z6MkjYP7Q8m862LdqNtExNMcH1BzgAkYTVp6QC5gHesktKbu {12D3KooWEuVEZ26PwP7f8mYxuQ97ChiqLvXL5JeDfmXmqN4ds3YT root}} to /public/hello
2024-09-27T12:48:30.479Z	DEBUG	node	node/public.go:59	hello from 12D3KooWEuVEZ26PwP7f8mYxuQ97ChiqLvXL5JeDfmXmqN4ds3YT
[...]

PR #3031, opened 2024-11-07
feat: eventbus: log error on slow consumers

### LINKED ISSUE #2361: eventbus: log a warning if an event channel is full (and continue to block)
We had [another case](https://github.com/libp2p/go-libp2p-kad-dht/issues/849) pop up where it wasn't obvious that an event channel from the eventbus was full and things were getting blocked. We could be nice here and log something for the user with the tradeoff of paying for an extra timer and select. I'm thinking something like this [here](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/eventbus/basic.go#L407).

### Before

```go
sink.ch <- evt
```

### After

```go
timer.Reset(time.Second)
select {
case <-timer:
	log.Errorf("Event to %s is being delayed because the event channel is full. Are you doing blocking work? If so, spawn a goroutine.", sink.name, len(sink.ch))
	sink.ch <- evt // still block
case sink.ch <- evt:
}
```
(and some extra logic around cleaning up and reseting the timer. Ideally keeping this timer as part of the `*node`)

MarcoPolo:
Slow consumers can stall libp2p in hard to debug ways. For example, a slow consumer of Identify events can prevent NewStreams from opening. In some cases we do want this back-pressure, but in other cases it is a bug in the user's application (or even in go-libp2p). The recommended approach has been to use metrics with Grafana and Prometheus to identify these issues. In practice, that's been a non-trivial task for users to setup. A simple log would help identify these issues.

Note that the codepath is relatively unchanged in the normal case. Only if a producer is stalled will the extra work be put in.

Closes #2361 . See my comment https://github.com/libp2p/go-libp2p/issues/2983#issuecomment-2463270292 for why this is brought up again.

@vyzo Could you please try to repro your issue on this branch?

### DIFF
--- a/p2p/host/eventbus/basic.go
+++ b/p2p/host/eventbus/basic.go
@@ -6,9 +6,19 @@ import (
 	"reflect"
 	"sync"
 	"sync/atomic"
+	"time"
 
+	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p/core/event"
 )
+
+type logInterface interface {
+	Errorf(string, ...interface{})
+}
+
+var log logInterface = logging.Logger("eventbus")
+
+const slowConsumerWarningTimeout = time.Second
 
 // /////////////////////
 // BUS
@@ -322,6 +332,8 @@ type wildcardNode struct {
 	nSinks        atomic.Int32
 	sinks         []*namedSink
 	metricsTracer MetricsTracer
+
+	slowConsumerTimer *time.Timer
 }
 
 func (n *wildcardNode) addSink(sink *namedSink) {
@@ -348,6 +360,8 @@ func (n *wildcardNode) removeSink(ch chan interface{}) {
 	n.Unlock()
 }
 
+var wildcardType = reflect.TypeOf(event.WildcardSubscription)
+
 func (n *wildcardNode) emit(evt interface{}) {
 	if n.nSinks.Load() == 0 {
 		return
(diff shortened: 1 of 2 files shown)


vyzo:
Yes, I will try it.

p2p/host/eventbus/basic.go
```diff
@@ -360,7 +374,11 @@ func (n *wildcardNode) emit(evt interface{}) {
...
 		sendSubscriberMetrics(n.metricsTracer, sink)
 
-		sink.ch <- evt
+		select {
+		case sink.ch <- evt:
+		default:
+			n.slowConsumerTimer = emitAndLogError(n.slowConsumerTimer, wildcardType, evt, sink)
+		}
 	}
 	n.RUnlock()
```
sukunrt, line 383: This won't work with an `RLock`, right? We need Lock since we're updating state on `n`

I think it's fine to create the timer right when we create the node. I don't think there are that many nodes in total.

MarcoPolo: I didn't realize this was an `RLock`. Thanks for highlighting that. I've updated the code to set the timer within a lock. It's technically possible to allocate two or more stopped timers, but that should be fine as it'll get cleaned up by the GC.
 
> I think it's fine to create the timer right when we create the node.

I considered this, but there's no way to create a stopped timer. You must create an armed one. So if you are going through the trouble of setting one, you may as well use it.

sukunrt (approved):

PR #3032, opened 2024-11-07
 [DRAFT] feat: conntracker

MarcoPolo:
This PR attempts to refactor how NewStream works. Instead of having the swarm be responsible for picking the best connection, and then waiting on identify to run on that connection. We have a separate service that subscribes to connection events and identify events, and the NewStream asks for the best connection that has been identified.

This has a couple of advantages:
1. We can wait for a Identify to complete on _any_ connection. Not just the one we happen to pick.
2. It moves connection tracking code out of the swarm into a smaller service that can be tested independently.
  a. I'd even like to run a fuzz test on the conntracker service itself.
3. It attaches a peer's supported protocols to the connection itself rather than the peer.
  a. This allows a peer to say they support protocol X and Y over a connection, and only support Z over another one. This solves a similar problem to labelling connections as "limited", but lets the remote have a say in how we use the connection rather than the local side assuming something. But note that we aren't actually using this just yet. See also https://github.com/libp2p/go-libp2p/issues/2693

This is a draft for now as there are a couple of TODOs in the code, some more tests I want to add, and some thought in how to roll this out (maybe default on, but can disable with an ENV flag for debugging purposes.)

This is also how I think we should build on and improve go-libp2p moving forward. Simple services that share information over an eventbus or channels (depending on the use case).

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -22,6 +22,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/transport"
 	"github.com/libp2p/go-libp2p/p2p/host/autonat"
 	"github.com/libp2p/go-libp2p/p2p/host/basic/internal/backoff"
+	"github.com/libp2p/go-libp2p/p2p/host/conntracker"
 	"github.com/libp2p/go-libp2p/p2p/host/eventbus"
 	"github.com/libp2p/go-libp2p/p2p/host/pstoremanager"
 	"github.com/libp2p/go-libp2p/p2p/host/relaysvc"
@@ -74,6 +75,7 @@ type BasicHost struct {
 	// keep track of resources we need to wait on before shutting down
 	refCount sync.WaitGroup
 
+	connTracker  *conntracker.ConnTracker
 	network      network.Network
 	psManager    *pstoremanager.PeerstoreManager
 	mux          *msmux.MultistreamMuxer[protocol.ID]
@@ -92,6 +94,7 @@ type BasicHost struct {
 	emitters struct {
 		evtLocalProtocolsUpdated event.Emitter
 		evtLocalAddrsUpdated     event.Emitter
+		evtProtoNegotiation      event.Emitter
 	}
 
 	addrChangeChan chan struct{}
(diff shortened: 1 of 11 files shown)


p2p/host/basic/basic_host.go
```diff
@@ -170,6 +173,8 @@ type HostOpts struct {
 	DisableIdentifyAddressDiscovery bool
 	EnableAutoNATv2                 bool
 	AutoNATv2Dialer                 host.Host
+
+	ConnTracker *conntracker.ConnTracker
 }
```
sukunrt, line 178: How can users of basic host get access to the ConnTracker?

sukunrt (commented):
I think this makes sense. 

I need to think more about any race condition involving swarm and identify events ordering. Not sure we've surely dealt with https://github.com/libp2p/go-libp2p/issues/2983

Should we add a NewStream method to ConnTracker? This allows us to check whether the connection we're returning is closed already and use some other connection. 

Overall, I like reducing the size of swarm and to move protocols to connection. Maybe one day we can even deprecate Peerstore peer protocols api.

p2p/host/basic/basic_host.go
MarcoPolo: from Fx and requiring it as a input parameter.

p2p/host/basic/basic_host.go
```diff
@@ -685,6 +708,80 @@ func (h *BasicHost) RemoveStreamHandler(pid protocol.ID) {
...
+
+	// Wait for a connection that works for us
+	connChan, err := h.connTracker.GetBestConnChan(ctx, p, conntracker.GetBestConnOpts{
+		OneOf:           requiredProtos,
+		FilterFn:        connFilter,
+		WaitForIdentify: true, // Old behavior
+	})
+	if err != nil {
+		return conntracker.ConnWithMeta{}, err
+	}
```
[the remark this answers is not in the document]
MarcoPolo: Not opposed, what do we do when we get back multiple connections?

MarcoPolo:
I probably won't get back to this for a while, so this is PR is pick-up-able.

As a current core dump:
- The PR description is still accurate.
- We need to review carefully if there are any implied assumptions about event ordering here that are wrong.
- A fuzz test should be required prior to merge.

MarcoPolo:
and fwiw, I think this is still a better way of picking a "best" connection to use than what we're currently doing.

MarcoPolo:

PR #3034, opened 2024-11-11
eventbus: dont panic on closing Subscription twice

sukunrt:


MarcoPolo:
I'm not sure. Why wouldn't this follow the same semantics as closing a channel twice? https://play.golang.com/p/S5frqwAAI1j. That panics, and this seems similar to me.

sukunrt:
Not a fan of that decision either. But channel closing twice panics so that it explicitly disallows broken concurrency primitives and avoids race conditions. In our case that's not a problem. There's actually no way to synchronize a subscriber with an emitter. In fact, when dealing with channels, the receiver *never* closes a channel, it's always the sender who closes the channel because otherwise the sender might panic.

MarcoPolo (approved):
I'm still not convinced by the argument, but I really don't like panics.

sukunrt:
I don't like them either!

PR #3035, opened 2024-11-12
fix: identify: push should not dial a new connection

MarcoPolo:
There a small window after a peer disconnects, but before identify has updated its conn map that would cause the host to dial a peer in order to send a push message. This prevents that

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -347,6 +347,10 @@ func (ids *idService) sendPushes(ctx context.Context) {
 			defer func() { <-sem }()
 			ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 			defer cancel()
+
+			// We only want to send an identify push if we already have an open
+			// connection.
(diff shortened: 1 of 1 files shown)


p2p/protocol/identify/id.go
```diff
@@ -347,6 +347,10 @@ func (ids *idService) sendPushes(ctx context.Context) {
 			defer func() { <-sem }()
 			ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 			defer cancel()
+
+			// We only want to send an identify push if we already have an open
+			// connection.
+			ctx = network.WithNoDial(ctx, "id push")
 			str, err := ids.Host.NewStream(ctx, c.RemotePeer(), IDPush)
```
sukunrt, line 354: Should we create the stream explicitly on the connection and do multistream select explicitly like we do for identify(non push). This might use a different connection to open the stream.

MarcoPolo: Yeah, that's better. done on b53c174b85a9

sukunrt (approved):

PR #3036, opened 2024-11-12
fix(tcpreuse): handle connection that failed to be sampled

aschmahmann:
There was an oversight here in the error reporting. A test should be added, but also probably this should've been caught by automated tooling (e.g. one of the linters), maybe there are more we should be enabling.

### DIFF
--- a/p2p/transport/tcpreuse/demultiplex.go
+++ b/p2p/transport/tcpreuse/demultiplex.go
@@ -40,35 +40,35 @@ func (t DemultiplexedConnType) IsKnown() bool {
 
 // identifyConnType attempts to identify the connection type by peeking at the
 // first few bytes.
-// It Callers must not use the passed in Conn after this
-// function returns. if an error is returned, the connection will be closed.
+// Its Callers must not use the passed in Conn after this function returns.
+// If an error is returned, the connection will be closed.
 func identifyConnType(c manet.Conn) (DemultiplexedConnType, manet.Conn, error) {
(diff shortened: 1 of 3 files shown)


p2p/transport/tcpreuse/demultiplex.go
```diff
@@ -40,35 +40,35 @@ func (t DemultiplexedConnType) IsKnown() bool {
...
-// function returns. if an error is returned, the connection will be closed.
+// Its Callers must not use the passed in Conn after this function returns.
+// If an error is returned, the connection will be closed.
 func identifyConnType(c manet.Conn) (DemultiplexedConnType, manet.Conn, error) {
 	if err := c.SetReadDeadline(time.Now().Add(identifyConnTimeout)); err != nil {
 		closeErr := c.Close()
 		return 0, nil, errors.Join(err, closeErr)
 	}
 
-	s, c, err := sampledconn.PeekBytes(c)
```
MarcoPolo, line 51: These changes aren't necessary, right? It's also not a peekableConn. It's just a manet.Conn. If you want to rename, i would rename the parameter.

MarcoPolo:
The issue was that the returned connection may be nil when an error is returned, right?

MarcoPolo:
https://github.com/uber-go/nilaway Seems promising, but not sure if it would have caught this issue.

p2p/transport/tcpreuse/demultiplex.go
MarcoPolo: Ah, I see. This shadows the `c` in L53.

MarcoPolo (approved):

PR #3039, opened 2024-11-13
fix(tcpreuse): remove windows specific code

MarcoPolo:
Unclear why, but this causes a lot of CPU usage according to @aschmahmann

I wonder if we are spinning in the for loop. Is it possible the Windows api returns no bytes read and no error?


MarcoPolo:
need to debug the failing test still

MarcoPolo:
just an incomplete read in the test. Not a real issue.

MarcoPolo:
Merging since this may be related to the issues in #3036

PR #3045, opened 2024-11-18
fix(eventbus): Idempotent wildcardSub close

MarcoPolo:
The tests expects the wildcardSub .Close to be idempotent, but it never checked this. The tests also weirdly asserted that there were orphaned events on the channel. That was wrong as the events should be flushed on close as to not indefinitely block an emitter.

The wrong test is what causes #3039 to fail.


MarcoPolo:
This is causing flakiness in other PRs, so merging this now.

sukunrt:
Thanks!

PR #3047, opened 2024-11-18
refactor(libp2phttp): don't require specific port for the HTTP host example

MarcoPolo:


MarcoPolo:
merging because it fixes a flaky test

PR #3048, opened 2024-11-19
refactor(core/routing): split ContentRouting interface

hsanjuan:
The ContentRouting interface does too much. One can think of systems that only Provide, and systems that only FindProviders.

Therefore I am proposing to split it further. While this can be done on user-code, core/routing interfaces are canonical and it is better to use interfaces coming from here.

As someone said: the best interface is that which does as little as possible.

### DIFF
--- a/core/routing/routing.go
+++ b/core/routing/routing.go
@@ -18,22 +18,33 @@ var ErrNotFound = errors.New("routing: not found")
 // type/operation.
 var ErrNotSupported = errors.New("routing: operation or key not supported")
 
+// ContentProviding is able to announce where to find content on the Routing
+// system.
+type ContentProviding interface {
+	// Provide adds the given cid to the content routing system. If 'true' is
+	// passed, it also announces it, otherwise it is just kept in the local
+	// accounting of which objects are being provided.
+	Provide(context.Context, cid.Cid, bool) error
+}
+
+// ContentDiscovery is able to retrieve providers for a given CID using
+// the Routing system.
+type ContentDiscovery interface {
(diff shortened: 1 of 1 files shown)


core/routing/routing.go
```diff
@@ -18,24 +18,35 @@ var ErrNotFound = errors.New("routing: not found")
...
+type ContentProvider interface {
 	// Provide adds the given cid to the content routing system. If 'true' is
 	// passed, it also announces it, otherwise it is just kept in the local
 	// accounting of which objects are being provided.
 	Provide(context.Context, cid.Cid, bool) error
+}
 
+// A ContentProviderFinder is able to retrieve providers for a given CID using
+// the Routing system.
+type ContentProviderFinder interface {
```
sukunrt, line 32:
```suggestion
type ContentDiscovery interface {
```

hsanjuan: Then should also rename ContentProvider to ContentProviding to have consistent language? Also consisting with PeerRouting

willscott:
Chiming in to say I agree with this direction - IPNI is a good example of a system where we've run into issues with there being a desire for the system to support provides as IPFS defines in delegated routing as part of being able to be used for queries.

core/routing/routing.go
hsanjuan: done

sukunrt (commented):
I'm fine with the change, but will let Marco opine as he has more understanding of this side of the codebase.

MarcoPolo (approved):

PR #3049, opened 2024-11-19
fix(holepunch/tracer): replace inline peer struct with peerInfo type

cpeliciari:
- Introduced the `peerInfo` struct to encapsulate peer metadata.
- Updated the `peers` map in the `tracer` struct to use `peerInfo`.
- Simplified code by avoiding inline struct definition for peers.


MarcoPolo (approved):

MarcoPolo:
thanks!

PR #3052, opened 2024-11-20
tcp: fix metrics test build directive

sukunrt:
This test works on all platforms. It tests for transport correctness in case we have metrics enabled. 
This removes the incorrect directive(whitespace in front of the name) and the platform specific file suffix.


p-shahi:
@sukunrt Do we still need this PR?

sukunrt:
@p-shahi looks reasonable to me. Updated the description to explain the change.

p-shahi (approved):

PR #3054, opened 2024-11-21
chore(dashboard/alertmanager): update api version from v1 to v2

cpeliciari:
The error message was:
Error loading config (--config.file=/etc/prometheus/prometheus.yml): expected Alertmanager API version to be one of [v2] but got v1

### DIFF
--- a/dashboards/prometheus.yml
+++ b/dashboards/prometheus.yml
@@ -6,7 +6,7 @@ alerting:
   alertmanagers:
   - scheme: http
     timeout: 10s
-    api_version: v1
+    api_version: v2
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Our config is still valid with just this version change?

cpeliciari:
Just changing this functioning normally. I thought of fixing the version of prometheus and grafana. But I see no need

cpeliciari:
Currently in the version, when you try to run the compose of the Prometheus initialization failure, it only reports the version.

MarcoPolo (approved):

Issue #3057, opened 2024-11-21
NATMap mapping failed, resulting in an observed address error


wlynxg:
<details>
<summary>Version Information</summary>
<pre>
github.com/libp2p/go-libp2p v0.37.0
</pre>
</details>

I found that my libp2p node's address list had a NATMap address with port 0 and an observed address with port 0.
```
2024-11-21T09:15:44.482Z        DEBUG   net/identify    identify/id.go:486      sending snapshot        {"seq": 7, "protocols": ["/ipfs/id/1.0.0","/ipfs/id/push/1.0.0","/ipfs/ping/1.0.0","/libp2p/circuit/relay/0.2.0/stop"], "addrs": [...,"/ip4/100.64.138.163/tcp/0","/ip4/100.64.138.163/udp/0/quic-v1/webtransport/certhash/uEiAO2S9URsYyw-LBGIJytw3zQbw8pvXgomESq3PLp31PVg/certhash/uEiCxf6Zi2SitEayX9jE50_APDrBMbfR5m-U6F7HcBMnkFw","/ip4/{observed ip}/tcp/0",...]}
```

At the same time, I found that my NAT mapping had the following error due to too many mapped addresses:
```
2024-11-21T09:15:42.049Z        WARN    nat     nat/nat.go:239  failed to establish port mapping: SOAP fault. Code: s:Client | Explanation: UPnPError | Detail: <UPnPError xmlns="urn:schemas-upnp-org:control-1-0"><errorCode>501</errorCode><errorDescription>Port mapping table id full !</errorDescription></UPnPError>
```

After debugging, I found that the problem is caused by:
1. When NAT mapping fails, port 0 will be returned
https://github.com/libp2p/go-libp2p/blob/288868c0d90e2d72a124da26caa14c77d35dbbdd/p2p/net/nat/nat.go#L245

2. When basic_host calls AllAddrs, it gets the mapping address from natmgr, and natmrg gets the mapping address from nat. However, nat does not determine the mapping failure, resulting in port 0 being returned.
https://github.com/libp2p/go-libp2p/blob/288868c0d90e2d72a124da26caa14c77d35dbbdd/p2p/host/basic/basic_host.go#L902
https://github.com/libp2p/go-libp2p/blob/288868c0d90e2d72a124da26caa14c77d35dbbdd/p2p/host/basic/natmgr.go#L269
[...]

PR #3058, opened 2024-11-21
fix: don't use invalid port 0 when NATMap fails

wlynxg:
fix: https://github.com/libp2p/go-libp2p/issues/3057

### DIFF
--- a/p2p/host/basic/basic_host.go
+++ b/p2p/host/basic/basic_host.go
@@ -905,8 +905,23 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
 				continue
 			}
 
+			// Drop the IP from the external maddr
+			_, extMaddrNoIP := ma.SplitFirst(extMaddr)
+
+			// Determine whether the mapped port is valid
+			isValidPort := true
+			ma.ForEach(extMaddrNoIP, func(c ma.Component) bool {
+				switch c.Protocol().Code {
+				case ma.P_TCP, ma.P_UDP:
+					isValidPort = c.Value() != "0"
+					return false
+				default:
+					return true
+				}
+			})
+
 			// if the router reported a sane address
-			if !manet.IsIPUnspecified(extMaddr) {
+			if !manet.IsIPUnspecified(extMaddr) && isValidPort {
 				// Add in the mapped addr.
 				finalAddrs = append(finalAddrs, extMaddr)
 			} else {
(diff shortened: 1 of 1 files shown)


p2p/net/nat/nat.go
```diff
@@ -107,7 +107,7 @@ func (nat *NAT) GetMapping(protocol string, port int) (addr netip.AddrPort, foun
 		return netip.AddrPort{}, false
 	}
 	extPort, found := nat.mappings[entry{protocol: protocol, port: port}]
-	if !found {
+	if !found || extPort == 0 {
 		return netip.AddrPort{}, false
 	}
```
sukunrt, line 112: Remove the check from here.

p2p/host/basic/basic_host.go
```diff
@@ -929,6 +931,8 @@ func (h *BasicHost) AllAddrs() []ma.Multiaddr {
 				continue
 			}
 
+			successMapCnt++
+
```
sukunrt, line 935: Add the port 0 check to line 911 where we check for is unspecifid address. If you do so, it removes the need for introducing this new successMapCnt variable. 

I would prefer keeping this change small because this code is pretty bad and we intend to rewrite it soon in #2229

sukunrt (changes requested):
Thanks for the PR @wlynxg

p2p/host/basic/basic_host.go
wlynxg: Does the check placed on line 911 mean that the `continue` is executed when the `extMaddr` port is 0? If so, when all `extMaddr` ports are 0, will `observedAddrs` not be added to the `finalAddrs` address list?

sukunrt: I don't think so. 
```
		if !manet.IsIPUnspecified(extMaddr) {
				// Add in the mapped addr.
				finalAddrs = append(finalAddrs, extMaddr)
} else {
```
If the nat provided address is not an unspecified addr and(new change) it's not port == 0, we will add it to final addrs. 
We still fall through and add the observed addresses.

wlynxg: It is OK to add the address whose port is not 0 to finalAddrs here.

However, at https://github.com/libp2p/go-libp2p/blob/f42120245634b2f6af12f9889859e9f92ac8e588/p2p/host/basic/basic_host.go#L952, it is also necessary to judge whether the port is 0. If the port is 0, should the obsMaddr address be added directly without splicing?

p-shahi:
@wlynxg thanks for the contribution - Would you like to address Sukun's feedback?

wlynxg:
> @wlynxg thanks for the contribution - Would you like to address Sukun's feedback?感谢您的贡献 - 您想回应 Sukun 的反馈吗？

I will refer to Sukun's feedback and re-edit the code later.

sukunrt:
Apologies for the confusion here. This specific method's code is too confusing. I think your original approach of erroring when port is 0 was correct. 

closing in favour of: #3094 
rest is handled by: #3075

PR #3059, opened 2024-11-21
chore(dashboard/pyroscope): integrate pyroscope database and dashboard

cpeliciari:
- Added Pyroscope database configuration in Grafana
- Included Pyroscope service in the docker-compose setup
- Created a new dashboard named "go-libp2p Profiling" for visualizing profiling data

image


sukunrt:
Thanks for the PR @cpeliciari. I'll take a look at this some time this week. I just need to think about whether we have the bandwidth to support and maintain this. cc @MarcoPolo

p-shahi:
@cpeliciari this seems useful; echo Sukun's concern about maintenance. How did you generate the json file, and could you add some more info into the readme about Pyroscope?

MarcoPolo:
This is pretty cool! This seems pretty well separated from the rest of the metrics/dashboards. Is there a reason to want it merged into this repo? I think if we had many users all wanting to reference this exact dashboard that would be a good case. If it's just a nice demo or proof-of-concept, maybe we could put some information on docs.libp2p.io?

cpeliciari:
When I came across the dashboard, I thought it was something to help developers visualize what is happening with libp2p. I had some consumption issues with the library when I was running some tests.

p-shahi:
@aschmahmann @sukunrt and I discussed this in the maintainers call last week, we are on board with merging this since it's a good contribution and can be useful. The consensus was that it doesn't seem like too much of a maintenance overhead and merging it here would be more useful that in another repo. In case, there will be too much overhead required to upkeep pyroscope, we'll remove it later. Thanks @cpeliciari 

@sukunrt let's review and merge?

dashboards/README.md
```diff
@@ -24,12 +24,44 @@ For local development and debugging, it can be useful to spin up a local Prometh
...
+http.Handle("/debug/metrics/prometheus", promhttp.Handler())
+log.Fatal(http.ListenAndServe(":5001", nil))
 }()
+
+pyroscope.Start(pyroscope.Config{
+    ApplicationName: "simple.golang.app",
+    
+    // replace this with the address of pyroscope server
+    ServerAddress:   "http://127.0.0.1:4040",
+
```
sukunrt, line 42: Can we replace this with the address of the server that we setup with docker?

dashboards/README.md
```diff
@@ -24,12 +24,44 @@ For local development and debugging, it can be useful to spin up a local Prometh
...
 
 go func() {
-    http.Handle("/debug/metrics/prometheus", promhttp.Handler())
-    log.Fatal(http.ListenAndServe(":5001", nil))
+http.Handle("/debug/metrics/prometheus", promhttp.Handler())
+log.Fatal(http.ListenAndServe(":5001", nil))
 }()
+
+pyroscope.Start(pyroscope.Config{
+    ApplicationName: "simple.golang.app",
```
sukunrt, line 38: This needs to go in the goroutine above so that users can copy paste this block. 

This will still not work because the go.mod doesn't have pyroscope, so we should add a line regarding it somewhere.

sukunrt (changes requested):
Apologies for getting to this so late. I was recovering from an accident. 

I have changed my mind regarding including this in the repo(cc @p-shahi @aschmahmann). Integrating pyroscope has nothing to do with go-libp2p. The steps mentioned here will work the same with any golang application. For prometheus we have metrics built in and we have this elaborate setup so users can explore the dashboards quickly. 

I have reviewed this in case we do decide to include this in the repo.

cpeliciari:
> Apologies for getting to this so late. I was recovering from an accident.
> 
> I have changed my mind regarding including this in the repo(cc @p-shahi @aschmahmann). Integrating pyroscope has nothing to do with go-libp2p. The steps mentioned here will work the same with any golang application. For prometheus we have metrics built in and we have this elaborate setup so users can explore the dashboards quickly.
> 
> I have reviewed this in case we do decide to include this in the repo.

No problem, I can see how dedicated you all are to the project. That was just a way for the developers to check the usage of the library. And I already thank you for working so hard on libp2p.

dashboards/README.md
cpeliciari: Yes, we can use docker DNS.

p-shahi:
Thanks for the update @sukunrt - let's close this PR then @cpeliciari ? Thanks for the contribution nonetheless 🙏

PR #3062, opened 2024-11-25
feat: allow passing options to memoryAddrBook

2color:
## Background

I'm adding an instance of the `MemoryAddrBook` to someguy to manage peer record caching/ttls locally in a similar fashion to https://github.com/libp2p/go-libp2p-pubsub/pull/555/. 

That requires passing passing options and using it without a PeerStore. For that being able to pass options is handy.

This PR also fixes the defaults in the factory.

### DIFF
--- a/p2p/host/peerstore/pstoremem/addr_book.go
+++ b/p2p/host/peerstore/pstoremem/addr_book.go
@@ -183,8 +183,8 @@ const (
 	defaultMaxUnconnectedAddrs  = 1_000_000
 )
 
-// MemoryAddrBook manages addresses.
-type MemoryAddrBook struct {
+// memoryAddrBook manages addresses.
+type memoryAddrBook struct {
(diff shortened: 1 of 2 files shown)


sukunrt:
Can't you just use it like https://github.com/libp2p/go-libp2p-pubsub/pull/555 does? By assuming the return value is an `AddrBook`. Otherwise, I'd rather change everything here to use the `AddrBook` interface and not export `MemoryAddrBook`.

2color:
> Can't you just use it like https://github.com/libp2p/go-libp2p-pubsub/pull/555 does? By assuming the return value is an AddrBook. Otherwise, I'd rather change everything here to use the AddrBook interface and not export MemoryAddrBook.

I did. If you look at the latest version of this PR, it solely adds the ability pass in options without using the PeerBook.

sukunrt (approved):

PR #3068, opened 2024-11-26
fix: avoid modifying slice returned by AddrsFactory

ivan4th:
In some cases, `AddrsFactory` may return a reusable slice. This slice should not be modified.
In `BasicHost.Addrs()`, the slice was already being cloned before being passed to `ma.Unique()` (updated this place to use `slices.Clone()` function), but in other places, the slice was being modified without being cloned, which in some cases could lead to crashes / unpredictable behavior.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -582,7 +582,7 @@ func (cfg *Config) addAutoNAT(h *bhost.BasicHost) error {
 	if cfg.AddrsFactory != nil {
 		addrFunc = func() []ma.Multiaddr {
 			return slices.DeleteFunc(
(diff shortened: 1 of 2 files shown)


sukunrt (approved):
Nice! Thanks for `slices.Clone` :heart:

ivan4th:
Could we merge this PR? (I have no permissions to do so...)

sukunrt:
Yeah. I was waiting for fixing the interop test. But this is fine. We can do that later.

ivan4th:
Thanks!

PR #3078, opened 2024-12-03
swarm: delay /webrtc-direct dials by 1 second

sukunrt:
Previously these addresses weren't delayed at all. When I initially did the ranker implementation, I was too conservative regarding what to delay. So, non QUIC, WebTransport, TCP, WS addresses were ignored in ranking. We only ever need to dial /webrtc-direct when there's no other address available for the peer, in which case we will dial the addresses immediately.

This would have helped with https://github.com/libp2p/js-libp2p/issues/2805 as there would have been fewer peers dialing webrtc and then cancelling because they connected on a better transport.

This also introduces an additional 1 second delay for any fancy non IP transports

### DIFF
--- a/p2p/test/basichost/basic_host_test.go
+++ b/p2p/test/basichost/basic_host_test.go
@@ -4,16 +4,19 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	"github.com/libp2p/go-libp2p"
+	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/p2p/net/swarm"
 	"github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/client"
 	"github.com/libp2p/go-libp2p/p2p/protocol/circuitv2/relay"
+	libp2pquic "github.com/libp2p/go-libp2p/p2p/transport/quic"
 	libp2pwebrtc "github.com/libp2p/go-libp2p/p2p/transport/webrtc"
 	libp2pwebtransport "github.com/libp2p/go-libp2p/p2p/transport/webtransport"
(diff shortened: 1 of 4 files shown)


p2p/test/basichost/basic_host_test.go
```diff
@@ -199,3 +203,72 @@ func TestAddrFactorCertHashAppend(t *testing.T) {
 		return hasWebRTC && hasWebTransport
 	}, 5*time.Second, 100*time.Millisecond)
 }
+
+func TestWebRTCDirectDialDelay(t *testing.T) {
```
MarcoPolo, line 207: I don't understand this test. Is it trying to make sure that if many peers are connected to a server (via QUIC?), that another peer can also connect to that server via WebRTC?

MarcoPolo (approved):
The changes make sense, just a question about the test

p2p/test/basichost/basic_host_test.go
sukunrt: There are two tests. 
1. TestOnlyWebRTCDirectDialNoDelay: Tests that dials to a peer with only webrtc address aren't delayed. 
2. TestWebRTCWithManyQUICConnections: Tests that in presence of quic connections webrtc dials are actually delayed. In case there's no happy eyeballs, the 200 initial dials would have taken up all the goroutines available for handshakes in the webrtc listener.

MarcoPolo: >  In case there's no happy eyeballs, the 200 initial dials would have taken up all the goroutines available for handshakes in the webrtc listener.

But you wait for the Connect to finish, so wouldn't the handshake be complete at this point?

sukunrt: Yes, the QUIC handshake will complete not the webrtc one. The webrtc handshake will be cancelled. As there's no way to signal this cancellation in webrtc, the server will have to wait for timeout on this handshake. So, at this point if we get a new webrtc handshake attempt, it'll fail because there were no free listener goroutines. This all happens when there's no happy eyeballs. 

With happy eyeballs, the quic handshake should complete without any webrtc dials.

MarcoPolo (approved):

PR #3080, opened 2024-12-04
fix(sampledconn): Correctly handle slow bytes and closed conns

MarcoPolo:


MarcoPolo:
Currently investigating why linux fails here.

MarcoPolo:
I've removed the OS specific things. The wrapper approach works fine and avoids a lot of complexity.

p2p/transport/tcpreuse/internal/sampledconn/sampledconn_test.go
```diff
@@ -63,7 +64,7 @@ func TestSampledConn(t *testing.T) {
 				assert.Equal(t, "hello", string(buf))
 			} else {
 				// Wrap the client connection in SampledConn
-				sample, sampledConn, err := newFallbackSampledConn(clientConn.(ManetTCPConnInterface))
```
sukunrt, line 66: should we rename `fallbackPeekingConn` to `sampledConn`?

sukunrt (approved):

Issue #3073, opened 2024-11-30
Allow users to create a New mDNS Discovery service with context as an argument to make it cancellable


phillipahereza:
Currently, the [NewMdnsService](https://github.com/libp2p/go-libp2p/blob/8423de3a64f17f6bec18bf57b472e5a3615883db/p2p/discovery/mdns/mdns.go#L53C6-L53C20) function creates a new context, meaning it cannot be cancelled from the calling function.

To maintain compatibility, I suggest adding a new function, `NewMdnsServiceWithContext(ctx context.Context, host host.Host, serviceName string, notifee Notifee)`, that accepts a context as an argument.

MOHANKUMAR-IT:
This will be useful , as currently there is no cancel option

sukunrt:
You can use the `Close` method to close the service. Explained here too: https://github.com/libp2p/go-libp2p/pull/3089#issuecomment-2522799377

PR #3089, opened 2024-12-06
added NewMdnsServiceWithContext

MOHANKUMAR-IT:
For issue [https://github.com/libp2p/go-libp2p/issues/3073](url)


marten-seemann:
This is wrong. Contexts should not be used for service cancellation. Use the `Close` method instead.

Issue #1163, opened 2021-08-18
advertise updated addresses via mDNS


marten-seemann:
At the moment we're just advertising a static set of addresses. Instead, we should be listening to "address change" events.

MarcoPolo:
We should use the eventbus to get notified about address change events.

MOHANKUMAR-IT:
Have raised pull request for the fix guys.

PR #3096, opened 2024-12-09
Fix : Announcing static addresses for mDNS 

MOHANKUMAR-IT:
Re-register proxy on local address change event from libp2p host

- Subscribe to local address update events from libp2p host
- On address change, shut down the existing mDNS server and re-register with updated address

FIxes : #1163


MarcoPolo:
Did you manage to test that this works for you? I wonder if we have to start the resolver as well.

MarcoPolo:
I think we do need to restart the resolver as well.

Could you please add a test case as well?

MOHANKUMAR-IT:
I'll also add a test case to ensure the behavior is covered and working as expected. Thank you for pointing it out!

MarcoPolo:

PR #3099, opened 2024-12-12
chore: use errors.New to replace fmt.Errorf with no parameters

ChengenH:


p-shahi:
From reading golang docs, it seems like both of these are almost identical under the hood. Is there a good reason to switch? cc @sukunrt or @MarcoPolo

marten-seemann:
I don’t think performance in the error code path matters at all. This seems like an unsubstantiated performance claim without a benchmark.

PR #3103, opened 2024-12-17
feat: add AutoTLS example

### LINKED ISSUE #1360: Universal Browser Connectivity using HTTPS/Secure Websockets
Every (public) node should have a Subdomain and a trusted TLS certificate, so that browser nodes can connect using `/wss` (Secure WebSockets transport).

We will support manual certificate configuration, and ACME automation for user’s domain names (see #1331 for details), but these setups require some manual step/input from the user. 

Here, we discuss a protocol for providing TLS cert for `/wss`  for all remaining nodes in a way that happens automatically, without user’s involvement.

## Obtaining a Certificate using the DNS01 ACME Challenge

For the purpose of this discussion, let’s assume that we have one PL-controlled "gatekeeper" node that has (programmatic) access to the DNS configuration of `libp2p.direct` (or some other domain).

1. The gatekeeper node is run by PL and has a fixed DNS name, e.g.`dns.libp2p.direct`.
2. When a node wants to get a certificate, it requests a subdomain name from the gatekeeper.
3. The gatekeeper then generates a new subdomain: `<subdomain>.libp2p.direct`.
    * TBD: this could be a random string, or a certain encoding of the peer ID.
4. The node then runs its ACME client and requests a DNS-01 challenge from LE. It forwards the ACME challenge to the gatekeeper via a yet-to-defined libp2p protocol.
5. The gatekeeper verifies the ACME challenge (i.e. checks the subdomain, that it is well-formed, etc.) and publishes the TXT record. It un-publishes the TXT record when the ACME challenge is done, or after the expiration of a short timer (10 minutes?).
6. LE now issues a certificate directly to the nodes’ ACME client.
7. The node can now advertise `<ip-multiaddr>.<subdomain>.libp2p.direct`. Our DNS resolver makes sure that this resolves to `<ip>`.

## Steps

* [x] implement a DNS server that can
    * resolve <ip-multiaddr>.<subdomain>.libp2p.direct` to `<ip>`
    * programmatically set the TXT records for the ACME challenge
    * Done in: https://github.com/marten-seemann/wildcard-dns
* [x] coordinate with LetsEncrypt (or ZeroSSL) that we can get an unlimited number of wildcard certificates for subdomains of `libp2p.direct` 
* [x] transfer ownership of `libp2p.direct` to the infrastructure team (earliest possible date: Apr 24)
* [x] specify and implement a libp2p protocol to request a subdomain from the gatekeeper and to set TXT records
* [ ] integrate into go-ws-transport, so we can use and advertise `/wss` addresses
  - [ ] https://github.com/libp2p/go-libp2p/pull/3103

2color:
This adds an example showing how to use the p2p-forge client library with the AutoTLS backend to issue a wild card certificate. 

Credits to @guillaumemichel for writing the initial code in this example.

Closes #1360

## Fix before merging

- [x] The libp2p.direct address is not announced by the peer via identify, even though the peer has successfully minted a TLS certificate. Fixed by https://github.com/libp2p/go-libp2p/pull/3103/commits/c16ebd9e186717466f9e00177ec58c3ed5cb5787
- [x] Store peer identity on disk to avoid generating a new certificate on each run


examples/autotls/README.md
```diff
@@ -0,0 +1,14 @@
+# libp2p host with AutoTLS
+
+This example builds on the [libp2p host example](../libp2p-host) example and demonstrates how to use [AutoTLS](https://blog.ipfs.tech/2024-shipyard-improving-ipfs-on-the-web/#autotls-with-libp2p-direct) to automatically generate a wildcard Let's Encrypt TLS certificate unique to the libp2p host (`*.<PeerID>.libp2p.direct`), enabling browsers to directly connect to the libp2p host.
+
+For this example to work, you need to have a public IP address and be publicly reachable.
```
lidel, line 5:
```suggestion
For this example to work, you need to have a public IP address and be publicly reachable. AutoTLS is guarded by connectivity check, and will not ask for a certificate unless your libp2p node emits `event.EvtLocalReachabilityChanged` with `network.ReachabilityPublic`.
```

2color:
@MarcoPolo Is it a problem this PR upgrades the go to 1.23 in go.mod?

I believe this happened due to [p2p-forge](https://github.com/ipshipyard/p2p-forge/blob/main/go.mod#L3)

examples/autotls/main.go
```diff
@@ -0,0 +1,117 @@
...
+		libp2p.NATPortMap(),   // Attempt to open ports using UPnP for NATed hosts.
+
+		libp2p.ListenAddrStrings(
+			"/ip4/0.0.0.0/tcp/5500",         // regular tcp connections
+			"/ip4/0.0.0.0/udp/5500/quic-v1", // a UDP endpoint for the QUIC transport
+
+			// AutoTLS will automatically generate a certificate for this host
+			// and use the forge domain (`libp2p.direct`) as the SNI hostname.
+			fmt.Sprintf("/ip4/0.0.0.0/tcp/5500/tls/sni/*.%s/ws", p2pforge.DefaultForgeDomain),
+			fmt.Sprintf("/ip6/::/tcp/5500/tls/sni/*.%s/ws", p2pforge.DefaultForgeDomain),
```
aschmahmann, line 63: Why ipv4 + ipv6 here, but just ipv4 for the tcp listener?

2color: Good point. Fixed. I believe we need ipv6 TCP anyways so that AutoNAT can confirm out addresses. (We can't do port reuse for outgoing connections with WS, making it hard for us to detect our ipv6 if we don't have ipv6 TCP)

guillaumemichel:
https://github.com/ipshipyard/p2p-forge/pull/29/ should solve the build issue

examples/autotls/main.go
```diff
@@ -0,0 +1,148 @@
...
+	// p2pforge is the AutoTLS client library.
+	// The cert manager handles the creation and management of certificate
+	certManager, err := p2pforge.NewP2PForgeCertMgr(
+		// Configure CA ACME endpoint
+		// NOTE:
+		// This example uses Let's Encrypt staging CA (p2pforge.DefaultCATestEndpoint)
+		// which will not work correctly in browser, but is useful for initial testing.
+		// Production should use Let's Encrypt production CA (p2pforge.DefaultCAEndpoint).
+		p2pforge.WithCAEndpoint(p2pforge.DefaultCATestEndpoint), // test CA endpoint
+		// TODO: p2pforge.WithCAEndpoint(p2pforge.DefaultCAEndpoint),  // production CA endpoint
```
lidel, line 52: @2color @guillaumemichel  would appreciate sanity check on this.

My "strong conviction held weakly" is that we should not use production endpoints in examples like this.

I switched this code to use staging endpoint Let's Encrypt provided for the purpose of testing and examples – https://letsencrypt.org/docs/staging-environment/.


But since `libp2p.direct` is on https://publicsuffix.org/ we don't really care about rate-limiting, mostly just dont want to pollute https://crt.sh/?q=libp2p.direct with noise produced by this example.


At the same time, perhaps its a lesser evil to use production endpoint in example – if people copy&paste blindly, they will get broken certs and unless they try to connect to WSS in browser, they will not know its broken. 


So the question is: should we switch this example  `p2pforge.DefaultCATestEndpoint` → `p2pforge.DefaultCAEndpoint` ?

examples/autotls/main.go
```diff
@@ -0,0 +1,148 @@
...
+
+		// Configure where to store certificate
+		p2pforge.WithCertificateStorage(&certmagic.FileStorage{Path: "p2p-forge-certs"}),
```
lidel, line 55: :information_source: added this to example, as it quite common for people to wonder  how to control this

examples/autotls/main.go
```diff
@@ -0,0 +1,148 @@
...
+
+		// Configure logger to use
+		p2pforge.WithLogger(rawLogger.Sugar().Named("autotls")),
```
lidel, line 58: :information_source:  this fixes p2p-forge/client logs, which were not displayed. the example is now way more chatty.

examples/autotls/main.go
2color: To be honest, I wasted a fair while trying to figure out what was wrong after I had deliberately switched to the staging env. I think we just need to log conspicuously when using staging.

examples/autotls/main.go
2color: That's what I was looking for!

examples/go.mod
```diff
@@ -1,27 +1,30 @@
 module github.com/libp2p/go-libp2p/examples
 
-go 1.22.0
+go 1.23
 
-toolchain go1.22.3
+toolchain go1.23.3
```
lidel, line 5: :thought_balloon:  I suspect this produces `staticcheck ./...` error:
- https://github.com/libp2p/go-libp2p/actions/runs/12676518662/job/35329879396?pr=3103#step:12:36

In the past, the fix was to:
1. check if `go install honnef.co/go/tools/cmd/staticcheck@latest` → `staticcheck ./...` passes locally
2. ask IPDX to update Unified GO CI to use new staticcheck

lidel: Confirmed, staticcheck@a093f7c2d3d45d5104fb3414ae939a98be37be02 used on CI does not support 1.23 correctly, and 1.22 is also bit out of date: https://github.com/dominikh/go-tools/releases

Will ping IPDX.

examples/autotls/main.go
lidel: This is a good idea. 

I will add the same "ERROR" log to p2p-forge/client when staging/testing endpoint is used to make sure folks remember to change it and it never sneaks into production. 

This way we dont pollute this example, and everyone using client will benefit.

lidel: See https://github.com/ipshipyard/p2p-forge/pull/33

examples/go.mod
galargh: I looked into this. In the `go-check` workflow, we're choosing the version of `staticcheck` to use based on the Go version that we're using. We already have the latest `staticcheck` version associated with the latest Go version. This PR changes the Go version in the `go.mod` file to `1.23`, but it didn't change it in the `go-check` workflow file. I updated it here - https://github.com/libp2p/go-libp2p/pull/3103/commits/5da91acf93f6685bb48251b06da5bd07806cb5dc

The reason why the Go version is explicitly set in the workflow file is because the maintainers of this repository preferred the version string to be present in the workflow run name, and this was the way to achieve that. You'll notice that in most other repos using Unified GitHub Workflows, the names of the workflow runs will contain things like `this`/`next`. In those, all we need to do to update the version used by the workflow is to update the `go.mod`.

We should also have a look at the `go-test` workflow. Should we remove testing with Go `1.22` in this PR from here - https://github.com/libp2p/go-libp2p/blob/add-autotls-example/.github/workflows/go-test.yml#L20 - as well?

lidel: Thank you for figuring this out. go-check fix should be good enough.
I don't think we should remove 1.22 support, the main `/go.mod` still uses that. We should  not change this without approval of @libp2p/go-libp2p-maintainers 

@2color I think the source of problem here in that this PR is bumping go version to 1.23 in `examples/go.mod` but the main `go.mod` remains on 1.22. 

Perhaps we could do `go mod edit -go=1.22` to pin `/examples/go.mod` to 1.22 to match `/go.mod`?

examples/autotls/main.go
```diff
@@ -0,0 +1,159 @@
...
+		),
+
+		// Configure the TCP transport
+		libp2p.Transport(tcp.NewTCPTransport),
+
+		// Share the same TCP listener between the TCP and WS transports
+		libp2p.ShareTCPListener(),
+
+		// Configure the QUIC transport
+		libp2p.Transport(quic.NewTransport),
```
guillaumemichel, line 110: nit: IIUC `quic` isn't required for the example to work. We can remove the transport if the goal is to show a minimal working example.

examples/go.mod
2color: As I suspected, if `go.mod` for the examples is updated to use `go 1.22`, it doesn't build.

```
go: updates to go.mod needed; to update it:
	go mod tidy
```

If I try to build with `go 1.221` I get:

```
go: github.com/ipshipyard/p2p-forge@v0.2.1 requires go >= 1.23 (running go 1.22.8; GOTOOLCHAIN=local)
```

MarcoPolo:
> @MarcoPolo Is it a problem this PR upgrades the go to 1.23 in go.mod?
> 
> I believe this happened due to dependency on [p2p-forge](https://github.com/ipshipyard/p2p-forge/blob/main/go.mod#L3)

Yes. We can't do that until the Go team releases 1.24

examples/go.mod
lidel: Thank you :pray: I believe that library should follow go-libp2p, filled:
- https://github.com/ipshipyard/p2p-forge/issues/37

lidel: Pushed https://github.com/libp2p/go-libp2p/pull/3103/commits/285d38f3902f2b3e576e32ec9a96fc7fac0c5919 with p2p-forge [v0.2.2 ](https://github.com/ipshipyard/p2p-forge/releases/tag/v0.2.2), this allows us to use the same version in `./examples` as main go-libp2p.

lidel:
Pushed  update to [p2p-forge/client v0.2.2](https://github.com/ipshipyard/p2p-forge/releases/tag/v0.2.2).

It allows this PR to use the same go1.22 version in `./examples/go.mod` as main go-libp2p in `/go.mod`.
Going forward, p2p-forge releases will aim to use same go version as go-libp2p to ensure users can try it without being forced to update go.

examples/autotls/main.go
2color: https://github.com/libp2p/go-libp2p/pull/3103/commits/c301ad7a25b9696494887af46aa868ab21ca4fba

examples/autotls/main.go
```diff
@@ -0,0 +1,153 @@
...
+		panic(err)
+	}
+
+	opts := []libp2p.Option{
+		libp2p.Identity(privKey), // Use the loaded identity key
+		libp2p.DisableRelay(),    // Disable relay, since we need a public IP address
+		libp2p.NATPortMap(),      // Attempt to open ports using UPnP for NATed hosts.
+
+		libp2p.ListenAddrStrings(
+			// Configure default catch-all listeners for TCP and UDP
```
2color, line 89:
```suggestion
```

lidel (approved):
Updated the example to use latest p2p-forge/client, lgtm.
Since this is still opt-in, I think its ok to merge and so we can unblock https://github.com/libp2p/blog/pull/137.

If we need to refine, ok to do it in future PR.

examples/autotls/main.go
```diff
@@ -0,0 +1,150 @@
...
+		dht.BootstrapPeers(dht.GetDefaultBootstrapPeerAddrInfos()...),
+	}
+	dht, err := dht.New(ctx, h, dhtOpts...)
+	if err != nil {
+		panic(err)
+	}
+
+	go dht.Bootstrap(ctx)
+
+	time.Sleep(5 * time.Second)
```
2color, line 133:
```suggestion
	// Wait to connect to some other peers so that out public address is confirmed with AutoNAT
	time.Sleep(5 * time.Second)
```

examples/autotls/identity.go
```diff
@@ -0,0 +1,49 @@
+package main
+
+// Borrowed from https://github.com/libp2p/go-libp2p-relay-daemon/blob/master/identity.go
```
2color, line 3:
```suggestion
```

examples/autotls/main.go
```diff
@@ -0,0 +1,151 @@
...
+	// p2pforge is the AutoTLS client library.
+	// The cert manager handles the creation and management of certificate
+	certManager, err := p2pforge.NewP2PForgeCertMgr(
+		// Configure CA ACME endpoint
+		// NOTE:
+		// This example uses Let's Encrypt staging CA (p2pforge.DefaultCATestEndpoint)
+		// which will not work correctly in browser, but is useful for initial testing.
+		// Production should use Let's Encrypt production CA (p2pforge.DefaultCAEndpoint).
+		p2pforge.WithCAEndpoint(p2pforge.DefaultCATestEndpoint), // test CA endpoint
+		// TODO: p2pforge.WithCAEndpoint(p2pforge.DefaultCAEndpoint),  // production CA endpoint
```
aschmahmann, line 50:
```suggestion
```

Should we remove these lines now (or at least the TODO) line given the comments above?

aschmahmann: Not disagreeing with using staging, just thinking these lines were meant to be removed given the comments added here https://github.com/libp2p/go-libp2p/pull/3103/files#diff-b413b2c049e0ed402c4b3fe4f1501acded0f28dea3a64d709e88809be9b3a867R45-R48

2color: Ah I think the idea was for this to be a DX hack for copy pasta out of this example

aschmahmann: ok, fair enough.

PR #3110, opened 2024-12-18
[exp] metrics over libp2p streams

MarcoPolo:
Allows a node to expose its metrics to another libp2p node. Only allow listed peer ids can query the metrics.

This enables a user to share their metrics with us, and for us to use our own grafana+prometheus stack to ingest those metrics.

A couple more things are left to do here. In draft until those are done. I may split the ergonomic auth to a separate PR.


MarcoPolo:

PR #3111, opened 2024-12-18
fix(httpauth): Correctly handle concurrent requests on server

MarcoPolo:
I originally wrote this code assuming the hmac object was copied by value, but it's an interface. Concurrent requests were using the same pointer.

### DIFF
--- a/p2p/http/auth/server.go
+++ b/p2p/http/auth/server.go
@@ -30,11 +30,12 @@ func newHmacPool(key []byte) *hmacPool {
 }
 
 func (p *hmacPool) Get() hash.Hash {
-	return p.p.Get().(hash.Hash)
+	h := p.p.Get().(hash.Hash)
+	h.Reset()
+	return h
 }
 
 func (p *hmacPool) Put(h hash.Hash) {
-	h.Reset()
 	p.p.Put(h)
 }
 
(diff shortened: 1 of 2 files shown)


p2p/http/auth/server.go
```diff
@@ -15,6 +15,29 @@ import (
...
+		},
+	}
+}
+
+func (p *hmacPool) Get() hash.Hash {
+	return p.p.Get().(hash.Hash)
+}
+
+func (p *hmacPool) Put(h hash.Hash) {
+	h.Reset()
```
aschmahmann, line 37: It seems fine either way, but it looks like the example sync.Pool usage has the reset on the Get rather than the Put https://github.com/golang/go/blob/master/src/sync/example_pool_test.go#L29-L40. Similarly with this string slice pool in go-libp2p https://github.com/libp2p/go-libp2p/blob/b07e3aa6c56a4d3e8254d6689e9b32db4f0e6f61/p2p/metricshelper/pool.go#L15-L19

aschmahmann (approved):
Left a brief comment, but LGTM

p2p/http/auth/server.go
MarcoPolo: Resetting in Get is better because you avoid the cost of `.Reset()` if the object is going to get GC'd anyways. In other words, you only reset things you actually use.

Thanks for highlighting this.

PR #3112, opened 2024-12-18
ci: Install specific protoc version when generating protobufs

MarcoPolo:
This fixes the issue where CI is using a different version of `protoc`/`protoc-gen-go` from our local environment. It downloads the specified version of the protobuf binary for the current system and verifies its hash. Current supported systems are macOS-aarch64 and linux-{x86_64,aarch64}. This is only used when we are regenerating the protobufs, so likely not something we will often run into.

Now the generated pbs should be reproducible locally and on CI.


MarcoPolo:
merging without another review since this is causing every other PR to fail the check.

PR #3117, opened 2024-12-20
ci: get back on the main release track of release checker

galargh:
@MarcoPolo Are we ready to get back to the main release track or is there some feature missing that you'd like us to add there?

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -16,4 +16,4 @@ concurrency:
 
 jobs:
   release-check:
-    uses: marcopolo/unified-github-workflows/.github/workflows/release-check.yml@e66cb9667a2e1148efda4591e29c56258eaf385b
+    uses: ipdxco/unified-github-workflows/.github/workflows/release-check.yml@v1.0


p-shahi:
The [reason to use the fork](https://github.com/libp2p/go-libp2p/pull/3018 ) was merged to master https://github.com/ipdxco/unified-github-workflows/pull/96 so I think we're good? @sukunrt

.github/workflows/release-check.yml
```diff
@@ -16,4 +16,4 @@ concurrency:
 
 jobs:
   release-check:
-    uses: marcopolo/unified-github-workflows/.github/workflows/release-check.yml@e66cb9667a2e1148efda4591e29c56258eaf385b
+    uses: ipdxco/unified-github-workflows/.github/workflows/release-check.yml@v1.0
```
p-shahi, line 19:
```suggestion
    uses: ipdxco/unified-github-workflows/.github/workflows/release-check.yml@v1.0.17
```
use latest?

2color: I think it automatically takes the latest patch with just 1.0

sukunrt (approved):

PR #3128, opened 2024-12-30
swarm: remove unnecessary error log

sukunrt:
There's no reason to log this error here, this is perfectly legal. We can make it a debug statement, but even that seems not useful.


MarcoPolo (approved):
Looks fine. But when would you setup no transports and expect anything to work? In other words, would this error log be helpful to users?

sukunrt:
We provide `ErrNoTransport` from `Listen` and `NoGoodAddresses` from `Dial` 

That should inform users. We can move this log to `Dial` and `Listen`. My problem is that this method is an arbitrary place to log this. We should return appropriate error from `Listen` or `Dial`.

Issue #3119, opened 2024-12-20
TCP connection multiplexing (tcpreuse package) doesn't work with pnet


aschmahmann:
The tcpreuse package was introduced so that the same TCP port could be reused across the TCP, WS, WSS, etc. transports like we have with UDP and QUIC + WebRTC (-Direct). Doing this requires looking at the first packet for the connection to figure out which transport to associate the connection with.

Unfortunately, this does not integrate nicely with the private networks / pnet feature since in the case of the TCP transport + pnet the first few bytes are encrypted which means we can't use our current check which checks for the start of the multistream protocol. Similarly, we also would no longer be able to rely on the WS, WSS, etc. checks because the encryption could be "unlucky" and collide with the first bytes from HTTP or TLS.

To reproduce this start 2 libp2p nodes with pnet and have one use the tcpreuse package and listen on TCP. Try to open a stream and send a message and you'll see a failure. The bug was discovered integrating the shared tcp listener into kubo https://github.com/ipfs/kubo/pull/10565#discussion_r1894207331.

Some options here:
1. Mark the tcpreuse feature as unsupported for pnet until enough people want it to warrant the effort, and add errors in the host constructor like we do for pnet + quic/webrtc.
2. Forbid pnet nonce's from coming from the list of supported starting bytes to switch on (e.g. the first 3 bytes that could come from multistream, HTTP, or TLS). If the inbound packet is "unknown" assume it's the TCP transport If in the future new ones are added pnet will need to be updated, and rely on the assumption that pnet networks are able to update in lockstep (not totally crazy given the keys could leak and need to be rotated, but some people might just be using pnet as a hack to create namespaced networks and not be able to rotate as easily).
3. Expand the number of bytes we sniff to figure out the transport to be enough that we have very high confidence that something that looks like HTTP or TLS is actually one of those and not just an unlucky set of random nonce bytes that happen to look like HTTP or TLS. Not sure if this is even really doable.

At the moment it seems like just marking the combination of tcpreuse + pnet as unsupported seems like way to go. The complexity of the other two options doesn't seem like a good use of time unless there's demand for it.

gammazero:
I agree we should consider the current version of pnet (v0) to be incompatible with tcpreuse. A future pnet (v1) can propose adding its own identifying first bytes. This will also allow for pnet to do version detection in the future.

lidel:
+1, given that 
[...]

PR #3129, opened 2024-12-30
tcpreuse: error on using tcpreuse with pnet

sukunrt:
Sharing the TCP Listen socket with WebSocket doesn't work with pnet.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -449,7 +449,22 @@ func (cfg *Config) newBasicHost(swrm *swarm.Swarm, eventBus event.Bus) (*bhost.B
 // NewNode constructs a new libp2p Host from the Config.
 //
 // This function consumes the config. Do not reuse it (really!).
-func (cfg *Config) NewNode() (host.Host, error) {
+func (cfg *Config) NewNode() (_ host.Host, createErr error) {
+
+	defer func() {
+		if createErr != nil {
+			if cfg.ResourceManager != nil {
+				cfg.ResourceManager.Close()
+			}
+			if cfg.ConnManager != nil {
+				cfg.ConnManager.Close()
+			}
+			if cfg.Peerstore != nil {
+				cfg.Peerstore.Close()
+			}
+		}
+	}()
+
 	if cfg.EnableAutoRelay && !cfg.Relay {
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -466,6 +481,10 @@ func (cfg *Config) NewNode() (host.Host, error) {
 		rcmgr.MustRegisterWith(cfg.PrometheusRegisterer)
 	}
 
+	if len(cfg.PSK) > 0 && cfg.ShareTCPListener {
+		return nil, errors.New("cannot use shared TCP and WebSocket listener with PSK")
+	}
```
aschmahmann, line 486: LGTM, although technically the shared listener feature isn't coupled to having TCP + WS since you can currently enable it with just TCP (albeit at no real benefit) and down the road it might be configurable with HTTP as well.

config/config.go
```diff
@@ -449,7 +449,22 @@ func (cfg *Config) newBasicHost(swrm *swarm.Swarm, eventBus event.Bus) (*bhost.B
 // NewNode constructs a new libp2p Host from the Config.
 //
 // This function consumes the config. Do not reuse it (really!).
-func (cfg *Config) NewNode() (host.Host, error) {
+func (cfg *Config) NewNode() (_ host.Host, createErr error) {
+
+	defer func() {
+		if createErr != nil {
```
aschmahmann, line 455: Does this cover everything not already stopped by fx and how does this play with the closing that happens when autonat fails to start?

aschmahmann (approved):
Left some questions/comments, but generally LGTM

config/config.go
MarcoPolo: This instantiation is so hacky :(. I believe if autonat fails this will end up closing things twice.

Maybe slightly better would be to move the config validation logic to a separate fn. Then cleanup on if that fn returns an error.

config/config.go
MarcoPolo: Changing this error test to just say "... shared TCP listener ...". This is in line with option one in https://github.com/libp2p/go-libp2p/issues/3119.

MarcoPolo (approved):

PR #3130, opened 2024-12-30
swarm_test: support more transports for GenSwarm

sukunrt:
This adds support for `/webtransport` and `/webrtc-direct` to GenSwarm.

Ideally, we should rewrite this to have the same semantics, opt into transports not opt out, as `libp2p.New`. But I need webtransport and webrtc support to write address inference tests for https://github.com/libp2p/go-libp2p/pull/3075

Depending on how disruptive this is to users, we can decide on whether to merge or drop this.


p2p/net/swarm/testing/testing.go
```diff
@@ -191,6 +209,75 @@ func GenSwarm(t testing.TB, opts ...Option) *swarm.Swarm {
 			if err := s.Listen(ma.StringCast("/ip4/127.0.0.1/udp/0/quic-v1")); err != nil {
 				t.Fatal(err)
 			}
+			for _, a := range s.ListenAddresses() {
+				if _, err := a.ValueForProtocol(ma.P_QUIC_V1); err == nil {
+					quicListenAddr = a
```
MarcoPolo, line 214: It's not obvious to me why you want to share this address with all the other transports. Can you explain? Also what happens if `ListenAddresses` contains a different address for WebTransport?

MarcoPolo (commented):
one question, but otherwise looks good.

p2p/net/swarm/testing/testing.go
sukunrt: You're right. We should replicate how the basic host works which just listens on port 0.

sukunrt: The intention was to be maximially similar to how people use this in production. That can be handled separately.

guillaumemichel (approved):
> Depending on how disruptive this is to users, we can decide on whether to merge or drop this.

I don't see how it could break existing users to support more transports in tests. Worst case failing tests may flag bugs in `webrtc` and `webtransport`.

Issue #286, opened 2018-02-27
(socks) proxy support


Stebalien:
In some environments, users need to be able to proxy all traffic through, e.g., a corporate socks proxy. This sucks but is sometimes inevitable.

Solution:

* Tunnel outbound connections through the socks proxy.
* If the proxy supports bind use it.
* Otherwise, connect to and advertise a relay address.

Problems:

* Proxy aware transports are going to be a royal pain.

Stebalien:
So, @rklaehn pointed out a simple solution for dialing: use the websocket transport. It should work just fine over *any* HTTP(s) proxy. All we need to do is expose this somehow.

Unfortunately, *listening* could be tricky.

rklaehn:
So it *should* work, but in my experiments I could not actually get it to work. Just to clarify: websocket is a complete transport that will work just as well as the native tcp or udp transports? So in theory you could have a swarm just communicating over ws? Or is it in some way limited?

Stebalien:
Yes, it's a full transport. I assumed it wouldn't work because we don't explicitly tell it to use the proxy from the environment. However, it looks like it does this *by default*.

So, this is a bit odd. It should just work. I'm wondering if it's a port issue. Does your proxy not support forwarding to random ports?

cretz:
I too need a transport proxy interface. In many cases in Go, proxies are implemented by accepting a `net.Dialer` and/or a `net.Listener`. There needs to be some way to downgrade a transport to these items or the logic that does things with the connections needs to be decoupled from the logic wrapping in manet dialier/listener and transports. My use case is for web sockets over Tor (which happens to be a socks proxy for dialing, but there is also a part I want to proxy for listening). However, there is enough unexposed logic in the websocket that I have to take the entire websocket transport dial/listen or none of it. I would prefer just the parts that operate on the net connections, or even better, let it take an externally provided tcp dialer/listener.

ConorTighe1995:
Did anyone find a solution that worked for this problem?

{"error":"failed to bootstrap. context deadline exceeded","event":"bootstrapError","peerID":"QmXL3Pe6USjNf5Vks2XBi7ryngCe8NguazLyCjeeeCVVe3","system":"core","time":"2018-07-17T09:32:28.447604435Z"}

Stebalien:
So, I've tested this and dialing does actually work. You just need to correctly configure your environment to use a proxy: https://wiki.archlinux.org/index.php/Proxy_settings#Environment_variables

[...]

PR #3136, opened 2025-01-14
feat(transport/websocket): enable pulling http proxy from environment variables

aschmahmann:
related to #286 it seems like there was an undocumented behavioral regression in https://github.com/libp2p/go-ws-transport/pull/116/commits/0ea4a8af2af12617080915a32326158bf8d95684 whereby we no longer respected various standard environment variables related to HTTP(S) proxies for the websocket transport. This puts the behavior back.

@parkan could you try this out in singularity and report back if it's good enough for your use case?

---

A few notes/areas for comment:
1. Any recommendations on testing to prevent regressions here? Importing a socks5 server library just for a test like this seems like overkill, but maybe folks have better ideas
2. Any objections to using the environment variables rather than making it an option?
3. Any recommendations on where to document this behavior? It seems like there's some documentation missing here that'd be nice (e.g. the environment variables go-libp2p users may want to set, even if they're from downstream libraries and we're not committed to them like for HTTP Proxies, go-log, etc.)

### DIFF
--- a/p2p/transport/websocket/websocket.go
+++ b/p2p/transport/websocket/websocket.go
@@ -191,7 +191,7 @@ func (t *WebsocketTransport) maDial(ctx context.Context, raddr ma.Multiaddr) (ma
 		return nil, err
 	}
 	isWss := wsurl.Scheme == "wss"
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Close in favor of #3137

PR #3137, opened 2025-01-14
feat(transport/websocket): support SOCKS proxy with wss

MarcoPolo:
closes https://github.com/libp2p/go-libp2p/issues/286 (a 7 year old issue??)

This PR allows users to use a SOCKS5 proxy when dialing `/wss` addresses by using the `HTTPS_PROXY` environment variable. 

Here's an example using vole to dial through a socks proxy:

```
HTTPS_PROXY="socks5://127.0.0.1:9081" vole libp2p ping /dns4/am6.bootstrap.libp2p.io/tcp/443/wss/p2p/QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb
```

### DIFF
--- a/p2p/transport/websocket/websocket.go
+++ b/p2p/transport/websocket/websocket.go
@@ -160,6 +160,8 @@ func (t *WebsocketTransport) Resolve(_ context.Context, maddr ma.Multiaddr) ([]m
 	return []ma.Multiaddr{parsed.toMultiaddr()}, nil
 }
 
+// Dial will dial the given multiaddr and expect the given peer. If an
+// HTTPS_PROXY env is set, it will use that for the dial out.
 func (t *WebsocketTransport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (transport.CapableConn, error) {
 	connScope, err := t.rcmgr.OpenConnection(network.DirOutbound, true, raddr)
 	if err != nil {
--- a/p2p/transport/websocket/websocket.go
+++ b/p2p/transport/websocket/websocket.go
@@ -191,7 +191,11 @@ func (t *WebsocketTransport) maDial(ctx context.Context, raddr ma.Multiaddr) (ma
 		return nil, err
 	}
 	isWss := wsurl.Scheme == "wss"
-	dialer := ws.Dialer{HandshakeTimeout: 30 * time.Second}
+	dialer := ws.Dialer{
+		HandshakeTimeout: 30 * time.Second,
+		// Inherit the default proxy behavior
+		Proxy: ws.DefaultDialer.Proxy,
+	}
 	if isWss {
 		sni := ""
 		sni, err = raddr.ValueForProtocol(ma.P_SNI)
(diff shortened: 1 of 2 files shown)


parkan:
lgtm :+1:

MarcoPolo:
I didn't really debug, but SetEnv might be surprising if other tests are run in parallel.

parkan:
> I didn't really debug, but SetEnv might be surprising if other tests are run in parallel.

huh I was fairly certain that SetEnv only affects the current process and children, though I guess the tests may run in the same process as separate goroutines?

p2p/transport/websocket/websocket.go
```diff
@@ -204,16 +210,22 @@ func (t *WebsocketTransport) maDial(ctx context.Context, raddr ma.Multiaddr) (ma
...
-			// Setting the NetDial because we already have the resolved IP address, so we don't want to do another resolution.
 			// We set the `.Host` to the sni field so that the host header gets properly set.
+			wsurl.Host = sni + ":" + wsurl.Port()
+			// Setting the NetDial because we already have the resolved IP address, so we can avoid another resolution.
 			dialer.NetDial = func(network, address string) (net.Conn, error) {
-				tcpAddr, err := net.ResolveTCPAddr(network, ipAddr)
+				var tcpAddr *net.TCPAddr
+				var err error
+				if address == wsurl.Host {
+					tcpAddr, err = net.ResolveTCPAddr(network, ipAddr) // Use our already resolved IP address
```
sukunrt, line 220: maybe rename ipAddr? It has the port too.

sukunrt (approved):
Do you think it's worth it to add a dependency on a socks5 server to test this end to end?

parkan:
I am going to make a test build of singularity against this for a functional test and report tmw.

parkan:
hmm actually can't verify correct behavior w/vole, hold please

parkan:
OK I can't say that this :100: works just yet because we're getting proxy errors, but the fact that proxy use is attempted is a big win:

```
104.202.252.41:24001: i/o timeout\n  |   * [/ip4/104.202.252.41/tcp/24002/ws] proxy: failed to read greeting from SOCKS5 proxy at http-proxy.us.archive.org:8080: read tcp 172.17.0.9:59326->207.241.238.251:8080: i/o timeout\n      └─ Wraps: (6) failed to dial /ip4/104.202.252.41/tcp/24002/ws\n        └─ Wraps: (7) proxy: failed to read greeting from SOCKS5 proxy at http-proxy.us.archive.org:8080: read tcp 172.17.0.9:59326->207.241.238.251:8080: i/o timeout\n      └─ Wraps: (8) failed to dial /ip4/104.202.252.41/tcp/24001\n        └─ Wraps: (9) dial tcp4 104.202.252.41:24001\n          └─ Wraps: (10) i/o timeout\n      └─ Wraps: (11) all dials failed\nError types: (1) *withstack.withStack (2) *withstack.withStack (3) *errutil.withPrefix (4) *fmt.wrapError (5) *swarm.DialError (6) 
```

draft PR for singularity here https://github.com/data-preservation-programs/singularity/pull/450

MarcoPolo:
I can make this work with HTTP_PROXY and `ws`. Can you give me a ws endpoint to check against?

Just curious, what is the reason you are using `/ws` without tls? Why can’t you open a tcp stream? Is the client on a browser?

parkan:
@MarcoPolo client is inside a secure network where all connections to the outside must go through SOCKS and only to allowlisted hosts -- this is actually not entirely uncommon in "enterprise" environments though it is a huge pain

for a `ws` endpoint you can check  /ip4/104.202.252.41/tcp/24002/ws/p2p/12D3KooWQnScgybqBcTv6f3yvgbF8pH7sLnrkrbYVB3FRyZY2NsR, I've confirmed it's healthy just now

MarcoPolo:
Testing this locally with [vole](https://github.com/ipfs-shipyard/vole) using this branch of go-libp2p I was able to connect to your endpoint using `HTTP_PROXY`.

```
HTTP_PROXY="socks5://127.0.0.1:9081" vole libp2p ping /ip4/104.202.252.41/tcp/24002/ws/p2p/12D3KooWQnScgybqBcTv6f3yvgbF8pH7sLnrkrbYVB3FRyZY2NsR             
```

`ws` uses `HTTP_PROXY` and `wss` uses `HTTPS_PROXY`.

parkan:
@MarcoPolo we're only sending deal proposals so the payload size is trivial, using websockets from that perspective is fine

separately I am finding that our proxy may not actually even have a *limited* SOCKS implementation... so :facepalm:

parkan:
ok, upon review it seems that we may be able to get raw tcp through the proxy after all (once underlying blockage is fixed), which will help us avoid all the ws rough edges... is it possible to ship the TCP_PROXY suggestion @MarcoPolo? (the HTTP(S)_PROXY fix will actually matter for us elsewhere so that effort is much appreciated)

MarcoPolo:
I think what makes the most sense is to allow a user to set a custom net.Dialer. Then they can use whatever custom logic they want. This would let you use https://github.com/mwitkow/go-http-dialer without tying the library to that specific case.

parkan:
Sounds like a good idea, it should allow us to pass in a http proxy supporting dialer. If we can ship TCP_PROXY support first I can get moving with an interim solution (like using a danted socks->http proxy chain shim)

PR #3146, opened 2025-01-21
feat(holepunch): add logging when DirectConnect execution fails

wlynxg:


wlynxg:
In `beginDirectConnect`, `ErrHolePunchActive` will be returned for a connection that is already hole punching. This error is not logged. I hope to observe this error when I observe hole drilling behavior

wlynxg:
On the other hand, we can't always rely on internal functions to print all errors once. Since DirectConnect is the outermost interface exposed, it seems to be a better solution to make a general print record of missed errors here (even if it may repeatedly print the error log that has been printed internally).

MarcoPolo (approved):
It's generally a worse pattern to log an error and then return. This was likely left after some debugging. It's better to return the error and let the parent that handles the error decide to log it.

PR #3153, opened 2025-01-23
Update quic-go to v0.49.0

MarcoPolo:
Small breaking change around Errors.Is for StreamError

### DIFF
--- a/p2p/transport/quic/stream.go
+++ b/p2p/transport/quic/stream.go
@@ -19,16 +19,20 @@ type stream struct {
 var _ network.MuxedStream = &stream{}
 
 func (s *stream) Read(b []byte) (n int, err error) {
+	var streamErr *quic.StreamError
+
 	n, err = s.Stream.Read(b)
-	if err != nil && errors.Is(err, &quic.StreamError{}) {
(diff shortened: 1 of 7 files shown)


p2p/transport/quic/stream.go
```diff
@@ -19,16 +19,20 @@ type stream struct {
 var _ network.MuxedStream = &stream{}
 
 func (s *stream) Read(b []byte) (n int, err error) {
+	var streamErr *quic.StreamError
+
 	n, err = s.Stream.Read(b)
-	if err != nil && errors.Is(err, &quic.StreamError{}) {
+	if err != nil && errors.As(err, &streamErr) {
```
aschmahmann, line 25: I see, this is due to a breaking change in quic-go to make errors.Is be an approximation for an equality check instead of for type checking as it was before. Given the situation LGTM.

aschmahmann (approved):
LGTM

p2p/transport/quic/stream.go
marten-seemann: That's correct. We implemented `errors.Is` incorrectly for the longest time, and now we finally fixed this.

PR #3154, opened 2025-01-24
chore: move go-nat to internal package

MarcoPolo:
I believe we are the main users of this library, and we don't have enough bandwidth to maintain this for other use cases besides our own.

This PR brings in the commit history from the go-nat repo, and does some refactoring to clean up the code and surface errors to help debug issues.

After merging this, I'll archive the old repo.

Before merge:
- Squash refactor commits
- merge this as a merge commit


MarcoPolo:
For reviewing, only the last commit is likely useful.

p2p/net/nat/internal/nat/upnp.go
```diff
@@ -0,0 +1,235 @@
...
+func discoverUPNP_GenIGDev(ctx context.Context) (nats []NAT, errs []error) {
+	DeviceList, err := ssdp.Search(ssdp.All, 5, "")
+	if err != nil {
+		errs = append(errs, err)
+		return
+	}
+	var gw ssdp.Service
+	for _, Service := range DeviceList {
+		if strings.Contains(Service.Type, "InternetGatewayDevice") {
+			gw = Service
```
MarcoPolo, line 54: Shouldn't we do this on all services that match this string?

marten-seemann:
> I believe we are the main users of this library, and we don't have enough bandwidth to maintain this for other use cases besides our own.

Please don't do this. It's totally ok to refuse PRs to reduce the maintenance burden (as you've been doing with zeroconf), and you can call this out in the README. But code should only be moved if the code logically belongs somewhere else. This clearly is not the case for go-nat.

MarcoPolo:
Are you using this library?

p2p/net/nat/internal/nat/natpmp.go
```diff
@@ -12,42 +12,48 @@ var (
...
-		defer close(res)
-		client := natpmp.NewClient(ip)
-		_, err := client.GetExternalAddress()
+		client, err := discoverNATPMPWithAddr(ctx, ip)
 		if err != nil {
+			errCh <- err
 			return
 		}
-		res <- client
+		clientCh <- client
```
aschmahmann, line 32: Should this be a buffered channel of size 1? Otherwise if the context gets cancelled this goroutine may hang forever.

p2p/net/nat/internal/nat/upnp.go
```diff
@@ -16,197 +16,115 @@ import (
...
+	DeviceList, err := ssdp.Search(ssdp.All, 5, "")
+	if err != nil {
+		errs = append(errs, err)
+		return
+	}
 
-		DeviceList, err := ssdp.Search(ssdp.All, 5, "")
-		if err != nil {
-			return
+	const maxIGDevs = 3
```
aschmahmann, line 55: Can we add some comments about why this is the case?

p2p/net/nat/internal/nat/nat.go
```diff
@@ -4,15 +4,33 @@ package nat
...
+
 var ErrNoExternalAddress = errors.New("no external address")
 var ErrNoInternalAddress = errors.New("no internal address")
-var ErrNoNATFound = errors.New("no NAT found")
+
+type ErrNoNATFound struct {
+	Errs []error
+}
+
+func (e ErrNoNATFound) Error() string {
```
aschmahmann, line 27: Should we also implement `Unwrap` here?

p2p/net/nat/internal/nat/nat.go
```diff
@@ -35,65 +53,91 @@ type NAT interface {
...
 func DiscoverGateway(ctx context.Context) (NAT, error) {
-	var nats []NAT
-	for nat := range DiscoverNATs(ctx) {
-		nats = append(nats, nat)
-	}
+	nats, errs := discoverNATs(ctx)
+
 	switch len(nats) {
 	case 0:
-		return nil, ErrNoNATFound
```
aschmahmann: We weren't equality checking this error anywhere, were we?

p2p/net/nat/internal/nat/upnp.go
```diff
@@ -16,197 +16,115 @@ import (
...
 
-		RootDevice.Device.VisitServices(func(srv *goupnp.Service) {
-			if ctx.Err() != nil {
-				return
+		RootDevice.Device.VisitServices(serviceVisitor(ctx, RootDevice, &nats, &errs))
+	}
+	return
+}
+
+func serviceVisitor(ctx context.Context, rootDevice *goupnp.RootDevice, outNats *[]NAT, outErrs *[]error) func(srv *goupnp.Service) {
```
aschmahmann, line 82: We need a comment somewhere (e.g. on this function or the top level caller of all the upnp discovery mechanisms) describing why we'll handle whatever service type gets sent to us even if we searched for a totally different one (e.g. searching for IG1 and getting IP2 which comes from IG2). This will help us later if there's a problem.

Note: do we want / need this behavior anymore now that we're doing the sequential lookups to deal with badly implemented routers (e.g. fritzbox)?

aschmahmann (changes requested):
@MarcoPolo I reviewed f8bd11c239c3f2a874d0e5b82244a92db03d2ab8. Is there anything else you wanted me to look at (e.g. the git merging things)?

MarcoPolo:
> @MarcoPolo I reviewed [f8bd11c](https://github.com/libp2p/go-libp2p/commit/f8bd11c239c3f2a874d0e5b82244a92db03d2ab8). Is there anything else you wanted me to look at (e.g. the git merging things)?

That should be it. I'll merge this a merge commit so the history should be preserved.

p2p/net/nat/internal/nat/nat.go
MarcoPolo: no

p2p/net/nat/internal/nat/upnp.go
MarcoPolo: This will now collect an error if it sees a v2 service on a v1 device.

aschmahmann:
Regarding whether we should bring this functionality into the monorepo I'm +1 with @MarcoPolo's assessment.

 - The library is small
 - The library is only used by us (and it's had a whole bunch of years of independence if it was going to get independent users)
 - It's much easier to work with and debug the library within the monorepo.... which is part of what inspired the move to the monorepo to begin with

FWIW if the above were not true (particularly, if the library was going to be used by others independent of go-libp2p and the Go module pruning was insufficient) I'd understand wanting to keep the repo separate. Additionally, this is not an uncommon thing to do. Other projects like Syncthing (https://github.com/syncthing/syncthing/tree/124673f7a89b73e7389dc664ce46441b41b7012b/lib/upnp) and Tailscale (https://github.com/tailscale/tailscale/tree/08dd4994d09cf4b960663321136b4fde3f5ccb18/net/portmapper) seem to take the same approach with bundling their portmapping code into their codebases (albeit they're less library focused than go-libp2p is).

If in the future say a group like Syncthing wanted to collaborate on / share a portmapping module and it made sense to do so in a separate repo that'd seem very reasonable to me.

aschmahmann (approved):
A couple small comments, but otherwise LGTM. Thanks, hopefully this will make managing the NAT code and fixing bugs (like the one already fixed here) easier 🙏

PR #3157, opened 2025-01-24
Release v0.38.3

MarcoPolo:
Includes `quic-go@v0.49.0`

Technically we are exporting a new type `ErrCertHashMismatch`, but the API is still returning the `error` interface so this should be okay.


go.mod
```diff
@@ -53,7 +53,7 @@ require (
 	github.com/pion/webrtc/v3 v3.3.5
 	github.com/prometheus/client_golang v1.20.5
 	github.com/prometheus/client_model v0.6.1
-	github.com/quic-go/quic-go v0.48.2
```
sukunrt, line 56: Isn't this breaking semver?

p2p/transport/webtransport/crypto.go
```diff
@@ -86,6 +86,15 @@ func generateCert(key ic.PrivKey, start, end time.Time) (*x509.Certificate, *ecd
 	return ca, caPrivateKey, nil
 }
 
+type ErrCertHashMismatch struct {
+	Expected []byte
+	Actual   [][]byte
+}
+
```
sukunrt, line 93: Why introduce this is the patch release?

sukunrt (commented):
I think this breaks Versioning
1. Will upgrade quic-go for all clients.
2. Exports new type from te webtransport transport.

Should we just make a new minor release?

PR #3159, opened 2025-01-27
Add support for network interfaces in mDNS configuration

cpeliciari:
This commit addresses an issue encountered on Android devices when using the [zeroconf] package, which results in the following error:

```
[zeroconf] no suitable IPv4 interface: udp4: failed to join any of these interfaces: []
```

The root cause of this issue is that the zeroconf package fails to properly detect network interfaces on Android. To resolve this, I have implementation with anet, which includes a fix specifically targeting Android compatibility.

Additionally, a similar issue exists in the package github.com/multiformats/go-multiaddr, where it fails to handle network interfaces correctly on Android. I worked around this issue with the following commit:

[8491a6c56a51cc4dec517388f4f044d939896984](https://github.com/cpeliciari/go-multiaddr/commit/8491a6c56a51cc4dec517388f4f044d939896984)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,6 +57,7 @@ require (
 	github.com/quic-go/webtransport-go v0.8.1-0.20241018022711-4ac2c9250e66
 	github.com/raulk/go-watchdog v1.3.0
 	github.com/stretchr/testify v1.10.0
+	github.com/wlynxg/anet v0.0.5
 	go.uber.org/fx v1.23.0
 	go.uber.org/goleak v1.3.0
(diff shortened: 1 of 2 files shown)


MarcoPolo:
For the same reasons that https://github.com/multiformats/go-multiaddr/pull/255 is open, we don't want to use the anet package here.

A preferred change would be to add an option to provide a custom `.Interfaces()` method. User's can then use anet to implement that method.

MarcoPolo:
A similar solution would be welcome in go-multiaddr btw.

Issue #3095, opened 2024-12-09
Add transport integration tests for circuitv2 transport


sukunrt:
We don't have integration tests for circuit v2 transport. It seems straightforward to add one. 

We need to start an unlimited relay and use address factory to provide the right addresses from `host.Addrs`. Ideally, it should just work.

badgooooor:
Hi, interested on working on this issue.

p-shahi:
@rose2221 There is a closed PR here: https://github.com/libp2p/go-libp2p/pull/3162 picking up where badgor left off and addressing Sukun's comments would be a good place to start!

PR #3162, opened 2025-01-29
Add host with circuitv2 to integration test

badgooooor:
### Description

- Resolve https://github.com/libp2p/go-libp2p/issues/3095
- Add host with `circuitv2` to transport integration test

### DIFF
--- a/p2p/test/transport/transport_test.go
+++ b/p2p/test/transport/transport_test.go
@@ -188,6 +188,22 @@ var transportsToTest = []TransportTestCase{
 			return h
 		},
 	},
+	{
+		Name: "Circuit V2",
+		HostGenerator: func(t *testing.T, opts TransportTestCaseOpts) host.Host {
+			libp2pOpts := transformOpts(opts)
+			libp2pOpts = append(libp2pOpts, libp2p.EnableRelay())
+			libp2pOpts = append(libp2pOpts, libp2p.EnableRelayService())
+			if opts.NoListen {
+				libp2pOpts = append(libp2pOpts, libp2p.NoListenAddrs)
+			} else {
+				libp2pOpts = append(libp2pOpts, libp2p.ListenAddrStrings(("/ip4/127.0.0.1/tcp/0/ws")))
+			}
+			h, err := libp2p.New(libp2pOpts...)
+			require.NoError(t, err)
+			return h
+		},
+	},
 }
 
 func TestPing(t *testing.T) {


p2p/test/transport/transport_test.go
```diff
@@ -830,3 +867,129 @@ func TestConnClosedWhenRemoteCloses(t *testing.T) {
 		})
 	}
 }
+
+func TestConnRelay(t *testing.T) {
+	for _, tc := range transportsToTest {
+		if tc.Name != RelayTestOnTransportName {
+			continue
```
sukunrt, line 874: We don't want a separate test for Relay. We want to run all integration tests with a Relay transport. 

This package has a bunch of tests that we want all transports to satisfy, like ConnGater Test, ResourceManager Test, PingPong on streams, etc. We should add another transport to the list of already present transports which connects hosts over a circuit relay address. 

At the moment it's fine to ignore which underlying transport is used to make the relay connection, and just use QUIC as the underlying transport. Ideally, any problem in the underlying transports should be caught when testing that specifc transport.

p2p/test/transport/transport_test.go
```diff
@@ -66,6 +69,16 @@ func transformOpts(opts TransportTestCaseOpts) []config.Option {
...
+	err := b.Connect(context.Background(), pi)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+const RelayTestOnTransportName = "QUIC"
+
 var transportsToTest = []TransportTestCase{
 	{
```
sukunrt, line 83: Add another item to this list with Name: "circuit-v2"

p2p/test/transport/transport_test.go
```diff
@@ -140,6 +165,9 @@ var transportsToTest = []TransportTestCase{
 			} else {
 				libp2pOpts = append(libp2pOpts, libp2p.ListenAddrStrings("/ip4/127.0.0.1/tcp/0/ws"))
 			}
+			if opts.EnabledRelay {
+				libp2pOpts = append(libp2pOpts, libp2p.EnableRelay(), libp2p.EnableRelayService())
+			}
```
sukunrt, line 170: It's better to just run the relay test separately. Where the dialer and the listener are both using p2p-circuit to connect.

p2p/test/transport/transport_test.go
```diff
@@ -868,6 +894,128 @@ func TestConnClosedWhenRemoteCloses(t *testing.T) {
 	}
 }
 
+func TestConnRelay(t *testing.T) {
+	for _, tc := range transportsToTest {
+		t.Run(tc.Name, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
```
sukunrt, line 901: We don't need this test. We want to run all the transport integration tests for the circuit relay transport.

p2p/test/transport/transport_test.go
```diff
@@ -225,6 +237,20 @@ var transportsToTest = []TransportTestCase{
 			return h
 		},
 	},
+	{
+		Name: "circuit-v2",
+		HostGenerator: func(t *testing.T, opts TransportTestCaseOpts) host.Host {
+			libp2pOpts := transformOpts(opts)
+			if opts.NoListen {
+				libp2pOpts = append(libp2pOpts, libp2p.NoListenAddrs)
```
sukunrt, line 245: You need to add the relay transport to the host, if not already present.

p2p/test/transport/transport_test.go
```diff
@@ -225,6 +237,20 @@ var transportsToTest = []TransportTestCase{
...
+			} else {
+				libp2pOpts = append(libp2pOpts, libp2p.ListenAddrStrings("/ip4/127.0.0.1/udp/0/quic-v1"))
+			}
```
sukunrt, line 248: You need this host to obtain a relay reservation with a relay server.  

For the listen side, you'll need to start 2 hosts. 1 relay host, 1 listener node. The listener node will obtain a relay reservation with the relay host and then advertise its address using `AddressFactory`(https://github.com/libp2p/go-libp2p/blob/master/options.go#L260)

sukunrt (changes requested):

sukunrt:
Since there's no update here for a few months, I'm closing this. Please reopen if you get around to this @badgooooor

PR #3166, opened 2025-02-01
feat: Implement Custom TCP Dialers

### LINKED ISSUE #3155: Custom Dialer for TCP transports
Some environments need to proxy TCP connections. Instead of trying to handle all cases in the library, allow a user to specify a custom `interface { DialContext(ctx context.Context, network, address string) (net.Conn, error) }` to use.

We should expose a new `Option` called `CustomTCPDialer` that accepts a `func(remoteAddr multiaddr.Multiaddr) (ContextDialer, error)`. If set, this function should be used instead of the standard net.Dialer.

cc @parkan (see https://github.com/libp2p/go-libp2p/pull/3137#issuecomment-2612294347 for some more background information)

MarcoPolo:
closes #3155. @parkan does this work for your use case? See the libp2p_test.go for an example on how to set this up.

### DIFF
--- a/p2p/transport/tcp/tcp.go
+++ b/p2p/transport/tcp/tcp.go
@@ -118,9 +118,11 @@ func WithMetrics() Option {
 	}
 }
 
-func WithCustomDialer(d CustomTCPDialer) Option {
+// WithDialerForAddr sets a custom dialer for the given address.
+// If set, it will be the *ONLY* dialer used.
+func WithDialerForAddr(d DialerForAddr) Option {
 	return func(tr *TcpTransport) error {
-		tr.customDialer = d
+		tr.overrideDialerForAddr = d
 		return nil
 	}
 }
@@ -128,7 +130,11 @@ func WithCustomDialer(d CustomTCPDialer) Option {
 type ContextDialer interface {
 	DialContext(ctx context.Context, network, address string) (net.Conn, error)
 }
-type CustomTCPDialer func(raddr ma.Multiaddr) (ContextDialer, error)
+
+// DialerForAddr is a function that returns a dialer for a given address.
+// Implementations must return either a ContextDialer or an error. It is
+// invalid to return nil, nil.
+type DialerForAddr func(raddr ma.Multiaddr) (ContextDialer, error)
 
 // TcpTransport is the TCP transport.
 type TcpTransport struct {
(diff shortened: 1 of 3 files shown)


p2p/transport/tcp/tcp.go
```diff
@@ -117,12 +118,29 @@ func WithMetrics() Option {
...
+	return func(tr *TcpTransport) error {
+		tr.customDialer = d
+		return nil
+	}
+}
+
+type ContextDialer interface {
+	DialContext(ctx context.Context, network, address string) (net.Conn, error)
+}
+type CustomTCPDialer func(raddr ma.Multiaddr) (ContextDialer, error)
```
sukunrt, line 131: NIT: `DialerForAddr`

p2p/transport/tcp/tcp.go
```diff
@@ -117,12 +118,29 @@ func WithMetrics() Option {
 	}
 }
 
+func WithCustomDialer(d CustomTCPDialer) Option {
+	return func(tr *TcpTransport) error {
+		tr.customDialer = d
+		return nil
+	}
+}
```
sukunrt, line 126: NIT: `WithDialerForAddr`

sukunrt (approved):
The change looks good except for some naming preferences. 

1. The pattern is to use the provided `Dialer` for dialling. So, do we need to name it `Custom`. 
2. `CustomTCPDialer` reads like it should be the `Dialer` not something that returns a `Dialer`.

parkan:
in practice I think this should work for us :+1: 

what's mostly on my mind is how similar functionality is already implemented in [net.proxy](https://cs.opensource.google/go/x/net/+/refs/tags/v0.34.0:proxy/proxy.go) and whether anything in there should be leveraged or used as an example to follow :thinking: 

looking at it closely [RegisterDialerType](https://cs.opensource.google/go/x/net/+/refs/tags/v0.34.0:proxy/proxy.go;l=72) is actually a good place to inject a http proxy dialer for us, though that doesn't generalize to other dialer behaviors that don't map to Url schemes

for ref the onion transport uses net/proxy already: https://github.com/OpenBazaar/go-onion-transport/blob/master/onion_transport.go#L136

parkan:
for example the [PerHost](https://pkg.go.dev/golang.org/x/net/proxy#PerHost ) functionality is very similar to `overrideDialerForAddr`

MarcoPolo:
I don't want to spend too much time bikeshedding an API that a handful of people will use. I see this as an escape hatch that enables your use case, but not something that most users should use or care about.

With some finangling, I believe the `PerHost` should work here without any go-libp2p changes.

parkan:
@MarcoPolo totally fair, thank you for getting this in :muscle:

PR #3175, opened 2025-02-06
chore: update pion/ice to v4

achingbrain:
Pion/ICE doesn't negotiate ICE role conflicts [as per RFC 8445](https://datatracker.ietf.org/doc/html/rfc8445#section-7.3.1.1).

The Pion maintainers are [open to a PR](https://github.com/pion/ice/issues/359#issuecomment-2613338072) but it won't be backported to v2 so the first step is to upgrade go-libp2p to v4.

This is needed for interop with js-libp2p's WebRTC-Direct listener.


sukunrt:
Does this fix the issue for you? Can you share the link for interop failure?   

1. The upgrade changes only where the interface is imported from. All the ice machinery should be coming from v4 on current master since that's what `pion/webrtc/v4` uses under the hood.
2. On the Dial side the ICE Role should always be controlling since servers are expected to be webrtc lite.

achingbrain:
> Can you share the link for interop failure?

Unfortunately no, because I'm testing locally with patches applied to the C/C++ WebRTC stack - I'm still waiting for these patches to be merged upstream.

> The upgrade changes only where the interface is imported from. All the ice machinery should be coming from v4

That's interesting, would it not be better to import the same version of the interface as is used by `pion/webrtc/v4`?

Either way, it turns out my issue was you have to invoke [libdatachannel](https://github.com/paullouisageneau/libdatachannel) methods *just so* in order to get it to be in controlled mode for ICE, so this isn't needed for interop after all but seems like the kind of thing you might want to do anyway?

sukunrt:
> but seems like the kind of thing you might want to do anyway?

Definitely! We should depend only on the latest ice package.

sukunrt (approved):

PR #3181, opened 2025-02-11
tcpreuse: fix Scope() for *tls.Conn

sukunrt:
Yet another interface embedding bug :sob:


p2p/transport/tcpreuse/connwithscope.go
```diff
@@ -17,6 +17,11 @@ func (c connWithScope) Scope() network.ConnManagementScope {
 	return c.scope
 }
 
+func (c *connWithScope) Close() error {
```
sukunrt, line 20: This is required for failures in the http tls handshake for wss. The http server closes the connection when handshake fails.

PR #3183, opened 2025-02-12
feat(swarm): logging waitForDirectConn return error

wlynxg:
When I tested the relay connection with the new version, I was able to successfully connect to the server peer, but a `context deadline exceeded` error occurred when calling `NewStream` because `WithAllowLimitedConn` was not used.

This error caused me trouble, and I finally found that it was a timeout in `waitForDirectConn` (because my environment could not successfully holepunch).

Logging the errors returned by waitForDirectConn can help more developers locate errors and know through code logic that WithAllowLimitedConn is needed.


sukunrt (commented):
LGTM. Can you make the line more verbose? I suggested an alternative.

sukunrt (approved):

Issue #3182, opened 2025-02-11
Refactor shared TCP Listener's resource manager scope usage


sukunrt:
We've been bitten twice by the broken Scope handling for the shared TCP Listener: #3181 #3142 

From: https://github.com/libp2p/go-libp2p/blob/master/p2p/transport/tcpreuse/listener.go#L198

> // Bugs around multiple calls to OpenConnection or InterceptAccept are prevented by the transport
> // integration tests.

This is clearly not true. So far, we've run into two configurations that we weren't testing for. 

Intuitively, we want to pass the scope explicitly between the layers: 
TCPReuse -> TCP -> Upgrader
TCPReuse -> WebSocket -> Upgrader

Currently TCPReuse sets up the Scope for the Upgrader but any wrapping done in the transports hides the Scope.

PR #3186, opened 2025-02-12
transport: add GatedMaListener type

sukunrt:
This introduces a new GatedMaListener type which gates conns
accepted from a manet.Listener with a gater and creates the rcmgr
scope for it. Explicitly passing the scope allows for many guardrails
that the previous interface assertion didn't.

This breaks the previous responsibility of the upgradeListener method
into two, one gating the connection initially, and the other upgrading
the connection with a security and muxer selection.

This split makes it easy to gate the connection with the resource
manager as early as possible. This is especially true for websocket
because we want to gate the connection just after the TCP connection is
established, and not after the tls handshake + websocket upgrade is
completed.

fixes: #3182

### DIFF
--- a/core/transport/transport.go
+++ b/core/transport/transport.go
@@ -129,11 +129,41 @@ type TransportNetwork interface {
 	AddTransport(t Transport) error
 }
 
+// GatedMaListener is listener that listens for raw(unsecured and non-multiplexed) incoming connections,
+// gates them with a `connmgr.ConnGater`and creates a resource management scope for them.
+// It can be upgraded to a full libp2p transport listener by the Upgrader.
+//
+// Compared to manet.Listener, this listener creates the resource management scope for the accepted connection.
+type GatedMaListener interface {
+	// Accept waits for and returns the next connection to the listener.
+	Accept() (manet.Conn, network.ConnManagementScope, error)
+
+	// Close closes the listener.
+	// Any blocked Accept operations will be unblocked and return errors.
+	Close() error
+
+	// Multiaddr returns the listener's (local) Multiaddr.
+	Multiaddr() ma.Multiaddr
+
+	// Addr returns the net.Listener's network address.
+	Addr() net.Addr
+}
+
 // Upgrader is a multistream upgrader that can upgrade an underlying connection
 // to a full transport connection (secure and multiplexed).
 type Upgrader interface {
 	// UpgradeListener upgrades the passed multiaddr-net listener into a full libp2p-transport listener.
+	//
+	// Deprecated: Use UpgradeGatedMaListener(upgrader.GateMaListener(manet.Listener)) instead.
 	UpgradeListener(Transport, manet.Listener) Listener
+
+	// GateMaListener creates a GatedMaListener from a manet.Listener. It gates the accepted connection
+	// and creates a resource scope for it.
+	GateMaListener(manet.Listener) GatedMaListener
+
+	// UpgradeGatedMaListener upgrades the passed GatedMaListener into a full libp2p-transport listener.
+	UpgradeGatedMaListener(Transport, GatedMaListener) Listener
+
 	// Upgrade upgrades the multiaddr/net connection into a full libp2p-transport connection.
 	Upgrade(ctx context.Context, t Transport, maconn manet.Conn, dir network.Direction, p peer.ID, scope network.ConnManagementScope) (CapableConn, error)
 }
--- a/p2p/transport/tcp/tcp.go
+++ b/p2p/transport/tcp/tcp.go
@@ -41,7 +41,7 @@ var ReuseportIsAvailable = tcpreuse.ReuseportIsAvailable
 func tryKeepAlive(conn net.Conn, keepAlive bool) {
 	keepAliveConn, ok := conn.(canKeepAlive)
 	if !ok {
-		log.Errorf("Can't set TCP keepalives.")
+		log.Errorf("can't set TCP keepalives. net.Conn of type %T doesn't support SetKeepAlive", conn)
 		return
 	}
 	if err := keepAliveConn.SetKeepAlive(keepAlive); err != nil {
@@ -76,23 +76,23 @@ func tryLinger(conn net.Conn, sec int) {
 	}
 }
 
-type tcpListener struct {
-	manet.Listener
+type tcpGatedMaListener struct {
+	transport.GatedMaListener
 	sec int
 }
 
-func (ll *tcpListener) Accept() (manet.Conn, error) {
-	c, err := ll.Listener.Accept()
+func (ll *tcpGatedMaListener) Accept() (manet.Conn, network.ConnManagementScope, error) {
+	c, scope, err := ll.GatedMaListener.Accept()
 	if err != nil {
-		return nil, err
+		if scope != nil {
+			log.Errorf("BUG: got non-nil scope but also an error: %s", err)
+			scope.Done()
+		}
+		return nil, nil, err
 	}
 	tryLinger(c, ll.sec)
 	tryKeepAlive(c, true)
-	// We're not calling OpenConnection in the resource manager here,
-	// since the manet.Conn doesn't allow us to save the scope.
-	// It's the caller's (usually the p2p/net/upgrader) responsibility
-	// to call the resource manager.
-	return c, nil
+	return c, scope, nil
 }
 
 type Option func(*TcpTransport) error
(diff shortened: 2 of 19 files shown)


core/transport/transport.go
```diff
@@ -129,11 +129,40 @@ type TransportNetwork interface {
...
+	Addr() net.Addr
+}
+
 // Upgrader is a multistream upgrader that can upgrade an underlying connection
 // to a full transport connection (secure and multiplexed).
 type Upgrader interface {
 	// UpgradeListener upgrades the passed multiaddr-net listener into a full libp2p-transport listener.
+	//
+	// Deprecated: Use UpgradeGatedManetListener(upgrader.GateManetListener(manet.Listener)) instead.
 	UpgradeListener(Transport, manet.Listener) Listener
```
MarcoPolo, line 157: Is anyone besides us using this? Should we just remove it?

p2p/transport/tcpreuse/listener.go
```diff
@@ -28,32 +28,36 @@ var log = logging.Logger("tcp-demultiplex")
...
 		reuse:           reuseport.Transport{},
-		connGater:       gater,
-		rcmgr:           rcmgr,
+		upgrader:        upgrader,
 		listeners:       make(map[string]*multiplexedListener),
 	}
 }
 
-func (t *ConnMgr) maListen(listenAddr ma.Multiaddr) (manet.Listener, error) {
+func (t *ConnMgr) UnsecuredListen(listenAddr ma.Multiaddr) (transport.GatedManetListener, error) {
```
MarcoPolo, line 46: Why is this called `UnsecuredListen`? The type doesn't sound like that's what it should be called.

MarcoPolo: Does this need to be public?

p2p/transport/websocket/http_listener.go
```diff
@@ -0,0 +1,90 @@
...
+		if scope != nil {
+			log.Errorf("BUG: scope non-nil when err is non nil: %v", err)
+			scope.Done()
+		}
+		return nil, err
+	}
+	l.mx.Lock()
+	l.scopes[conn] = scope
+	// TODO: This should be 10 seconds?
+	time.AfterFunc(time.Minute, func() {
```
MarcoPolo, line 46: Is this in case the connection is closed or dropped? This is kind of hacky

p2p/transport/websocket/http_listener.go
```diff
@@ -0,0 +1,90 @@
...
+	l.mx.Unlock()
+	return conn, nil
+}
+
+func (l *httpNetListener) ConnContext(ctx context.Context, c net.Conn) context.Context {
+	return context.WithValue(ctx, connKey{}, c)
+}
+
+func (l *httpNetListener) PopScope(c net.Conn) (network.ConnManagementScope, error) {
+	if tc, ok := c.(*tls.Conn); ok {
```
MarcoPolo, line 66: `Interface{NetConn() ...}` check instead?

p2p/transport/websocket/http_listener.go
```diff
@@ -0,0 +1,90 @@
...
+		scope.Done()
+		conn.Close()
+		delete(l.scopes, conn)
+	})
+	l.mx.Unlock()
+	return conn, nil
+}
+
+func (l *httpNetListener) ConnContext(ctx context.Context, c net.Conn) context.Context {
+	return context.WithValue(ctx, connKey{}, c)
```
MarcoPolo, line 62: Is this used?

p2p/transport/websocket/http_listener.go
```diff
@@ -0,0 +1,90 @@
...
+	"net"
+	"sync"
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/core/transport"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+type httpNetListener struct {
```
MarcoPolo, line 16: You may find [`Server.ConnState`](https://pkg.go.dev/net/http#ConnState).

MarcoPolo (commented):
brief review. Overall I like being explicit here.

p2p/transport/tcpreuse/listener.go
sukunrt: left over from a refactoring.

p2p/transport/websocket/http_listener.go
sukunrt: I'm using `tls.Conn` incase we ever implement a NetConn() method for maConn and that breaks this. `*tls.Conn` api is guaranteed by the go 1 compatibility guarantee now.

p2p/transport/websocket/http_listener.go
sukunrt: it was. But I've significantly changed this.

p2p/transport/websocket/http_listener.go
sukunrt: Do you think we need that? 
Our purpose here is similar to how we handle webtransport upgrade. I've written this similar to that code using a `*negotiatingConn`

p2p/transport/websocket/listener.go
```diff
@@ -62,107 +71,160 @@ func newListener(a ma.Multiaddr, tlsConf *tls.Config, sharedTcp *tcpreuse.ConnMg
...
-		lnet, lnaddr, err := manet.DialArgs(parsed.restMultiaddr)
-		if err != nil {
-			return nil, err
-		}
-		nl, err = net.Listen(lnet, lnaddr)
+		mal, err := manet.Listen(parsed.restMultiaddr)
 		if err != nil {
 			return nil, err
 		}
+		gmal = upgrader.GateMaListener(mal)
```
sukunrt, line 80: Note: This changes the address reported to the resource manager and connection gater from:
`/ip4/a.b.c.d/tcp/xx/ws` to `/ip4/a.b.c.d/tcp/xx`
We are not providing the `/ws` component anymore. This is the same behavior as what the shared tcp listener does as it doesn't know whether the connection will be a websocket or tcp one. 

This is easy enough to fix, but that would make the behavior between the shared tcp listener and unshared one inconsistent. 

I don't think the resource manager cares, we don't have a limit of connections per transport. 
The conn gater might care about this. If that's the case, we can move the conngater.InterceptAccept into GateMaListener and have the gatedMaListener only manage resource manager scope.

p2p/transport/websocket/http_listener.go
MarcoPolo: I meant to say you may find that helpful, but let me take a look at the latest changes.

MarcoPolo (approved):
Looks good. Just one question around the multiaddr not including "ws". I think it probably should, if only to help debugging.

p2p/transport/tcpreuse/listener.go
```diff
@@ -80,8 +83,9 @@ func getTCPAddr(listenAddr ma.Multiaddr) (ma.Multiaddr, error) {
 // DemultiplexedListen returns a listener for laddr listening for `connType` connections. The connections
 // accepted from returned listeners need to be upgraded with a `transport.Upgrader`.
 // NOTE: All listeners for port 0 share the same underlying socket, so they have the same specific port.
-func (t *ConnMgr) DemultiplexedListen(laddr ma.Multiaddr, connType DemultiplexedConnType) (manet.Listener, error) {
+func (t *ConnMgr) DemultiplexedListen(laddr ma.Multiaddr, connType DemultiplexedConnType) (transport.GatedMaListener, error) {
 	if !connType.IsKnown() {
+		fmt.Println("connType", connType)
```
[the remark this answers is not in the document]
melroy89: Seems like logging pollution to me

p2p/transport/websocket/listener.go
sukunrt: We'll need a method on the Scope / Rcmgr for this. We don't know whether it'll be a tcp or a ws connection just when we've accepted the tcp connection.

p2p/transport/tcp/tcp.go
```diff
@@ -316,22 +316,26 @@ func (t *TcpTransport) unsharedMAListen(laddr ma.Multiaddr) (manet.Listener, err
...
+		mal, err := t.unsharedMAListen(laddr)
+		if err != nil {
+			return nil, err
+		}
+		list = t.upgrader.GateMaListener(mal)
 	}
 
 	if t.enableMetrics {
-		list = newTracingListener(&tcpListener{list, 0}, t.metricsCollector)
+		// TODO: Fix this: The tcpListener wrapping should happen on both enableMetrics and disabledMetrics path
```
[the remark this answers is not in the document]
sukunrt: there are two wrappings. One is for metrics. That shouldn't happen on metrics path. The other is for setting socket options like SoLinger(0). That should happen on both parts.

p2p/transport/websocket/conn.go
```diff
@@ -167,6 +148,7 @@ func (c *Conn) closeOnceFn() error {
 		time.Now().Add(GracefulCloseTimeout),
 	)
 	err2 := c.Conn.Close()
+	c.Scope.Done()
```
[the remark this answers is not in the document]
sukunrt: removed this.

Issue #3179, opened 2025-02-09
Fix Flaky test TestServerDataRequest


sukunrt:
Most recent failure here:

sukunrt:
fixed by #3187

PR #3187, opened 2025-02-13
autonatv2: allow multiple concurrent requests per peer

sukunrt:
This fixes the flaky test:
TestServerDialData(https://github.com/libp2p/go-libp2p/issues/3179) The exact sequence of events that flakiness is:

The client rejects the dial data request by resetting the stream. The client makes a new request to the server. The server rejects this request as it hasn't read the previous stream's reset yet.

It is better to allow multiple concurrent requests per peer. 5 concurrent requests per peer is okay.

### DIFF
--- a/p2p/protocol/autonatv2/server_test.go
+++ b/p2p/protocol/autonatv2/server_test.go
@@ -149,7 +149,7 @@ func TestServerDataRequest(t *testing.T) {
 			}
 			return false
 		}),
-		WithServerRateLimit(10, 10, 10),
+		WithServerRateLimit(10, 10, 10, 2),
 		withAmplificationAttackPreventionDialWait(0),
 	)
 	defer an.Close()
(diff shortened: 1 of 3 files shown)


p2p/protocol/autonatv2/server_test.go
```diff
@@ -187,6 +187,64 @@ func TestServerDataRequest(t *testing.T) {
...
+
+	c := newAutoNAT(t, nil, allowPrivateAddrs)
+	defer c.Close()
+	defer c.host.Close()
+
+	idAndWait(t, c, an)
+
+	errChan := make(chan error)
+	const N = 10
+	// num concurrnetRequests will stall and N will fail
```
MarcoPolo, line 218: nit typo

MarcoPolo (commented):
Why 5? Would 2 be enough to fix the test? 2 also seems the more conservative choice here.

sukunrt:
No science to this. I guess 2 is okay too.

p2p/protocol/autonatv2/server_test.go
```diff
@@ -187,6 +187,64 @@ func TestServerDataRequest(t *testing.T) {
...
+	// check N failures
+	for i := 0; i < N; i++ {
+		select {
+		case err := <-errChan:
+			require.Error(t, err)
+		case <-time.After(10 * time.Second):
+			t.Fatalf("expected %d errors: got: %d", N, i)
+		}
+	}
+
```
MarcoPolo, line 235: Assert there's no waiting error in the chan?
```go
select {
    case <-errChan:
      // fail
    default:
}
```

MarcoPolo (approved):

PR #3188, opened 2025-02-13
feat(libp2phttp): More ergonomic auth

MarcoPolo:
Auth is now uniform across HTTP and stream transports and easier to use. See `ExampleHost_authenticatedHTTP` for an example. This will also show up in the docs page.

Adds:
- `ServerPeerID()` function to get a server's peer id from an `*http.Response`. Used by clients.
- `ClientPeerID()` function to get a client's peer id from an `*http.Request`. Used by servers.
- `ServerPeerIDAuth` and `ClientPeerIDAuth` fields in the http.Host struct, to specify the key used to authenticate.

### DIFF
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -820,8 +820,6 @@ func (h *Host) RoundTrip(r *http.Request) (*http.Response, error) {
 			}
 
 			serverID, resp, err := h.ClientPeerIDAuth.AuthenticateWithRoundTripper(rt, r)
-			// c := http.Client{Transport: rt}
-			// serverID, resp, err := h.ClientPeerIDAuth.AuthenticatedDo(&c, r)
 			if err != nil {
 				return nil, err
 			}
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -25,6 +25,7 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/core/protocol"
+	httpauth "github.com/libp2p/go-libp2p/p2p/http/auth"
 	gostream "github.com/libp2p/go-libp2p/p2p/net/gostream"
 	ma "github.com/multiformats/go-multiaddr"
 )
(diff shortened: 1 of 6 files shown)


p2p/http/libp2phttp.go
```diff
@@ -743,6 +810,35 @@ func (h *Host) RoundTrip(r *http.Request) (*http.Response, error) {
...
+			}
+
+			if r.Host == "" {
+				// Missing a host header. Default to what we parsed earlier
+				r.Host = u.Host
+			}
+
+			serverID, resp, err := h.ClientPeerIDAuth.AuthenticateWithRoundTripper(rt, r)
+			// c := http.Client{Transport: rt}
+			// serverID, resp, err := h.ClientPeerIDAuth.AuthenticatedDo(&c, r)
```
sukunrt, line 826: commented code.

p2p/http/libp2phttp.go
```diff
@@ -529,10 +573,16 @@ func relativeMultiaddrURIToAbs(original *url.URL, relative *url.URL) (*url.URL,
...
+	withoutPath, afterAndIncludingPath := ma.SplitFunc(originalMa, func(c ma.Component) bool {
 		return c.Protocol().Code == ma.P_HTTP_PATH
 	})
 	withNewPath := withoutPath.Encapsulate(relativePathComponent)
+	_, afterPath := ma.SplitFirst(afterAndIncludingPath)
+
+	// Include after path since it may include other relevant parts (/p2p?)
+	if afterPath != nil {
+		withNewPath = withNewPath.Encapsulate(afterPath)
+	}
```
sukunrt, line 585: NIT:

I guess the `/p2p` bit in the comment is wrong, as the spec says p2p must come before:
 
https://github.com/multiformats/multiaddr/blob/master/protocols/http-path.md?plain=1#L27

But the logic seems okay.

sukunrt (approved):

sukunrt:
Can you add some detail to the PR description for the change. Would be nice to be able to git blame and see what's happening.

p2p/http/libp2phttp.go
MarcoPolo: Thanks. Looking at this method, it isn't handling relative uris properly. I'll
fix it now.

PR #3208, opened 2025-02-25
fix(libp2phttp): Fix relative to absolute multiaddr URI logic

MarcoPolo:
The previous version did not correctly handle the case of dot segments in the path. For example if you requested `/foo/bar/` and the response `Location` header redirected you to `../baz`, you should go to `/foo/baz/`.

### DIFF
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -523,19 +523,18 @@ func (rt *streamRoundTripper) RoundTrip(r *http.Request) (*http.Response, error)
 	}
 	resp.Body = &streamReadCloser{resp.Body, s}
 
-	locUrl, err := resp.Location()
-	if err == nil {
-		// Location url in response. Is this a multiaddr uri? and is it relative?
-		// If it's relative we want to convert it to an absolute multiaddr uri
-		// so that the next request knows how to reach the endpoint.
-		if locUrl.Scheme == "multiaddr" && resp.Request.URL.Scheme == "multiaddr" {
-			// Check if it's a relative URI and turn it into an absolute one
-			u, err := relativeMultiaddrURIToAbs(resp.Request.URL, locUrl)
-			if err == nil {
-				// It was a relative URI and we were able to convert it to an absolute one
-				// Update the location header to be an absolute multiaddr uri
-				resp.Header.Set("Location", u.String())
+	if resp.Request.URL.Scheme == "multiaddr" {
+		// This was a multiaddr uri, we may need to convert relative URI
+		// references to absolute multiaddr ones so that the next request
+		// knows how to reach the endpoint.
+		locationHeader := resp.Header.Get("Location")
+		if locationHeader != "" {
+			u, err := locationHeaderToMultiaddrURI(resp.Request.URL, locationHeader)
+			if err != nil {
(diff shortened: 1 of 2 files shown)


p2p/http/libp2phttp.go
```diff
@@ -523,19 +523,18 @@ func (rt *streamRoundTripper) RoundTrip(r *http.Request) (*http.Response, error)
...
-				resp.Header.Set("Location", u.String())
+	if resp.Request.URL.Scheme == "multiaddr" {
+		// This was a multiaddr uri, we may need to convert relative URI
+		// references to absolute multiaddr ones so that the next request
+		// knows how to reach the endpoint.
+		locationHeader := resp.Header.Get("Location")
+		if locationHeader != "" {
+			u, err := locationHeaderToMultiaddrURI(resp.Request.URL, locationHeader)
+			if err != nil {
+				return nil, err
```
sukunrt, line 534: wrap the err with the request URL and locationHeader.

p2p/http/libp2phttp.go
```diff
@@ -523,19 +523,18 @@ func (rt *streamRoundTripper) RoundTrip(r *http.Request) (*http.Response, error)
...
-		// If it's relative we want to convert it to an absolute multiaddr uri
-		// so that the next request knows how to reach the endpoint.
-		if locUrl.Scheme == "multiaddr" && resp.Request.URL.Scheme == "multiaddr" {
-			// Check if it's a relative URI and turn it into an absolute one
-			u, err := relativeMultiaddrURIToAbs(resp.Request.URL, locUrl)
-			if err == nil {
-				// It was a relative URI and we were able to convert it to an absolute one
-				// Update the location header to be an absolute multiaddr uri
-				resp.Header.Set("Location", u.String())
+	if resp.Request.URL.Scheme == "multiaddr" {
```
sukunrt, line 526: You have the request here. Can we just do `r.URL.Scheme`?

p2p/http/libp2phttp.go
```diff
@@ -523,19 +523,18 @@ func (rt *streamRoundTripper) RoundTrip(r *http.Request) (*http.Response, error)
...
+		// This was a multiaddr uri, we may need to convert relative URI
+		// references to absolute multiaddr ones so that the next request
+		// knows how to reach the endpoint.
+		locationHeader := resp.Header.Get("Location")
+		if locationHeader != "" {
```
sukunrt, line 531: I'm curious, why didn't this work in the previous version where we used 
`resp.Location`
This under the hood resolves the relative path(or at least claims to)
https://github.com/golang/go/blob/master/src/net/http/response.go#L137

sukunrt (approved):

p2p/http/libp2phttp.go
MarcoPolo: It does, but it doesn't work with how we encode http paths in the
multiaddr URI. For multiaddr URIs in general we only use URI path components. A relative
URI reference should be relative to the http-path component not the whole
multiaddr. For example multiaddr:/ip4/1.2.3.4/http-path/foo with relative reference /bar
should resolve to multiaddr:/ip4/1.2.3.4/http-path/bar not multiaddr:/bar.

PR #3213, opened 2025-02-27
Make Certificate Validation / Creation Configurable

ethan-gallant:


p2p/security/tls/crypto.go
```diff
@@ -236,32 +266,6 @@ func GenerateSignedExtension(sk ic.PrivKey, pubKey crypto.PublicKey) (pkix.Exten
...
-
-	certDER, err := x509.CreateCertificate(rand.Reader, certTmpl, certTmpl, certKey.Public(), certKey)
-	if err != nil {
-		return nil, err
-	}
-	return &tls.Certificate{
-		Certificate: [][]byte{certDER},
-		PrivateKey:  certKey,
-	}, nil
-}
```
ethan-gallant, line 263: This now lives as a part of the `DefaultCertManager`

sukunrt:
Can you explain how it enables Pnet? 
Looks like you'll run into this specific problem: 
https://github.com/libp2p/go-libp2p/issues/1432#issuecomment-1106666968

p2p/transport/quic/listener.go
```diff
@@ -108,7 +107,7 @@ func (l *listener) wrapConnWithScope(qconn quic.Connection, connScope network.Co
 	// Since we don't have any way of knowing which tls.Config was used though,
 	// we have to re-determine the peer's identity here.
 	// Therefore, this is expected to never fail.
-	remotePubKey, err := p2ptls.PubKeyFromCertChain(qconn.ConnectionState().TLS.PeerCertificates)
+	remotePubKey, err := l.transport.identity.CertManager().VerifyCertChain(qconn.ConnectionState().TLS.PeerCertificates)
```
ethan-gallant, line 110: @MarcoPolo is it the case that the peers identity has already been verified at this point? 

e.g. is the authentication done here invalid aside from ensuring an already authenticated peer don't sneakily change its LibP2P identity?

MarcoPolo: We first calculate the the Peer's ID from the TLS handshake here: https://github.com/libp2p/go-libp2p/blob/master/p2p/transport/quic/transport.go#L288-L292 using the ConfigForPeer function.

This part gets peer's id so we can include it in the returned connection.

We are on the listening side here, so we don't have an expectation of what the peer should be (as opposed to dialing where you expect a certain peer id). We just return the peer id we learn about from the TLS handshake

sukunrt:
In its present form, allowing arbitrary TLS configuration, this will be too much maintenance burden. 
I am willing to review / maintain a solution that works for: https://github.com/libp2p/specs/issues/489

Note, the problem with signing certificates with psk is that you now have to protect the psk as well as *all* the TLS private keys that are produced, since compromising the signed certificate & the private key will allow anyone to join the private network. 
See: https://github.com/libp2p/go-libp2p/issues/1432#issuecomment-1106666968

PR #3215, opened 2025-02-27
refactor: use a more straightforward return value

shenpengfeng:
The error here must be nil because a check for err != nil has already been performed earlier.

Use a more straightforward return value.


sukunrt:
Closing this as it has little to no impact on behavior or readability.

PR #3219, opened 2025-02-28
fix: fix incorrect nil return value

huochexizhan:
Since we have already checked err before and returned != nil, err must be nil here. In fact, it should return `r.Error`.


huochexizhan:
Thanks for your review.

sukunrt (approved):

sukunrt (approved):

Issue #3165, opened 2025-02-01
UDP QUIC Hole Punching Fails due to Port Reuse


ethan-gallant:
We're leveraging LibP2Ps hole punching functionality to mesh together several private networks. The network layout looks something like this

We have multiple apps running in Kubernetes clusters, each cluster has:
- 1 Relay Node with
    -  A public address with 443 open
    - A private address with 443 also open (for local peers to register with the Relay for relaying)
- 2 Peer Nodes with
    - A public address that is different than that of the Relay where ports cannot be manually opened / forwarded
    - A private address on the same network as the cluster relay

Below visualizes how this looks in actuality
```mermaid
flowchart TD
  subgraph Cluster1
    R1["Relay Node<br>Public: 443 open<br>Private: 443 open"]
    P1["Peer Node 1<br>Public: no ports<br>Private: 443"]
    P2["Peer Node 2<br>Public: no ports<br>Private: 443"]
    P1 -- "register via Private:443" --> R1
    P2 -- "register via Private:443" --> R1
  end

  subgraph Cluster2
    R2["Relay Node<br>Public: 443 open<br>Private: 443 open"]
    P3["Peer Node 1<br>Public: no ports<br>Private: 443"]
    P4["Peer Node 2<br>Public: no ports<br>Private: 443"]
    P3 -- "register via Private:443" --> R2
    P4 -- "register via Private:443" --> R2
  end

  subgraph Cluster3
    R3["Relay Node<br>Public: 443 open<br>Private: 443 open"]
    P5["Peer Node 1<br>Public: no ports<br>Private: 443"]
    P6["Peer Node 2<br>Public: no ports<br>Private: 443"]
    P5 -- "register via Private:443" --> R3
    P6 -- "register via Private:443" --> R3
  end
```

The connection flow looks something like below:
1. The peers connect locally over the network to eachother
```mermaid
graph TD
    subgraph "Initial State & Local Discovery"
        subgraph "Cluster A"
            PA1[Peer A1]
            PA2[Peer A2]
            PA1 --- PA2
        end
        subgraph "Cluster B"
            PB1[Peer B1]
            PB2[Peer B2]
            PB1 --- PB2
        end
    end
```

2. The peers connect through the Relays to eachother
```mermaid
graph TD
    subgraph "Relay Connection Phase"
        subgraph "Cluster A"
            PA1[Peer A1]
            PA2[Peer A2]
            RA[Relay A]
        end
        subgraph "Cluster B"
[...]

PR #3221, opened 2025-03-04
connmgr: fix transport association bug

sukunrt:
The second listen for the same IP:Port combination doesn't get tagged with the association. So if you listen on webtransport after quic, the wt dials won't reuse the transport. 

We should refactor this is bit more, but any change will probably change the refCountedTransport interface which is returned from one public method, so I'd like to do that in a separate PR, and use this for a patch release.  

This probably impacts: #3165

### DIFF
--- a/p2p/transport/quicreuse/connmgr.go
+++ b/p2p/transport/quicreuse/connmgr.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"sync"
@@ -190,6 +191,18 @@ func (c *ConnManager) ListenQUICAndAssociate(association any, addr ma.Multiaddr,
 		}
 		key = tr.LocalAddr().String()
 		entry = quicListenerEntry{ln: ln}
+	} else if c.enableReuseport && association != nil {
+		reuse, err := c.getReuse(netw)
+		if err != nil {
+			return nil, fmt.Errorf("reuse error: %w", err)
+		}
+		err = reuse.AssertTransportExists(entry.ln.transport)
+		if err != nil {
+			return nil, fmt.Errorf("reuse assert transport failed: %w", err)
+		}
+		if tr, ok := entry.ln.transport.(*refcountedTransport); ok {
+			tr.associate(association)
+		}
 	}
 	l, err := entry.ln.Add(tlsConf, allowWindowIncrease, func() { c.onListenerClosed(key) })
 	if err != nil {
--- a/p2p/transport/quicreuse/reuse.go
+++ b/p2p/transport/quicreuse/reuse.go
@@ -303,6 +303,10 @@ func (r *reuse) transportForDialLocked(association any, network string, source *
 					return tr, nil
 				}
 			}
+			// We don't have a transport with the association, use any one
+			for _, tr := range trs {
+				return tr, nil
+			}
 		}
 	}
 
@@ -312,6 +316,10 @@ func (r *reuse) transportForDialLocked(association any, network string, source *
 		if tr.hasAssociation(association) {
 			return tr, nil
 		}
+	}
+	// We don't have a transport with the association, use any one
+	for _, tr := range r.globalListeners {
+		return tr, nil
 	}
 
 	// Use a transport we've previously dialed from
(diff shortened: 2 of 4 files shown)


guillaumemichel (approved):

ethan-gallant:
@sukunrt can we merge this one?

p2p/transport/quicreuse/reuse.go
```diff
@@ -360,6 +368,35 @@ func (r *reuse) AddTransport(tr *refcountedTransport, laddr *net.UDPAddr) error
 	return nil
 }
 
+func (r *reuse) AssociateListenTransport(association any, tr refCountedQuicTransport) error {
```
MarcoPolo, line 371: Do we need this method? It's doing some sanity checks, but ultimately it just
calls `tr.associate`

p2p/transport/quicreuse/connmgr.go
```diff
@@ -190,6 +191,15 @@ func (c *ConnManager) ListenQUICAndAssociate(association any, addr ma.Multiaddr,
...
 		entry = quicListenerEntry{ln: ln}
+	} else if c.enableReuseport && association != nil {
+		reuse, err := c.getReuse(netw)
+		if err != nil {
+			return nil, fmt.Errorf("reuse error: %w", err)
+		}
+		err = reuse.AssociateListenTransport(association, entry.ln.transport)
+		if err != nil {
+			return nil, fmt.Errorf("reuse associate failed: %w", err)
+		}
```
MarcoPolo, line 202:
```suggestion
	} else if c.enableReuseport {
		if tr, ok := entry.ln.transport.(*refcountedTransport); ok {
			tr.associate(association)
		}
```

sukunrt: What's wrong with sanity checks? 
- The reuse type isn't exported, so this method is effectively private to this package. 

I'd argue for one of the two:
1. We move all associate inside reuse.
or 
2. We add Associate on the refCountedTransport interface with a NOOP implementation in singleOwnerTransport to avoid the type assertion. 

But for a patch release, I think this method with sanity checks is fine.

MarcoPolo: Sanity checks are fine in general. These seem to be in the wrong place and thus add noise to the change. You are checking the reuse struct's invariants. Is there a reason you don't trust these invariants? Would it be better to do these checks where we would break the invariants instead ([here](https://github.com/libp2p/go-libp2p/blob/master/p2p/transport/quicreuse/reuse.go#L415-L429))?

If you do trust these invariants, then the sanity checks here are just noise. and the above suggestion is a clearer change.

MarcoPolo: I think something like this would also be okay:

```suggestion
	} else if c.enableReuseport {
	    // ... get reuse
	    err := reuse.AssertTransportExists(tr) // This method _only_ checks invariants
		// ... handle err
		if tr, ok := entry.ln.transport.(*refcountedTransport); ok {
			tr.associate(association)
		}
```

This split seems clearer to me because you have your sanity check step, and then your normal association step.

Note that I believe I would have also made the same initial design, and am mostly able to make this point as a fresh reviewer of this change.

sukunrt: >Is there a reason you don't trust these invariants?

The input transport is provided from outside reuse. That user might pass in anything. 

On the optimal design though, I've changed my mind. I think we should put associate on `refCountedTransport` and have a NO-Op implementation in `singleOwnerTransport`. Then we move the associate bit out of `transportForListen` in to this function. So associate always happens on all code paths no matter the input. 
I'm not sure how semver compliant that is so I'll do that in a separate PR.

MarcoPolo: > The input transport is provided from outside reuse. That user might pass in anything.

I don't follow. The transport here is from the `quicListeners` map which is controlled by this quicreuse package.

> On the optimal design though, I've changed my mind. I think we should put associate on refCountedTransport and have a NO-Op implementation in singleOwnerTransport.

No strong opinion here, but I think it's fine to keep this method only on refCountedTransport as it's only meaningful there.

MarcoPolo (approved):

PR #3225, opened 2025-03-07
fix(libp2phttp): bound NewStream timeout

algorandskiy:
`libp2phttp.streamRoundTripper.RoundTrip` uses a context from `http.Request` when establishing a new stream.
Suppose a user expects to fetch very large files with multiplexed http so they set very large timeout into `http.Request` context. Then suppose one of the servers does not respond to protocols negotiation for some reason (for example by reaching out max connections set via resource manager).
This makes `NewStream` to stuck in reading server response for potentially for very long time until the context with high timeout value expires.

Tested manually.

### DIFF
--- a/p2p/http/libp2phttp.go
+++ b/p2p/http/libp2phttp.go
@@ -25,7 +25,6 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/core/protocol"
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Thanks.

I like this change, but I don't love that it brings in basichost as a dependency. It might be better to simply copy the default timeout to this package instead.

algorandskiy:
Thanks for the review. Fixed.

MarcoPolo (approved):

PR #3222, opened 2025-03-05
optimize: using t.TempDir() replace os.MkdirTemp in testfile

argentpapa:
logs:
```
➜  go-libp2p git:(master) ✗ go test -v ./p2p/transport/quicreuse/  -run '' 

=== RUN   TestListenOnSameProto
=== RUN   TestListenOnSameProto/with_reuseport
=== RUN   TestListenOnSameProto/without_reuseport
--- PASS: TestListenOnSameProto (0.02s)
    --- PASS: TestListenOnSameProto/with_reuseport (0.01s)
    --- PASS: TestListenOnSameProto/without_reuseport (0.01s)
=== RUN   TestConnectionPassedToQUICForListening
--- PASS: TestConnectionPassedToQUICForListening (0.00s)
=== RUN   TestAcceptErrorGetCleanedUp
    connmgr_test.go:113: num goroutines: 7
    connmgr_test.go:127: num goroutines: 7
--- PASS: TestAcceptErrorGetCleanedUp (0.00s)
=== RUN   TestConnectionPassedToQUICForDialing
--- PASS: TestConnectionPassedToQUICForDialing (0.00s)
=== RUN   TestListener
=== RUN   TestListener/with_reuseport
=== RUN   TestListener/without_reuseport
--- PASS: TestListener (0.04s)
    --- PASS: TestListener/with_reuseport (0.02s)
    --- PASS: TestListener/without_reuseport (0.02s)
=== RUN   TestExternalTransport
--- PASS: TestExternalTransport (0.02s)
=== RUN   TestConvertToQuicMultiaddr
--- PASS: TestConvertToQuicMultiaddr (0.00s)
=== RUN   TestConvertToQuicV1Multiaddr
--- PASS: TestConvertToQuicV1Multiaddr (0.00s)
=== RUN   TestConvertFromQuicV1Multiaddr
--- PASS: TestConvertFromQuicV1Multiaddr (0.00s)
=== RUN   TestReuseListenOnAllIPv4
--- PASS: TestReuseListenOnAllIPv4 (0.05s)
=== RUN   TestReuseListenOnAllIPv6
--- PASS: TestReuseListenOnAllIPv6 (0.05s)
=== RUN   TestReuseCreateNewGlobalConnOnDial
--- PASS: TestReuseCreateNewGlobalConnOnDial (0.00s)
=== RUN   TestReuseConnectionWhenDialing
--- PASS: TestReuseConnectionWhenDialing (0.00s)
[...]

PR #3226, opened 2025-03-08
refactor(test): using `t.TempDir` to auto manage temp file in tests

tinyfoxy:
Similar to #3222

BTW: I have checked the whole project and there is no more tests need to be replaced into `t.TempDir`.


sukunrt:
I think this is not used at all.

sukunrt:
closing in favor of #3252

Issue #1437, opened 2019-06-07
Record X-Real-IP header


magik6k:
We currently don't support TLS here, so a reverse proxy needs to be set up, which has the side effect of loosing real ip addresses in `ipfs swarm peers`.

We could check if the `X-Real-IP` header is set when accepting connections and change the ip address we report as the remote (and probably only do that when we proxy from localhost)

MarcoPolo:
Here's how I think we could implement this:

* Add option to always trust this header
* By default, trust this header only of private network/localhost

snichols:
Another issue that comes from using a proxy is that the host thinks that remote websocket connections are coming from localhost which causes the identity protocol to share private addresses with websocket peers.  When this happens, browser peers react poorly and refuse to connect to the proxied peer after identify is completed. I'm still narrowing down that piece and will report to the libp2p-js repo.

Relevant filtering code in identify: https://github.com/libp2p/go-libp2p/blob/be32b5ba0d7fc08135a27193d00f94808f0d828c/p2p/protocol/identify/id.go#L999

That being said, this issue is a thorn in the side of anyone that wants to easily provide browser-friendly websocket transports.  I'm using ngrok -> nginx -> libp2p to wrangle my web traffic.  This would be exacerbated if I were deploying to Kubernetes as web ingress relies on standard proxying to communicate details about the external connection.  Accepting standard headers seems like a necessary feature to me and would help make adoption of libp2p easier in production environments, IMO.

PR #3235, opened 2025-03-14
wip: feat: support forwarded headers for websocket

jclab-joseph:
Fixes https://github.com/libp2p/go-libp2p/issues/1437
Also https://github.com/ipfs/kubo/issues/8193

Kubo Log:
```
DEBUG   upgrader        upgrader/listener.go:99 resource manager blocked accept of new connection       {"error": "connections per ip limit exceeded for /ip4/10.42.111.111/tcp/38108/ws"}
```

### DIFF
--- a/p2p/transport/websocket/listener.go
+++ b/p2p/transport/websocket/listener.go
@@ -29,6 +29,8 @@ type listener struct {
 	// so we can't rely on checking if server.TLSConfig is set.
 	isWss bool
 
+	allowForwardedHeader bool
+
 	laddr ma.Multiaddr
 
 	incoming chan *Conn
@@ -52,7 +54,7 @@ func (pwma *parsedWebsocketMultiaddr) toMultiaddr() ma.Multiaddr {
 
 // newListener creates a new listener from a raw net.Listener.
 // tlsConf may be nil (for unencrypted websockets).
-func newListener(a ma.Multiaddr, tlsConf *tls.Config, sharedTcp *tcpreuse.ConnMgr) (*listener, error) {
+func newListener(a ma.Multiaddr, tlsConf *tls.Config, sharedTcp *tcpreuse.ConnMgr, allowForwardedHeader bool) (*listener, error) {
 	parsed, err := parseWebsocketMultiaddr(a)
 	if err != nil {
 		return nil, err
@@ -106,6 +108,8 @@ func newListener(a ma.Multiaddr, tlsConf *tls.Config, sharedTcp *tcpreuse.ConnMg
 		laddr:    parsed.toMultiaddr(),
 		incoming: make(chan *Conn),
 		closed:   make(chan struct{}),
+		
+		allowForwardedHeader: allowForwardedHeader,
 	}
 	ln.server = http.Server{Handler: ln, ErrorLog: stdLog}
 	if parsed.isWSS {
@@ -130,7 +134,13 @@ func (l *listener) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// The upgrader writes a response for us.
 		return
 	}
-	nc := NewConn(c, l.isWss)
+
+	var overrideRemoteAddr string
+	if l.allowForwardedHeader {
+		overrideRemoteAddr = GetRealIP(c.RemoteAddr(), r.Header)
+	}
+
+	nc := NewConn(c, l.isWss, overrideRemoteAddr)
 	if nc == nil {
 		c.Close()
 		w.WriteHeader(500)
--- a/p2p/transport/websocket/real_ip.go
+++ b/p2p/transport/websocket/real_ip.go
@@ -1,0 +1,82 @@
+package websocket
+
+import (
+	"fmt"
+	"net"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+func GetRealIP(addr net.Addr, h http.Header) string {
+	remoteAddr := addr.String()
+	remoteIp := GetRealIPFromHeader(h)
+	if remoteIp != nil {
+		remoteTcpAddr, ok := addr.(*net.TCPAddr)
+		if ok {
+			remoteAddr = IpPort(remoteIp, remoteTcpAddr.Port)
+		} else {
+			p := strings.LastIndex(remoteAddr, ":")
+			if p > 0 {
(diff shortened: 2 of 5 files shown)


p2p/transport/websocket/listener.go
```diff
@@ -52,7 +55,7 @@ func (pwma *parsedWebsocketMultiaddr) toMultiaddr() ma.Multiaddr {
 
 // newListener creates a new listener from a raw net.Listener.
 // tlsConf may be nil (for unencrypted websockets).
-func newListener(a ma.Multiaddr, tlsConf *tls.Config, sharedTcp *tcpreuse.ConnMgr) (*listener, error) {
+func newListener(a ma.Multiaddr, tlsConf *tls.Config, sharedTcp *tcpreuse.ConnMgr, allowForwardedHeader bool, trustedProxies []*net.IPNet) (*listener, error) {
```
sukunrt, line 58: Can we use `WithNetworkPrefixLimit` to allow many connections from trustedProxies?  

https://github.com/libp2p/go-libp2p/blob/master/p2p/host/resource-manager/conn_limiter.go#L75

jclab-joseph: WithNetworkPrefixLimit is a feature that adjusts connection limits based on specific network prefixes, and it should be applied to the real IPs obtained from trusted proxies.
If use Ingress, since there are only a few (2-3) ingress points, increasing the WithNetworkPrefixLimit indefinitely does not seem like a good solution.

jclab-joseph: My PR replaces RemoteAddr() with a real IP. When used with WithNetworkPrefixLimit , proxy connections will not be affected and will be applied to the real IP range.

MarcoPolo: I've not reviewed the code yet, but I think @jclab-joseph is right here. We don't want to just allow many connections from the trusted proxy as that allows a single malicious actor to use all those connections. Instead, these proxied connections, for the resource manager, should refer to the "real ip" address.

jclab-joseph: @MarcoPolo The trusted proxy must be inside the firewall. Nodes in the public network must not have allowForwardedHeader enabled. This is common in most programs.

p2p/transport/websocket/listener.go
```diff
@@ -124,13 +130,46 @@ func (l *listener) serve() {
 	}
 }
 
+// isProxyTrusted checks if the given IP address belongs to a trusted proxy
+func (l *listener) isProxyTrusted(remoteAddr string) bool {
+	if len(l.trustedProxies) == 0 {
+		// allow any
+		return true
```
MarcoPolo, line 137: This seems surprising. I'd rather have an explicit match everything mask (i.e. 0.0.0.0/0)

p2p/transport/websocket/real_ip.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+	return remoteAddr
+}
+
+// GetRealIPFromHeader extracts the client's real IP address from HTTP request header.
+// It checks various proxy header to find the actual IP.
+func GetRealIPFromHeader(h http.Header) net.IP {
+	// Check X-Real-IP header (used by Nginx and others)
+	ipStr := h.Get("X-Real-IP")
```
MarcoPolo, line 32: This function seems hacky. It doesn't seem like this is standardized. It might be preferable to force the user to define this function rather than try solving all cases.

One problem here is what happens if my reverse proxy is only built to handle one of these header keys, and the client can smuggle a different value.


For example, imagine if my proxy only knows about "True-Client-IP" header, but it doesn't know it should strip "X-Real-IP", "X-Forwarded-For", or "CF-Connecting-IP". That means that a client can simply set one of those other headers and trick the server.

p2p/transport/websocket/websocket.go
```diff
@@ -81,6 +83,45 @@ func WithTLSConfig(conf *tls.Config) Option {
 	}
 }
 
+// WithAllowForwardedHeader configures whether to allow the usage of
+// Forwarded and X-Forwarded-For headers to determine the real client
+// IP address when behind a proxy or load balancer.
+// configurable a list of trusted proxy IP addresses or CIDR ranges.
+// Only when a connection is from a trusted proxy, Forwarded headers will be used.
+// If trustedProxies is empty, everything is trusted.
+func WithAllowForwardedHeader(trustedProxies []string) Option {
```
MarcoPolo, line 92: I'd accept a `[]netip.Prefix` so you get type safety and avoid all this parsing logic.

p2p/transport/websocket/listener.go
```diff
@@ -29,6 +29,9 @@ type listener struct {
 	// so we can't rely on checking if server.TLSConfig is set.
 	isWss bool
 
+	allowForwardedHeader bool
+	trustedProxies       []*net.IPNet
```
MarcoPolo, line 33: nit: prefer `Prefix`, `Addr`, and `AddrPort` from the `netip` package. Generally fewer allocations and memory friendly.

MarcoPolo (changes requested):
I like the general idea here. I think we need to be careful not to open up security vulnerabilities here.

p2p/transport/websocket/real_ip.go
MarcoPolo: Ah the standard seems to be: https://datatracker.ietf.org/doc/html/rfc7239

I think it makes sense to support this standard. But allow users to override the standard behavior sense it seems not all proxies support this standard(?)

p2p/transport/websocket/listener.go
jclab-joseph: I agree. I removed allow any.

p2p/transport/websocket/real_ip.go
jclab-joseph: The default is rfc7239 only, but I modified it to be customizable.

p2p/transport/websocket/listener.go
sukunrt: Makes sense. It would be a security issue to allow many connections through this trustedProxy using network limits.

ligustah:
What's needed to get this in? I'd very much like to use this in my setup. Thanks!

PR #3237, opened 2025-03-17
fix: Add wrapped error type in `stream.go`

### LINKED ISSUE #3156: Return wrapped `network.ErrReset` error type with more info
Right now we replace specific stream errors with `network.ErrReset` ([example](https://github.com/libp2p/go-libp2p/blob/fde0e3a487fc18fc22eb5dfc97d81572b79b5d95/p2p/transport/webtransport/stream.go#L40)). We should instead return a new error type that unwraps to a  `network.ErrReset` with a [`.Unwrap() []error` method](https://pkg.go.dev/errors) and includes addition error information gleaned from the transport. This is at least a problem in the WebTransport and QUIC transports. Note that we probably want the multi-error return version and return both a network.ErrReset and the underlying stream error.

Prabhat1308:
fixes #3156

### DIFF
--- a/p2p/transport/webtransport/stream.go
+++ b/p2p/transport/webtransport/stream.go
@@ -13,18 +13,6 @@ const (
 	reset webtransport.StreamErrorCode = 0
 )
 
-type StreamResetError struct {
-	underlying webtransport.StreamError
-}
-
-func (e *StreamResetError) Error() string {
-	return "stream reset: " + e.underlying.Error()
-}
-
-func (e *StreamResetError) Unwrap() []error {
-	return []error{network.ErrReset, &e.underlying}
-}
-
 type webtransportStream struct {
 	webtransport.Stream
 	wsess *webtransport.Session
@@ -51,7 +39,11 @@ func (s *stream) Read(b []byte) (n int, err error) {
 	if err != nil && errors.Is(err, &webtransport.StreamError{}) {
 		var streamErr *webtransport.StreamError
 		if errors.As(err, &streamErr) {
-			err = &StreamResetError{underlying: *streamErr}
+			err = &network.StreamError{
+				ErrorCode:      0,
+				Remote:         true,
+				TransportError: streamErr,
+			}
 		}
 	}
 	return n, err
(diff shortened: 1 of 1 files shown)


sukunrt (changes requested):
Can we just use: 
`StreamError` from:

https://github.com/libp2p/go-libp2p/blob/master/core/network/mux.go#L17

sukunrt:
You can just use the remote field on `webtransport.StreamError`

p2p/transport/webtransport/stream.go
```diff
@@ -37,15 +37,29 @@ var _ network.MuxedStream = &stream{}
...
 	if err != nil && errors.Is(err, &webtransport.StreamError{}) {
-		err = network.ErrReset
+		var streamErr *webtransport.StreamError
+		if errors.As(err, &streamErr) {
+			err = &network.StreamError{
+				ErrorCode:      0,
+				Remote:         streamErr.Remote,
+				TransportError: streamErr,
+			}
+		}
```
sukunrt, line 62: Does this work? 
```
if err != nil {
    var streamErr *webtransport.StreamError
    if errors.As(err, &streamErr) {
        err = &network.StreamError {...}
    }
}
```

We avoid the traversing the error chain twice with both `errors.Is` and `errors.As`.

p2p/transport/webtransport/stream.go
```diff
@@ -37,15 +37,29 @@ var _ network.MuxedStream = &stream{}
 func (s *stream) Read(b []byte) (n int, err error) {
 	n, err = s.Stream.Read(b)
 	if err != nil && errors.Is(err, &webtransport.StreamError{}) {
-		err = network.ErrReset
+		var streamErr *webtransport.StreamError
+		if errors.As(err, &streamErr) {
+			err = &network.StreamError{
+				ErrorCode:      0,
+				Remote:         streamErr.Remote,
+				TransportError: streamErr,
```
sukunrt, line 45: Should be `err`. We want the complete / unparsed / wrapped error in TransportError.

p2p/transport/webtransport/stream.go
Prabhat1308: Ahh missed this. Fixed !

p2p/transport/webtransport/stream.go
Prabhat1308: fixed.

sukunrt (approved):

sukunrt:
Thanks @Prabhat1308

MarcoPolo (approved):

PR #3247, opened 2025-03-21
autonatv2: fix server dial data request policy

sukunrt:
The policy was comparing the connection local addr to the observed addr. This should be comparing the connection remote addr to the requested dial addr.

The impact here is:
if we refresh reachability for addresses every hour, we will be spending 100kB per address per hour. That's equivalent to 30 B/s. For 10 addrs  this will be 300 B/s or 3kb/s

### DIFF
--- a/p2p/protocol/autonatv2/server.go
+++ b/p2p/protocol/autonatv2/server.go
@@ -30,7 +30,7 @@ var (
 	errDialDataRefused       = errors.New("dial data refused")
 )
 
-type dataRequestPolicyFunc = func(s network.Stream, dialAddr ma.Multiaddr) bool
+type dataRequestPolicyFunc = func(observedAddr, dialAddr ma.Multiaddr) bool
 
 type EventDialRequestCompleted struct {
 	Error            error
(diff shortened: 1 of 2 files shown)


p2p/protocol/autonatv2/server.go
```diff
@@ -517,11 +517,11 @@ func (r *rateLimiter) Close() {
 
 // amplificationAttackPrevention is a dialDataRequestPolicy which requests data when the peer's observed
 // IP address is different from the dial back IP address
-func amplificationAttackPrevention(s network.Stream, dialAddr ma.Multiaddr) bool {
```
MarcoPolo, line 520: dialAddr was previously unused? That seems like it should have been caught by a lint check.

p2p/protocol/autonatv2/server_test.go
```diff
@@ -520,6 +520,75 @@ func TestReadDialData(t *testing.T) {
...
+		Reachability: network.ReachabilityPublic,
+		Status:       pb.DialStatus_OK,
+	}, res)
+
+	// ipv6 address should require dial data
+	_, err = c.GetReachability(context.Background(), []Request{{Addr: quicv6Addr, SendDialData: false}})
+	require.Error(t, err)
+
+	// ipv6 address should work fine with dial data
+	res, err = c.GetReachability(context.Background(), []Request{{Addr: quicv6Addr, SendDialData: true}})
```
MarcoPolo, line 570: I don't think this check actually asserts that dial data was requested, right?

Is there a way to assert that we were requested dial data?

MarcoPolo (commented):
one thing around the test, but otherwise good

p2p/protocol/autonatv2/server_test.go
sukunrt: added this to the checks:
`	require.ErrorContains(t, err, "invalid dial data request: low priority addr")`

MarcoPolo (approved):

PR #3255, opened 2025-03-26
webrtc: support receiving 256kB messages

sukunrt:
In experiments with js we've found that increasing the message size
increases throughput. See: https://github.com/libp2p/specs/pull/628#issuecomment-2451491070
for details. 

This changes the protobuf reader for the stream to read 256kB messages.
This also forces a change to the connection SCTP read buffer to be
increased to about 2.5 MB, to support 1 message being buffered for 10
streams.

This isn't enough to support larger messages. We most likely need to
change the inferred SDP of the server to use 256kB maxMessageSize, and
need some backwards compatible mechanism in the handshake to opt in to
large messages. See: https://github.com/libp2p/specs/pull/628 for
details


p2p/transport/webrtc/stream_write.go
```diff
@@ -110,11 +110,19 @@ func (s *stream) SetWriteDeadline(t time.Time) error {
 	return nil
 }
 
+func (s *stream) sendBufferSize() int {
```
guillaumemichel, line 113: moving comments
```suggestion
// sendBufferSize() is the maximum data we enqueue on the underlying data channel for writes.
// The underlying SCTP layer has an unbounded buffer for writes. We limit the amount enqueued
// per stream is limited to avoid a single stream monopolizing the entire connection.
func (s *stream) sendBufferSize() int {
```

p2p/transport/webrtc/stream_write.go
```diff
@@ -110,11 +110,19 @@ func (s *stream) SetWriteDeadline(t time.Time) error {
...
+	return 2 * s.maxSendMessageSize
+}
+
+func (s *stream) sendBufferLowThreshold() int {
```
guillaumemichel, line 117: moving comments
```suggestion
// sendBufferLowThreshold() is the threshold below which we write more data on the underlying
// data channel. We want a notification as soon as we can write 1 full sized message.
func (s *stream) sendBufferLowThreshold() int {
```

guillaumemichel (approved):

PR #3258, opened 2025-03-30
fix(relay): fix data-race in relayFinder

b00f:
# Description  

This PR fixes a data race condition in the relay finder. It was observed [here](https://github.com/pactus-project/pactus/actions/runs/14100227555/job/39494949146?pr=1713#step:4:511).  

The `cleanupDisconnectedPeers` function deletes `rf.relays` while, at the same time, `getCircuitAddrs` iterates over the map.  

This PR adds a lock in `getCircuitAddrs` to ensure it iterates over the map only when it is unlocked.


sukunrt:
Thanks @b00f

Issue #3250, opened 2025-03-24
Avoid dependency on concreate datastore implementations


Wondertan:
I was looking at why certain dependencies are in dep tree and realized that go-libp2p imports some seemingly unnecessary ones. Recently, go-watchdog was removed which is nice, as it pulls a lot, but also we could avoid importing go-ds-badger and go-ds-leveldb. Both of these are [imported solely to test peerstoreds](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/peerstore/pstoreds/ds_test.go), but its seems totally redundant.

Also, its worth bumping go-datastore as it obliterates go-process out of deptree

PR #3260, opened 2025-03-31
peerstore: remove leveldb tests

sukunrt:
This also removes the go-leveldb-datastore dependency. There's no
reason to test with levelDB. This code should work with any compliant
go-datastore.

Bumps go-datastore to latest as it removes the go-process dependency. 

Fixes: https://github.com/libp2p/go-libp2p/issues/3250


Wondertan:
* There is a https://github.com/ipfs/go-datastore/blob/master/sync/sync.go wrapper.
* Can you also bump go-datastore version? Then #3250 is going to be entirely fixed

guillaumemichel (approved):

PR #3262, opened 2025-04-03
feat(simlibp2p): Simulated libp2p Networks

MarcoPolo:
This adds a new Simulated Link to extend the Simulated UDP Packet conn. Simulated links can emulate bandwidth and latency over a connection, and can be used to connect two simulated conns together.

This also adds a simulated network helper to easily create a libp2p network and route the various links.

As-is, this works as a simple real-time simulator that may lose precision as computation gets bigger. The real benefit is when this is combined with Go 1.24's new `testing/synctest` package that allows the simulated network to run as a discrete event simulator. For this use case, only QUIC is supported.

This should serve to replace mocknet with a more accurate environment.

In draft until I clean up the tests a bit and use it a bit more.

### DIFF
--- a/x/simlibp2p/libp2p.go
+++ b/x/simlibp2p/libp2p.go
@@ -41,7 +41,7 @@ type MockSourceIPSelector struct {
 	ip atomic.Pointer[net.IP]
 }
 
-func (m *MockSourceIPSelector) PreferredSourceIPForDestination(dst *net.UDPAddr) (net.IP, error) {
+func (m *MockSourceIPSelector) PreferredSourceIPForDestination(_ *net.UDPAddr) (net.IP, error) {
 	return *m.ip.Load(), nil
 }
 
@@ -53,7 +53,7 @@ func QUICSimnet(simnet *simnet.Simnet, linkSettings simnet.NodeBiDiLinkSettings,
 		quicreuse.OverrideSourceIPSelector(func() (quicreuse.SourceIPSelector, error) {
 			return m, nil
 		}),
-		quicreuse.OverrideListenUDP(func(network string, address *net.UDPAddr) (net.PacketConn, error) {
+		quicreuse.OverrideListenUDP(func(_ string, address *net.UDPAddr) (net.PacketConn, error) {
 			m.ip.Store(&address.IP)
 			c := simnet.NewEndpoint(address, linkSettings)
 			return c, nil
@@ -133,7 +133,7 @@ func newBlankHost(opts BlankHostOpts) (*wrappedHost, error) {
 		quicreuse.OverrideSourceIPSelector(func() (quicreuse.SourceIPSelector, error) {
 			return m, nil
 		}),
-		quicreuse.OverrideListenUDP(func(network string, address *net.UDPAddr) (net.PacketConn, error) {
+		quicreuse.OverrideListenUDP(func(_ string, address *net.UDPAddr) (net.PacketConn, error) {
 			m.ip.Store(&address.IP)
 			c := opts.simnet.NewEndpoint(address, opts.linkSettings)
 			return c, nil


MarcoPolo:
hah these tests will be infinitely flaky in CI. but should be perfect once we can use go 1.24's synctest here (as a buildtag, then another command for just tests with those build tags, just as we do with nocover)

MarcoPolo:
Fixed flaky tests by using synctest. Set an extra CI step to run these synctest tests, like we do for NoCover.

MarcoPolo:
cc @Wondertan since you are (were?) a user of mocknet. I'm curious if you would find this useful. It's more realistic than mocknet (you can set bandwidth limits and latencies) and you exercise all the same quic code you would in production. Combining this with synctest means you should be able to run large simulated networks accurately.

Wondertan:
Hey @MarcoPolo, thanks for the ping. Yes, we rely on mocknet and I was hoping that in memory transport is gonna replace the need for it.

I guess the simulated network has even more capabilities with bandwidth/latency tunables as its a lower layer construct, but I wouldnt say this is something we need atm. Although, definetly useful

Afaiu, the synctest would help with flakes for both.

MarcoPolo:
> Afaiu, the synctest would help with flakes for both.

Correct

MarcoPolo:
I've been playing with this a lot, and I think it is quite useful. Worth merging

p2p/net/simconn/libp2p/libp2p.go
```diff
@@ -0,0 +1,258 @@
...
+		}))
+	return libp2p.QUICReuse(
+		func(l fx.Lifecycle, statelessResetKey quic.StatelessResetKey, tokenKey quic.TokenGeneratorKey, opts ...quicreuse.Option) (*quicreuse.ConnManager, error) {
+			cm, err := quicreuse.NewConnManager(statelessResetKey, tokenKey, opts...)
+			if err != nil {
+				return nil, err
+			}
+			l.Append(fx.StopHook(func() error {
+				// When we pass in our own conn manager, we need to close it manually (??)
+				// TODO: this seems like a bug
```
sukunrt: Is this true? Can you open an issue?

p2p/net/simconn/libp2p/libp2p.go
```diff
@@ -0,0 +1,258 @@
...
+const OneMbps = 1_000_000
+
+func QUICSimConnSimpleNet(router *simconn.SimpleSimNet, linkSettings simconn.NodeBiDiLinkSettings, quicReuseOpts ...quicreuse.Option) libp2p.Option {
+	m := &MockSourceIPSelector{}
+	quicReuseOpts = append(quicReuseOpts,
+		quicreuse.OverrideSourceIPSelector(func() (quicreuse.SourceIPSelector, error) {
+			return m, nil
+		}),
+		quicreuse.OverrideListenUDP(func(network string, address *net.UDPAddr) (net.PacketConn, error) {
+			m.ip.Store(&address.IP)
```
sukunrt: Does this mean we will only dial out from the last Listened IP?

p2p/net/simconn/libp2p/libp2p.go
```diff
@@ -0,0 +1,258 @@
...
+					router:          nw,
+					linkSettings:    l.LinkSettings,
+					quicReuseOpts:   quicReuseOpts,
+					ConnMgr:         opts.ConnMgr,
+				})
+			} else {
+				h, err = libp2p.New(
+					libp2p.ListenAddrStrings(addr),
+					QUICSimConnSimpleNet(nw, l.LinkSettings, quicReuseOpts...),
+					libp2p.DisableIdentifyAddressDiscovery(),
```
sukunrt: Why do we need to disable address discovery?

p2p/net/simconn/router.go
```diff
@@ -12,18 +13,101 @@ type PacketReciever interface {
...
+
+	var k ipPortKey
+	if err := k.FromNetAddr(addr); err != nil {
+		return err
+	}
+	m.nodes[k] = v
+	return nil
+}
+
+func (m *addrMap[V]) Remove(addr net.Addr) error {
```
sukunrt:
```suggestion
func (m *addrMap[V]) Delete(addr net.Addr) error {
```

p2p/net/simconn/simconn.go
```diff
@@ -114,19 +145,31 @@ func (c *SimConn) ReadFrom(p []byte) (n int, addr net.Addr, err error) {
...
 	}
 
 	var pkt Packet
 	if !deadline.IsZero() {
 		select {
 		case pkt = <-c.packetsToRead:
+		case <-c.closedChan:
+			return 0, nil, io.EOF
 		case <-time.After(time.Until(deadline)):
 			return 0, nil, ErrDeadlineExceeded
```
sukunrt: deadline update should be handled in this block too, right? You can change the deadline to a different deadline.

p2p/net/simconn/simlink.go
```diff
@@ -0,0 +1,244 @@
+package simconn
+
+import (
+	"math"
+	"sync"
+	"time"
+)
+
+// A rateLimiter is used by a link to determine how long to wait before sending
+// data given a bandwidth cap.
```
sukunrt: Does this package work for this? https://pkg.go.dev/golang.org/x/time/rate
If not, can you add a comment explaining why?

p2p/transport/quicreuse/options.go
```diff
@@ -24,6 +24,13 @@ func OverrideSourceIPSelector(f func() (SourceIPSelector, error)) Option {
 	}
 }
 
+func WithQlogDir(dir string) Option {
```
sukunrt:
```suggestion
func WithQlogTracerDir(dir string) Option {
```

p2p/net/simconn/libp2p/libp2p.go
MarcoPolo: We likely don't need to. This was probably me reducing the number of goroutines per peer.

go.mod
```diff
@@ -44,7 +44,7 @@ require (
 	github.com/multiformats/go-multibase v0.2.0
 	github.com/multiformats/go-multicodec v0.9.0
 	github.com/multiformats/go-multihash v0.2.3
-	github.com/multiformats/go-multistream v0.6.0
+	github.com/multiformats/go-multistream v0.6.1-0.20250402000044-f8b2fec1e8f0
```
sukunrt: adding this so we remember to update this.

p2p/net/simconn/router.go
```diff
@@ -12,18 +13,101 @@ type PacketReciever interface {
...
+	return nil
+}
+
 // PerfectRouter is a router that has no latency or jitter and can route to
 // every node
 type PerfectRouter struct {
-	mu    sync.Mutex
-	nodes map[net.Addr]PacketReciever
+	nodes addrMap[PacketReciever]
 }
```
sukunrt: Can you document these entities a bit more? In particular adding some hints about how to use these. I'm looking at `Router`, `SimConn`, `Link`, and `SimNet`.

sukunrt (approved):

p2p/net/simconn/libp2p/libp2p.go
MarcoPolo: True. Not sure it's a bug though: https://github.com/libp2p/go-libp2p/blob/master/config/config.go#L412

maybe the lifecycle should be handled here (?)

p2p/net/simconn/libp2p/libp2p.go
MarcoPolo: Yes

p2p/net/simconn/libp2p/libp2p.go
MarcoPolo: Just tried it. We do need to disable it because something in that service stalls it.

Going to leave it disabled for now because there are a lot of changes happening there. I'll leave it as a todo to come back to it later.

PR #3264, opened 2025-04-03
rcmgr: use netip.Prefix as map key instead of string

sukunrt:
We're converting this prefix to string for storing in maps. There's no need to, as netip.prefix is a value type that can be used as a key. 

No functional change expected.


MarcoPolo (approved):
Nice. This was likely a holdover from when I was using the `net` package instead of `netip`. Thank you

guillaumemichel (approved):

Issue #3265, opened 2025-04-03
Introduce Rate Limiting


sukunrt:
We should rate limit new connections and streams created by a host. go-libp2p should rate limit new connections, and streams used by libp2p itself, like identify-push, autonat, etc. Users should be able to configure these limits as well as set rate limits for their protocols. 

This is very similar to what the resource manager does. It limits resource consumption at a point in time. Rate limits limit resource consumption over a specific window. From an implementation perspective we can choose to extend the resource manager with this functionality.

There are two specific benefits I'm interested in:
1. Security benefits of being able to exploit some vulnerability or memory leak as many times as physically possible vs only being able to exploit it a few times a minute. 
2. From a systems perspective, there's a limit beyond which things will not work smoothly, especially in a heterogeneous p2p environment where you're running a bunch of things on the same node.
Note: 2 makes things harder to configure as the limits would necessarily be different depending on the machine the node's run on. These are all the same issues that make the resource manager hard to configure.

As creating a peerID is cheap, we need to rely on limiting by IP subnets. Connection rate limiting can extend the IP specific Conn limit setup in the resource manager. Stream rate limiting can also borrow most of this setup, allowing users to rate limit streams from IPs sharing the same prefix.

Similar to ConnLimits in resource manager we can allow configuring rate limits at three levels
1. Global limit: applicable to all events
2. Subnet Specific rates: events grouped by ip subnet
3. Specific Network Prefix rates: events from specific network prefixes

gerceboss:
@sukunrt , would like to work on this with you. Can you you mention where can we discuss about it?

PR #3266, opened 2025-04-03
identify: rate limit id push protocol

sukunrt:
The rate limits id pushes from peers to one every five second with an allowed burst of 10 pushes. This should be enough for all but malfunctioning and malicious peers. 

We can use the exact same code for autonat, autonatv2, circuit v2, etc. 

Introducing limits to identify separately to get some feedback for #3265. For this PR, I'd like to ignore issues regarding where should this piece of code go, and focus on how specifically it should behave. See the long comment in `rateLimiter.allow` for example. We should probably modify `x/time/rate` for our usecase.

Part of: #3265


p2p/protocol/identify/ratelimit.go
```diff
@@ -0,0 +1,325 @@
+package identify
```
MarcoPolo, line 1: Any reason this is in the identify package?

guillaumemichel (commented):
Rate limiting mechanism looks good overall. Left a few comments

p2p/protocol/identify/ratelimit.go
sukunrt: I'll move this when introducing it for connections.

sukunrt: >For this PR, I'd like to ignore issues regarding where should this piece of code go, and focus on how specifically it should behave.

p2p/protocol/identify/ratelimit.go
```diff
@@ -0,0 +1,325 @@
...
+	if err != nil {
+		return r.globalBucket.Allow()
+	}
+	ipAddr, ok := netip.AddrFromSlice(ip)
+	if !ok {
+		return r.globalBucket.Allow()
+	}
+
+	isWithinNetworkPrefix := false
+	for i, limit := range r.NetworkPrefixLimits {
```
MarcoPolo, line 106: These need to be sorted by most specific to least specific as well, right?

If so, can we set that invariant in init?

p2p/protocol/identify/ratelimit.go
```diff
@@ -0,0 +1,325 @@
...
+			oldest := h.Top()
+			if oldest.Expiry.After(now) {
+				break
+			}
+			h.RemoveTop()
+		}
+	}
+}
+
+// tokenBucket is a *rate.Limiter with a `FullAfter` method.
```
MarcoPolo, line 229: Out of date comment?

p2p/protocol/identify/ratelimit.go
```diff
@@ -0,0 +1,325 @@
...
+	return prefixBucketWithExpiry{}
+}
+
+func (h *bucketHeap) RemoveTop() {
+	if h.Len() > 0 {
+		heap.Pop(h)
+	}
+}
+
+// Upsert updates or inserts the prefix's bucket with bucket.
```
MarcoPolo, line 283: Wrong comment

p2p/protocol/identify/ratelimit.go
```diff
@@ -0,0 +1,325 @@
...
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/network"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+	"golang.org/x/time/rate"
+)
+
+// rateLimit is the configuration for a token bucket rate limiter.
+// The bucket has a capacity of Burst, and is refilled at a rate of RPS tokens per second.
```
MarcoPolo, line 17: What is the initial capacity of the bucket? Let's add this to the comment

MarcoPolo (approved):
Looks good!

p2p/protocol/identify/ratelimit.go
```diff
@@ -0,0 +1,325 @@
...
+	for _, h := range s.ipv6Heaps {
+		for h.Len() > 0 {
+			oldest := h.Top()
+			if oldest.Expiry.After(now) {
+				break
+			}
+			h.RemoveTop()
+		}
+	}
+}
```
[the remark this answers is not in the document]
MarcoPolo: This will possibly allocate more space for s.ipv4Heaps which I don't think we want. Can also cause unecessary GC pressure.

p2p/protocol/identify/ratelimit.go
```diff
@@ -0,0 +1,325 @@
...
+
+// Upsert updates or inserts the prefix's bucket with bucket.
+func (h *bucketHeap) Top() prefixBucketWithExpiry {
+	if h.Len() == 0 {
+		return prefixBucketWithExpiry{}
+	}
+	return h.prefixBucket[0]
+}
```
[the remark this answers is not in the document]
MarcoPolo: That comes with the downside of paying twice for the reshuffling of the heap (log(n)) when you pop and have to push it back.

p2p/protocol/identify/ratelimit.go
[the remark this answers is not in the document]
sukunrt: added a `Expire` method to the heap. That reduces the code duplication.

p2p/protocol/identify/ratelimit.go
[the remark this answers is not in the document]
sukunrt: Added an `Expire` method. So all of this is internal to `bucketHeap` and we don't need to export `RemoveTop` and `Top`.

p2p/protocol/identify/ratelimit.go
```diff
@@ -0,0 +1,325 @@
...
+	*rate.Limiter
+}
+
+// FullAt returns the duration from now after which the tokens will be full.
+func (b *tokenBucket) FullAt(now time.Time) time.Time {
+	tokensNeeded := float64(b.Limiter.Burst()) - b.Limiter.TokensAt(now)
+	refillRate := float64(b.Limiter.Limit())
+	eta := time.Duration((tokensNeeded / refillRate) * float64(time.Second))
+	return now.Add(eta)
+}
```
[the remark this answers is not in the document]
sukunrt: Keeping this as is because this is not exported and it reads simpler. Ideally `rate.Limiter` should have exposed this method.

p2p/protocol/identify/ratelimit.go
sukunrt: That's better. Moved to `p2p/internal/rate`

p2p/protocol/identify/ratelimit.go
sukunrt: yes. Fixed.

p2p/protocol/identify/ratelimit.go
sukunrt: It's equal to `Burst`. Added to comment.

p2p/protocol/identify/ratelimit.go
```diff
@@ -0,0 +1,325 @@
...
+		if limit.Prefix.Contains(ipAddr) {
+			if !r.networkPrefixBuckets[i].Allow() {
+				return false
+			}
+			isWithinNetworkPrefix = true
+		}
+	}
+	if isWithinNetworkPrefix {
+		return true
+	}
```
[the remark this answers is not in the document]
sukunrt: Added a comment.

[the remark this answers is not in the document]
sukunrt: This was also a bug. We must sort the prefix list from most to least generic.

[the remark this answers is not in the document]
sukunrt: Added a test too.

Issue #3248, opened 2025-03-24
Enable (more?) lint rules in ci


sukunrt:
Some simple lint rules in ci will be helpful in catching bugs.
For example: Unused function params would have caught this bug: https://github.com/libp2p/go-libp2p/pull/3247

We can start with a small set of checks and increase the list incrementally.

Issue #1843, opened 2022-10-30
Lint with golangci-lint


estensen:
Running [golangci-lint](https://github.com/golangci/golangci-lint) on my machine I find quite a few improvements. Would it be an idea to add the linter to CI and gradually remove warnings? I can add an Action for it.

marten-seemann:
We’re using our Unified CI setup. Using golangci-lint was discussed and we decided against it.

MarcoPolo:
I looked and I think I found these two references:

https://github.com/protocol/.github/issues/41#issuecomment-812623506

and

https://github.com/protocol/.github/issues/6#issuecomment-759220140 

it seems like we don’t want everything from golangci-lint since sometimes it can be annoying. And it’s may lag behind releases on the underlying linters.

marten-seemann:
There’s also https://github.com/ipld/ipld/pull/92, which contains a lot of discussion. 

In addition to not being up to date, golangci-lint also sometimes breaks the underlying linters. See https://twitter.com/dominikhonnef/status/1530695446007521280 for an example.

PR #3269, opened 2025-04-07
ci: add golangci-lint for linting

sukunrt:
Configures ci to use golang-ci for linting. 

The workflow is copied from quic-go. The os specific checks don't add too much value, I guess. I am fine with keeping only linux. 

I couldn't get the unparam linter to detect unused parameters in functions. I guess it is aggressive about not raising failse positives. Using revive for unparam check. 

I fear errcheck may be too much some times. We can disable / configure that one when we come across a change that warrants it. 

See: https://github.com/libp2p/go-libp2p/actions/runs/14307422981/job/40094362132?pr=3269
for a failure. 

See: #1843 for past discussion on this.  

Fixes: #3248

### DIFF
--- a/.github/workflows/go-check.yml
+++ b/.github/workflows/go-check.yml
@@ -19,3 +19,42 @@ jobs:
     with:
       go-version: "1.24.x"
       go-generate-ignore-protoc-version-comments: true
+  
+  golangci-lint:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        go: [  "1.23.x", "1.24.x" ]
+    env:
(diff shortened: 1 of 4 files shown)


guillaumemichel (commented):
@galargh any reason `golangci-lint` isn't part of unified ci?

Other repos could also benefit from it

MarcoPolo (approved):

galargh:
> @galargh any reason golangci-lint isn't part of unified ci?

The current plan is to add it as an optional additional step that could be easily enabled via Unified CI's config. Once it's out, I'll make sure to update the configuration here.

PR #3274, opened 2025-04-24
fix(pstoreds): Correct cyclic batch threshold and handle peer ID decoding errors

tomasandroil:
1.  **Fix cyclic batch threshold initialization:** 
In `newCyclicBatch`, the provided `threshold` parameter was not being assigned to the struct field, causing the batch to potentially commit on every operation instead of respecting the threshold. This change correctly assigns the `threshold`.

2.  **Handle errors during peer ID decoding:** 
In `uniquePeerIds`, errors returned by `base32.RawStdEncoding.DecodeString` and `peer.IDFromBytes` were ignored using `_`. This could lead to panics or inclusion of invalid peer IDs if corrupted data is encountered in the datastore. This change adds proper error checking and skips invalid entries, logging a debug message.

### DIFF
--- a/p2p/host/peerstore/pstoreds/cyclic_batch.go
+++ b/p2p/host/peerstore/pstoreds/cyclic_batch.go
@@ -28,7 +28,7 @@ func newCyclicBatch(ds ds.Batching, threshold int) (ds.Batch, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &cyclicBatch{Batch: batch, ds: ds}, nil
+	return &cyclicBatch{Batch: batch, ds: ds, threshold: threshold}, nil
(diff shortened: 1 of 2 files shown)


tomasandroil:
Hi @MarcoPolo!
Quick reminder about PR #3274 (“fix(pstoreds): Fix cyclic batch threshold handling and add error checking for peer ID decoding”):

- Correctly initialize `cyclicBatch` threshold to avoid premature commits  
- Add error handling in `uniquePeerIds` to skip invalid peer IDs instead of panicking  
- New unit tests for batch threshold behavior and invalid-peer-ID cases  

Let me know if you need anything else! 🔧

PR #3275, opened 2025-04-25
fix: Ensure tcpGatedMaListener wrapping on both metrics and non-metri…

GarmashAlex:
…cs paths

This commit fixes the TCP listener wrapping issue identified in the TODO comment.
Previously, the tcpGatedMaListener was only being applied when metrics were
enabled, but it should be applied consistently on both paths.

The fix ensures that the TCP-specific configurations (such as linger and keepalive
settings) are always applied regardless of whether metrics are enabled, by:
1. Creating a tcpGatedMaListener wrapper first
2. Conditionally adding metrics tracing only when enabled
3. Using the properly wrapped listener in both code paths

This change maintains the same functionality but makes behavior consistent 
between metrics-enabled and metrics-disabled configurations, resolving the TODO.

### DIFF
--- a/p2p/transport/tcp/tcp.go
+++ b/p2p/transport/tcp/tcp.go
@@ -331,11 +331,16 @@ func (t *TcpTransport) Listen(laddr ma.Multiaddr) (transport.Listener, error) {
 		list = t.upgrader.GateMaListener(mal)
 	}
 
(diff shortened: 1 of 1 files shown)


sukunrt (approved):

sukunrt:
Thanks @GarmashAlex. Apologies for the very late review.

PR #3279, opened 2025-05-02
quicreuse: add some documentation for the package

sukunrt:
This adds some helpful documentation for the package and explicitly specifies the reuse and association logic for `ListenQUICAndAssociate` and `DialQUIC`. 

There's a minor refactoring where I've removed a redundant packet conn which is supposed to always be `quic.Transport.Conn`

I'd prefer renaming `ConnManager` to `SharedQUIC` which is more suggestive of the functionality, but that's for later.

### DIFF
--- a/p2p/transport/quicreuse/connmgr.go
+++ b/p2p/transport/quicreuse/connmgr.go
@@ -1,3 +1,7 @@
+// Package quicreuse provides `quicreuse.ConnManager`, which provides functionality
+// for reusing QUIC transports for various purposes, like listening & dialing, having
+// multiple QUIC listeners on the same address with different ALPNs, and sharing the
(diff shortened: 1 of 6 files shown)


p2p/transport/quicreuse/connmgr.go
```diff
@@ -1,3 +1,7 @@
+// Package quicreuse provides `quicreuse.ConnManager`, which provides functionality
+// for reusing QUIC tranports for listening & dialing, having multiple QUIC listeners
+// on the same address with differetn ALPNs, and sharing the same address with non QUIC
```
guillaumemichel, line 3:
```suggestion
// on the same address with different ALPNs, and sharing the same address with non QUIC
```

guillaumemichel (approved):

PR #3281, opened 2025-05-04
Make go-libp2p compile to WASM

talentlessguy:
Makes go-libp2p compile by omitting pion/webrtc which explicitly does not support WASM for certain APIs.

Additionally

- TCP and QUIC transports are disabled for WASM in default settings, as they are not supported by browsers, leaving only WebSocket and WebRTC
- Some utilities from webrtc that are still used elsewhere are moved into *_shared.go files

I locally tested and verified that both linux and wasm targets build properly

This is a more basic approach to https://github.com/libp2p/go-libp2p/pull/3280, whose goal is to bring WASM support to WebRTC, but it's blocked by upstream dependencies


MarcoPolo:
Thank you! I think this would be great to merge. For reference, did you see https://github.com/libp2p/go-libp2p/pull/2191?

I'll try to review this this week.

defaults_js.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	ws "github.com/libp2p/go-libp2p/p2p/transport/websocket"
+	webtransport "github.com/libp2p/go-libp2p/p2p/transport/webtransport"
+)
+
+// Only WebSocket and WebTransport are supported in the browser.
+var DefaultTransports = ChainOptions(
+	Transport(ws.New),
+	Transport(webtransport.New),
```
MarcoPolo, line 14: I assume you didn't actually test these, correct?

MarcoPolo:
As a rough guideline, I think this would be a good way to merge Wasm support:

1. Make changes to be able to compile go-libp2 without any transport.
2. Add transports support in one by one.
2a. Extend libp2p's transport interop tests to test the transport. (I can help with the specifics here)

talentlessguy:
> Thank you! I think this would be great to merge. For reference, did you see #2191?
> 
> I'll try to review this this week.

I didn't know about this PR, neat! Since then a lot of things have progressed, browsers support WebTransport much better now and Go has better WASM interop

MarcoPolo:
> > As a rough guideline, I think this would be a good way to merge Wasm support:
> > ```
> > 1. Make changes to be able to compile go-libp2 without any transport.
> > ```
> 
> Do you mean to not include any transports by default for wasm target?

Yep. Just to get everything else to compile.

defaults_js.go
talentlessguy: I made the list empty by default for WASM

talentlessguy (commented):
Explicitly disabling wasm compilation for some unsupported APIs managed to remove 6 megabytes from the binary, mainly due to exclusion of quic-go

pasha1980:
@MarcoPolo Any update on that? I assume it's on hold?

MarcoPolo:
I think the core go-libp2p can compile to wasm. It's more about building a useful transport within wasm.

Some work here: https://github.com/libp2p/go-libp2p/pull/3386 that facilitates building libp2p hosts without needing packages that can't compile to wasm.

v1rtl:
> I think the core go-libp2p can compile to wasm. It's more about building a useful transport within wasm.
> 
> Some work here: #3386 that facilitates building libp2p hosts without needing packages that can't compile to wasm.

it can't last time i checked. fails because of transport libraries that get imported as a side effect. I still have to patch stuff in go-lib2p to make it compile without trying to pull non-wasm-compatible stuff

pancsta:
Id be very interested in this - currently working with WebSockets, WebWorkers, MessageChannels, and TCP relays. WS and worker ports are represented as plain `net.Conn`. You can take the same approach as I did and inject an existing connection (tcpreuse transport?). I also have a libp2p-pubsub integration and would love to bring it to the browser.

Useful links
- [working demo](https://asyncmachine.dev/wasm-workflow)
- [WS-netconn bits](https://github.com/pancsta/asyncmachine-go/blob/3cbfb6da8e569b8064e2637b7420746ef46ccd6c/pkg/rpc/rpc_server.go#L387)
- [WS-to-TCP relay](https://github.com/pancsta/asyncmachine-go/tree/main/tools/relay)

PR #3285, opened 2025-05-07
WebTransport WASM support

talentlessguy:
Implements WebAssembly support for the WebTranport transport based on https://github.com/libp2p/go-libp2p/pull/2191

I've also added an example to demonstrate connecting to a remote libp2p WT node, it works in Chrome


p-shahi:
@talentlessguy - Thank you for this, this is really cool! To get this merged, a few things that would give more confidence here is to add more tests here and add support for this in the interop test runner (libp2p/test-plans). cc @MarcoPolo @sukunrt who will review when time permits

MarcoPolo:
I can help add to the interop test runner too. (Maybe we can pair on it?)

v1rtl:
I'm not sure how to run interoperability tests from that repo. can someone from the team help?

also FYI it works just fine in chrome but on FF it fails during dial I think. Seems like FF is missing certain APIs

p-shahi:
@talentlessguy are you able to join this call tomorrow https://lu.ma/oel95994 or in the future? We can discuss this and other PRs

PR #3286, opened 2025-05-07
eventbus: change slow consumer event from error to warn

guillaumemichel:
Change event bus slow consumer event logging from `ERROR` to `WARN`.

`WARN` seems more appropriate since this isn't an error.

Closes https://github.com/ipfs/kubo/issues/10778


MarcoPolo (approved):
Maybe we should also increase the `slowConsumerWarningTimeout` time form 1s to 5s(?)

Issue #3289, opened 2025-05-15
move away from `benjohnson/clock`


sukunrt:
It's archived. We can use 
`filecoin-project/clock`

              Let's do this in a separate PR. 
Separately, should we ask them to move this to the libp2p org?

_Originally posted by @sukunrt in https://github.com/libp2p/go-libp2p/pull/3231#discussion_r2054497297_

guillaumemichel:
What would be the reason for moving the repo to the libp2p org?

guillaumemichel:
See https://github.com/libp2p/go-libp2p/pull/3293#issuecomment-2913913207

sukunrt:
Let's keep this as it is for now. It has been working fine for us. Let's move it to the libp2p org when we need to make a change to the clock package.

MarcoPolo:
Sounds good. The next Go version may also have synctest which will change how we fake time.

PR #3293, opened 2025-05-21
go-clock migration

guillaumemichel:
## Context
* [`github.com/benbjohnson/clock`](https://github.com/benbjohnson/clock) has been archived 2 years ago
* Migrate to the maintained fork [`github.com/filecoin-project/go-clock`](https://github.com/filecoin-project/go-clock)
* Fixes https://github.com/libp2p/go-libp2p/issues/3289

## Depends on
* [x] https://github.com/libp2p/go-flow-metrics/pull/36

## Notes
This PR is only removing direct dependencies. Some indirect dependency on `github.com/benbjohnson/clock` are remaining, but they are trickier to remove, since some are circular dependencies (e.g `go-libp2p-pubsub` depends on `go-libp2p/net/p2p/connmgr` [here](https://github.com/libp2p/go-libp2p-pubsub/blob/9e5145fb29c9df968bbec842fcb4cbab64f47b7f/gossipsub_connmgr_test.go#L49-L52)).

I suggest we migrate the dependency iteratively, first remove direct dependencies from `go-libp2p`, after the release migrate libraries to use the latest `go-libp2p` and `filecoin-project/go-clock`, and finally update `go-libp2p` to use the latest libs depending on its latest version.


sukunrt (approved):

MarcoPolo:
I previously thought this only affected our tests. Now I realize this affects libp2p users using the connmgr. This means downstream users now must add `filecoin-project/go-clock` as a dependency. I don't think this is good.

I suggest we:
1. Revert this change.
2. Make a new package in `core/clock` that has the Clock interface we [are already using](https://github.com/libp2p/go-libp2p/blob/853011ae7349d892dc8271ddc74355e13b19608e/p2p/host/autorelay/options.go#L165-L177) elsewhere that fixes some [subtle footguns](https://github.com/libp2p/go-libp2p/blob/853011ae7349d892dc8271ddc74355e13b19608e/core/test/mockclock.go#L53-L60) in the go-clock library .
3. Update our usages of go-clock to use that new package and consolidate our existing usages.

cc @guillaumemichel

guillaumemichel:
> downstream users now must add filecoin-project/go-clock as a dependency. I don't think this is good.

Why? They already depended on [`github.com/benbjohnson/clock`](https://github.com/benbjohnson/clock), so this doesn't add a new dependency.

> Update our usages of go-clock to use that new package and consolidate our existing usages.

I am not familiar with the go-libp2p Clock interface, but you say it has feature parity for our use cases with [`github.com/benbjohnson/clock`](https://github.com/benbjohnson/clock)?

MarcoPolo:
> Why? They already depended on [github.com/benbjohnson/clock](https://github.com/benbjohnson/clock), so this doesn't add a new dependency.

Even if it was simply a 1:1 copy, it changes the imports. It's effectively a new dependency. There also may be users that don't want to import from the filecoin-project org.

> I am not familiar with the go-libp2p Clock interface, but you say it has feature parity for our use cases with [github.com/benbjohnson/clock](https://github.com/benbjohnson/clock)?

It should, or, if not, should be very easy to add missing methods.

MarcoPolo:
Note that I think this is a blocker for the next release.

PR #3305, opened 2025-06-09
autonatv2: add Unknown addrs to event

sukunrt:
Providing unknown addresses to the user, helps them infer whether autonatv2 has any more addresses it might confirm.

There's no upside to not sending this information.

This also changes `host.ConfirmedAddrs` to return all the three categories, Reachable, Unreachable, and Unknown addrs.


sukunrt:
The deadlock is tracked here: https://github.com/quic-go/quic-go/pull/5220

PR #3310, opened 2025-06-12
mocknet: notify listeners on listen

sukunrt:
This is causing a regression in kubo(https://github.com/ipfs/kubo/pull/10835#pullrequestreview-2914948501) with the new addrs manager which depends on being notified of new listen addresses instead of recomputing all the host's addresses every time on every `host.Addrs` call.

### DIFF
--- a/p2p/net/mock/mock_peernet.go
+++ b/p2p/net/mock/mock_peernet.go
@@ -361,6 +361,11 @@ func (pn *peernet) BandwidthTotals() (in uint64, out uint64) {
 // Listen tells the network to start listening on given multiaddrs.
 func (pn *peernet) Listen(addrs ...ma.Multiaddr) error {
 	pn.Peerstore().AddAddrs(pn.LocalPeer(), addrs, peerstore.PermanentAddrTTL)
+	for _, a := range addrs {
+		pn.notifyAll(func(n network.Notifiee) {
+			n.Listen(pn, a)
+		})
+	}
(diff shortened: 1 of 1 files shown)


p-shahi (approved):

p-shahi:
👀 Why is the Tag Push Checker workflow running & failing on this PR @galargh

galargh:
> 👀 Why is the Tag Push Checker workflow running & failing on this PR @galargh

Thanks for the ping! I looked into this. The tag push workflow was failing because we were providing wrong arguments to the `gh issues create` in that workflow. We were providing assignees argument prefixed with `@` which is not needed and using a null value for the repo argument. I fixed both of these issues in [v1.0.25](https://github.com/ipdxco/unified-github-workflows/releases/tag/v1.0.25) release of unified CI and rerun the workflow which now correctly created https://github.com/libp2p/go-libp2p/issues/3311

The reason why that run showed up in this pull request is because the manually created tag - https://github.com/libp2p/go-libp2p/tree/v0.42.0-alpha-2 - is associated with https://github.com/libp2p/go-libp2p/pull/3310/commits/0cf29bdaf0d28de8c85489d1d483e09b5cde4536, which is currently the HEAD commit of this PR. The workflow itself is triggered on a tag push event, but if that tag happens to be for a commit that is the HEAD of a PR, then the status checks from that workflow will appear in the PR as well.

PR #3320, opened 2025-06-20
refactor(webtransport): Use keygen package for deterministic ecdsa key

MarcoPolo:
We can finally get rid of this ugly hack for two reasons:
- Go 1.24 supports deterministic ecdsa signatures
- We can rely on https://pkg.go.dev/filippo.io/keygen#ECDSA to do the key generation.

Since this PR requires Go 1.24, we can't merge until Go 1.25 is released. We also likely want to wait for the keygen package to release v1.0.0 to avoid churn in its key generation.


MarcoPolo:
rebased on top of master. This is ready for review and merge @sukunrt

PR #3323, opened 2025-06-27
quic: upgrade quic-go to v0.53

sukunrt:
quic-go v0.53 has changed its api to return concrete structs and not interfaces. Only the WebTransport Transport requires some special handling, rest of the changes are straightforward.


p2p/transport/quic/stream.go
```diff
@@ -14,7 +14,7 @@ const (
 )
 
 type stream struct {
-	quic.Stream
+	*quic.Stream
```
MarcoPolo, line 17: Can this be `quic.Stream` instead of a pointer? We already use this as a pointer everywhere. Using this as a value avoids the double pointer lookup.

p2p/transport/webtransport/stream.go
```diff
@@ -29,7 +29,7 @@ func (s *webtransportStream) RemoteAddr() net.Addr {
 }
 
 type stream struct {
-	webtransport.Stream
+	*webtransport.Stream
```
MarcoPolo, line 32: same comment here

p2p/transport/webtransport/listener.go
```diff
@@ -125,13 +85,61 @@ func newListener(reuseListener quicreuse.Listener, t *transport, isStaticTLSConf
...
+			ln.startHandshake(conn)
+			go ln.server.ServeQUICConn(conn)
 		}
 	}()
 	return ln, nil
 }
 
+func (l *listener) startHandshake(conn *quic.Conn) {
+	ctx, cancel := context.WithTimeout(l.ctx, handshakeTimeout)
+	l.mx.Lock()
```
MarcoPolo: Can we move the lock to below the context.AfterFunc.

It's fine as is unless we ctx implements its own `AfterFunc` that doesn't use a separate goroutine, in which case it would deadlock.

sukunrt: It needs an extra context.Err check but that seems more idiomatic.

p2p/transport/quic/stream.go
sukunrt: Fixing it the other way round. Using stream and not *stream. We can't copy quic.Stream as it has a mutex.

PR #3327, opened 2025-07-04
fix: improve resource management and monitoring

pranavkonde:
# Resource Management Improvements

## Problem
- Memory leaks in resource manager error handling paths
- Negative resource counts being silently corrected
- Static file descriptor limits not system-aware
- Insufficient resource monitoring

## Changes
1. Fixed memory leaks by adding proper cleanup in error paths
2. Added validation to prevent negative resource counts
3. Made file descriptor limits dynamically configurable
4. Added ResourceMonitor for usage tracking and early warning
5. Implemented comprehensive resource limit validation

### DIFF
--- a/p2p/host/resource-manager/limit.go
+++ b/p2p/host/resource-manager/limit.go
@@ -11,6 +11,8 @@ package rcmgr
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


sukunrt:
This PR is too large. Can you break it into separate changes preferably with a corresponding issue that this aims to solve?

PR #3329, opened 2025-07-05
Fix/protocol diff complexity

### LINKED ISSUE #3328: Performance Optimization: Improve Protocol Diff Algorithm Complexity
# Optimize protocol diff implementation from O(n²) to O(n)

## Problem
The current `diff()` function in `p2p/protocol/identify/id.go` uses nested loops with O(n²) complexity, which doesn't scale well for larger protocol lists.

## Solution
Replace with map-based approach for O(n) complexity:

```go
func diff(a, b []protocol.ID) (added, removed []protocol.ID) {
    seen := make(map[protocol.ID]struct{}, len(a))
    
    // Mark all protocols in 'a'
    for _, x := range a {
        seen[x] = struct{}{}
    }
    
    // Find added protocols and remove existing ones from map
    for _, x := range b {
        if _, exists := seen[x]; !exists {
            added = append(added, x)
        }
        delete(seen, x)
    }
    
    // Remaining items are removed protocols
    for x := range seen {
        removed = append(removed, x)
    }
    
    return added, removed
}
```

## Benefits
- **Performance**: O(n²) → O(n) time complexity
- **Scalability**: Better performance as protocol lists grow
- **Memory**: Uses empty struct{} for minimal memory overhead

pranavkonde:
## Description
Fixes #3328

Optimizes the protocol diff implementation in `identify/id.go` from O(n²) to O(n) by using a map-based approach instead of nested loops.

## Changes
```go
// Before: O(n²)
for _, x := range b {
    for _, y := range a {
        if x == y {
            found = true
```

```go
// After: O(n)
seen := make(map[protocol.ID]struct{}, len(a))
for _, x := range a {
    seen[x] = struct{}{}
}

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -303,6 +303,7 @@ func (ids *idService) loop(ctx context.Context) {
 	// That way, we can end up with
 	// * this Go routine busy looping over all peers in sendPushes
 	// * another push being queued in the triggerPush channel
+	// and we'll only send one push at a time.
 	triggerPush := make(chan struct{}, 1)
 	ids.refCount.Add(1)
 	go func() {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: p2p/protocol/identify/id.go [hunk 1]
```diff
@@ -753,31 +754,28 @@ func (ids *idService) getSignedRecord(snapshot *identifySnapshot) []byte {
 
 // diff takes two slices of strings (a and b) and computes which elements were added and removed in b
 func diff(a, b []protocol.ID) (added, removed []protocol.ID) {
```
sukunrt, line 756: Can you do sort -> unique -> diff? 
That'll make it not alloc for the map.

marten-seemann:
Just documenting it here, the requested benchmark was still not provided: https://github.com/libp2p/go-libp2p/issues/3328#issuecomment-3039733306.

I'm highly skeptical that this PR provides the benefits promised.

vyzo:
this is AI slop.

PR #3330, opened 2025-07-07
swarm: move AddCertHashes to swarm

sukunrt:
The simplified idea here is: if we can listen on the addr, we should try adding certhashes to the addr. 
This removes dependency on WebTransport and WebRTC Transports.

### DIFF
--- a/p2p/net/swarm/swarm.go
+++ b/p2p/net/swarm/swarm.go
@@ -961,3 +961,31 @@ func (r ResolverFromMaDNS) ResolveDNSComponent(ctx context.Context, maddr ma.Mul
 	}
 	return addrs, nil
 }
+
+// AddCertHashes adds certificate hashes to relevant transport addresses, if there
+// are no certhashes already present on the method. It mutates `listenAddrs`.
+// This method is useful for adding certhashes to public addresses discovered
+// via identify, nat mapping, or provided by the user.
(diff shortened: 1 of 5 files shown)


p2p/net/swarm/swarm.go
```diff
@@ -961,3 +961,31 @@ func (r ResolverFromMaDNS) ResolveDNSComponent(ctx context.Context, maddr ma.Mul
 	}
 	return addrs, nil
 }
+
+// AddCertHashes adds certificate hashes to relevant transport addresses, if there
+// are no certhashes already present on the method. It mutates `listenAddrs`.
+// This method is useful for adding certhashes to public addresses discovered
+// via identify, nat mapping, or provided by the user.
```
MarcoPolo, line 968: This comment needs some editing. Maybe:

```suggestion
// AddCertHashes mutates listenAddrs. It adds applicable certificate hashes to
// the multiaddrs. It will not modify a multiaddr if it already contains a
// certificate hash. This method is useful for adding certhashes to public
// addresses discovered via identify, nat mapping, or provided by the user.
```

MarcoPolo (approved):

p2p/net/swarm/swarm.go
sukunrt: Missed this. I'll open a separate PR.

Issue #2941, opened 2024-08-29
Failure to meet activation threshold for observed address without DHT client 


2color:
## Preface 

This is not a bug per se, but an issue that arises in specific circumstances. 

TL;DR:
- To start hole punching you need to determine your public IP, which generally requires other peers if you're behind NAT
- We determine our own observed address by running the `identify` protocol against peers
- We "activate" an observed address once 4 peers return tell us they've observed an address. This reduces the trust assumptions on any particular peer and provides some level of security
  https://github.com/libp2p/go-libp2p/blob/3deee921046b2b184d93b4ca64c56d85df1e197a/p2p/protocol/identify/obsaddr.go#L17-L22
- The activation of an observed address is transport and IP specific. That means that with QUIC and TCP enabled with both IPv4 and IPv6 addrs, you will need 16 connections for the 4 addresses to be activated. By corollary, as the number of transports and IPs increases, more connections are necessary to activate a given addr. (Can you confirm this @sukunrt)
- When you are behind NAT, don't start a DHT client, and create a new host with the default configuration, there's a good chance that you won't be able to hole punch into other peers behind NAT, because you don't have any activated public addresses.
- Even if you connect to the 4 libp2p bootstrap nodes (without starting a DHT client), depending on the transports over it connects, none of your public addrs will activate. 

## Background

[ipfs-check](https://github.com/ipfs/ipfs-check), is a retrievability diagnostics tool. You provide it with a CID and a multiaddr and it checks —amongst other things— whether the peer is dialable and whether it has the block of the CID. This is especially useful to determine whether a peer behind NAT is actually "hole-punchable", which you can only know by trying to connect to.

The tool creates a short-lived libp2p test host for each check, which connects to one or two peers (two if it needs circuit relay reservation if the other peer is behind NAT), however hole punching will never kick off and it will hang on the following line:

```
2024-08-29T13:40:04.468+0200	DEBUG	p2p-holepunch	holepunch/svc.go:98	waiting until we have at least one public addresspeer12D....`
```

I was able to work around this by adding a DHT client to the test host, however, it's not clear whether this is ideal. There are some situations (like resource constrained environments) where you don't want the overhead of increased connections and bandwidth associated with a DHT clients.

MarcoPolo:
> By corollary, as the number of transports and IPs increases, more connections are necessary to activate a given addr.

[...]

PR #3332, opened 2025-07-09
basichost: move observed address manager to basichost

sukunrt:
Changes in the PR:

- Removes normalization of addresses. Mapping only thinwaist addresses removes the need for normalization.
- If the NATDevice reports a private IP mapping, we add observed addrs mappings. Previously we tried to keep the port from NAT and use the IP from observed address. There's no reason to do this. 
- Removes localAddr from observedAddrsManager. The object can work fine without it.
- There was some special case logic when checking whether any NATType Changed events were pending. Previously we checked if we ran identify in the last minute, and only then computed the new nat type. This removes this special case and checks for nat type change every one minute. It isn't expensive.
- Change to how we calculate addresses for hole punching. We lower the activation threshold to check for addresses for hole punching. This fixes: #2941

### DIFF
--- a/p2p/host/basic/addrs_manager.go
+++ b/p2p/host/basic/addrs_manager.go
@@ -161,33 +161,27 @@ func (a *addrsManager) updateAddrsSync() {
 	}
 }
 
-func closeIfError(err error, closer io.Closer, name string) error {
-	if err != nil {
-		err1 := closer.Close()
-		if err1 != nil {
-			err1 = fmt.Errorf("error closing %s: %w", name, err1)
-		}
-		return errors.Join(err, err1)
-	}
-	return nil
-}
-
 func (a *addrsManager) startBackgroundWorker() (retErr error) {
-	autoRelayAddrsSub, err := a.bus.Subscribe(new(event.EvtAutoRelayAddrsUpdated), eventbus.Name("addrs-manager"))
+	autoRelayAddrsSub, err := a.bus.Subscribe(new(event.EvtAutoRelayAddrsUpdated), eventbus.Name("addrs-manager autorelay sub"))
 	if err != nil {
 		return fmt.Errorf("error subscribing to auto relay addrs: %s", err)
 	}
-	defer func() { retErr = closeIfError(retErr, autoRelayAddrsSub, "autorelay subscription") }()
-
-	autonatReachabilitySub, err := a.bus.Subscribe(new(event.EvtLocalReachabilityChanged), eventbus.Name("addrs-manager"))
+	mc := multiCloser{autoRelayAddrsSub}
+	autonatReachabilitySub, err := a.bus.Subscribe(new(event.EvtLocalReachabilityChanged), eventbus.Name("addrs-manager autonatv1 sub"))
 	if err != nil {
-		return fmt.Errorf("error subscribing to autonat reachability: %s", err)
+		return errors.Join(
+			fmt.Errorf("error subscribing to autonat reachability: %s", err),
+			mc.Close(),
+		)
 	}
-	defer func() { retErr = closeIfError(retErr, autonatReachabilitySub, "autonatReachability subscription") }()
+	mc = append(mc, autonatReachabilitySub)
 
 	emitter, err := a.bus.Emitter(new(event.EvtHostReachableAddrsChanged), eventbus.Stateful)
 	if err != nil {
-		return fmt.Errorf("error creating reachability subscriber: %s", err)
+		return errors.Join(
+			fmt.Errorf("error creating reachability subscriber: %s", err),
+			mc.Close(),
+		)
 	}
 
 	var relayAddrs []ma.Multiaddr
@@ -659,4 +653,22 @@ func removeNotInSource(addrs, source []ma.Multiaddr) []ma.Multiaddr {
 		}
 	}
 	return addrs[:i]
+}
+
+type multiCloser []io.Closer
+
+func (mc *multiCloser) Close() error {
+	var errs []error
+	for _, closer := range *mc {
+		if err := closer.Close(); err != nil {
+			var closerName string
+			if named, ok := closer.(interface{ Name() string }); ok {
+				closerName = named.Name()
+			} else {
+				closerName = fmt.Sprintf("%T", closer)
+			}
+			errs = append(errs, fmt.Errorf("error closing %s: %w", closerName, err))
+		}
+	}
+	return errors.Join(errs...)
 }
--- a/p2p/host/observedaddrs/manager.go
+++ b/p2p/host/observedaddrs/manager.go
@@ -167,26 +167,17 @@ func NewManager(eventbus event.Bus, net network.Network) (*Manager, error) {
 		return nil, err
 	}
 
-	// Maybe this should be done in start, but that makes testing without a swarm difficult.
-	nb := &network.NotifyBundle{
-		DisconnectedF: func(_ network.Network, c network.Conn) {
-			o.RemoveConn(c)
-		},
-	}
-	o.stopNotify = func() {
-		net.StopNotify(nb)
-	}
 	return o, nil
 }
 
 // newManagerWithListenAddrs uses the listenAddrs directly to simplify creation in tests.
-func newManagerWithListenAddrs(eventbus event.Bus, listenAddrs func() []ma.Multiaddr) (*Manager, error) {
+func newManagerWithListenAddrs(bus event.Bus, listenAddrs func() []ma.Multiaddr) (*Manager, error) {
 	o := &Manager{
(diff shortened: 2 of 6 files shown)


p2p/host/basic/obsaddr_test.go
```diff
@@ -44,62 +40,62 @@ func newConn(local, remote ma.Multiaddr) *mockConn {
 	return &mockConn{local: local, remote: remote}
 }
 
-func normalize(addr ma.Multiaddr) ma.Multiaddr {
-	for {
-		out, last := ma.SplitLast(addr)
-		if last == nil {
-			return addr
-		}
-		if _, err := last.ValueForProtocol(ma.P_CERTHASH); err != nil {
```
MarcoPolo: Could we just remove the certhash and everything after as the normalize function? We might need to be careful around circuit addresses.

That would remove the dependency on the webtransport/webrtc transports

p2p/host/basic/addrs_manager.go
```diff
@@ -262,6 +334,38 @@ func (a *addrsManager) background(autoRelayAddrsSub, autonatReachabilitySub even
 	}
 }
 
+func (a *addrsManager) observedAddrsWorker(identifySub event.Subscription, natTypeEmitter event.Emitter) {
```
MarcoPolo: As we discussed, we should move this to observed address manager

p2p/host/basic/addrs_manager.go
```diff
@@ -159,34 +208,50 @@ func (a *addrsManager) updateAddrsSync() {
 	}
 }
 
-func (a *addrsManager) startBackgroundWorker() error {
+func closeIfError(err error, closer io.Closer, name string) error {
```
MarcoPolo: Here's an alternative that I think is more idiomatic:

```diff
diff --git a/p2p/host/basic/addrs_manager.go b/p2p/host/basic/addrs_manager.go
index 5c8936d63..eeca76c2c 100644
--- a/p2p/host/basic/addrs_manager.go
+++ b/p2p/host/basic/addrs_manager.go
@@ -200,27 +200,15 @@ func (a *addrsManager) triggerAddrsUpdate() {
 	}
 }
 
-func closeIfError(err error, closer io.Closer, name string) error {
-	if err != nil {
-		err1 := closer.Close()
-		if err1 != nil {
-			err1 = fmt.Errorf("error closing %s: %w", name, err1)
-		}
-		return errors.Join(err, err1)
-	}
-	return nil
-}
-
 func (a *addrsManager) startObservedAddrsWorker() (retErr error) {
 	identifySub, err := a.bus.Subscribe(new(event.EvtPeerIdentificationCompleted), eventbus.Name("addrs-manager"))
 	if err != nil {
 		return fmt.Errorf("error subscribing to autonat reachability: %s", err)
 	}
-	defer func() { retErr = closeIfError(retErr, identifySub, "identify subscription") }()
 
 	natTypeEmitter, err := a.bus.Emitter(new(event.EvtNATDeviceTypeChanged), eventbus.Stateful)
 	if err != nil {
-		return fmt.Errorf("error creating nat type emitter: %s", err)
+		return errors.Join(fmt.Errorf("error creating nat type emitter: %s", err), identifySub.Close())
 	}
 
 	a.wg.Add(1)
@@ -228,22 +216,34 @@ func (a *addrsManager) startObservedAddrsWorker() (retErr error) {
 	return nil
 }
 
+type multiCloser []io.Closer
+
+func (mc *multiCloser) Close() error {
+	var errs []error
+	for _, closer := range *mc {
+		if err := closer.Close(); err != nil {
+			errs = append(errs, fmt.Errorf("error closing %T: %w", closer, err))
+		}
+	}
+	return errors.Join(errs...)
+}
+
 func (a *addrsManager) startBackgroundWorker() (retErr error) {
 	autoRelayAddrsSub, err := a.bus.Subscribe(new(event.EvtAutoRelayAddrsUpdated), eventbus.Name("addrs-manager"))
 	if err != nil {
 		return fmt.Errorf("error subscribing to auto relay addrs: %s", err)
 	}
-	defer func() { retErr = closeIfError(retErr, autoRelayAddrsSub, "autorelay subscription") }()
+	mc := multiCloser{autoRelayAddrsSub}
 
 	autonatReachabilitySub, err := a.bus.Subscribe(new(event.EvtLocalReachabilityChanged), eventbus.Name("addrs-manager"))
 	if err != nil {
-		return fmt.Errorf("error subscribing to autonat reachability: %s", err)
+		return errors.Join(fmt.Errorf("error subscribing to autonat reachability: %s", err), mc.Close())
 	}
-	defer func() { retErr = closeIfError(retErr, autonatReachabilitySub, "autonatReachability subscription") }()
+	mc = append(mc, autonatReachabilitySub)
 
 	emitter, err := a.bus.Emitter(new(event.EvtHostReachableAddrsChanged), eventbus.Stateful)
 	if err != nil {
-		return fmt.Errorf("error creating reachability subscriber: %s", err)
+		return errors.Join(fmt.Errorf("error creating reachability subscriber: %s", err), mc.Close())
 	}
 
 	var relayAddrs []ma.Multiaddr
```

p2p/host/basic/addrs_manager.go
```diff
@@ -114,6 +145,15 @@ func (a *addrsManager) Start() error {
 			return fmt.Errorf("error starting addrs reachability tracker: %s", err)
 		}
 	}
+	if a.observedAddrsManager != nil {
+		a.observedAddrsManager.Start()
```
MarcoPolo: As mentioned in our call, this should take a Notifier interface so the obsaddr manager can register its Notifee. And the obsaddr worker should be move to the obsaddr manager.

p2p/host/basic/obsaddr_test.go
sukunrt: I'll do this in a separate PR. This happens in autonatv2 now.

p2p/host/basic/addrs_manager.go
sukunrt: The type of all subscribers is `event.Subscription` so `%T` will not have the name info.

sukunrt:
I've moved the manager to its own package. This makes the observed addrs manager only depend on swarm and we can inject it in the host.

p2p/host/basic/addrs_manager.go
MarcoPolo: Ah right. That's an easy fix though:

```go
func (mc *multiCloser) Close() error {
	var errs []error
	for _, closer := range *mc {
		if err := closer.Close(); err != nil {
			var closerName string
			if named, ok := closer.(interface{ Name() string }); ok {
				closerName = named.Name()
			} else {
				closerName = fmt.Sprintf("%T", closer)
			}
			errs = append(errs, fmt.Errorf("error closing %s: %w", closerName, err))
		}
	}
	return errors.Join(errs...)
}
```

p2p/protocol/identify/opts.go
```diff
@@ -43,14 +42,6 @@ func WithMetricsTracer(tr MetricsTracer) Option {
 	}
 }
 
-// DisableObservedAddrManager disables the observed address manager. It also
-// effectively disables the nat emitter and EvtNATDeviceTypeChanged
-func DisableObservedAddrManager() Option {
```
MarcoPolo, line 48: Please update the release notes to call this change out.

p2p/host/observedaddrs/manager.go
```diff
@@ -138,112 +128,131 @@ type observation struct {
...
+	}
+
+	// Maybe this should be done in start, but that makes testing without a swarm difficult.
+	nb := &network.NotifyBundle{
+		DisconnectedF: func(_ network.Network, c network.Conn) {
+			o.RemoveConn(c)
+		},
+	}
+	o.stopNotify = func() {
+		net.StopNotify(nb)
```
MarcoPolo: Where did you register net.Notify?

MarcoPolo: And yes, this should happen in Start

p2p/host/observedaddrs/manager.go
```diff
@@ -138,112 +128,131 @@ type observation struct {
...
+		eventbus:            eventbus,
+		stopNotify:          func() {},
 	}
 	o.ctx, o.ctxCancel = context.WithCancel(context.Background())
+	return o, nil
+}
 
-	o.wg.Add(1)
+// Start tracking addrs
+func (o *Manager) Start() {
```
MarcoPolo: This should take network.Network so it can properly register the notifier

MarcoPolo (commented):
Looks good mod a couple of comments.

I think we should change the errdefer pattern to the multiCloser. It's a bit more idiomatic in Go.

Issue #3196, opened 2025-02-16
Proposal: Provide Unwrap an method on `manet.Conn`, `transport.CapableConn`, and `network.Conn`


sukunrt:
Currently wrapping any of these connection types is very error prone. We've run in to interface assertion errors a bunch of times around these connections. 
The last few were with the introduction of the shared tcp listener for tcp and websocket. 

The problem is that any wrapping breaks interface assertions downstream. 
Most recently, when implementing #3186 I ran into it again when the `tcp.tracingListener` rejected the wrapped `manet.Conn` as it didn't expose the tcp metrics tracing methods. Here's the code where this happens: https://github.com/libp2p/go-libp2p/blob/master/p2p/transport/tcp/metrics.go#L292

If we add `Unwrap() Conn` methods to the most used types(`manet.Conn`, `transport.Conn`, `network.Conn`), similar to the `Unwrap() []error` method on error wrappers, we will have an idiomatic way of asserting if any connection in the whole wrapping chain implements the expected method. 

This will still be brittle. It depends on correctly implementing Unwrap always. But it should be better than the situation right now.

MarcoPolo:
I think we can close this as we have conn.As now

PR #3338, opened 2025-07-17
feat(network): Add ConnAs

MarcoPolo:
ConnAs works in a similar way to errors.As. It allows a user to cut through the interface layers and extract a specific type of connection if available.

This serves as a sort of escape hatch to allow users to leverage some connection specific feature without having to support that feature for all connections. Getting RTT information is one example.

It also allows us, within the library, to get specific types of connections out of the interface box. This would have been useful in the recent changes in tcpreuse. See https://github.com/libp2p/go-libp2p/pull/3181 and https://github.com/libp2p/go-libp2p/pull/3142.

Getting access to the underlying type can lead to hard to debug issues. For example, if a user mutates connection state on the underlying type, hooks that relied on only mutating that state from the wrapped connection would never be called.

It is up to the user to ensure they are using this safely.

### DIFF
--- a/core/network/conn.go
+++ b/core/network/conn.go
@@ -87,6 +87,20 @@ type Conn interface {
 	// IsClosed returns whether a connection is fully closed, so it can
 	// be garbage collected.
 	IsClosed() bool
+
+	// As finds the first conn in Conn's wrapped types that matches target, and
+	// if one is found, sets target to that conn value and returns true.
+	// Otherwise, it returns false. Similar to errors.As.
+	//
+	// target must be a pointer to the type you are matching against.
+	//
+	// This is an EXPERIMENTAL API. Getting access to the underlying type can
+	// lead to hard to debug issues. For example, if you mutate connection state
+	// on the underlying type, hooks that relied on only mutating that state
+	// from the wrapped connection would never be called.
+	//
+	// You very likely do not need to use this method.
+	As(target any) bool
 }
 
 // ConnectionState holds information about the connection.
--- a/p2p/net/upgrader/conn.go
+++ b/p2p/net/upgrader/conn.go
@@ -23,6 +23,10 @@ type transportConn struct {
 
 var _ transport.CapableConn = &transportConn{}
 
+func (c *transportConn) As(target any) bool {
+	return c.MuxedConn.As(target)
+}
+
 func (t *transportConn) Transport() transport.Transport {
 	return t.transport
 }
--- a/p2p/transport/quic/conn.go
+++ b/p2p/transport/quic/conn.go
@@ -25,6 +25,15 @@ type conn struct {
 	remoteMultiaddr ma.Multiaddr
 }
 
+func (c *conn) As(target any) bool {
+	if t, ok := target.(**quic.Conn); ok {
+		*t = c.quicConn
+		return true
+	}
+
+	return false
+}
+
 var _ tpt.CapableConn = &conn{}
 
 // Close closes the connection.
--- a/core/network/mux.go
+++ b/core/network/mux.go
@@ -137,6 +137,20 @@ type MuxedConn interface {
 
 	// AcceptStream accepts a stream opened by the other side.
 	AcceptStream() (MuxedStream, error)
+
+	// As finds the first conn in MuxedConn's wrapped types that matches target,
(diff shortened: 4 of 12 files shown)


sukunrt:
How about adding `Unwrap` too?
https://github.com/libp2p/go-libp2p/issues/3196

Though that proposal forces to unwrap to a `manet.Conn` always.

sukunrt:
Alternatively, for your use case I think the better approach is to add `RTT() network.Duration` to the Conn and setting the value really high or negative for transports that don't support this.

Providing RTT seems useful.

marten-seemann:
> Alternatively, for your use case I think the better approach is to add `RTT() network.Duration` to the Conn and setting the value really high or negative for transports that don't support this.
> 
> Providing RTT seems useful.

You probably also want RTT variance, min RTT, etc.

MarcoPolo:
> I think the better approach is to add RTT() network.Duration to the Conn

This is certainly something we can do. It's better in the sense that it doesn't expose the knives to users. But on the other hand sometimes you want the knives.

We could easily add the RTT() method to conn and implement it with ConnAs, so I think we probably still want ConnAs.

sukunrt:
The underlying transport method exposing the RTT might have any signature, like exposing the RTT, RTT variance, minRTT from a single method. Why would you want to implement the RTT method with `ConnAs` as opposed to exposing a RTT method from `transport.CapableConn` which wraps whatever that specific transport does?

MarcoPolo:
> ... as opposed to exposing a RTT method from transport.CapableConn which wraps whatever that specific transport does?

Maybe there's a misunderstanding? I'm saying that even with this method you would still want `ConnAs` so that you can cut through the layers of abstraction and match against a specific transport.

core/network/conn.go
```diff
@@ -141,3 +141,10 @@ type ConnScoper interface {
 	// Scope returns the user view of this connection's resource scope
 	Scope() ConnScope
 }
+
+func ConnAs(c Conn, target any) bool {
+	if x, ok := c.(interface{ As(any) bool }); ok && x.As(target) {
+		return true
+	}
+	return false
+}
```
sukunrt: Can you mark this as experimental and add the warning: 
```
Getting access to the underlying type can lead to hard to debug issues. For example, if a user mutates connection state on the underlying type, hooks that relied on only mutating that state from the wrapped connection would never be called.
```

p2p/net/upgrader/conn.go
```diff
@@ -23,6 +23,13 @@ type transportConn struct {
 
 var _ transport.CapableConn = &transportConn{}
 
+func (c *transportConn) As(target any) bool {
+	if x, ok := c.MuxedConn.(interface{ As(any) bool }); ok && x.As(target) {
+		return true
+	}
+	return false
+}
```
sukunrt: Should we add this method to all the interfaces `MuxedConn` and `CapableConn`, otherwise any wrapping will disable this functionality. This has happened many times before.

sukunrt: In the present state, this wouldn't have helped with: 

>It also allows us, within the library, to get specific types of connections out of the interface box. This would have been useful in the recent changes in tcpreuse. See https://github.com/libp2p/go-libp2p/pull/3181 and https://github.com/libp2p/go-libp2p/pull/3142.

Because the wrapping struct could forget to implement `As`.

p2p/transport/quic/conn.go
```diff
@@ -25,6 +25,15 @@ type conn struct {
 	remoteMultiaddr ma.Multiaddr
 }
 
+func (c *conn) As(target any) bool {
```
sukunrt, line 28: Can you implement these for WebTransport, WebSocket, and WebRTC?

core/network/conn.go
sukunrt: Some documentation with an example would be nice.

p2p/transport/quic/conn.go
MarcoPolo: Skipping websocket, but will do it for webtransport and webrtc

p2p/net/upgrader/conn.go
MarcoPolo: done

sukunrt (approved):

Issue #3340, opened 2025-07-21
Using transport reuse results in an extra listener on top of the original listener instead of replacing it


derrandz:
When using quic reuse to use your own udp listener and transport to intercept raw quic connections, libp2p still creates its own quic transport and doesn't replace it with the one from reuse. This results in having either errors with quic port reported as already being used by the original transport, or it results in having two separate quic addresses with different ports... And that's not what we are trying to achieve with reuse I assume. At least for our use case which inspired the whole quic reuse thing, we should end up with a single quic address that is intercepting raw connections and giving non-raw ones back to libp2p.

You'd think that in order to result in one quic transport, you'd skip giving transport option to the new libp2p instance, but that causes the transport from reuse to not be used and we end up with no quic listening addresses.

derrandz:
Actually now, when I give a listen address with one the same quic port, and I use the re-use feature, libp2p can't even start I run into this bug:

```
2025/07/22 16:00:59 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 7168 kiB, got: 416 kiB). See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes for details.
panic: unexpected chan already found in accept muxer

goroutine 148 [running]:
github.com/libp2p/go-libp2p/p2p/transport/quic.(*acceptLoopRunner).AcceptForVersion(0xc000437580, 0x1)
        /home/gustavo/go/pkg/mod/github.com/libp2p/go-libp2p@v0.42.1-0.20250702212416-7b7c3ed4ce04/p2p/transport/quic/virtuallistener.go:61 +0x125
github.com/libp2p/go-libp2p/p2p/transport/quic.(*transport).Listen(0xc000172480, {0xc000416580, 0x3, 0x4})
        /home/gustavo/go/pkg/mod/github.com/libp2p/go-libp2p@v0.42.1-0.20250702212416-7b7c3ed4ce04/p2p/transport/quic/transport.go:338 +0x595
github.com/libp2p/go-libp2p/p2p/net/swarm.(*Swarm).AddListenAddr(0xc000356a00, {0xc000416580, 0x3, 0x4})
        /home/gustavo/go/pkg/mod/github.com/libp2p/go-libp2p@v0.42.1-0.20250702212416-7b7c3ed4ce04/p2p/net/swarm/swarm_listen.go:116 +0x70
[...]

PR #3347, opened 2025-07-28
fix: remove duplicate listen addresses to avoid muxer panic

derrandz:
Fixes this part of the issue: https://github.com/libp2p/go-libp2p/issues/3340#issuecomment-3103612088

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -475,6 +475,19 @@ func (cfg *Config) validate() error {
 		return errors.New("cannot use shared TCP listener with PSK")
 	}
 
+	// check for duplicate listen addresses and remove them
+	listenAddresses := cfg.ListenAddrs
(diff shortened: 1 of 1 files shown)


MarcoPolo:
It is an error to try to Listen on the same address twice. This is probably not what you want.

derrandz:
hmmmm, what should we do when passed two identical listen addresses? that muxer panic is very ambiguous. maybe we should handle this case? passing two addresses is a user bug ofc but it was disorientating enough to see that panic

MarcoPolo:
a better error would be a welcome change

derrandz:
@MarcoPolo I went with a solution to simply ignore and discard duplicate listen addresses. Wdyt? or shall we return an error on validate?

PR #3354, opened 2025-08-02
chore: fix some typo

chengxilo:
1. In examples/libp2p-host/host.go, the comment incorrectly uses lowercase for w in water; I’ve corrected the case. Additionally, I’ve updated the phrasing to "Low watermark" and "High watermark" for improved clarity.

2. In p2p/net/connmgr.go, the comment refers to the low watermark argument as lo, but the actual argument is named low in the code. I’ve updated the comment to match the code for consistency.


sukunrt:
https://github.com/libp2p/go-libp2p?tab=readme-ov-file#contribute

PR #3355, opened 2025-08-04
basichost: move EvtLocalAddrsChanged to addrs_manager

sukunrt:
We'll deprecate this event, but we still have to keep sending this for a few more releases. More importantly, we need to update the peerstore with the host's addresses and it's better to do this *before* sending update events so that consumers of the event can rely on the host addrs being updated in the peerstore.

### DIFF
--- a/p2p/host/basic/addrs_manager.go
+++ b/p2p/host/basic/addrs_manager.go
@@ -11,8 +11,12 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/libp2p/go-libp2p/core/crypto"
 	"github.com/libp2p/go-libp2p/core/event"
 	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peerstore"
+	"github.com/libp2p/go-libp2p/core/record"
 	"github.com/libp2p/go-libp2p/p2p/host/basic/internal/backoff"
 	"github.com/libp2p/go-libp2p/p2p/host/eventbus"
 	"github.com/libp2p/go-netroute"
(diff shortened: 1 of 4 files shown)


sukunrt:
I've made this change to adding a method on the peerstore. I'm not convinced about it though. 

1. This adds host context to the peerstore, which at the moment can be used by any object, like Gossipsub and kaddht, to store data against peerids.
1. This method must be on peerstore and not on any of the component interfaces like `AddrBook` because we need the Private Key from the `KeyBook`.
1. Not sure if creating a PeerRecord should be with peerstore. At the moment `AddrManager` seems like the best owner for this logic since the only thing in the PeerRecord are the addresses.

core/peerstore/peerstore.go
```diff
@@ -70,6 +70,11 @@ type Peerstore interface {
 	// RemovePeer removes all the peer related information except its addresses. To remove the
 	// addresses use `AddrBook.ClearAddrs` or set the address ttls to 0.
 	RemovePeer(peer.ID)
+
+	// UpdateHostAddrs updates the addresses of the host in the peerstore. It also adds the signed
+	// peer record for the host if there's a private key available for the host in the KeyBook.
+	// This method should only be used to update the addresses of the Host that owns the peerstore.
+	UpdateHostAddrs(hostID peer.ID, currentAddrs, removedAddrs, peerRecordAddrs []ma.Multiaddr) error
```
sukunrt: Keeping the signature this way to avoid attaching a peerstore to a host. That would complicate the host creation even more.

p2p/host/peerstore/pstoremem/peerstore.go
```diff
@@ -98,6 +100,33 @@ func (ps *pstoremem) PeerInfo(p peer.ID) peer.AddrInfo {
 	}
 }
 
+func (ps *pstoremem) UpdateHostAddrs(hostID peer.ID, currentAddrs, removedAddrs, peerRecordAddrs []ma.Multiaddr) error {
```
sukunrt: There's a small change here. All host's now will store the signed peer record in th e peerstore. This I think is okay because identify is configured to disable sending signed peer records.

MarcoPolo:
> I've made this change to adding a method on the peerstore. I'm not convinced about it though.
> 
> 1. This adds host context to the peerstore, which at the moment can be used by any object, like Gossipsub and kaddht, to store data against peerids.
> 2. This method must be on peerstore and not on any of the component interfaces like `AddrBook` because we need the Private Key from the `KeyBook`.
> 3. Not sure if creating a PeerRecord should be with peerstore. At the moment `AddrManager` seems like the best owner for this logic since the only thing in the PeerRecord are the addresses.

I think you're right. I'm sorry for the false lead. I'm not a fan of the PeerStore changes in the second commit.

I've reviewed the first commit again, and thinking about it a bit more I think it's fine.

Please drop the second commit, and I'll approve

sukunrt:
Done. move it backed to the previous commit.

MarcoPolo (approved):

PR #3356, opened 2025-08-04
basichost: improve autonatv2 reachability logic

sukunrt:
This improves the reachability detection logic by introducing the concept of primary and secondary addresses. If we have a webtransport address which shares the IP and Port with a QUIC address, the WebTransport address will be considered secondary and the QUIC address will be considered primary. 

If the Primary is reachable or unreachable, we require only one confirmation for the Secondary address. This speeds up address verification considerably.

### DIFF
--- a/p2p/host/basic/addrs_reachability_tracker.go
+++ b/p2p/host/basic/addrs_reachability_tracker.go
@@ -361,9 +361,15 @@ const (
 	// and then a success(...S S S S F S). The confidence in the targetConfidence window  will be equal to
 	// targetConfidence, the last F and S cancel each other, and we won't probe again for maxProbeInterval.
 	maxRecentDialsWindow = targetConfidence + 2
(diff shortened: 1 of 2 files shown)


MarcoPolo:
> If we have a webtransport address which shares the IP and Port with a QUIC address, the WebTransport address will be considered secondary and the QUIC address will be considered primary.

Why isn't this based on just the "thinwaist" aspect rather than the transport? If my IP + UDP port is reachable, then it very likely doesn't matter if I'm running QUIC, WebTransport, or WebRTC

sukunrt:
It tests the other addresses once to ensure that there are no issues with the transport / node configuration etc. 
I've modified the PR to only refresh the reachability for these every 3 hours as opposed to every 1 hour for the primary thin waist address.

MarcoPolo (approved):
Looks good, just one comment: https://notes.marcopolo.io/view/5c574f20d5994a47d95487416be9505c

Please forgive the notes off github, it's much easier for me to review this locally, and the github UI is often quite slow for me.

PR #3364, opened 2025-08-09
feat: Migrate to log/slog

### LINKED ISSUE #3362: Migrate to slog
Migrate to the standard library structured logging package. Also remove the global logger and pass the logger in. 

This will let us (optionally) add a peer id to every log statement, which will make debugging tests much easier. 

It also lets users ingest the logs into a structured log database such as Loki.

We can add a shim to continue supporting the GOLOG_LOG_LEVEL

MarcoPolo:
closes #3362 

This is the first step in a larger refactor to be able to pass in a custom slog down to all components and get rid of the global logger.

Also removes the ipfs/go-log dependency with a gologshim. It makes slog behave close enough to ipfs/go-log with respect to environment variables. Users who are parsing the text version will notice differences, but the JSON version is identical except the caller source location key.

Users can use gologshim from go-libp2p to remove their dependency on ipfs/go-log and migrate to slog as well.

The patch is straightforward if not large. The general rules are:

- If we were doing some form of logf, change the text to the structured key-value pairs.
- logw becomes just log
- Try to normalize the log attribute keys.

The most interesting thing to review here is likely the shim itself. But a once over on the log changes would be appreciated.


gologshim/gologshim.go
```diff
@@ -0,0 +1,145 @@
...
+			} else if a.Key == slog.LevelKey {
+				// ipfs go-log uses lowercase level names
+				if lvl, ok := a.Value.Any().(slog.Level); ok {
+					if s, ok := lvlToLower[lvl]; ok {
+						a.Value = s
+					}
+				}
+			}
+			return a
+		},
```
sukunrt, line 44: Should we just remove this?

core/canonicallog/canonicallog.go
```diff
@@ -1,25 +1,54 @@
 package canonicallog
```
sukunrt, line 1: Not needed in this PR, but should we just remove this package?

core/canonicallog/canonicallog.go
```diff
@@ -1,25 +1,54 @@
...
+}
 
 // LogMisbehavingPeer is the canonical way to log a misbehaving peer.
 // Protocols should use this to identify a misbehaving peer to allow the end
 // user to easily identify these nodes across protocols and libp2p.
 func LogMisbehavingPeer(p peer.ID, peerAddr multiaddr.Multiaddr, component string, err error, msg string) {
-	log.Warnf("CANONICAL_MISBEHAVING_PEER: peer=%s addr=%s component=%s err=%q msg=%q", p, peerAddr.String(), component, err, msg)
+	logWithSkip(context.Background(), log, slog.LevelWarn, 1, "CANONICAL_MISBEHAVING_PEER",
+		"peer", p,
+		"addr", peerAddr.String(),
```
sukunrt: Can we pass `peerAddr.String()` as just `peerAddr` and then let the logger convert it to string?

libp2p_test.go
```diff
@@ -432,7 +432,7 @@ func TestMain(m *testing.M) {
 		goleak.IgnoreAnyFunction("github.com/koron/go-ssdp.Search"),
 		goleak.IgnoreAnyFunction("github.com/pion/sctp.(*Stream).SetReadDeadline.func1"),
 		// Logging & Stats
-		goleak.IgnoreTopFunction("github.com/ipfs/go-log/v2/writer.(*MirrorWriter).logRoutine"),
+		goleak.IgnoreTopFunction("github.com/libp2p/go-libp2p/gologshim/writer.(*MirrorWriter).logRoutine"),
```
sukunrt: Do we need this now?

gologshim/gologshim.go
```diff
@@ -0,0 +1,145 @@
...
+		logFmt = os.Getenv("GOLOG_LOG_FMT")
+	}
+	if logFmt == "json" {
+		c.format = logFormatJSON
+	}
+
+	logFmt = os.Getenv("GOLOG_LOG_ADD_SOURCE")
+	if logFmt != "" && logFmt != "0" {
+		c.addSource = true
+	}
```
sukunrt: This should be true by default and opt out. 

Currently when I do: `GOLOG_LOG_LEVEL='autonatv2=debug'` I get the source in the log lines: 

- old log line:
``` 
2025-08-20T18:46:45.433+0530    DEBUG   autonatv2       autonatv2/autonat.go:225        reachability check with 12D3KooWJYQQ1EbsGX3TBr1uE9MRGd3T6gPD2c5gDVoSVLKvg2FK successful
```
 
- new log line:
```
ts=2025-08-20T18:48:46.349+05:30 level=debug msg="reachability check successful" logger=autonatv2 peer=12D3KooWP9hPt4wrMw27o8XJR34uGanNZFsCUZfYmsxN9qxEHNCd
```

sukunrt: We're also losing color in the output. I am okay with that, but you may consider keeping ipfs/go-log and only removing it when we allow users to pass a custom slog handler.

libp2p_test.go
MarcoPolo: No, thanks for the callout. I forgot to come back to this.

gologshim/gologshim.go
MarcoPolo: On source, I went back and forth on this, but I think you're right. I'll bring it back in.

On color output, I'm not worried about this because it's not something that would break users.

core/canonicallog/canonicallog.go
MarcoPolo: Maybe. I'm not sure if anyone is using this. The original idea behind this was to allow connecting libp2p applications to things like fail2ban via the logger. I'd leave it unless there's a good reason to remove it.

gologshim/gologshim.go
MarcoPolo: What is this? The change in key and values? I want to keep this the same if users are already processing the logs (I expect some ipfs users are)

Issue #3361, opened 2025-08-07
enhancement: include non public relay addresses to be advertised too


shoriwe:
## What is the issue

Private based relays are unable to populate the `Reservation.Addrs` when non of its addresses are public.

## Expected behavior.

Permit the relay to respond with all its public addresses which is already filtered with the following line:

https://github.com/libp2p/go-libp2p/blob/f6c14a215b2012f3839f1b7157dfec70a772143a/p2p/protocol/circuitv2/relay/relay.go#L596

But also include those addresses shared with the connected peer. This could be performed by comparing `stream.Conn().RemoteMultiaddr()` and `stream.Conn().LocalMultiaddr()`. 

Even so, **what about including all of its addresses with filtering?** Is there any security consideration. Because they could anyway be guessed based on the statements of bellow.

## Why this is a issue

A valid circuit multiaddress can be successfully constructed and used based on the known addresses found on the `AddrInfo` of the connected peer. The problem is, then why we even care of `Reservation.Addrs` if that is possible this field is latency in the packet. We need to enforce the use of `Reservation.Addrs` by including information that can be actually used by the relay client.

shoriwe:
https://github.com/libp2p/go-libp2p/pull/3365

sukunrt:
Why do you want to advertise non publicly accessible relay addresses?

shoriwe:
Hi, @sukunrt. The reason of this came from using libp2p in private networks. Airgap networks to be precise. This way internal peers will be able to advertise services by the DMZ's running relay. In this scenario DMZ has no public internet address but rather machines from other local network can directly talk to him.

By just removing non public addresses we remove the chance of advertising in private networks.

Maybe the approach needs to be more granular. My direct approach was directly removing the line. But we could tell the relay service to block non public addresses by default and add some kind of filtering to include extra entries

MarcoPolo:
We could add an `Option` to add custom filtering to the relay.

shoriwe:
Hi, @MarcoPolo. Thanks for your comments.

I have pushed more changes related to your suggestion of using new Options for doing the final address filtering. https://github.com/libp2p/go-libp2p/pull/3365/files#r2274517907

Added two new functions. One using subnets for doing the address checking the other one using direct address comparison:

https://github.com/libp2p/go-libp2p/blob/223e120162d1f737477ee56b33baad2422d00843/p2p/protocol/circuitv2/relay/options.go#L27-L46

[...]

PR #3365, opened 2025-08-09
enhancement: include non public relay addresses to be advertised too #3361

shoriwe:
## Reference

https://github.com/libp2p/go-libp2p/issues/3361

## What is the issue

Private based relays are unable to populate the `Reservation.Addrs` when non of its addresses are public.

## Expected behavior.

Permit the relay to respond with all its public addresses which is already filtered with the following line:

https://github.com/libp2p/go-libp2p/blob/f6c14a215b2012f3839f1b7157dfec70a772143a/p2p/protocol/circuitv2/relay/relay.go#L596

But also include those addresses shared with the connected peer. This could be performed by comparing `stream.Conn().RemoteMultiaddr()` and `stream.Conn().LocalMultiaddr()`. 

Even so, **what about including all of its addresses with filtering?** Is there any security consideration. Because they could anyway be guessed based on the statements of bellow.

## Why this is a issue

A valid circuit multiaddress can be successfully constructed and used based on the known addresses found on the `AddrInfo` of the connected peer. The problem is, then why we even care of `Reservation.Addrs` if that is possible this field is latency in the packet. We need to enforce the use of `Reservation.Addrs` by including information that can be actually used by the relay client.

### DIFF
--- a/p2p/host/resource-manager/README.md
+++ b/p2p/host/resource-manager/README.md
@@ -47,7 +47,7 @@ limits := cfg.Build(scaledDefaultLimits)
 // The resource manager expects a limiter, se we create one from our limits.
 limiter := rcmgr.NewFixedLimiter(limits)
 
-// Metrics are enabled by default. If you want to disable metrics, use the 
+// Metrics are enabled by default. If you want to disable metrics, use the
 // WithMetricsDisabled option
 // Initialize the resource manager
 rm, err := rcmgr.NewResourceManager(limiter, rcmgr.WithMetricsDisabled())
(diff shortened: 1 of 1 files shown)


p2p/protocol/circuitv2/relay/options.go
```diff
@@ -18,6 +24,27 @@ func WithLimit(limit *RelayLimit) Option {
...
+// It can specify those addresses directly here.
+// This is useful when we already know our address and we want to directly specify it
+// /ip4/192.168.0.10/...
+func WithRelayAddresses(addresses ...multiaddr.Multiaddr) (option Option) {
+	return func(r *Relay) (err error) {
+		r.publicAddresses = append(r.publicAddresses, addresses...)
+		return nil
+	}
+}
+
```
shoriwe: https://github.com/libp2p/go-libp2p/issues/3361#issuecomment-3177472563

PR #3368, opened 2025-08-17
Fix:QUIC connection rate limiting with DoS protection

irajgill:


sukunrt:
resource manager already does this.

PR #3376, opened 2025-08-26
p2p/autorelay: use reservation addrs, remove addrsplosion

sashass1315:
- Updated getCircuitAddrs to build /p2p-circuit addrs from rsvp.Addrs  
- Inlined isRelayAddr into relay_finder.go  
- Removed cleanupAddressSet and deleted addrsplosion.go + its tests  
- Dropped obsolete heuristic in favor of relay-provided addresses


MarcoPolo:
Can you please share the motivation for this change?

sashass1315:
> Can you please share the motivation for this change?

// TODO: Remove this, we don't need this. The current method tries to select the
// best address for the relay. Instead we should rely on the addresses provided by the
// relay in response to the reservation request

PR #3377, opened 2025-08-31
feat: Add Post-Quantum Signature Schemes Support (Dilithium, ML-DSA, Hybrids)

JonyBepary:
## Description

This PR adds comprehensive support for post-quantum (PQ) digital signature schemes to go-libp2p, implementing 8 different PQ algorithms using the Cloudflare CIRCL library. This enhances libp2p's cryptographic capabilities to prepare for the quantum computing era while maintaining full backward compatibility.

### Core Implementation
- **Added 8 PQ signature schemes:**
  - Dilithium2/3/5 (NIST Round 3 finalists)
  - ML-DSA-44/65/87 (FIPS 204 standard)
  - Ed25519-Dilithium2 hybrid
  - Ed448-Dilithium3 hybrid

- **Protobuf integration:**
  - Extended `KeyType` enum in `core/crypto/pb/crypto.proto`
  - Added unmarshaller registrations for all PQ types
  - Full protobuf marshaling/unmarshaling support

- **API compatibility:**
  - All PQ types implement the standard `PrivKey`/`PubKey` interfaces
  - Maintains existing `GenerateKeyPairWithReader` API
  - Added PQ types to generic test harness (`KeyTypes` slice)

### Files Added/Modified
- `core/crypto/pb/crypto.proto` - Added PQ key type enums
- `core/crypto/key.go` - Updated registries and key generation
- 8 new implementation files (dilithium2.go, mldsa44.go, etc.)
- 6 comprehensive test files with full coverage
- `examples/pq_demo.go` - Working demonstration

### Testing
- **24 PQ-specific tests** covering key generation, signing, verification, marshaling
- **Generic harness integration** - PQ types tested by `TestKeys`
- **Enhanced test coverage** matching RSA/Ed25519 parity:
  - Protobuf marshal/unmarshal round-trips
  - Sign zero message tests
  - Negative unmarshal error cases
  - KeyEqual function tests
- **All existing tests pass** - no regressions

## Standards Compliance

- **NIST FIPS 204** (ML-DSA)
- **CIRCL library** v1.6.1 for PQ crypto implementation
- **libp2p crypto interface** compliance
- **Protobuf wire format** backward compatibility

## Backward Compatibility

✅ **Fully backward compatible:**
- All existing RSA, Ed25519, ECDSA, Secp256k1 functionality unchanged
- Existing APIs work identically
- Protobuf enums are additive (no breaking changes)
- No impact on existing applications

## Usage Example

```go
import "github.com/libp2p/go-libp2p/core/crypto"

// Generate PQ key pair
priv, pub, err := crypto.GenerateKeyPair(crypto.Dilithium2, -1)
if err != nil {
    return err
}

// Sign and verify
signature, err := priv.Sign(message)
valid, err := pub.Verify(message, signature)
```


JonyBepary:
### Some Clarifications

- **Interfaces and transports:** All key types (Ed25519, Dilithium2, Dilithium3/5, ML-DSA, hybrids) implement the same ic.PrivKey/ic.PubKey interface. TLS and Noise only call sk.Sign and pub.Verify and do not depend on any stdlib type conversions. The TLS code signs a locally generated ECDSA certificate key with the libp2p key (GenerateSignedExtension); therefore, any libp2p key type will work as long as both ends can Unmarshal/Verify it. 

- **Serialization and IDs:** The protobuf schema includes Dilithium2 and hybrid enums. Public keys are marshaled with a KeyType and bytes; therefore, on-wire encoding/decoding works as long as the remote has the latest libp2p. 
PQ-based public keys are large; therefore, they will not be identity-inlined in peer IDs. As a result, the peer ID will be a regular multihash (SHA2-256 of the marshaled public key). ExtractPublicKey() from a peer ID only works when the key is identity-inlined, which is true for Ed25519 but not for PQ-based algorithms. This is expected and acceptable in this study. For the record, existing RSA keys (2048-bit) also use hash-based peer IDs.

- **Backwards compatibility:** Older peers that only know Ed25519 will not be able to verify Dilithium2 or hybrid signatures or unmarshal these key types. There is no negotiation step for identity key algorithms; therefore, both sides must support the chosen key type. If any system needs to remain compatible with Ed25519-only peers, then they must continue using Ed25519 keys for those deployments. 

- **Hybrid option:** The Ed25519-Dilithium2 hybrid type provides dual signature security (classical + PQ). It does not magically make it compatible with Ed25519-only peers; peers still need to recognize and verify the hybrid scheme. It is best for migration when all peers have been upgraded, giving a post-secured system without compromising classical compatibility. 

- **RNG:** CIRCL’s GenerateKey() uses crypto/rand internally and does not accept an RNG parameter; hence, the ignored io.Reader. This is acceptable for cryptographic soundness. If you need deterministic or externally seeded generation, the current post-quantum scheme does not expose that; you would need to save and reload keys via Marshal/Unmarshal for reproducibility.

MarcoPolo:
Was this AI Generated? Note our Assistance notice policy: https://github.com/libp2p/go-libp2p?tab=readme-ov-file#ai-assistance-notice

JonyBepary:
> Was this AI Generated? Note our Assistance notice policy: https://github.com/libp2p/go-libp2p?tab=readme-ov-file#ai-assistance-notice

I have read the AI assistance notice and apologize for not disclosing it. The description was AI-generated, but I thoroughly checked the information. The ‘clarification’ comments I have written are raw ideas and concerns, and then I used AI to improve the text and bullet points for better understanding.  Regarding the code, I used Copilot (Grok Fast 1) to assist and improve my codes, but it is not an AI-generated blob where I did the prompt, and AI did everything for me.  I have previously used libp2p and circl for many of my projects, and I have some ideas about this project. However, it is possible that I missed updating any other folder except the core/crypto folder. In addition, I feel like I should have a PR with one algo at a time to reduce the burden on the maintainer. The reason for pushing multiple algorithms is that Circl shares a common interface, which is bit easier and redundant, and I follow the existing rsa and ed25199 implementation interface, test.  If the entire push is overwhelming, you can close these pr, and I can again pull these in one algo at a time and proper disclosure.

MarcoPolo:
Thanks for the honesty.

I haven't looked closely at the current set of standards, but it doesn't seem like we have a standard yet for Hybrid or Post-Quantum algorithms in TLS 1.3. That needs to happen before we make any changes to go-libp2p. After that is standardized we would discuss in libp2p/specs, and then make go-libp2p changes.

That said, you're welcome to experiment with this branch as much as you like. I would ask that you use a 4 byte number for the crypto protobuf fields so as not to conflict with any future algorithm.

JonyBepary:
Thank you for your understanding.
I'm definitely going to work that branch with 4 byte number for the crypto protobuf fields. I hadn't considered the standard TLS 1.3 compatibility with PQ until you brought it up, and now I've looked into it. Interestingly, I came across a draft from last year on _ML-KEM Post-Quantum Key Agreement for TLS 1.3_ [1]. Additionally, Microsoft Edge is currently working on this, and Windows, Mac, and Android will soon support post-quantum keys [2]. With the FIPS 203 standard now established, having a fully post-quantum (not hybrid) key agreement option will be needed for moving beyond hybrids and for users post-quantum security.

1. https://datatracker.ietf.org/doc/draft-connolly-tls-mlkem-key-agreement/05/
2. http://learn.microsoft.com/en-us/deployedge/microsoft-edge-browser-policies/postquantumkeyagreementenabled

PR #3379, opened 2025-09-01
 enhancement: include non public relay addresses to be advertised too #3361 #3365 

shoriwe:
Moved the code in my fork to its own branch. GitHub automatically closed [old](https://github.com/libp2p/go-libp2p/pull/3365)

## Reference

https://github.com/libp2p/go-libp2p/issues/3361

## What is the issue

Private based relays are unable to populate the `Reservation.Addrs` when non of its addresses are public.

## Expected behavior.

Permit the relay to respond with all its public addresses which is already filtered with the following line:

https://github.com/libp2p/go-libp2p/blob/f6c14a215b2012f3839f1b7157dfec70a772143a/p2p/protocol/circuitv2/relay/relay.go#L596

But also include those addresses shared with the connected peer. This could be performed by comparing `stream.Conn().RemoteMultiaddr()` and `stream.Conn().LocalMultiaddr()`. 

Even so, **what about including all of its addresses with filtering?** Is there any security consideration. Because they could anyway be guessed based on the statements of bellow.

## Why this is a issue

A valid circuit multiaddress can be successfully constructed and used based on the known addresses found on the `AddrInfo` of the connected peer. The problem is, then why we even care of `Reservation.Addrs` if that is possible this field is latency in the packet. We need to enforce the use of `Reservation.Addrs` by including information that can be actually used by the relay client.

### DIFF
--- a/p2p/protocol/circuitv2/relay/options.go
+++ b/p2p/protocol/circuitv2/relay/options.go
@@ -1,5 +1,9 @@
 package relay
 
+import (
+	"github.com/multiformats/go-multiaddr"
+)
+
 type Option func(*Relay) error
 
 // WithResources is a Relay option that sets specific relay resources for the relay.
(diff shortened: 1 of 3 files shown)


shoriwe:
Thanks for you review, @MarcoPolo. I will have those points patched based on your review for tonight.

shoriwe:
@MarcoPolo, I have pushed the changes based on your recommendations. Very good suggestions. This new change will permit anyone add custom filters for the addresses the relay can promote.

Waiting for a second review

PR #3381, opened 2025-09-07
feat: implement net.Conn for network.Stream ergonomics

jimmale:
This PR attempts to improve the ergonomics of using go-libp2p by implementing the `net.Conn` interface on `network.Stream`. This makes it easier to use go-libp2p with other components in the go ecosystem that are not written with go-libp2p in mind. 

In my specific use case, I'm building a REST API with the gin framework.

This PR also makes a small change to a test relating to file descriptors. See commits for full details.

AI disclosure: The code in this PR was written with traditional line-by-line autocomplete in my IDE. No large language models were involved. The trivial nature of this change makes me confident that there are no hidden IP issues. The commit messages were written entirely by hand, so all spelling mistakes and other errors are 100% my own.

One question I struggled with while writing this is what the `Network()` value of a libp2p address should be in go... the net package defines `tcp`, `tcp4`, `tcp6`, `udp`, `udp4`, `udp6`, `ip`, `ip4`, `ip6`, `unix`, `unixgram`, and `unixpacket` for the more traditional network types. `libp2p` seemed to be a good choice, but I'm very much open to suggestions.

### DIFF
--- a/core/network/stream.go
+++ b/core/network/stream.go
@@ -1,6 +1,8 @@
 package network
 
 import (
+	"net"
+
 	"github.com/libp2p/go-libp2p/core/protocol"
 )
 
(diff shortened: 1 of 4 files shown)


marten-seemann:
There's no need to change this interface and add even more method (as well as break tons of users). The same behavior is absolutely trivial to achieve by implementing a tiny wrapper in the application code.

PR #3384, opened 2025-09-18
Separate Core Module

MarcoPolo:
This lets protocols, such as Gossipsub and Maybe Kad-DHT, to be more agnostic to the specific version of go-libp2p they are running on. Instead they depend only on a specific version of an interface.

The core module should remain fairly stable. After some time we could consider cutting a v1 release. But for now it will follow a patch or match versioning. If it’s a patch compatible change only the patch number will change, if not it will match the version of go-libp2p.


MarcoPolo:
The pain with the previous setup was the separate repo. This keeps this mono repo, which allows for atomic changes across modules.

PR #3389, opened 2025-09-21
refactor: using maps.Clone

cuiweixie:
It's better to use maps.Clone here.
https://github.com/golang/go/commit/b5a7f2eef7cb17255cb396cd4ff7df04957dd21e

### DIFF
--- a/core/routing/options.go
+++ b/core/routing/options.go
@@ -1,5 +1,7 @@
 package routing
 
+import "maps"
+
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Are you actually running into this as a hotspot?

I don’t think this change is necessary and I’m hesitant to merge changes for little to no reasons.

PR #3391, opened 2025-09-22
fix(dcutr): Fix client/server roles to match DCUtR specification

### LINKED ISSUE #3171: Fix dcutr roles in holepunching
We did the opposite thing of what the spec says to do. This meant that we would fail to hole-punch with rust nodes because both sides would attempt to be the dialer.

The [spec](https://github.com/libp2p/specs/blob/master/relay/DCUtR.md#the-protocol) is a bit confusing since roles get flipped in the middle. But essentially:

1. A connects to B.
2. B initiates the `/libp2p/dcutr` protocol.
3. After synchronizing, A and B dial each other. A is the client and B is the server.

This means the side handling the stream is the client, and the side initiating the stream should be the server.

#3044 exposes a flag to enable the spec behavior, but it is default set to legacy behavior. There are no backwards compatible migrations in that PR.

To close this issue we should implement a backwards compatible change. One suggestion by Sukun is:

> On the side that in the new code will be the Server, can we add a random wait time after which the server will switch the role to Client and initiate the security negotiation?

asmit27rai:
- In svc.go: Stream receiver (A) now acts as client per spec
- In holepuncher.go: Stream initiator (B) now acts as server per spec
- This fixes hole punching compatibility with Rust nodes
- Backwards compatibility preserved via legacyBehavior flag

Fixes #3171

### DIFF
--- a/p2p/protocol/holepunch/holepuncher.go
+++ b/p2p/protocol/holepunch/holepuncher.go
@@ -161,9 +161,9 @@ func (hp *holePuncher) directConnect(rp peer.ID) error {
 			hp.tracer.StartHolePunch(rp, addrs, rtt)
 			hp.tracer.HolePunchAttempt(pi.ID)
 			ctx, cancel := context.WithTimeout(hp.ctx, hp.directDialTimeout)
-			isClient := true
+			isClient := false
 			if hp.legacyBehavior {
-				isClient = false
+				isClient = true
 			}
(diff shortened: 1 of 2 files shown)


MarcoPolo:
Did you read #3171? This PR does the wrong thing. It flips the behavior of the legacy flag.

This is also not an easy change. We need a rollout strategy and be careful about it.

PR #3394, opened 2025-09-23
 Fix DCUtR roles in holepunching

### LINKED ISSUE #3171: Fix dcutr roles in holepunching

asmit27rai:
## Fix DCUtR roles in holepunching

Fixes #3171

### DIFF
--- a/p2p/protocol/holepunch/backwards_compat_test.go
+++ b/p2p/protocol/holepunch/backwards_compat_test.go
@@ -1,0 +1,44 @@
+package holepunch_test
+
+import (
+	"testing"
+	"github.com/libp2p/go-libp2p"
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/p2p/protocol/holepunch"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBackwardsCompatibility(t *testing.T) {
+	t.Run("spec-compliant to legacy", func(t *testing.T) {
+		specHost := MustNewHost(t,
+			libp2p.EnableHolePunching(holepunch.WithSpecCompliantBehavior()),
+		)
+		defer specHost.Close()
+
+		legacyHost := MustNewHost(t,
+			libp2p.EnableHolePunching(SetLegacyBehavior(true)),
+		)
+		defer legacyHost.Close()
+
+		testHolePunchConnection(t, specHost, legacyHost)
+	})
+
+	t.Run("legacy to spec-compliant", func(t *testing.T) {
+		legacyHost := MustNewHost(t,
+			libp2p.EnableHolePunching(SetLegacyBehavior(true)),
+		)
+		defer legacyHost.Close()
+
+		specHost := MustNewHost(t,
+			libp2p.EnableHolePunching(holepunch.WithSpecCompliantBehavior()),
+		)  
+		defer specHost.Close()
+
+		testHolePunchConnection(t, legacyHost, specHost)
+	})
+}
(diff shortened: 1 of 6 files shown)


asmit27rai:
@MarcoPolo Please look at this PR i think this will work well.

p2p/protocol/holepunch/holepuncher.go
```diff
@@ -107,82 +106,7 @@ func (hp *holePuncher) DirectConnect(p peer.ID) error {
...
-		case <-hp.ctx.Done():
-			timer.Stop()
-			return hp.ctx.Err()
-		}
-		if i == maxRetries {
-			hp.tracer.HolePunchFinished("initiator", maxRetries, addrs, obsAddrs, nil)
-		}
-	}
-	return fmt.Errorf("all retries for hole punch with peer %s failed", rp)
+	return hp.directConnectFixed(rp)
```
MarcoPolo, line 109: Why the indirection? It makes it hard to understand what the diff is.

p2p/protocol/holepunch/backwards_compat_test.go
```diff
@@ -0,0 +1,44 @@
...
+			libp2p.EnableHolePunching(holepunch.WithSpecCompliantBehavior()),
+		)  
+		defer specHost.Close()
+
+		testHolePunchConnection(t, legacyHost, specHost)
+	})
+}
+
+func testHolePunchConnection(t *testing.T, h1, h2 host.Host) {
+	require.NotNil(t, h1)
```
MarcoPolo, line 42: This is not testing holepunch connection, as I'm sure you know. You are doing nothing except testing the constructor.

asmit27rai:
> Hey @asmit27rai,
> 
> thanks for taking the time to draft this PR and looking into the issue. However, the proposed changes don't make sense, and I'm not convinced you've really thought about the problem.
> 
> If you want to fix this issue, please take the time to understand the problem deeply and evaluate potential solutions. We can discuss potential solutions on the issue itself. We should discuss the overall plan before making significant code changes.
> 
> Before you open your next PR, I ask that you critically review the code yourself first. If you submit another low quality PR I will take it as a sign that you are not interested in contributing quality work and I will stop reviewing your changes.

@MarcoPolo 
I just try to implement Sukun's suggestion, we'll implement a random wait mechanism where the server-side peer waits for a random period before switching roles to client if the connection isn't established.
In https://github.com/libp2p/go-libp2p/pull/3044.

Can we discuss about the solution for this issue at any communication channel ? I want to implement the solution.

MarcoPolo:
Let's discuss at [#libp2p](https://web.libera.chat/gamja/?channel=#libp2p) on Libera.chat

PR #3409, opened 2025-10-11
refactor: use reflect.TypeFor

stemlaud:
Try to use better api `reflect.TypeFor`  recommend by Go Team instead of `reflect.TypeOf` when we have known the type.

More info https://github.com/golang/go/issues/60088


stemlaud:
@MarcoPolo Hi, Could you please review this PR at your convenience? Thank you very much.

PR #3418, opened 2025-10-27
feat(gologshim): Add SetDefaultHandler

lidel:
## Problem

The migration to slog in 

- https://github.com/libp2p/go-libp2p/pull/3364 

broke go-log's ability to adjust subsystem log levels at runtime via `SetLogLevel(`) and control output formatting (colors, json). This was a key debugging feature that allowed changing log verbosity without restarting daemons.

Things like `ipfs log tail`, `ipfs log level` and Diagnostic screen in ipfs-webui (https://github.com/ipfs/ipfs-webui/pull/2392) are broken with go-libp2p 0.44 – no go-libp2p logs are available, users are unable to, for example adjust `nat` or `net/identify` to `DEBUG` level at runtime to do diagnostics.

Second problem was that logs that worked due to env-variable being set to DEBUG ignored the user-preferred formatting. 

Long term we may want to move `go-log` from `zap` backend to `slog`, but in the meantime we need to unblock software that uses it to be able to use go-libp2p without regressions. 

cc @MarcoPolo  @gammazero  for visibility

## Solution

First of all, a lot of time went into restoring this. It was not easy to fix without introducing `go-log` as a dependency.

To avoid depending on `go-log`, this PR detects go-log's slog bridge via duck typing and uses it when available, restoring dynamic level control and unified formatting. When go-log isn't present, `gologshim` falls back to standalone slog handlers as before.

~The lazy handler pattern solves initialization order issues where package-level loggers are created before go-log's bridge is installed.~

The intention of this PR is to fix regression allowing existing software to update to latest go-libp2p.

With this approach, users that depend on both `go-log` and `go-libp2p` just need to update to this version of `go-libp2p` and `go-log` with https://github.com/ipfs/go-log/pull/176 - ~no code changes or `init()` hacks required, and it does not matter in thish order `go-log` and `go-libp2p` are imported and initialized by existing apps. Things "just work" without breaking existing userbase.~

> [!IMPORTANT]
> Original version of this PR was rejected by go-libp2p maintainers, and instead go-libp2p 0.45 requires go-log users to manually set up wiring. See https://github.com/ipfs/go-log/releases/tag/v2.9.0 and https://github.com/libp2p/go-libp2p/pull/3419#issuecomment-3487634114

## Reference 

- This PR depends on https://github.com/ipfs/go-log/pull/176 to be release first
  - We will do that if this PR is merged first. 
- Kubo bug https://github.com/ipfs/kubo/issues/11035
  - Same bug occurs in every project that uses `go-log` and `go-libp2p`
  - Kubo PR with end-to-end fixes and tests: https://github.com/ipfs/kubo/pull/11039

### DIFF
--- a/gologshim/gologshim.go
+++ b/gologshim/gologshim.go
@@ -1,16 +1,32 @@
-// Package gologshim provides slog-based logging that integrates with go-log
-// when available, without requiring go-log as a dependency.
+// Package gologshim provides slog-based logging for go-libp2p that works
+// standalone or integrates with go-log for unified log management across
+// IPFS/libp2p applications, without adding go-log as a dependency.
 //
-// Usage:
+// # Usage
 //
-//	var log = logging.Logger("subsystem")
+// Create loggers using the Logger function:
+//
+//	var log = gologshim.Logger("subsystem")
 //	log.Debug("message", "key", "value")
 //
-// When go-log's slog bridge is detected (via duck typing), gologshim
-// automatically integrates for unified formatting and dynamic level control.
-// Otherwise, it creates standalone slog handlers writing to stderr.
+// # Integration with go-log
 //
-// Environment variables (see go-log README for full details):
+// Applications can optionally connect go-libp2p to go-log by calling SetDefaultHandler:
+//
+//	func init() {
+//		gologshim.SetDefaultHandler(slog.Default().Handler())
+//	}
+//
+// When integrated, go-libp2p logs use go-log's formatting and can be controlled
+// programmatically via go-log's SetLogLevel("subsystem", "level") API to adjust
+// log verbosity per subsystem at runtime without restarting.
+//
+// # Standalone Usage
+//
+// Without calling SetDefaultHandler, gologshim creates standalone slog handlers
+// writing to stderr. This mode is useful when go-log is not present or when you
+// want independent log configuration via backward-compatible (go-log) environment variables:
+//
 //   - GOLOG_LOG_LEVEL: Set log levels per subsystem (e.g., "error,ping=debug")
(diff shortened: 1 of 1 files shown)


gologshim/gologshim.go
```diff
@@ -15,25 +40,48 @@ var lvlToLower = map[slog.Level]slog.Value{
...
-func Logger(system string) *slog.Logger {
-	var h slog.Handler
-	c := ConfigFromEnv()
-	handlerOpts := &slog.HandlerOptions{
-		Level:     c.LevelForSystem(system),
-		AddSource: c.addSource,
+// goLogBridge detects go-log's slog bridge via duck typing, without import dependency
+type goLogBridge interface {
+	GoLogBridge()
+}
```
lidel: ℹ️  this is the gist of the solution that avoids depending on go-log in go.mod

MarcoPolo:
Thanks lidel, I'll have a review by eod tomorrow.

gologshim/gologshim.go
```diff
@@ -43,16 +91,56 @@ func Logger(system string) *slog.Logger {
...
-	attrs = append(attrs, slog.String("logger", system))
-	attrs = append(attrs, c.labels...)
-	h = h.WithAttrs(attrs)
-	return slog.New(h)
+
+	// Use lazy handler for init order issues
+	return slog.New(&lazyBridgeHandler{
+		system: system,
+		config: c,
+	})
```
gammazero, line 146: Yeah, cannot rely on `init()` order across multiple packages.
👍

gammazero (approved):

MarcoPolo:
See #3419

PR #3419, opened 2025-10-29
Gologshim runtime controlled handler

MarcoPolo:
lidel, take a look and let me know your thoughts. This is targeting your branch.

This removes the goLogBridge interface check and having to modify the default slog logger. Instead you explicitly set the handler via `gologshim.SetDefaultHandler`.

### DIFF
--- a/gologshim/gologshim.go
+++ b/gologshim/gologshim.go
@@ -1,16 +1,32 @@
-// Package gologshim provides slog-based logging that integrates with go-log
-// when available, without requiring go-log as a dependency.
+// Package gologshim provides slog-based logging for go-libp2p that works
+// standalone or integrates with go-log for unified log management across
+// IPFS/libp2p applications, without adding go-log as a dependency.
 //
-// Usage:
+// # Usage
 //
-//	var log = logging.Logger("subsystem")
+// Create loggers using the Logger function:
+//
+//	var log = gologshim.Logger("subsystem")
 //	log.Debug("message", "key", "value")
 //
-// When go-log's slog bridge is detected (via duck typing), gologshim
-// automatically integrates for unified formatting and dynamic level control.
-// Otherwise, it creates standalone slog handlers writing to stderr.
+// # Integration with go-log
 //
-// Environment variables (see go-log README for full details):
+// Applications can optionally connect go-libp2p to go-log by calling SetDefaultHandler:
+//
+//	func init() {
+//		gologshim.SetDefaultHandler(slog.Default().Handler())
+//	}
+//
+// When integrated, go-libp2p logs use go-log's formatting and can be controlled
+// programmatically via go-log's SetLogLevel("subsystem", "level") API to adjust
+// log verbosity per subsystem at runtime without restarting.
+//
+// # Standalone Usage
+//
+// Without calling SetDefaultHandler, gologshim creates standalone slog handlers
+// writing to stderr. This mode is useful when go-log is not present or when you
+// want independent log configuration via backward-compatible (go-log) environment variables:
+//
 //   - GOLOG_LOG_LEVEL: Set log levels per subsystem (e.g., "error,ping=debug")
 //   - GOLOG_LOG_FORMAT/GOLOG_LOG_FMT: Output format ("json" or text)
 //   - GOLOG_LOG_ADD_SOURCE: Include source location (default: true)
@@ -64,6 +80,8 @@ func (h *dynamicHandler) ensureHandler() slog.Handler {
 			h.handler = h.createFallbackHandler()
 		}
 		attrs := make([]slog.Attr, 0, 1+len(h.config.labels))
+		// Use "logger" attribute for compatibility with go-log's Zap-based format
+		// and existing IPFS/libp2p tooling and dashboards.
 		attrs = append(attrs, slog.String("logger", h.system))
 		attrs = append(attrs, h.config.labels...)
 		h.handler = h.handler.WithAttrs(attrs)


gammazero (approved):
I think this makes the code cleaner/clearer. Your call @lidel

lidel (approved):
This works for us also, pushed docs + some thoughts below:

## On the trade-off here

Original #3418 required no changes in existing apps, just updating both `go-log` and `go-libp2p` to latest versions. 

This PR requires every legacy application that used both `go-log` and `go-libp2p` to add this extra code:

```go
  import (
      "github.com/libp2p/go-libp2p/gologshim"
  )

  func init() {
      // After go-log's init() runs and installs slog bridge,
      // tell gologshim to use it      
      handler := slog.Default().Handler()      
      // Verify it's go-log's bridge
      type goLogBridge interface {
	      GoLogBridge()
      }
      if _, ok := handler.(goLogBridge); !ok {
	      panic("aborting startup: slog.Default() is not go-log's bridge, logs would be missing due to incorrect wiring")
      }
      // use the implicit go-log bridge      
      gologshim.SetDefaultHandler(handler)
  }
```

or 

```go
  import (
      logging "github.com/ipfs/go-log/v2"
      "github.com/libp2p/go-libp2p/gologshim"
  )

  func init() {
      // use the explicit go-log bridge      
      gologshim.SetDefaultHandler(golog.SlogHandler())
  }
```

If this is the price to pay to fix logs, we can do the above in Kubo, Rainbow and every other application Shipyard maintains, but everyone else will be missing logs unless they do the same. 

Just a trade-off decision.

## Next steps

@MarcoPolo lgtm, i've tested this PR in https://github.com/ipfs/kubo/pull/11039 and it also works for us (with the above extra code).

Up to you if you prefer implicit or explicit.

- If you prefer explicit, feel free to merge this PR into #3418, and then merge that (#3419+#3418).
- If you are ok with implicit detection, this PR can be closed and only merge #3418

Either way, we would appreciate a release with a fix :pray:

MarcoPolo:
> This PR requires every legacy application that used both go-log and go-libp2p to add this extra code:

I don't follow, do you mind explaining a bit more? Where are you calling `slog.SetDefault`? And why not directly call `gologshim.SetDefaultHandler` instead?

lidel:
@MarcoPolo I wrote an (wip draft) explainer in https://github.com/ipfs/go-log/pull/176 for go-log users, check "Slog integration" section in `README` in that PR and especially the ["For library authors"](https://github.com/ipfs/go-log/blob/eea1e4119c495cce5e8ac0c210d12c0f13853f9d/README.md#for-library-authors) section, which has two approaches and their ramifications illustrated with sample code:

- [`Approach 1: Duck-typing detection (automatic)` ](https://github.com/ipfs/go-log/blob/eea1e4119c495cce5e8ac0c210d12c0f13853f9d/README.md#approach-1-duck-typing-detection-automatic)
  - this is how https://github.com/libp2p/go-libp2p/pull/3418 worked for go-libp2p as well 
- [`Approach 2: Explicit handler passing (manual)`](https://github.com/ipfs/go-log/blob/eea1e4119c495cce5e8ac0c210d12c0f13853f9d/README.md#approach-2-explicit-handler-passing-manual) 
  - this is what this PR (https://github.com/libp2p/go-libp2p/pull/3419) implies for end users of go-libp2p (requires extra code, but still works)

> Where are you calling `slog.SetDefault`?

go-log already called `SetupLogging` in own [go-log's `init()` here](https://github.com/ipfs/go-log/blob/1ac7bd1c28b25b0a0af528b4c97976fabcf730f5/setup.go#L20) to correctly handle ENV-based configuration. This already happened automatically if code imported go-log.

What we added to that `init` in https://github.com/ipfs/go-log/pull/176 is an opt-out slog support which calls [`captureSlog` here](https://github.com/ipfs/go-log/blob/1ac7bd1c28b25b0a0af528b4c97976fabcf730f5/setup.go#L162-L167) that sets up bridge and [sets it via `slog.SetDefault` here](https://github.com/ipfs/go-log/blob/1ac7bd1c28b25b0a0af528b4c97976fabcf730f5/setup.go#L162)

This has to be done because we dont want to lose any slog entries in go-lang apps, nor want to have different formatting from libraries that use slog.  Going forward, various libraries may switch to slog, and we dont want to have regressions like the one go-libp2p caused when it switched to slog. 

In the future, we may decide to switch the go-log backend from zap to slog, but the fact we call `slog.SetDefault` would not change. If app choses to use go-log, they expect it to handle all logs.

> And why not directly call `gologshim.SetDefaultHandler` instead?

The `go-libp2p/gologshim`  only takes care of `go-libp2p` and ignores any other application that use `slog`.

`go-log` is how applications in ecosystem control log levels and we do not want to have regressions like https://github.com/ipfs/kubo/issues/11035 anywhere else.

The goal was to unbreak existing users without asking them to change any code.

If go-libp2p forces our hand to add extra call for `gologshim.SetDefaultHandler` to have per subsystem visibility in Kubo, that is fine, but unfortunate for turn-key experience across ecosystem. Ideally, we would not require end users to do extra work to fix every application.

MarcoPolo:
>  Going forward, various libraries may switch to slog, and we dont want to have regressions like the one go-libp2p caused when it switched to slog.

Are you expecting to have all other libraries add the goLogBridge interface check? I'm not sure that makes sense.

I think setting the default log handler is something that should be owned by the application, not by a library that could possibly be a transitive dependency of another library.

I don't mind merging this change, but this will change again if/when we move away from global loggers to passing loggers explicitly. At that point applications would need to pass go-log explicitly to go-libp2p anyways, and any init trick won't work.

gologshim/gologshim.go
```diff
@@ -1,16 +1,50 @@
...
 //
-// When go-log's slog bridge is detected (via duck typing), gologshim
-// automatically integrates for unified formatting and dynamic level control.
-// Otherwise, it creates standalone slog handlers writing to stderr.
+// # Integration with go-log
+//
+// Applications can optionally connect go-libp2p to go-log by calling SetDefaultHandler:
+//
+//	func init() {
+//		gologshim.SetDefaultHandler(slog.Default().Handler())
```
MarcoPolo, line 17: I think it would be less error prone and more explicit if go-log exposed a slog.Handler that could be referenced by the user.

This would then be
`gologshim.SetDefaultHandler(golog.Handler)`

In contrast, the example code above is confusing. There's a redirection here that's not obvious by looking at the code.

MarcoPolo: tagging @lidel so you don't miss this despite this being merged.

lidel: thanks, having `golog.SlogHandler()` for manual wiring when catch-all slog bridge is disabled is a good idea. added `SlogHandler()` in https://github.com/ipfs/go-log/pull/176/commits/101e2c1ac85a912be5b6b9e2e9b729fe09007b9c

this is does not break anything nor require changes in go-libp2p, but I've updated godoc in https://github.com/libp2p/go-libp2p/pull/3424 to use simplified `golog.SlogHandler()` example in godoc

Issue #3076, opened 2024-12-01
Consider advertising all interface addresses and not just the primary interface addrs


sukunrt:
When a host is listening on `0.0.0.0` we only advertise the primary interface address of the node. The primary interface is selected by querying the routing table for the default route. But if the node is listening on `0.0.0.0`, we should advertise all the interface addresses.

Specifically, I want the nodes to advertise the tailscale interface addrs so I can connect to nodes running over my different devices. 

The obvious downside here is that we'll start advertising a lot more addresses to the network.

dozyio:
Also have the same issue when running on a docker swarm network

MarcoPolo:
done in #3421

PR #3421, opened 2025-10-30
update go-netroute dependency

MarcoPolo:


sukunrt:
This breaks the `AllAddrs` method which was previously only returning the 'primary' interface and now returns all interface addrs.

The problem is that the new netroute implementation fails when we query `0.0.0.0` while the previous one returned the primary(default) interface. To be clear just removing the early error in netroute code wont fix this because then the os will return 127.0.0.1 as the interface because it considers 0.0.0.0 to be localhost. https://github.com/libp2p/go-netroute/blob/master/netroute_linux.go#L42 

We've inadvertently ended up implementing https://github.com/libp2p/go-libp2p/issues/3076

PR #3432, opened 2025-11-30
rcmgr: centralize blocked resource logging through trace reporter

### LINKED ISSUE #1744: rcmgr: debug logs around blocked resources should come from the same place
I think there's some confusing when users look at the blocked resources log from the resource manager. Specifically the logs don't always line up with the metrics. I think the metrics are more accurate because they are being hooked in via the traces, while the logs are ad-hoc throughout the code.

The fix here is to create a trace reporter that consumes the trace event and logs blocked resource events.

sneaxhuh:
Fixes #1744

This change consolidates debug logging for blocked resources into a centralized trace reporter system. Previously, debug logs for blocked resources were scattered throughout the codebase and didn't always align with the metrics, which were reliably tracked through trace events.

Changes:
- Add LogReporter that consumes trace events and logs blocked resources
- Remove ad-hoc log.Debug() calls from scope.go
- Install LogReporter by default in NewResourceManager
- Add tests for LogReporter functionality

Benefits:
- Logs and metrics now use the same source (trace events)
- More maintainable centralized logging approach
- Consistent blocked resource reporting

### DIFF
--- a/p2p/host/resource-manager/log_reporter.go
+++ b/p2p/host/resource-manager/log_reporter.go
@@ -1,0 +1,85 @@
+package rcmgr
+
+import (
+	"github.com/libp2p/go-libp2p/core/network"
+)
+
+// LogReporter is a trace reporter that logs blocked resource events.
+// It consumes trace events and produces debug logs, ensuring consistency
+// between logs and metrics by using the same trace event source.
+type LogReporter struct{}
+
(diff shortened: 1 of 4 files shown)


MarcoPolo:
This looks good, but we should have this use a parameterized slogger. We can default to the default log in the package as you currently do. 

We should also have the tests actually verify the log output.

Thanks!

PR #3442, opened 2025-12-20
fix(peer): skip nil addresses in AddrInfo.MarshalJSON

lidel:
Defensive fix to prevent panic if Addrs slice contains nil entries due to data corruption from concurrent access.

- Related: https://github.com/ipfs/kubo/issues/11116
  - https://github.com/libp2p/go-libp2p-kad-dht/pull/1208

### DIFF
--- a/core/peer/addrinfo_serde.go
+++ b/core/peer/addrinfo_serde.go
@@ -2,6 +2,8 @@ package peer
 
 import (
 	"encoding/json"
+	"fmt"
+	"os"
 
 	"github.com/libp2p/go-libp2p/core/internal/catch"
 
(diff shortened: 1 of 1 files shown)


MarcoPolo:
@lidel would it be more useful to return an error here to help identify the misbehaving code paths? The proposed change is silently hiding the issue.

lidel:
@MarcoPolo good idea, but we really dont want to break API by returning err here.

as a pragmatic compromise I've pushed f2979bd2 which adds a stderr warning when nil is encountered (like `HandlePanic` does, but less noisy/scary), keeping graceful degradation

hope this is acceptable

MarcoPolo:
Reading the linked issue for more context. This results in a caught panic and a returned error. I think that's the correct behavior. I don't think we should work around buggy callers.

The caller can do the same thing as this proposed change by deep cloning the addresses. But if there's concurrent access to the same memory, there's a deeper bug that needs to be fixed. If this is happening in go-libp2p I'd like to know about it.

Issue #3446, opened 2026-01-06
Client-Side QUIC Connection Migration


willianpaixao:
---
name: 'Client-Side QUIC Connection Migration'
about: 'Implement the possibility to switch between primary and secondary connections'
labels: feature
---

Currently, when the underlying network path changes, libp2p connections are dropped and must be re-established. This causes:
- Stream interruptions and data loss
- Reconnection latency
- Application-level disruption

QUIC (RFC 9000) natively supports connection migration, allowing endpoints to change their IP address/port while maintaining the cryptographic connection state. The quic-go library (v0.51.0+) already exposes the necessary APIs (`AddPath`, `Probe`, `Switch`).

#### API Design

**1. Check if a connection supports migration:**
```go
var migratable network.MigratableConn
if conn.As(&migratable) && migratable.SupportsMigration() {
    // This connection supports migration
}
```

**2. Perform a migration:**
```go
// Add a new path using a different local address
path, err := migratable.AddPath(ctx, newLocalAddr)
if err != nil {
    return err
}

// Probe the path to verify connectivity
if err := path.Probe(ctx); err != nil {
    path.Close()
    return err
}

// Switch to the new path
if err := path.Switch(); err != nil {
    return err
}
// Connection is now using the new path
```

**3. Query path information:**
```go
// Get active path info
activeInfo := migratable.ActivePath()
fmt.Printf("Active path: %s -> %s (RTT: %s)\n",
    activeInfo.LocalAddr, activeInfo.RemoteAddr, activeInfo.RTT)

// Get all available paths
for _, pathInfo := range migratable.AvailablePaths() {
    fmt.Printf("Path: %s -> %s (active: %v)\n",
        pathInfo.LocalAddr, pathInfo.RemoteAddr, pathInfo.Active)
}
```

**4. Monitor migration events:**
```go
sub, _ := host.EventBus().Subscribe([]any{
    new(event.EvtConnectionMigrationStarted),
    new(event.EvtConnectionMigrationCompleted),
    new(event.EvtConnectionMigrationFailed),
    new(event.EvtPathAdded),
    new(event.EvtPathRemoved),
})
```

#### Proposed Interfaces

```go
// In core/network/migration.go

type PathInfo struct {
    LocalAddr  ma.Multiaddr
    RemoteAddr ma.Multiaddr
    Active     bool
    RTT        time.Duration
}

type Path interface {
    Probe(ctx context.Context) error
    Switch() error
    Info() PathInfo
[...]

PR #3447, opened 2026-01-06
feat(quic): add experimental client-side connection migration support

willianpaixao:
# This is for testing purposes, do not merge

#3446 explains what I'm trying to archive with this PR.

### Enable migration support

```go
import libp2pquic "github.com/libp2p/go-libp2p/p2p/transport/quic"

host, err := libp2p.New(
    libp2p.Transport(libp2pquic.NewTransport,
        libp2pquic.EnableExperimentalConnectionMigration()),
)
```

### Perform a migration

```go
// Get MigratableConn from any connection
var migratable network.MigratableConn
if conn.As(&migratable) && migratable.SupportsMigration() {
    // Add a new path using a different local address
    path, err := migratable.AddPath(ctx, newLocalAddr)
    if err != nil {
        return err
    }

    // Probe the path to verify connectivity
    if err := path.Probe(ctx); err != nil {
        path.Close()
        return err
    }

    // Switch to the new path
    if err := path.Switch(); err != nil {
        return err
    }
    // Connection is now using the new path, all streams preserved
}
```

### Monitor migration events

```go
sub, _ := host.EventBus().Subscribe([]any{
    new(event.EvtConnectionMigrationStarted),
    new(event.EvtConnectionMigrationCompleted),
    new(event.EvtConnectionMigrationFailed),
})
```

## Constraints

Per the QUIC specification:
- **Only client-initiated (outbound) connections support migration**: server-side connections return `SupportsMigration() == false`
- **Single active path at a time**: QUIC maintains one active path per connection
- **Requires separate UDP socket per interface**: each local address needs its own transport

## Failure Handling

If migration fails:
1. The connection attempts to continue on the original path
2. If the original path is also unavailable, the connection is closed
3. `EvtConnectionMigrationFailed` event is emitted with error details

## Future Work (Out of scope)

- **Automatic migration** based on OS network interface events
- **Automatic retry** with exponential backoff on transient failures
- **Policy configuration** for migration preferences

### DIFF
--- a/core/event/migration.go
+++ b/core/event/migration.go
@@ -1,0 +1,96 @@
+package event
+
+import (
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+// EvtConnectionMigrationStarted is emitted when a connection migration begins.
+// This event would be emitted when AddPath() completes successfully, before
+// the user calls Probe() on the returned path.
+//
+// Note: Event emission is not yet implemented. These types are defined for
+// future use and API stability.
+//
+// This is an EXPERIMENTAL event and may change in future versions.
+type EvtConnectionMigrationStarted struct {
+	// Peer is the remote peer ID of the connection being migrated.
+	Peer peer.ID
+	// ConnID is the unique identifier of the connection.
+	ConnID string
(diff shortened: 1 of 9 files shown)


MarcoPolo:
Thanks for the PR. Let's use the existing Conn.As functionality for this use case for now.

Issue #2496, opened 2023-08-16
UPNP alternative for v6 statefull nats


Jorropo:
Even tho NAT was originally meant as a tool to make IT management easier when merging different company networks after merger and acquisitions, it has been extremely widely deployed due to the IPv4 starvation and people started relying on it for security as they play it loosy goosy on a private network without being exposed to internet (which is terrible horizontal security but whatever).

Asking people to change the config option of their router is not an option for the same reason asking people to port forward does not work and why we have upnp and dcutr.

To meet this security expectation from customers, here where I live (and I expect in most places) all off the shelf residential and business IPv6 deployments ship with stateful IPv6 firewalls by default. It would be nice if we could not rely on DCUTR if the router allows us to do so because the extra hops of DCUTR add latency and make it unsuitable for use-cases like DHT servers.
Alternatives:
- RFC3633 would allow us to request a delegated range, FWIW the 2 routers I have tried this with also add a firewall exception to the delegated range, maybe this is a flag idk ? Overall I have only ever seen it used with `/64` which is a bit overkill, I don't know if it's possible to just request a `/128`. I think this also requires `root` (or equivalent) permissions or some advanced networking capabilities for the libp2p process.
- ... ? ideas wanted

---

Priority: low ~ very low

MarcoPolo:
We should give some love to our upnp components. We could update our go-nat dependency to include support for IPv6.

prkbuilds:
Hey @MarcoPolo , took a little liberty to draft a PR on this work as I thought the requirements were clear. I’ve implemented UPnP WANIPv6FirewallControl1 pinhole support, including lease keepalive and health checks, plus a combined IPv4/IPv6 NAT manager as described in the issue. I’d really appreciate feedback on the overall approach, how this fits into the existing NAT management abstractions, and whether there are any edge cases or configuration scenarios I might have overlooked. If there’s anything I should keep in mind before this is merged (API surface, defaults, test coverage, or docs), please let me know so I can adjust the PR ( #3455 ) accordingly.

PR #3455, opened 2026-01-27
Add UPnP IPv6 pinhole support with keepalive and test tool

prkbuilds:
# Addressing the problem with v6 stateful routers mentioned in #
## Summary
- Add UPnP WANIPv6FirewallControl1 pinhole manager with lease keepalive and health checks
- Add combined IPv4/IPv6 NAT manager for opt‑in use
- Add a small pinhole-check tool and README for manual validation
## Context
This addresses the UPnP AddPinhole approach discussed in #2496 . RFC3633 prefix delegation is intentionally out of scope in this PR.
## Usage
Opt‑in via:
- libp2p.NATManager(basichost.NewUPnPCombinedNATManager)
# AI Assistance
AI assistance used to draft and refine the implementation over my direction.
## My thinking / direction
- The core issue is inbound reachability for IPv6 nodes behind stateful firewalls where DCUTR adds latency.
- UPnP pinholes are a practical, user‑transparent option when routers expose WANIPv6FirewallControl.
- This PR focuses on the pinhole path only; RFC3633 is larger and needs OS‑level privileges and different workflows.
- I kept it opt‑in to avoid changing default behavior until there’s broader validation.
## Question
Is this direction OK, and should I proceed with additional work (e.g., RFC3633 exploration, more robust address change handling, or optional auto‑enable)?


prkbuilds:
@MarcoPolo can you guide me if there are any changes that i need make or something I need to revisit?

examples/go.mod
```diff
@@ -1,72 +1,68 @@
...
-go 1.24
+go 1.24.6
+
+toolchain go1.24.12
 
 require (
 	github.com/caddyserver/certmagic v0.21.6
 	github.com/gogo/protobuf v1.3.2
 	github.com/google/uuid v1.6.0
-	github.com/ipfs/go-datastore v0.6.0
```
MarcoPolo, line 9: please backout all the dependency update changes

p2p/net/nat/upnpv6/cmd/pinhole-check/main.go
```diff
@@ -0,0 +1,57 @@
+package main
```
MarcoPolo, line 1: Wouldn't it make more sense to use the go-libp2p logic for this?

MarcoPolo:
Have you tested this locally with your own router? Can you share your router details?

It would help to see if this works with other home router implementations as well, if you could try others and compile a list that would be great.

PR #3457, opened 2026-01-29
add backwards compatibility for doctur with legacy behavior

jpserrat:
Adding dctur backward compatibility with legacy behavior:
- holepuncher (svc in the new code) now switch to client on third retry.

### DIFF
--- a/p2p/protocol/holepunch/holepuncher.go
+++ b/p2p/protocol/holepunch/holepuncher.go
@@ -136,7 +136,8 @@ func (hp *holePuncher) directConnect(rp peer.ID) error {
 	for i := 1; i <= maxRetries; i++ {
 		isClient := false
 		// On the last attempt we switch roles in case the connection is
-		// being made with a client with switched roles (https://github.com/libp2p/go-libp2p/issues/3171)
+		// being made with a client with switched roles. Common for peers
+		// running go-libp2p prior to v0.41.
 		if i == maxRetries {
 			isClient = true
 		}


jpserrat:
@MarcoPolo I've added the switch in holepuncher. Just a few points:
- I didn't add a random time for the switch, I didn't understand why it is needed. Can you elaborate?
- I've added the switch only on the server (holepuncher.go) as suggested on the issue, but for both clients we have to also add this on `svc.go`. I was think of making a peer store to track peer attempt and switch after some attempts. Do you have any suggestions?

p2p/protocol/holepunch/holepuncher.go
```diff
@@ -7,14 +7,15 @@ import (
 	"sync"
 	"time"
 
+	"github.com/libp2p/go-msgio/pbio"
+	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr/net"
+
```
MarcoPolo: Can you please undo the diffs from the import?

p2p/protocol/holepunch/holepuncher.go
```diff
@@ -141,6 +134,12 @@ func (hp *holePuncher) directConnect(rp peer.ID) error {
 
 	// hole punch
 	for i := 1; i <= maxRetries; i++ {
+		isClient := false
+		// On the last attempt we switch roles in case the connection is
+		// being made with a client with switched roles (https://github.com/libp2p/go-libp2p/issues/3171)
```
MarcoPolo: I like this strategy.
```suggestion
		// On the last attempt we switch roles in case the connection is
		// being made with a client with switched roles. Common for peers
		// running go-libp2p prior to v0.41.
```

Just to make the comment a bit more self contained and not require reading a full issue thread. Git blame will take someone here anyways.

MarcoPolo (approved):

MarcoPolo:
Thank you @jpserrat!

jpserrat:
Thank you @MarcoPolo!

PR #3463, opened 2026-02-13
refactor: apply go fix modernizers from Go 1.26

gammazero:
This PR applies automated refactoring from go 1.26 (go fix ./...):

- interface{} to any, slices.Contains, range loops and other idiomatic updates.

These changes are not necessary, but may be nice to have for improved code readability.

### DIFF
--- a/p2p/host/autorelay/autorelay_test.go
+++ b/p2p/host/autorelay/autorelay_test.go
@@ -317,7 +317,7 @@ func TestConnectOnDisconnect(t *testing.T) {
 	}
 	require.EventuallyWithT(t, func(collect *assert.CollectT) {
 		relaysInUse = usedRelays(h)
-		assert.Len(collect, relaysInUse, 1)
+		require.Len(collect, relaysInUse, 1)
 		assert.NotEqualf(collect, oldRelay, relaysInUse[0], "old relay should not be used again")
 	}, 10*time.Second, 100*time.Millisecond)
 }


sukunrt:
We should do this with go1.25, right? We support go1.25 and go1.26

MarcoPolo:
check failing because the CI is using an older version?

gammazero:
> check failing because the CI is using an older version?

Assuming so. I guess #3462 did not run CI tests?

MarcoPolo:
@galargh, could ipdx update the patch version of Go 1.25 it uses? Otherwise I think we have to fork that workflow :/

MarcoPolo:
ty @gammazero

galargh:
> @galargh, could ipdx update the patch version of Go 1.25 it uses? Otherwise I think we have to fork that workflow :/

FYI, it was https://github.com/actions/setup-go resolving 1.25.x (from https://github.com/libp2p/go-libp2p/blob/master/.github/workflows/go-check.yml#L20) to 1.25.6. If that input was skipped 1.25.7 would have been passed to setup-go action instead.

MarcoPolo:
By skipped you mean if it the `go-version: "1.25.x"` line was removed? Would ipdxco/unified-github-workflows/.github/workflows/go-check.yml use 1.25 in that case and not 1.26?

galargh:
Yeap, it should use the version from https://github.com/libp2p/go-libp2p/blob/master/go.mod#L3 by default

Issue #3460, opened 2026-02-02
#3421 changes the contract of `AllAddrs` method to return all interface addrs and not just the primary interface 


sukunrt:
This breaks the `AllAddrs` method which was previously only returning the 'primary' interface and now returns all interface addrs.

The problem is that the new netroute implementation fails when we query `0.0.0.0` while the previous one returned the primary(default) interface. To be clear just removing the early error in netroute code wont fix this because then the os will return 127.0.0.1 as the interface because it considers 0.0.0.0 to be localhost. https://github.com/libp2p/go-netroute/blob/master/netroute_linux.go#L42 

We've inadvertently ended up implementing https://github.com/libp2p/go-libp2p/issues/3076

_Originally posted by @sukunrt in https://github.com/libp2p/go-libp2p/issues/3421#issuecomment-3834531530_

MarcoPolo:
More context on the change: https://github.com/libp2p/go-netroute/pull/64

sukunrt:
I'm in favor of going with the current change and let users fix whatever problems they have with `AddrsFactory`. AddrsFactory is awkward, but to me it's still better than the surprise of not having all your interfaces reported after listening on `0.0.0.0`. 

pinging @lidel since IPFS would be the most impacted by this.

MarcoPolo:
tl;dr example call from CI env:

Old behavior:
```
h.AllAddrs() // [/ip4/127.0.0.1/tcp/38417 /ip4/127.0.0.1/tcp/39915 /ip4/172.17.0.1/tcp/38417]

```

Current behavior:
```
h.AllAddrs() // [/ip4/10.1.0.204/tcp/38417 /ip4/127.0.0.1/tcp/38417 /ip4/127.0.0.1/tcp/39915 /ip4/172.17.0.1/tcp/38417]
```

I think the latter behavior is correct if you listen on `0.0.0.0`. It is worth noting that you will get more listen addresses returned. 

I'll update the release notes.

lidel:
From Kubo's perspective, fixing VPN/WireGuard interop is a net positive(?)

The main thing to be aware of: many Kubo users run inside Docker, so `AllAddrs()` now includes internal ranges like the Docker bridge (`172.17.0.0/16`). Tailscale users will also advertise `100.64.0.0/10` addresses, though that seems like desired behavior.

[...]

PR #3468, opened 2026-02-26
basichost: advertise all interface addrs for unspecified listen addrs

sukunrt:
commit to #3460


MarcoPolo:
Please update the release notes https://github.com/libp2p/go-libp2p/issues/3470 as well

PR #3471, opened 2026-02-27
fix(mocknet): make stream deadline methods noop instead of returning error 

walldiss:
## Summary                                                                                                                                     
  - Make `SetDeadline`, `SetReadDeadline`, and `SetWriteDeadline` on mocknet streams return nil instead of an error
  - Mocknet uses `io.Pipe` which doesn't support deadlines, but returning an error breaks callers that treat `SetWriteDeadline` failure as fatal
  - Since writes on in-memory pipes complete immediately, deadline enforcement is unnecessary and the noop is safe

  ## Context
  go-libp2p-pubsub added `SetWriteDeadline` calls in [`handleSendingMessages`](https://github.com/libp2p/go-libp2p-pubsub/commit/6587e8e72b35957a53d61dcf91f05b1b9a78bfdf) (part of https://github.com/libp2p/go-libp2p-pubsub/pull/631). The deadline error is treated as fatal. The stream is reset and the hello packet is never sent. This prevents gossipsub peers from discovering each other's subscriptions in any mocknet-based test.

### DIFF
--- a/p2p/net/mock/mock_stream.go
+++ b/p2p/net/mock/mock_stream.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"errors"
 	"io"
-	"net"
 	"strconv"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


p2p/net/mock/mock_stream.go
```diff
@@ -174,17 +173,12 @@ func (s *stream) Conn() network.Conn {
...
-	return &net.OpError{Op: "set", Net: "pipe", Source: nil, Addr: nil, Err: errors.New("deadline not supported")}
-}
-
-func (s *stream) SetWriteDeadline(_ time.Time) error {
-	return &net.OpError{Op: "set", Net: "pipe", Source: nil, Addr: nil, Err: errors.New("deadline not supported")}
-}
+// SetDeadline is a noop for mocknet streams since the underlying pipe
+// transport does not support deadlines. Callers should not treat the
+// absence of deadline support as an error.
+func (s *stream) SetDeadline(_ time.Time) error      { return nil }
```
Wondertan, line 179: Might be worth logging out that its unsupported, if there are any logger available

MarcoPolo:
We recently released the simlibp2p package under `x/simlibp2p`. That's the preferred way to simulate networks in tests as it supports everything a real QUIC transport supports and properly models congestion.

I think we should deprecate our mocknet package. We aren't using it anywhere within go-libp2p, and I don't want to spend time maintaining it when we have a better alternative.

That said, I don't strongly oppose the general idea here, and I'm a bit surprised it's taken this long to find this issue.

> Since writes on in-memory pipes complete immediately, deadline enforcement is unnecessary and the noop is safe

Deadlines aren't only about how fast writes happen. They protect a node from peers that are (possibly maliciously) stalling a stream. 

These changes wouldn't work for testing those cases. But maybe that's fine

Wondertan:
This is a quick way to solve the issue that doesn't require major testing refactoring. We are aware of simnet and that mocknet looms with deprecation.

MarcoPolo:
Please gofmt the file @walldiss, and we can merge this.

walldiss:
There seems to be some CI issue related to testing setup: `go: go.mod requires go >= 1.25.7 (running go 1.24.13;`. @MarcoPolo anything needed from my end?

MarcoPolo:
nope, that's a known issue that I haven't had a chance to fix it.

MarcoPolo:
thanks for the PR

PR #3481, opened 2026-03-23
Add ability to pass through TLS options to the QUIC transport.

MarcoPolo:
Useful to set the KeyLog writer so you can debug the session with Wireshark

### DIFF
--- a/p2p/transport/quic/options.go
+++ b/p2p/transport/quic/options.go
@@ -1,0 +1,19 @@
+package libp2pquic
+
+import p2ptls "github.com/libp2p/go-libp2p/p2p/security/tls"
+
+// Option is a function that configures the QUIC transport.
+type Option func(o *transportConfig) error
+
+type transportConfig struct {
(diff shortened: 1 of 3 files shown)


p2p/transport/quic/options.go
```diff
@@ -0,0 +1,19 @@
...
+// Option is a function that configures the QUIC transport.
+type Option func(o *transportConfig) error
+
+type transportConfig struct {
+	tlsIdentityOpts []p2ptls.IdentityOption
+}
+
+// WithTLSIdentityOption passes the given p2ptls.IdentityOption through to the
+// TLS identity used by the QUIC transport.
+func WithTLSIdentityOption(opt p2ptls.IdentityOption) Option {
```
lidel, line 14: nit: callers wanting several must call it N times. A variadic `WithTLSIdentityOptions(opts ...p2ptls.IdentityOption)` would  read slightly better and match the surrounding variadic style (but fine either way)

sukunrt (approved):

PR #3484, opened 2026-04-04
Improve peer ID parse error context and wrap error

bobsingh-dev:
What does this PR do?

Improve error wrapping and include the peer ID input value in the error message to improve debugging.

Why was this needed?

Including the input improves observability when diagnosing invalid peer IDs. Using %w preserves the error chain.

How was this tested?

go test ./...


gammazero:
Duplicate of #3485

PR #3485, opened 2026-04-04
Fix/error wrap envelope

bobsingh-dev:
What does this PR do?

Wraps error using %w instead of %v to allow proper error unwrapping.

Why is this needed?

Using %w allows callers to use errors.Is and errors.As.

How to test?

go test ./...

### DIFF
--- a/core/peer/peer.go
+++ b/core/peer/peer.go
@@ -127,14 +127,14 @@ func Decode(s string) (ID, error) {
 		// base58 encoded sha256 or identity multihash
 		m, err := mh.FromB58String(s)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


gammazero:
This kind of change should be done more consistently through out the code base and not just limited these few lines. Searching for simple patterns using grep yields many more places where this change could be made.

Use the following command to see more locations where this change could apply:
```
grep -r 'fmt.Errorf' * | grep -v '%w' | grep ', err' | grep -e '%s' -e '%v'
```

sukunrt:
Closing this. it's not a meaningful enough change. See the contributor guidelines.

PR #3486, opened 2026-04-14
feat(pstore): cap unconnected addrs per peer

lidel:
> [!NOTE]
> This is another fix that aims to reduce stale addr accumulation (defensive, on Consumer side).

> [!NOTE]

## Problem

Third-party peers (e.g. via DHT gossip) can flood the peerstore with stale or misconfigured addresses for a single peer id. Real-world delegated-routing responses top out around 26 addrs per well-connected node, but a misbehaving peer can push hundreds before the global peerstore cap kicks in, crowding out good addrs during dial selection.

## This PR

Bound the stored unconnected addrs per peer to 64 by default and evict the unconnected entry with the nearest expiry when the cap is reached. This naturally sheds short-TTL gossip (TempAddrTTL) before longer-lived identify-backed entries (RecentlyConnectedAddrTTL) without tracking provenance. Addrs held by a live connection are never counted toward the cap and never evicted.

- `WithMaxAddressesPerPeer(n)` lets callers tune the cap.
- Enforcement lives in `addAddrsUnlocked` and `SetAddrs`, the two write paths that can grow the per-peer set.
- Shared-suite `SetNegativeTTLClears` adds 100 addrs, above the new default; the inmem test factory raises the cap to 10000 so the shared suite keeps exercising general behavior rather than the cap path.

### Open  Questions

- Is this the right approach? 
- Is 64 the right implicit default limit?

### DIFF
--- a/p2p/host/peerstore/pstoreds/addr_book.go
+++ b/p2p/host/peerstore/pstoreds/addr_book.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"slices"
 	"sort"
 	"sync"
 	"time"
@@ -557,7 +558,7 @@ func (ab *dsAddrBook) setAddrs(p peer.ID, addrs []ma.Multiaddr, ttl time.Duratio
 			return false
 		}
 		delete(addrsMap, string(pr.Addrs[victim].Addr))
-		pr.Addrs = append(pr.Addrs[:victim], pr.Addrs[victim+1:]...)
+		pr.Addrs = slices.Delete(pr.Addrs, victim, victim+1)
 		return true
 	}
 
--- a/p2p/host/peerstore/pstoreds/addr_book.go
+++ b/p2p/host/peerstore/pstoreds/addr_book.go
@@ -389,9 +389,9 @@ func (ab *dsAddrBook) supersededSignedAddrs(p peer.ID, newAddrs []ma.Multiaddr) 
 	return superseded
 }
 
-// ttlIsConnected reports whether the given TTL marks the address as held by
-// a live connection. Such entries are not subject to the per-peer cap and are
-// never evicted to make room for incoming addrs.
+// ttlIsConnected reports whether the TTL marks the address as held by a
+// live connection. Such entries bypass the per-peer cap and survive
+// eviction.
 func ttlIsConnected(ttl time.Duration) bool {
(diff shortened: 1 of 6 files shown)


p2p/host/peerstore/pstoreds/addr_book.go
```diff
@@ -470,6 +470,13 @@ func (ab *dsAddrBook) ClearAddrs(p peer.ID) {
 	}
 }
 
+// ttlIsConnected reports whether the given TTL marks the address as held by
+// a live connection. Such entries are not subject to the per-peer cap and are
+// never evicted to make room for incoming addrs.
+func ttlIsConnected(ttl time.Duration) bool {
+	return ttl >= pstore.ConnectedAddrTTL
+}
+
```
lidel, line 479: 💭 not a fan of this, mostly a placeholder for discussion if we should care about  ttl/connected state, or just simplify code of this PR to simply sort by TTL and drop the ones that will expire the fastest

p2p/host/peerstore/pstoreds/addr_book.go
```diff
@@ -517,6 +524,43 @@ func (ab *dsAddrBook) setAddrs(p peer.ID, addrs []ma.Multiaddr, ttl time.Duratio
...
+			if victim == -1 || a.Expiry < soonest {
+				victim = i
+				soonest = a.Expiry
+			}
+		}
+		if victim == -1 {
+			return false
+		}
+		delete(addrsMap, string(pr.Addrs[victim].Addr))
+		pr.Addrs = append(pr.Addrs[:victim], pr.Addrs[victim+1:]...)
```
MarcoPolo, line 560:
```suggestion
		pr.Addrs = slices.Delete(pr.Addrs, victim, victim+1)
```

MarcoPolo (approved):

p2p/host/peerstore/pstoreds/addr_book.go
MarcoPolo: This is fine and consistent with how pstoremem does it: https://github.com/libp2p/go-libp2p/blob/254e1a995af283f67adc2be911bc467e6faf4c64/p2p/host/peerstore/pstoremem/addr_book.go#L42

MarcoPolo:
@lidel, could you please take a look at the conflicts?

lidel:
@MarcoPolo done :)

PR #3487, opened 2026-04-15
fix(peerstore): replace stale addrs on newer signed peer record

lidel:
> [!NOTE]
> This is another fix that aims to reduce stale addr accumulation (defensive, on Consumer side).

By looking at commented code, tt seems there was intention to do it in the past.

## This PR

When we deal with signed peer records that have monotonic `Seq`, treat `ConsumePeerRecord` as replace-semantics rather than merge.

When a peer publishes a newer signed peer record, addrs that were present in the previously stored signed record but omitted from the new one are evicted, so the peerstore reflects the peer's current self-advertised set instead of the accumulated union. Unsigned addrs (e.g. DHT gossip or identify exchanges without a signed record) are untouched, and addrs held by a live connection (TTL >= ConnectedAddrTTL) are kept so active sessions are not torn down.

Both backends recover the prior addr set by decoding the stored envelope: `pstoremem` reuses the Envelope already kept on `peerRecordState`; `pstoreds` fetches it via `GetPeerRecord` and drops superseded addrs with `deleteAddrs`. `Envelope.Record()` caches on first access, so the diff is cheap. The shared `CertifiedAddresses` assertion is updated to match the new behavior.

> [!NOTE]

### DIFF
--- a/p2p/host/peerstore/pstoreds/addr_book.go
+++ b/p2p/host/peerstore/pstoreds/addr_book.go
@@ -282,7 +282,21 @@ func (ab *dsAddrBook) AddAddrs(p peer.ID, addrs []ma.Multiaddr, ttl time.Duratio
 
 // ConsumePeerRecord adds addresses from a signed peer.PeerRecord (contained in
 // a record.Envelope), which will expire after the given TTL.
-// See https://godoc.org/github.com/libp2p/go-libp2p/core/peerstore#CertifiedAddrBook for more details.
+// See https://godoc.org/github.com/libp2p/go-libp2p/core/peerstore#CertifiedAddrBook
+// for more details.
+//
+// The signed peer record's Seq is treated as monotonic per peer: a record
+// with a Seq lower than the last accepted one is rejected. Equal Seq is
+// accepted as a TTL refresh.
+//
+// When a newer signed record is accepted, addrs that were present in the
+// previously stored signed record but absent in the new one are evicted, so
+// the peerstore reflects the peer's current self-advertised set instead of
+// the union of every record we have ever seen. Unsigned addrs (added via
+// AddAddr / SetAddr from sources like DHT gossip, or from an identify
+// exchange where the peer did not send a signed record) are not touched, and
+// addrs held by a live connection (TTL >= ConnectedAddrTTL) are also kept so
+// active sessions are not dropped.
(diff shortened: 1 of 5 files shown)


p2p/host/peerstore/test/addr_book_suite.go
```diff
@@ -475,8 +475,8 @@ func testCertifiedAddresses(m pstore.AddrBook, clk *mockClock.Mock) func(*testin
 			t.Error("unable to retrieve signed routing record from addrbook")
 		}
 
-		// Adding a new envelope should clear existing certified addresses.
-		// Only the newly-added ones should remain
```
lidel, line 479: ℹ️ intention was documented, but never implemented. this PR aims to implement it

this PR does not purge all addrs, only one matching previous signed record, but we could also pivot it to do full clean so only signed addrs remain

(not feeling strongly, whatever feels better)

MarcoPolo: replacing it makes sense. The full purge is a bigger riskier change I think

p2p/host/peerstore/pstoreds/addr_book.go
```diff
@@ -315,6 +338,62 @@ func (ab *dsAddrBook) ConsumePeerRecord(recordEnvelope *record.Envelope, ttl tim
...
+	pr.RLock()
+	connected := make(map[string]struct{})
+	for _, a := range pr.Addrs {
+		if ttlIsConnected(time.Duration(a.Ttl)) {
+			connected[string(a.Addr)] = struct{}{}
+		}
+	}
+	pr.RUnlock()
+
+	superseded := make([]ma.Multiaddr, 0, len(prevRec.Addrs))
```
MarcoPolo, line 377: A very minor nit: if you return an iter.Seq you avoid this allocation.

MarcoPolo (approved):

p2p/host/peerstore/pstoreds/addr_book.go
lidel: I'd lean toward keeping the slice here: caller gates on `len(superseded) > 0` and `deleteInPlace` iterates the set per surviving entry, so `iter.Seq` would just push the alloc into `deleteAddrs` or force a signature change.. feels like not worth the noise.

PR #3489, opened 2026-04-16
feat:  NonPublicAddrPublishing option

lidel:
> [!NOTE]
> This is another fix that aims to reduce stale addr accumulation (proactive, on Producer side).

new opt-in `libp2p.NonPublicAddrPublishing(bool)` lets hosts exclude addresses that are not in a globally-routable range from the peerstore self-entry and the signed peer record so they are not advertised via identify or DHT.

scope mirrors `!manet.IsPublicAddr`: RFC 1918 private, RFC 6598 CGNAT, link-local, loopback, ULA, IPv6 documentation/multicast/reserved space (including unallocated prefixes like `1e::/16` observed in DHT records). multiaddrs without an IP component such as /p2p-circuit are kept; `.local` and other special-use DNS names are stripped.

default is `true` (current behavior preserved). when set to false the `addrsManager` strips matching addrs before `SetAddrs` and before signing the peer record.

## Alternative

Without this PR Kubo has to do the filtering in userland

## Context

I think this all started with #3421 (cc @sukunrt @MarcoPolo)

Since PR #3421 (update go-netroute dependency), `AllAddrs` now returns every interface address (including `127.0.0.1`?) instead of just the primary. That feeds straight into `addrsManager.updatePeerStore`, which is what this PR aims to filter out. #3460 tracks this too.

In IPFS Mainnet we see peers reporting their loopback addrs via DHT or identify, noticed it in error reports like:

> https://check.ipfs.network/?cid=QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb&multiaddr=%2Fp2p%2FQmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb&ipniIndexer=https%3A%2F%2Fcid.contact&timeoutSeconds=30&httpRetrieval=on:
> image

If you let it run for a while, we will also see garbage(?) addresses `1e::` from IANA-reserved `0000::/3` appear:

> image

> [!NOTE]

### DIFF
--- a/p2p/host/basic/addrs_manager.go
+++ b/p2p/host/basic/addrs_manager.go
@@ -395,11 +395,10 @@ func filterPublicAddrs(addrs []ma.Multiaddr) []ma.Multiaddr {
 // guard, filterPublicAddrs would also drop multiaddrs that have no
 // network-layer address, such as /p2p-circuit/p2p/<id>.
 func hasIPOrDNSComponent(addr ma.Multiaddr) bool {
-	first, _ := ma.SplitFirst(addr)
-	if first == nil {
+	if len(addr) == 0 {
 		return false
 	}
-	switch first.Protocol().Code {
+	switch addr[0].Protocol().Code {
 	case ma.P_IP4, ma.P_IP6, ma.P_IP6ZONE, ma.P_DNS, ma.P_DNS4, ma.P_DNS6, ma.P_DNSADDR:
 		return true
 	}
--- a/p2p/host/basic/addrs_manager.go
+++ b/p2p/host/basic/addrs_manager.go
@@ -71,10 +71,11 @@ type addrsManager struct {
 	addrsMx      sync.RWMutex
 	currentAddrs hostAddrs
 
-	signKey           crypto.PrivKey
-	addrStore         addrStore
-	signedRecordStore peerstore.CertifiedAddrBook
-	hostID            peer.ID
+	signKey                        crypto.PrivKey
+	addrStore                      addrStore
+	signedRecordStore              peerstore.CertifiedAddrBook
+	hostID                         peer.ID
+	disableNonPublicAddrPublishing bool
 
 	wg        sync.WaitGroup
 	ctx       context.Context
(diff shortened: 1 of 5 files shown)


MarcoPolo:
As I mentioned in  https://github.com/libp2p/go-libp2p/issues/3460, If you listen on 0.0.0.0, and then ask for all addresses it is correct to return all addresses including 127.0.0.1 as those are all addresses that you are listening on.

That said, I'm not opposed to this option. Especially since I don't love the AddrsFactory interface. I'll review soon, thanks for the PR

p2p/host/basic/addrs_manager.go
```diff
@@ -343,9 +346,19 @@ func (a *addrsManager) updateAddrs(prevHostAddrs hostAddrs, relayAddrs []ma.Mult
...
+		publishedAddrs, excludedAddrs = splitNonPublicAddrs(currentAddrs)
+	}
+	// Set the addrs we want to publish with a permanent TTL.
+	a.addrStore.SetAddrs(a.hostID, publishedAddrs, peerstore.PermanentAddrTTL)
+	// Clear previously published addrs that are no longer current, as well as
+	// any addrs being withheld because they are non-public. Use separate calls
+	// to avoid appending into the caller's removedAddrs backing array, which
+	// could race with or overwrite its state.
 	a.addrStore.SetAddrs(a.hostID, removedAddrs, 0)
+	a.addrStore.SetAddrs(a.hostID, excludedAddrs, 0)
```
MarcoPolo, line 361: Why is this needed? When would these addresses have been added in the first place?

p2p/host/basic/addrs_manager.go
```diff
@@ -366,6 +379,40 @@ func (a *addrsManager) updatePeerStore(currentAddrs []ma.Multiaddr, removedAddrs
...
+	return
+}
+
+// hasIPOrDNSComponent reports whether addr contains an IP or DNS component.
+// Without this guard, splitNonPublicAddrs would also drop multiaddrs that
+// have no network-layer address for manet.IsPublicAddr to evaluate, such as
+// /p2p-circuit/p2p/<id>.
+func hasIPOrDNSComponent(addr ma.Multiaddr) bool {
+	found := false
+	ma.ForEach(addr, func(c ma.Component) bool {
```
MarcoPolo, line 405: Do we need this to be a for each? Would comparing the first component suffice?

lidel: Switched to `ma.SplitFirst` in 01ea6502. 
Should be fine, there was only zoned link-local ipv6, which is encoded as `/ip6zone/<zone>/ip6/fe80::1/...`, so I included `P_IP6ZONE` in the accepted codes (matching `manet.IsPublicAddr`, which continues past it to the wrapped ip6) and added an `ip6zone` case to the test.

p2p/host/basic/addrs_manager.go
lidel: The intention was to cover a persistent peerstore carried over from a prior run that didn't have the option enabled, but now that I think about it it's unlikely to be useful +only clears addrs that happen to appear in the current `excludedAddrs`, so if the node moved networks between runs the stale ones aren't matched anyway. 

Dropped the call in 01ea6502. 
Also renamed `splitNonPublicAddrs` to `filterPublicAddrs` since it now returns only the kept set.

p2p/host/basic/addrs_manager.go
```diff
@@ -366,6 +372,40 @@ func (a *addrsManager) updatePeerStore(currentAddrs []ma.Multiaddr, removedAddrs
...
+// at the front, so the leading component is sufficient to tell whether
+// manet.IsPublicAddr can meaningfully evaluate the addr. Without this
+// guard, filterPublicAddrs would also drop multiaddrs that have no
+// network-layer address, such as /p2p-circuit/p2p/<id>.
+func hasIPOrDNSComponent(addr ma.Multiaddr) bool {
+	first, _ := ma.SplitFirst(addr)
+	if first == nil {
+		return false
+	}
+	switch first.Protocol().Code {
```
MarcoPolo:
```suggestion
	if len(addr) == 0 {
		return false
	}
	switch addr[0].Protocol().Code {
```

MarcoPolo: SplitFirst does a defensive copy on the second arg that you are ignoring. This is more efficient and direct.

MarcoPolo: changed in 6e25992c44d1

MarcoPolo (approved):

PR #3490, opened 2026-04-19
fix(routing): prevent QueryEvent publish races

lidel:
## The bug

`PublishQueryEvent` forwards `QueryEvent.Responses` to subscribers by pointer. A publisher that keeps mutating its `[]*peer.AddrInfo` (or any `AddrInfo.Addrs`) after the call returns races with every subscriber reading the event.

This is the canonical shape in routing systems, for example DHT implementations: a worker aggregates "closer peers" into a slice, publishes it as a  PeerResponse` event, then continues processing the same slice (e.g. enriching `Addrs` from a peerstore).

This class of race was reported against in ipfs/kubo#11287 and ipfs/kubo#11116.

## Proof

To make review easier, the first commit (`test: PublishQueryEvent Responses race`) adds a minimal test that fails under `-race`. CI caught it on `go-test` before the fix:

https://github.com/libp2p/go-libp2p/actions/runs/24641378021/job/72045816212?pr=3490#step:19:580

## The fix

> [!NOTE]
> This is a general fix, not a kad-dht-specific one. Any current or future routing/DHT implementation that emits `QueryEvent`s can hit the same race; centralizing the copy in `core/routing` covers all of them at once. I feel this is the right call to avoid similar bugs surfacing every few months when someone moves something around, or if golang changes something to make race more probable.

`PublishQueryEvent` now deep-copies `Responses` (and each `AddrInfo.Addrs`) before handing the event to subscribers. Callers get a strengthened contract: mutate your own copy freely after publishing.

- Zero cost when nobody subscribes (the existing `ctx.Value` early return is untouched)
- `Multiaddr` values stay shared; they are by-convention immutable
- No API break

> [!NOTE]

### DIFF
--- a/core/routing/query.go
+++ b/core/routing/query.go
@@ -2,6 +2,7 @@ package routing
 
 import (
 	"context"
+	"slices"
 	"sync"
 
 	"github.com/libp2p/go-libp2p/core/peer"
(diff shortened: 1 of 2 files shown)


guillaumemichel (approved):
The fix looks sound!

MarcoPolo (approved):

PR #3494, opened 2026-04-27
fix(peer): drop empty addrs in peer records

lidel:
A zero-value Multiaddr (since go-multiaddr v0.15 is a slice type) has no components and encodes to zero bytes on the wire. go-libp2p was sealing such empty entries into signed peer records, where peers that skip the empty-input check render them as "/" and reject the address.

## Changes

- core/peer/record.go: skip empty addrs in addrsToProtobuf, and in addrsFromProtobuf as a defensive read
- p2p/host/basic/addrs_manager.go: drop empty addrs in makeSignedPeerRecord before sealing the envelope

go-libp2p's own decoder errors on zero-byte input via NewMultiaddrBytes, but other implementations such as js-libp2p do not.

- See https://github.com/libp2p/js-libp2p/issues/3478#issuecomment-4322093929

> [!NOTE]


guillaumemichel (approved):
Fix looks good

Test failure seem unrelated to recent changes

MarcoPolo (approved):

PR #3497, opened 2026-05-01
Allow using go-libp2p with fewer mandatory dependencies

Jmohak2004:
## Problem

At the moment, simply importing go-libp2p (or a project that uses it) pulls in a VERY large set of indirect dependencies. These dependencies are all used even if only a single transport is enabled with `libp2p.ChainOptions(libp2p.NoTransports, libp2p.Transport(tcp.NewTCPTransport))`.

For example, even if only `tcp` is used, dependencies for all available transports are pulled in due to all transports being referenced in `DefaultTransports` variable in defaults.go.

Even if `DefaultPrivateTransports` and `DefaultTransports` are modified to avoid importing unwanted transports, it was impossible to avoid pulling in `github.com/quic-go/quic-go` as the `quicreuse` transport was baked deeply into the base and config packages.

## Solution

This PR refactors the codebase to make transport dependencies optional:

### Changes Made:

1. **Separated transport imports**: Moved transport imports from `defaults.go` to a new `default_transports.go` file. This prevents forcing all transport dependencies when users specify custom transports.

2. **Isolated QUIC-specific code**: 
   - Moved `QUICReuse` function from `options.go` to new `options_quic.go`
   - Moved QUIC-related code from `config/config.go` to new `config/config_quic.go`
   - Moved `PrivKeyToStatelessResetKey` and `PrivKeyToTokenGeneratorKey` functions to `config_quic.go`

3. **Removed mandatory imports**: Eliminated direct imports of `quic-go` and `quicreuse` from main package files (`options.go`, `config/config.go`)

### Benefits:

- Users who only use TCP transport no longer pull in QUIC, WebSocket, WebTransport, and WebRTC dependencies
- Eliminates mandatory `quic-go` dependency (~1.6MB) when QUIC is not used
- Reduces binary size by ~8MB when unused transports are excluded
- **Maintains full backward compatibility** - `DefaultTransports` still works exactly as before

### Impact:

The transport abstractions no longer leak into the base Options and Config packages. Users can now use go-libp2p with significantly fewer dependencies when they only need a subset of transports.

## Testing

The changes maintain backward compatibility. All existing code using `DefaultTransports` will continue to work. Users who want to use only specific transports can now do so without pulling in unnecessary dependencies:

```go
// Only TCP - no QUIC dependencies pulled in
h, err := libp2p.New(
    libp2p.NoTransports,
    libp2p.Transport(tcp.NewTCPTransport),
)
```

Fixes the issue described in: Allow using go-libp2p with fewer mandatory dependencies


MarcoPolo:
Take a look at this PR: https://github.com/libp2p/go-libp2p/pull/3386 and share feedback there please. I think it better addresses your goals.

That code does import all the transports, but it would be trivial to split that into a separate package do get the same effect.

PR #3498, opened 2026-05-06
fix(config): emit fx DI registration events at debug level

rvagg:
`fxevent.SlogLogger` defaults to `slog.LevelInfo` for non-error events (`Provided`, `Supplied`, `Replaced`, etc.), so each constructor in the host's DI graph emits an INFO line at `host.New(...)`, appro 70 lines per host construction under logger `p2p-config`.

Move them to DEBUG via `UseLogLevel`. Errors keep their default ERROR level. Graph remains available via `GOLOG_LOG_LEVEL=p2p-config=debug`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
+	"log/slog"
 	"net"
 	"slices"
 	"time"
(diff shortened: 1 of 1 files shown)


MarcoPolo (approved):

MarcoPolo:
Thank you

PR #3501, opened 2026-05-07
p2p/protocol/identify: bound the number of protocols accepted from peers

Sahil-4555:
### Problem
The identify protocol supports chunked message delivery for large payloads - when a signed peer record exceeds the legacy 2KB limit, `writeChunkedIdentifyMsg` splits it across multiple protobuf messages. On the receiving side, `readAllIDMessages` reads up to 10 chunks and merges them using `proto.Merge`. The issue is that `proto.Merge` appends repeated fields rather than replacing them, so the Protocols repeated field in the final merged message grows cumulatively across all chunks.

A remote peer can send 9 chunked identify responses (the 10th read slot is consumed by the EOF), each containing hundreds of short protocol strings like `/a/0`, `/a/1`, etc. After merging, all protocols are passed directly to `consumeMessage`, which calls `peerstore.SetProtocols(peer, protocols...)` with no upper bound on the protocol count. This stores an arbitrarily large protocol list in the victim's peerstore, wasting memory and degrading the performance of `GetProtocols()` and `SupportsProtocols()` lookups for that peer.

The address list was already correctly bounded - `consumeMessage` checks `if len(addrs) > connectedPeerMaxAddrs` and truncates to 500. No equivalent bound existed for protocols. A malicious peer could store 1,800+ protocol entries from a single identify exchange while addresses are capped at 500.

### Fix
This patch adds a `maxPeerProtocols` constant (set to 1024) alongside the existing address constants in `id.go`, and inserts a truncation check in `consumeMessage` immediately after converting the protocol strings. If the count exceeds the limit, a warning is logged and the list is truncated to `maxPeerProtocols`. The limit of 1024 is generous - real-world libp2p nodes typically advertise fewer than 50 protocols - but provides a hard upper bound that prevents memory amplification while leaving headroom for legitimate use cases.

### Tests
The accompanying test file `protocol_limit_test.go` contains four tests that reproduce the issue and verify the fix:
- `TestProtocolListFromSingleChunk`: A normal single-chunk identify message with 100 protocols passes through unchanged. This is a sanity check that the truncation doesn't break the common case.
- `TestChunkedIdentifyAmplifiesProtocolList`: Sends 9 chunks of 200 protocols each through `readAllIDMessages` and confirms the merge produces 1,800 protocols. This directly demonstrates the amplification - before this fix, all 1,800 would go straight into the peerstore with no cap.
- `TestProtocolListTruncation`: Verifies that a protocol list exceeding `maxPeerProtocols` is correctly truncated to exactly 1,024 entries, exercising the same guard we added in `consumeMessage`.
- `TestChunkedMergeMemoryCost`: Measures the serialized protobuf size before and after truncation. With 1,800 protocols the message is ~17KB; after capping to 1,024 it drops to ~9.6KB, saving ~7.3KB per peer.

### DIFF
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -733,6 +733,7 @@ func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn, isPush bo
 	if len(mesProtocols) > maxPeerProtocols {
 		log.Warn("peer advertises too many protocols, truncating",
 			"peer", p, "advertised", len(mesProtocols), "limit", maxPeerProtocols)
+		clear(mesProtocols[maxPeerProtocols:])
 		mesProtocols = mesProtocols[:maxPeerProtocols]
 	}
 	added, removed := diff(supported, mesProtocols)
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -731,7 +731,7 @@ func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn, isPush bo
 	supported, _ := ids.Host.Peerstore().GetProtocols(p)
 	mesProtocols := protocol.ConvertFromStrings(mes.Protocols)
 	if len(mesProtocols) > maxPeerProtocols {
-		log.Warn("peer advertises too many protocols, truncating",
+		log.Debug("peer advertises too many protocols, truncating",
 			"peer", p, "advertised", len(mesProtocols), "limit", maxPeerProtocols)
 		clear(mesProtocols[maxPeerProtocols:])
 		mesProtocols = mesProtocols[:maxPeerProtocols]
--- a/p2p/protocol/identify/id.go
+++ b/p2p/protocol/identify/id.go
@@ -55,6 +55,10 @@ const (
 	// localhost, private IP or public IP address
 	recentlyConnectedPeerMaxAddrs = 20
 	connectedPeerMaxAddrs         = 500
+	// maxPeerProtocols is the maximum number of protocols we store for a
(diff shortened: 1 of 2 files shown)


p2p/protocol/identify/id.go
```diff
@@ -726,6 +730,11 @@ func (ids *idService) consumeMessage(mes *pb.Identify, c network.Conn, isPush bo
 
 	supported, _ := ids.Host.Peerstore().GetProtocols(p)
 	mesProtocols := protocol.ConvertFromStrings(mes.Protocols)
+	if len(mesProtocols) > maxPeerProtocols {
+		log.Warn("peer advertises too many protocols, truncating",
+			"peer", p, "advertised", len(mesProtocols), "limit", maxPeerProtocols)
+		mesProtocols = mesProtocols[:maxPeerProtocols]
```
MarcoPolo, line 737: Does this actually do what you want? You still keep a reference to the full backing array. At very least you probably want to `clear(mesProtocols[maxPeerProtocols:])` first.

Sahil-4555: you're right that the re-slice keeps the full backing array alive. ConvertFromStrings allocates with `make([]ID, 0, len(ids))`, so the backing array is sized to the original (potentially 1800+) protocol count. The unused slots past index 1024 still hold `protocol.ID` string pointers, pinning their data.

Added `clear(mesProtocols[maxPeerProtocols:])` before the re-slice to zero those out. Worth noting that `mes.Protocols` (the protobuf source) also references the same strings, so the GC can't fully collect until `mes` goes out of scope too - but no reason for our slice to hold redundant references either way.

MarcoPolo (approved):

PR #3502, opened 2026-05-07
p2p/net/swarm: make connectedness event emitter AddConn non-blocking to prevent stalls

Sahil-4555:
### Problem
While reviewing the swarm's connection handling, we spotted a bug where a new connection could completely freeze up. When a peer connects to us (or we connect to them), the swarm tries to announce this by emitting a "Connectedness Changed" event. To do this, it calls `AddConn`, which pushes the peer's ID into a small queue (a channel that only holds 32 items). A background worker then processes this queue.

The problem was that the push to this queue was a blocking operation. If the background worker got bogged down—maybe because something else in the node was slow to handle the event - the queue would fill up. Once it hit 32 items, the next connection trying to announce itself would just sit there waiting forever.

Because the swarm locks the specific connection while it tries to make this announcement, waiting forever means the connection itself is totally frozen. Even worse, it also messes up the swarm's internal counter, permanently leaking a reference. This leaked reference actually prevents the node from ever shutting down cleanly because it's endlessly waiting for a connection that's stuck.

### Fix
e changed `AddConn` so it no longer waits around if the queue is full. Instead, it checks if there is room. If there is, it pushes the event like normal. If the queue is already packed at 32 items, it simply drops the event, prints a warning message to the logs, and immediately moves on. It's much safer to drop a late notification than to completely lock up the connection and permanently break the node's ability to shut down cleanly.

### Tests
We added a new test called `TestAddConnDoesNotBlockWhenChannelIsFull` inside `connectedness_event_emitter_test.go`. The test simulates the exact conditions of the bug: it creates a queue, deliberately ignores the background worker, and blasts 100 connection events into it as fast as possible. This proves that once the queue hits its limit of 32, our code safely drops the remaining 68 events and finishes instantly, rather than locking up the test.

### DIFF
--- a/p2p/net/swarm/connectedness_event_emitter.go
+++ b/p2p/net/swarm/connectedness_event_emitter.go
@@ -48,17 +48,19 @@ func newConnectednessEventEmitter(connectedness func(peer.ID) network.Connectedn
 	return c
 }
 
-func (c *connectednessEventEmitter) AddConn(p peer.ID) {
+func (c *connectednessEventEmitter) AddConn(p peer.ID) bool {
 	c.mx.RLock()
 	defer c.mx.RUnlock()
 	if c.ctx.Err() != nil {
-		return
+		return false
 	}
 
 	select {
 	case c.newConns <- p:
+		return true
 	default:
-		log.Warn("connectedness event emitter channel full, dropping event", "peer", p)
+		log.Error("connectedness event emitter channel full, dropping event", "peer", p)
+		return false
 	}
 }
 
--- a/p2p/net/swarm/connectedness_event_emitter.go
+++ b/p2p/net/swarm/connectedness_event_emitter.go
@@ -14,9 +14,11 @@ import (
 // the peer disconnects. This is because peers can observe a connection before they are notified
 // of the connection by a peer connectedness changed event.
 type connectednessEventEmitter struct {
-	mx sync.RWMutex
-	// newConns is the channel that holds the peerIDs we recently connected to
-	newConns      chan peer.ID
+	mx         sync.RWMutex
+	newConnsMx sync.Mutex
+	// newConns is a slice of peerIDs we have recently connected to
+	newConns []peer.ID
+
 	removeConnsMx sync.Mutex
 	// removeConns is a slice of peerIDs we have recently closed connections to
 	removeConns []peer.ID
@@ -27,6 +29,7 @@ type connectednessEventEmitter struct {
 	// emitter is the PeerConnectednessChanged event emitter
 	emitter         event.Emitter
 	wg              sync.WaitGroup
+	newConnNotif    chan struct{}
 	removeConnNotif chan struct{}
 	ctx             context.Context
 	cancel          context.CancelFunc
@@ -35,8 +38,8 @@ type connectednessEventEmitter struct {
 func newConnectednessEventEmitter(connectedness func(peer.ID) network.Connectedness, emitter event.Emitter) *connectednessEventEmitter {
 	ctx, cancel := context.WithCancel(context.Background())
 	c := &connectednessEventEmitter{
-		newConns:        make(chan peer.ID, 32),
 		lastEvent:       make(map[peer.ID]network.Connectedness),
+		newConnNotif:    make(chan struct{}, 1),
 		removeConnNotif: make(chan struct{}, 1),
 		connectedness:   connectedness,
 		emitter:         emitter,
@@ -48,19 +51,24 @@ func newConnectednessEventEmitter(connectedness func(peer.ID) network.Connectedn
 	return c
 }
 
-func (c *connectednessEventEmitter) AddConn(p peer.ID) bool {
+func (c *connectednessEventEmitter) AddConn(p peer.ID) {
 	c.mx.RLock()
 	defer c.mx.RUnlock()
 	if c.ctx.Err() != nil {
-		return false
+		return
 	}
 
+	c.newConnsMx.Lock()
+	// This queue is roughly bounded by the total number of added connections we
+	// have. If consumers of connectedness events are slow, we queue connection
+	// notifications without blocking the connection setup path on a fixed-size
+	// channel.
+	c.newConns = append(c.newConns, p)
+	c.newConnsMx.Unlock()
+
 	select {
-	case c.newConns <- p:
-		return true
+	case c.newConnNotif <- struct{}{}:
 	default:
-		log.Error("connectedness event emitter channel full, dropping event", "peer", p)
-		return false
 	}
 }
 
@@ -73,8 +81,8 @@ func (c *connectednessEventEmitter) RemoveConn(p peer.ID) {
 
 	c.removeConnsMx.Lock()
 	// This queue is roughly bounded by the total number of added connections we
-	// have. If consumers of connectedness events are slow, we apply
-	// backpressure to AddConn operations.
+	// have. If consumers of connectedness events are slow, queued connection
+	// notifications remain bounded by the swarm's connection limits.
 	//
 	// We purposefully don't block/backpressure here to avoid deadlocks, since it's
 	// reasonable for a consumer of the event to want to remove a connection.
@@ -97,20 +105,17 @@ func (c *connectednessEventEmitter) runEmitter() {
 	defer c.wg.Done()
 	for {
 		select {
-		case p := <-c.newConns:
-			c.notifyPeer(p, true)
+		case <-c.newConnNotif:
+			c.sendConnAddedNotifications()
 		case <-c.removeConnNotif:
 			c.sendConnRemovedNotifications()
 		case <-c.ctx.Done():
 			c.mx.Lock() // Wait for all pending AddConn & RemoveConn operations to complete
 			defer c.mx.Unlock()
 			for {
-				select {
-				case p := <-c.newConns:
-					c.notifyPeer(p, true)
-				case <-c.removeConnNotif:
-					c.sendConnRemovedNotifications()
-				default:
+				added := c.sendConnAddedNotifications()
+				removed := c.sendConnRemovedNotifications()
+				if !added && !removed {
 					return
 				}
 			}
@@ -138,7 +143,18 @@ func (c *connectednessEventEmitter) notifyPeer(p peer.ID, forceNotConnectedEvent
 	}
 }
 
-func (c *connectednessEventEmitter) sendConnRemovedNotifications() {
+func (c *connectednessEventEmitter) sendConnAddedNotifications() bool {
+	c.newConnsMx.Lock()
+	newConns := c.newConns
+	c.newConns = nil
+	c.newConnsMx.Unlock()
+	for _, p := range newConns {
+		c.notifyPeer(p, true)
+	}
+	return len(newConns) > 0
+}
+
+func (c *connectednessEventEmitter) sendConnRemovedNotifications() bool {
 	c.removeConnsMx.Lock()
 	removeConns := c.removeConns
 	c.removeConns = nil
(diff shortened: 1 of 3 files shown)


MarcoPolo:
We should probably tear down the connection if this happens as well as log an error. It is an application error to have a slow consumer of this event.

Sahil-4555:
Yeah, that's a fair point, a dropped connectedness event basically creates a ghost connection that lives in the conn map but that nobody upstream knows about. That's arguably worse than rejecting the connection outright.

Updated `AddConn` to return a `bool` so the caller can react. If it returns `false`, `swarm.addConn` now releases the notification lock, closes the connection, and returns an error. Also bumped the log level to `Error` since hitting this path means something is genuinely wrong with event consumption.

The teardown is safe from a lock ordering perspective - `s.conns` is already released at line 425, and `c.Close()` -> `doClose` spawns the disconnect notification goroutine which waits on `notifyLk` (which we unlock right before closing).

sukunrt:
>Because the swarm locks the specific connection while it tries to make this announcement, waiting forever means the connection itself is totally frozen.

That's intentional we want to wait for the announcement before starting the connection. Maybe it's okay to not wait here, but this has been the behavior since forever. Note, a few lines afer the call to `AddConn` we make a call to `Connected` notifees which might also stall for a while.  

>This leaked reference actually prevents the node from ever shutting down cleanly because it's endlessly waiting for a connection that's stuck.

Do you have a repro for this? What reference is being leaked. All connections added to the swarm conns map must be notified to the application via eventbus because they're notified via the `Connected` callback. `swarm.Close` will call the `Disconnected` callback so we need to ensure that we've sent the `Connected` 

@MarcoPolo 
>It is an application error to have a slow consumer of this event.
A slow event consumer should be okay. Why is it an error?. 

Even if it's an error the channel being full is fine. The total number of waiting goroutines are bounded by the rcmgr limits. It's not okay to drop the connections with a channel capacity of 32. If you make 200 dials in parallel and 100 of those connections connect concurrently, you'll drop (100-32) connections here. Maybe increasing the channel capacity to 10k is fine as the total number is bounded by the rcmgr limits.

Sahil-4555:
I reworked this after looking more closely at the ordering here. I think you’re right that dropping the event, or closing the connection from this path, is the wrong tradeoff. We should keep the existing behavior where the connectedness event is queued before the connection is started and before `Connected` notifiees run.

Rather than making `AddConn` non-blocking, or just changing the channel size from 32 to some larger fixed number, I changed `newConns` to follow the same shape as `removeConns`: keep pending peers in a mutex-protected slice, and use a 1-slot notification channel to wake the emitter goroutine. That gets rid of the fixed channel capacity without dropping events, rejecting valid connections, or allocating a large buffer per swarm.

On the ref point, the ref I meant is the second one from `s.refs.Add(2)`, which is normally consumed when `Conn.start` exits. If AddConn blocked forever before `c.start()` was reached, that ref would never be consumed. I am attaching a small repro for that , this change removes the fixed-size queue as the thing that can cause that stall.

I also updated the test to match the new behavior: it queues a burst while the emitter is not draining, checks that `AddConn` doesn’t block, then drains the queue and verifies all connectedness events are emitted.

Sahil-4555:
```go
package swarm

import (
	"context"
	"errors"
	"fmt"
	"sync"
	"testing"
	"time"

	ic "github.com/libp2p/go-libp2p/core/crypto"
	"github.com/libp2p/go-libp2p/core/event"
	"github.com/libp2p/go-libp2p/core/network"
	"github.com/libp2p/go-libp2p/core/peer"
	"github.com/libp2p/go-libp2p/core/transport"
	"github.com/libp2p/go-libp2p/p2p/host/eventbus"
	"github.com/libp2p/go-libp2p/p2p/host/peerstore/pstoremem"
	"github.com/stretchr/testify/require"
)

type blockingConnectednessEmitter struct {
	firstEmit sync.Once
	emitted   chan struct{}
	unblock   chan struct{}
}

var _ event.Emitter = (*blockingConnectednessEmitter)(nil)

func newBlockingConnectednessEmitter() *blockingConnectednessEmitter {
	return &blockingConnectednessEmitter{
		emitted: make(chan struct{}),
		unblock: make(chan struct{}),
	}
}

func (b *blockingConnectednessEmitter) Emit(any) error {
	b.firstEmit.Do(func() {
		close(b.emitted)
	})
	<-b.unblock
	return nil
}

func (b *blockingConnectednessEmitter) Close() error {
	return nil
}

type startTrackingCapableConn struct {
	localPeer  peer.ID
	remotePeer peer.ID

	acceptStarted chan struct{}
	startOnce     sync.Once
	closed        chan struct{}
	closeOnce     sync.Once
}

var _ transport.CapableConn = (*startTrackingCapableConn)(nil)

func newStartTrackingCapableConn(localPeer, remotePeer peer.ID) *startTrackingCapableConn {
	return &startTrackingCapableConn{
		localPeer:     localPeer,
		remotePeer:    remotePeer,
		acceptStarted: make(chan struct{}),
		closed:        make(chan struct{}),
	}
}

func (c *startTrackingCapableConn) Close() error {
	c.closeOnce.Do(func() {
		close(c.closed)
	})
	return nil
}

func (c *startTrackingCapableConn) CloseWithError(network.ConnErrorCode) error {
	return c.Close()
}

func (c *startTrackingCapableConn) IsClosed() bool {
	select {
	case <-c.closed:
		return true
	default:
		return false
	}
}

func (c *startTrackingCapableConn) OpenStream(context.Context) (network.MuxedStream, error) {
	return nil, errors.New("start tracking connection does not open streams")
}

func (c *startTrackingCapableConn) AcceptStream() (network.MuxedStream, error) {
	c.startOnce.Do(func() {
		close(c.acceptStarted)
	})
	return nil, errors.New("start tracking connection closed")
}

func (c *startTrackingCapableConn) As(any) bool {
	return false
}

func (c *startTrackingCapableConn) LocalPeer() peer.ID {
	return c.localPeer
}

func (c *startTrackingCapableConn) RemotePeer() peer.ID {
	return c.remotePeer
}

func (c *startTrackingCapableConn) RemotePublicKey() ic.PubKey {
	return nil
}

func (c *startTrackingCapableConn) ConnState() network.ConnectionState {
	return network.ConnectionState{}
}

func (c *startTrackingCapableConn) LocalMultiaddr() ma.Multiaddr {
	return ma.StringCast("/ip4/127.0.0.1/tcp/1")
}

func (c *startTrackingCapableConn) RemoteMultiaddr() ma.Multiaddr {
	return ma.StringCast("/ip4/127.0.0.1/tcp/2")
}

func (c *startTrackingCapableConn) Scope() network.ConnScope {
	return &network.NullScope{}
}

func (c *startTrackingCapableConn) Transport() transport.Transport {
	return nil
}

// TestAddConnBurstReachesConnStartWhenEmitterIsBlocked covers the
// ref-sensitive addConn path with the connectedness emitter stalled. With the
// old fixed-size newConns channel, this burst could fill the channel and stop
// addConn before Conn.start consumed its swarm ref.
func TestAddConnBurstReachesConnStartWhenEmitterIsBlocked(t *testing.T) {
	const burstSize = 100

	localPeer := peer.ID("local-peer-for-connectedness-test")
	peerStore, err := pstoremem.NewPeerstore()
	require.NoError(t, err)
	t.Cleanup(func() { peerStore.Close() })

	sw, err := NewSwarm(localPeer, peerStore, eventbus.NewBus())
	require.NoError(t, err)

	blockingEmitter := newBlockingConnectednessEmitter()
	sw.connectednessEventEmitter.Close()
	sw.connectednessEventEmitter = newConnectednessEventEmitter(sw.Connectedness, blockingEmitter)

	unblockEmitter := sync.OnceFunc(func() {
		close(blockingEmitter.unblock)
	})
	t.Cleanup(func() {
		unblockEmitter()
		require.NoError(t, sw.Close())
	})

	initialConn := newStartTrackingCapableConn(localPeer, peer.ID("remote-peer-for-connectedness-test-initial"))
	_, err = sw.addConn(initialConn, network.DirOutbound)
	require.NoError(t, err)

	select {
	case <-blockingEmitter.emitted:
	case <-time.After(time.Second):
		t.Fatal("connectedness emitter did not block on the initial event")
	}

	type addConnBurstResult struct {
		conns []*startTrackingCapableConn
		err   error
	}

	addConnBurstDone := make(chan addConnBurstResult, 1)
	go func() {
		conns := make([]*startTrackingCapableConn, 0, burstSize)
		for i := range burstSize {
			remotePeer := peer.ID(fmt.Sprintf("remote-peer-for-connectedness-test-%03d", i))
			conn := newStartTrackingCapableConn(localPeer, remotePeer)
			conns = append(conns, conn)

			if _, err := sw.addConn(conn, network.DirOutbound); err != nil {
				addConnBurstDone <- addConnBurstResult{err: fmt.Errorf("addConn failed for burst connection %d: %w", i, err)}
				return
			}
		}
		addConnBurstDone <- addConnBurstResult{conns: conns}
	}()

	var conns []*startTrackingCapableConn
	select {
	case result := <-addConnBurstDone:
		require.NoError(t, result.err)
		conns = result.conns
	case <-time.After(time.Second):
		t.Fatal("addConn burst did not complete while connectedness emitter was blocked")
	}

	for i, conn := range conns {
		select {
		case <-conn.acceptStarted:
		case <-time.After(time.Second):
			t.Fatalf("burst connection %d did not reach Conn.start", i)
		}
	}

	unblockEmitter()
}
```

sukunrt:
Firstly, I'm sorry about the state of this code. It's very complicated and we should simplify the whole connection management logic. The suggested change doesn't work either though. 

>I changed newConns to follow the same shape as removeConns: keep pending peers in a mutex-protected slice, and use a 1-slot notification channel to wake the emitter goroutine. That gets rid of the fixed channel capacity without dropping events, rejecting valid connections, or allocating a large buffer per swarm.

The problem with this is that this doesn't bound the outstanding number of goroutines. If you open and close connections quickly, these slices will grow unbounded. 

>If AddConn blocked forever before c.start() was reached, that ref would never be consumed.

This is an application bug. It should really never happen. If the consumer is blocked, why should the goroutine stalled on `AddConn` make progress? I think we should first get a handle on what we're trying to fix exactly. If a stalled consumer is the problem maybe we should log it this way? 

```
select {
case c.newConns <- p:
default:
  select {
  case c.newConns <- p:
  case <-time.After(30 * time.Second):
	  c.logger.Error("slow connectedness event consumer, conn stalled")
  }
}
```

Sahil-4555:
> The problem with this is that this doesn't bound the outstanding number of goroutines. If you open and close connections quickly, these slices will grow unbounded.

Yeah, that’s a good point. The slice version removes the fixed channel limit, but it also removes the useful bound that the channel was giving us. If the connectedness emitter is stuck and connections keep opening/closing, the pending slices can keep growing with that churn. That’s not the right tradeoff for this code path.

What if we keep the bounded `newConns` channel and treat this as an observability issue instead of changing the queueing semantics?

The shape I’m thinking is:

- `AddConn` keeps the fast path send into `newConns`.
- If the queue is full, it waits like it does today, so the existing backpressure and ordering behavior stays intact.
- If it remains blocked for a slow-consumer timeout, say 30s, we log an error like `slow connectedness event consumer, connection stalled`.
- After logging, it keeps waiting for capacity rather than dropping the event or closing the connection.
- If the emitter is closed while waiting, `AddConn` returns.

That keeps the queue bounded, preserves the current semantics, and makes the stalled-consumer case visible when it actually happens.

Sahil-4555:
After the discussion with @sukunrt, I agree this PR is not the right direction. The bounded `AddConn` backpressure is intentional, and the eventbus already logs slow consumers at the actual blocking point with better context. The alternatives I tried either dropped important events, rejected valid connections, or removed the useful bound on pending work.

I’m going to close this one rather than push a change that doesn’t preserve the intended behavior. Thanks for walking through the connection ordering here.

PR #3503, opened 2026-05-11
refactor(swarm): move the connected / disconnected callbacks into the events emitter

sukunrt:
removes the notification lock on connectiosn with a single lock on the events emitter. This consolidates all the connection events into one file making it easier to reason about them. 

A lot of it was written by Claude. I've reviewed everything.

### DIFF
--- a/p2p/net/swarm/connection_events_emitter.go
+++ b/p2p/net/swarm/connection_events_emitter.go
@@ -9,6 +9,18 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
+type peerConnectednessEventType int
+
+const (
+	removeConnEvent peerConnectednessEventType = iota
+	addConnEvent
+)
+
+type peerConnectednessEvent struct {
+	PeerID peer.ID
+	Type   peerConnectednessEventType
+}
+
 // connectionEventsEmitter emits PeerConnectednessChanged events and dispatches
 // per-conn Connected / Disconnected callbacks supplied by the owner.
 //
@@ -26,15 +38,11 @@ import (
 // from a goroutine; AddConn always runs onConnected synchronously so callers
 // see Connected fire before they continue.
 type connectionEventsEmitter struct {
-	// newConns is the channel that holds the peerIDs we recently connected to
-	newConns      chan peer.ID
-	removeConnsMx sync.Mutex
-	// removeConns is a slice of peerIDs we have recently closed connections to
-	removeConns []peer.ID
-	// removeConnNotif is used to notify the event loop on an addition to `removeConns`.
-	removeConnNotif chan struct{}
-	// lastEvent is the last connectedness event sent for a particular peer.
-	lastEvent map[peer.ID]network.Connectedness
+	// peerConnectednessCh carries add and remove events to the run loop, which
+	// uses them to emit PeerConnectednessChanged events.
+	peerConnectednessCh chan peerConnectednessEvent
+	// lastConnectednessEvent is the last connectedness event sent for a particular peer.
+	lastConnectednessEvent map[peer.ID]network.Connectedness
 	// connectedness is the function that gives the peers current connectedness state
 	connectedness func(peer.ID) network.Connectedness
 	// onConnected fires synchronously when a conn becomes connected.
@@ -72,17 +80,16 @@ func newConnectionEventsEmitter(
 ) *connectionEventsEmitter {
 	loopCtx, loopCancel := context.WithCancel(context.Background())
 	c := &connectionEventsEmitter{
-		newConns:          make(chan peer.ID, 32),
-		lastEvent:         make(map[peer.ID]network.Connectedness),
-		removeConnNotif:   make(chan struct{}, 1),
-		connectedness:     connectedness,
-		emitter:           emitter,
-		ctx:               loopCtx,
-		cancel:            loopCancel,
-		connected:         make(map[*Conn]struct{}),
-		pendingDisconnect: make(map[*Conn]struct{}),
-		onConnected:       onConnected,
-		onDisconnected:    onDisconnected,
+		peerConnectednessCh:    make(chan peerConnectednessEvent, 32),
+		lastConnectednessEvent: make(map[peer.ID]network.Connectedness),
+		connectedness:          connectedness,
+		emitter:                emitter,
+		ctx:                    loopCtx,
+		cancel:                 loopCancel,
+		connected:              make(map[*Conn]struct{}),
+		pendingDisconnect:      make(map[*Conn]struct{}),
+		onConnected:            onConnected,
+		onDisconnected:         onDisconnected,
 	}
 	c.loopWG.Add(1)
 	go c.runEmitter()
@@ -99,7 +106,10 @@ func (c *connectionEventsEmitter) AddConn(conn *Conn) {
 	c.closeMu.Unlock()
 	defer c.wg.Done()
 
-	c.newConns <- conn.RemotePeer()
+	c.peerConnectednessCh <- peerConnectednessEvent{
+		PeerID: conn.RemotePeer(),
+		Type:   addConnEvent,
+	}
 
 	// Dispatch onConnected before touching notifsLk so that a concurrent
 	// RemoveConn cannot observe `connected[conn]` and fire onDisconnected
@@ -108,7 +118,7 @@ func (c *connectionEventsEmitter) AddConn(conn *Conn) {
 
 	var dispatchDisconnect bool
 	c.notifsLk.Lock()
-	if _, parked := c.pendingDisconnect[conn]; parked {
+	if _, ok := c.pendingDisconnect[conn]; ok {
 		delete(c.pendingDisconnect, conn)
 		dispatchDisconnect = true
 	} else {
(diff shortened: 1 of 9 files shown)


p2p/net/swarm/swarm_conn.go
```diff
@@ -90,28 +86,19 @@ func (c *Conn) doClose(errCode network.ConnErrorCode) {
...
-	c.swarm.connectednessEventEmitter.RemoveConn(c.RemotePeer())
-
 	// This is just for cleaning up state. The connection has already been closed.
 	// We *could* optimize this but it really isn't worth it.
 	for s := range streams {
 		s.Reset()
 	}
 
-	// do this in a goroutine to avoid deadlocking if we call close in an open notification.
+	// Dispatch the close notifications in a goroutine: a Notifiee.Disconnected
```
MarcoPolo: Do you mean Notifiee.Connected handler? Why would you call Close inside a Disconnected handler?

MarcoPolo (approved):
Thank you. This is much nicer than the old code. One suggestion that isn't new from these changes but would be good to fix while we are at it:

Why do we need both `newConns` and `removeConnNotif`. Having two channels makes this transition hard to synchronize properly. I think it would be better to have a single channel that is emitted to with different variants. Then we can better uphold the invariant that a not connected event always proceeds a connected event.

p2p/net/swarm/swarm_conn.go
sukunrt: it's only to prevent a deadlock if someone does it accidently. 
`notifee.Connected` is also safe but that's not related to why this call MUST happen in a different goroutine. 

I'm open to removing this freedom and make it illegal to call conn.Close in a `Disconnected` notification.

sukunrt:
That's a good point. The problem is that both sides cannot block to prevent a deadlock where closing from the open connection event handler blocks on the push to the channel. 

Not sure if we can't work around this still. Lemme see if I can make it work.

p2p/net/swarm/swarm_conn.go
MarcoPolo: It seems better to return an error if a user tries to close on Disconnected event. That seems like a bug.

sukunrt: It definitely is an "error" or at best useless. 

how would you identify that the Close is from the Disconnected callback?

```
conn.Close 
-> OnDisconnect(network, conn) // The callback in question
-> conn.Close // deadlocks because close happens in a sync.Once
```

How would you error from the second `conn.Close` call?

p2p/net/swarm/swarm.go
```diff
@@ -416,15 +422,12 @@ func (s *Swarm) addConn(tc transport.CapableConn, dir network.Direction) (*Conn,
...
-	// enqueued the peer connectedness changed notification.
-	// TODO: Fix this fragility by taking a swarm ref for dial worker loop
-	c.notifyLk.Lock()
 	s.conns.Unlock()
 
-	s.connectednessEventEmitter.AddConn(p)
+	// AddConn dispatches PeerConnectednessChanged and Notifiee.Connected before
+	// c.start() spawns the AcceptStream loop, so handlers see the conn before
+	// any inbound stream arrives.
+	s.connectionEventsEmitter.AddConn(c)
```
Sahil-4555: I think this may have accidentally changed the ordering for direct-connection waiters. Previously, once a non-limited conn was added to `s.conns`, we woke `directConnNotifs` before running the `Connected` notifiees. In the new flow, `AddConn(c)` runs those notifiees first, so a slow or blocked `Connected` handler could leave `waitForDirectConn` waiting even though the direct conn is already available. In practice that could make a stream waiting on a direct conn time out unnecessarily.

I see why we want `Connected` to run before `c.start()`, so handlers see the conn before inbound streams arrive. Could we still keep that property while waking `directConnNotifs` before the connected callbacks?

sukunrt: I don't think it'd make much difference in practice. We're not processing any incoming streams before `c.start` anyway.

sukunrt:
>Why do we need both newConns and removeConnNotif

I think we don't. I've removed the removeConnNotif and done some renaming. There's just one channel for connectedness events now.

p2p/net/swarm/swarm.go
Sahil-4555: The case I had in mind was the outbound side. If something is already sitting in `waitForDirectConn`, it only needs to be woken so it can re-check `s.conns` and use the direct conn for `NewStream`. That doesn’t depend on the AcceptStream loop being started. So my concern is mostly that a slow `Connected` notifiee can now delay waking those outbound waiters even though the direct conn is already there. It may be rare in practice, but since the old ordering avoided that coupling, I think it’d be safer to keep it.

p2p/net/swarm/connection_events_emitter.go
```diff
@@ -0,0 +1,224 @@
...
+	c.closeMu.Lock()
+	if c.closed {
+		c.closeMu.Unlock()
+		return
+	}
+	c.wg.Add(1)
+	c.closeMu.Unlock()
+	defer c.wg.Done()
+
+	c.peerConnectednessCh <- peerConnectednessEvent{
```
Sahil-4555, line 152: I’m a bit worried this makes disconnect cleanup depend on the connectedness event loop. Before, `RemoveConn` used the side queue, so it could continue to the `Notifiee.Disconnected` bookkeeping without blocking behind `PeerConnectednessChanged` delivery.

With this change, a slow connectedness subscriber could stall `Emit`, fill `peerConnectednessCh`, and then block `RemoveConn` here. That would delay `Notifiee.Disconnected` and keep the close ref from `doClose` alive. Is that coupling intentional?

Could we keep the remove path non-blocking here, either by preserving the side queue for remove events or by moving the notifiee bookkeeping before the channel send?

sukunrt: if the event handler is slow, I think it's okay to delay disconnected. The events emitter has a slow consumer log for affected users.

p2p/net/swarm/swarm_conn.go
MarcoPolo: You could use an atomic.bool to detect if you've entered this method before.

p2p/net/swarm/connection_events_emitter.go
```diff
@@ -196,32 +199,26 @@ func (c *connectionEventsEmitter) runEmitter() {
...
-	if c.lastEvent[p] == network.NotConnected {
-		delete(c.lastEvent, p)
+// notifyPeer emits a PeerConnectednessChanged event when the peer's
+// connectedness has changed since the last emit. For add events, it additionally
+// emits NotConnected if the peer is already disconnected by the time we get
+// here (i.e. a RemoveConn raced ahead of the matching AddConn): the peer may
+// have observed the connection, so subscribers must still learn about it.
+func (c *connectionEventsEmitter) notifyPeer(pce peerConnectednessEvent) {
+	p := pce.PeerID
+	forceNotConnectedEvent := pce.Type == addConnEvent
```
MarcoPolo, line 209: do we need this forceNotConnectedEvent logic? Shouldn't we always get a connected followed by disconnected events?

p2p/net/swarm/connection_events_emitter.go
```diff
@@ -0,0 +1,224 @@
...
+	c.notifsLk.Lock()
+	if _, ok := c.pendingDisconnect[conn]; ok {
+		delete(c.pendingDisconnect, conn)
+		dispatchDisconnect = true
+	} else {
+		c.connected[conn] = struct{}{}
+	}
+	c.notifsLk.Unlock()
+
+	if dispatchDisconnect {
```
Sahil-4555, line 129: One small timing change I noticed here: in the parked remove case, `Disconnected` runs inline from `AddConn`. The ordering is correct, but if a `Connected` notifiee closes the conn, then a slow `Disconnected` notifiee can now block `addConn` before it returns.

Earlier this close notification would run from the close goroutine after the connected callback finished, so `addConn` itself was not waiting on the disconnected notifiees. Is this change intentional, or should we dispatch the parked disconnect asynchronously after `onConnected` finishes?

p2p/net/swarm/swarm_conn.go
sukunrt: but it must happen from the `Disconnected` callback, no? 
concurrently closing the same connection from two different places is fine.

p2p/net/swarm/connection_events_emitter.go
sukunrt: it's fine to block. if the connection is closed, there's nothing to do post the connected callback anyway.

Sahil-4555 (approved):

p2p/net/swarm/connection_events_emitter.go
sukunrt: This was the behavior before, so I didn't change it. 

The existing behavior is to send at least one connectedness event for the peer on making a connection. This was so because the connection could be observed in other ways, like making a `swarm.ConnsToPeer` call. 

It's a peer connectedness event not a connection state event. This requires some form of point in time connectedness state event because we have Limited and Connected, two Connectedness states. Say we get two concurrent connections to the peer, one Limited one non-Limited(Connected), and both of these are immediately closed. Same reason if we get two connections to the peer, both non-limited full connections, we only want to send one Connected event to the peer as the connectedness hasn't changed.

PR #3504, opened 2026-05-12
fix holepunch shutdown deadlock on canceled address discovery

Sahil-4555:
## Summary

This PR fixes a shutdown deadlock in the holepunch service.

`Service.waitForPublicAddr` waits until the host has a public address, then initializes the hole puncher while holding `holePuncherMx`. If the service context is canceled after public address discovery but before hole puncher initialization, the function could return without unlocking `holePuncherMx`.

That can make `Service.Close()` block forever because `Close()` also needs the same mutex.

## Impact

This can cause shutdown or teardown to hang when hole punching is enabled.

Possible effects:

- `Host.Close()` or service shutdown can get stuck
- goroutines can remain blocked during teardown
- process restart can stall
- tests or applications that create and close hosts repeatedly can hang

## Fix

Unlock `holePuncherMx` with a `defer` immediately after locking it:

```go
s.holePuncherMx.Lock()
defer s.holePuncherMx.Unlock()
```
This guarantees the mutex is released on both the normal path and the cancellation path.

### Test
The test reproduces the race by:
1. Starting `waitForPublicAddr`.
2. Blocking inside `listenAddrs`.
3. Canceling the service context.
4. Returning a public address from `listenAddrs`.
5. Verifying `holePuncherMx` is not left locked.
6. Verifying `Close()` returns instead of blocking.

### DIFF
--- a/p2p/protocol/holepunch/svc.go
+++ b/p2p/protocol/holepunch/svc.go
@@ -143,13 +143,13 @@ func (s *Service) waitForPublicAddr() {
 	}
 
 	s.holePuncherMx.Lock()
+	defer s.holePuncherMx.Unlock()
 	if s.ctx.Err() != nil {
 		// service is closed
 		return
 	}
 	s.holePuncher = newHolePuncher(s.host, s.ids, s.listenAddrs, s.tracer, s.filter)
 	s.holePuncher.directDialTimeout = s.directDialTimeout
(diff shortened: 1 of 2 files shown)


sukunrt (approved):

sukunrt:
Thanks!

PR #3505, opened 2026-05-13
config: disable black hole detector on AutoNAT v2 dialerHost

### LINKED ISSUE #3491: AutoNAT v2 dialerHost should not have the UDP black hole detector
This issue distills findings from a systematic cross-implementation evaluation of AutoNAT v2, conducted by [ProbeLab](https://probelab.io/). The final report of the investigation is available in the following [link](https://github.com/probe-lab/autonat-perf-audit/blob/main/docs/final-report.md).

The AutoNAT v2 `dialerHost` shares the main host's UDP black hole counter in read-only mode [config.go#L240-L246](https://github.com/libp2p/go-libp2p/blob/v0.48.0/config/config.go#L240-L246). Read-only mode treats any state other than `Allowed` as `Blocked` [black_hole_detector.go#L254-L262](https://github.com/libp2p/go-libp2p/blob/v0.48.0/p2p/net/swarm/black_hole_detector.go#L254-L262), so when the main host is in `Probing` (zero history), the dialerHost refuses every QUIC dial-back with `E_DIAL_REFUSED`.

Observed behaviour:

- Fresh servers, low-traffic nodes, and isolated testbeds can stay in `Probing` → every QUIC dial-back is refused.
- `E_DIAL_REFUSED` is discarded from v2 confidence by [AddRefusal](https://github.com/libp2p/go-libp2p/blob/v0.48.0/p2p/host/basic/addrs_reachability_tracker.go#L689-L692) (it does not call `AddOutcome`), so the client's QUIC address doesn't flip to Unreachable — but it never accumulates positive evidence either, and stays stuck in `Unknown`.
- AutoNAT v1 had the same problem and was fixed in [[#2529](https://github.com/libp2p/go-libp2p/pull/2529)](https://github.com/libp2p/go-libp2p/pull/2529) by disabling the detector on the v1 dialer entirely.

Proposal: apply the v1 approach — pass `nil` counters to `makeAutoNATV2Host()` so the detector doesn't run on the dialerHost. The dialerHost only dials addresses clients explicitly request to test; the detector has no useful role there.

Deln0r:
Closes #3491.

## Problem

The AutoNAT v2 `dialerHost` constructed by `makeAutoNATV2Host()` shares the main host's UDP/IPv6 black-hole counters in read-only mode. Read-only mode treats anything other than `Allowed` as `Blocked` (see `swarm/black_hole_detector.go`), so while the main host is still in `Probing` — the natural state for fresh servers, low-traffic nodes, and isolated testbeds — every QUIC dial-back is refused with `E_DIAL_REFUSED`.

`AddRefusal` (`p2p/host/basic/addrs_reachability_tracker.go`) does not feed `AddOutcome`, so the client's QUIC address never flips to `Unreachable`, but it never accumulates positive evidence either and stays stuck in `Unknown`.

This was systematically observed in the cross-implementation [ProbeLab AutoNAT v2 audit](https://github.com/probe-lab/autonat-perf-audit/blob/main/docs/final-report.md) and reported in #3491.

## Fix

Apply the same approach already used for AutoNAT v1 (#2529): stop inheriting `UDPBlackHoleSuccessCounter` / `IPv6BlackHoleSuccessCounter` from the parent config, drop `WithReadOnlyBlackHoleDetector()`, and pass `nil` counters to swarm via `SwarmOpts`. The dialerHost only dials addresses clients explicitly asked us to test, so the detector has no useful role on it.

```go
SwarmOpts: []swarm.Option{
    swarm.WithUDPBlackHoleSuccessCounter(nil),
    swarm.WithIPv6BlackHoleSuccessCounter(nil),
},
```

This mirrors the existing AutoNAT v1 dialer construction in the same file.

## Test plan

```
$ go test -short -count=1 -timeout 120s ./p2p/protocol/autonatv2/... ./p2p/host/autonat/... ./config
ok  	github.com/libp2p/go-libp2p/p2p/protocol/autonatv2	23.380s
ok  	github.com/libp2p/go-libp2p/p2p/host/autonat	7.079s
ok  	github.com/libp2p/go-libp2p/p2p/host/autonat/test	0.840s
ok  	github.com/libp2p/go-libp2p/config	0.595s
```

No new tests added; the change is a configuration delta inside a private host-construction helper and is covered by the AutoNAT v2 integration tests that already exercise the dial-back path. If you want a more targeted regression test (assertion that the dialerHost's swarm has black-hole detection disabled), happy to add one.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -230,22 +230,28 @@ func (cfg *Config) makeAutoNATV2Host() (host.Host, error) {
 	}
 
 	autoNatCfg := Config{
-		Transports:                  cfg.Transports,
-		Muxers:                      cfg.Muxers,
-		SecurityTransports:          cfg.SecurityTransports,
-		Insecure:                    cfg.Insecure,
-		PSK:                         cfg.PSK,
-		ConnectionGater:             cfg.ConnectionGater,
-		Reporter:                    cfg.Reporter,
-		PeerKey:                     autonatPrivKey,
-		Peerstore:                   ps,
-		DialRanker:                  swarm.NoDelayDialRanker,
-		UDPBlackHoleSuccessCounter:  cfg.UDPBlackHoleSuccessCounter,
-		IPv6BlackHoleSuccessCounter: cfg.IPv6BlackHoleSuccessCounter,
-		ResourceManager:             cfg.ResourceManager,
+		Transports:         cfg.Transports,
+		Muxers:             cfg.Muxers,
+		SecurityTransports: cfg.SecurityTransports,
+		Insecure:           cfg.Insecure,
+		PSK:                cfg.PSK,
+		ConnectionGater:    cfg.ConnectionGater,
+		Reporter:           cfg.Reporter,
+		PeerKey:            autonatPrivKey,
+		Peerstore:          ps,
+		DialRanker:         swarm.NoDelayDialRanker,
+		ResourceManager:    cfg.ResourceManager,
 		SwarmOpts: []swarm.Option{
-			// Don't update black hole state for failed autonat dials
-			swarm.WithReadOnlyBlackHoleDetector(),
+			// The dialerHost only dials addresses clients explicitly ask
+			// us to test, so the black hole detector has no useful role
+			// here. Sharing the main host's counters in read-only mode
+			// causes every dial-back to be refused with E_DIAL_REFUSED
+			// while the main host is still in the Probing state, which
+			// in turn keeps the client's QUIC address stuck in Unknown
+			// (see #3491). Disable the detectors entirely, matching the
+			// approach used for the AutoNAT v1 dialer (#2529).
+			swarm.WithUDPBlackHoleSuccessCounter(nil),
+			swarm.WithIPv6BlackHoleSuccessCounter(nil),
 		},
 	}
 	fxopts, err := autoNatCfg.addTransports()


sukunrt:
```
AddRefusal (p2p/host/basic/addrs_reachability_tracker.go) does not feed AddOutcome, so the client's QUIC address never flips to Unreachable, but it never accumulates positive evidence either and stays stuck in Unknown.
```

The client is supposed to use other servers to verify their reachability in such cases.

Deln0r:
> ```
> The client is supposed to use other servers to verify their reachability in such cases.
> ```
Fair point, "stuck in Unknown" was overstated. The client does fall back to other servers, so a single refusing server doesn't strand a healthy client. The accurate version: from this particular server the client gets no signal at all (refusal isn't counted as positive or negative evidence), so the server contributes zero to the reachability verdict despite having had the slot.

The fix still seems worth landing for the v1/v2 asymmetry (#2529 already did the same nil-counter shape on v1) and to avoid wasting the dial-back attempt on a CIDR the dialerHost has no actual history for. Happy to reword the PR description / commit message accordingly - want me to push an updated description, or just fix it on squash?

sukunrt:
we didn't port the "fix" to autonatv1 because we wanted to just move forward with autonat v2. 

https://github.com/libp2p/go-libp2p/pull/2529#issuecomment-1695090307

sukunrt:
I'm closing this as the existing behavior is correct. We don't want to dial as a server unless we're sure we have the capability to dial such addresses.

Deln0r:
Thanks for the context - understood, the v2 strategy is intentional and not the v1 nil-counter pattern carried forward. Appreciate the time on the review.

PR #3509, opened 2026-05-15
feat(websocket): share port with HTTP handler

lidel:
Adds `websocket.WithHTTPHandler(http.Handler)` so a `/ws` or `/tls/ws` listener answers non-WebSocket requests with the supplied handler instead of 404, on the same TCP port, plus `websocket.WithHTTPServerConfig(func(*http.Server))` to configure that server's timeouts and HTTP/2 settings.

:point_right:  Opt-in, backward-compatible, default behaviour is unchanged, but go-libp2p gains important regression tests.

## Two uses

- Share a libp2p WebSocket port with an ordinary HTTPS site. The port looks like a normal HTTPS endpoint that also accepts WebSocket upgrades. Libp2p on port 443 with regular website becomes way harder to censor.
- Expose kubo's trustless gateway on the AutoTLS-secured `/tls/ws` port. HTTP retrieval clients (`boxo/bitswap/network/httpnet`) fetch blocks directly over HTTPS using the AutoTLS cert. Improved interop of IPFS stack.

## Routing

WebSocket upgrades go to libp2p as before; everything else goes to the handler. The handler runs over HTTP/1.1 and HTTP/2 on TLS listeners (ALPN inside `http.Server.ServeTLS`) and over HTTP/1.1 and HTTP/2 cleartext on plaintext listeners (`net/http` `Protocols` with `SetUnencryptedHTTP2`). Also in `tcpreuse`: `IsHTTP` matches the `PRI` h2c preface so it routes to the HTTP demux branch.

```go
tpt, err := websocket.New(upgrader, rcmgr, sharedTCP,
    websocket.WithTLSConfig(tlsConfig),
    websocket.WithHTTPHandler(httpHandler),
    websocket.WithHTTPServerConfig(func(s *http.Server) {
        s.IdleTimeout = 30 * time.Second
        s.HTTP2 = &http.HTTP2Config{MaxConcurrentStreams: 256}
    }),
)
```

> [!IMPORTANT]
> Real-world example in ipfs/kubo#11333, where kubo uses both opt-in funcs to expose its trustless gateway on the AutoTLS-secured `/ws` and `/tls/ws` TCP ports, on one Let's Encrypt cert:
> - `WithHTTPHandler` installs the trustless gateway handler (h2 required over TLS, h1+h2c allowed cleartext).
> - `WithHTTPServerConfig` sets streaming-safe timeouts and HTTP/2 limits, deriving `WriteByteTimeout`/`SendPingTimeout` from `Gateway.RetrievalTimeout` so the gateway's own 504 fires before the connection is dropped.
>
> An in-process AutoTLS canary (Pebble + p2p-forge) exercises the full path end to end.

## Regression tests

:point_right: tests added here protect ecosystem from silently breaking `/ws`.

**gorilla/websocket has no HTTP/2 / RFC 8441 server support today.** WebSocket-over-h2 needs extended CONNECT (`:method=CONNECT, :protocol=websocket`); gorilla's `Upgrader.Upgrade` still reads HTTP/1.1 hijack-based headers, and Go's h2 stack does not advertise `SETTINGS_ENABLE_CONNECT_PROTOCOL` (gated on process-global `GODEBUG=http2xconnect=1`; per-server API tracked in `golang/go#71128`). Two tripwires added by this PR catch any future upstream shift so go-libp2p maintainers can deliberately embrace WS-over-h2 or pin the dispatch:

- `TestFallbackHTTPHandler_HTTP2NeverReachesWSPath`: every HTTP/2 request reaches the fallback handler, never the WebSocket hijack path. Fails if a future gorilla release extends `IsWebSocketUpgrade` to recognise ext-CONNECT.
- `TestModernBrowserWSSFlow`: the server's initial h2 SETTINGS frame omits `SETTINGS_ENABLE_CONNECT_PROTOCOL`. Browsers rely on its absence to fall back to HTTP/1.1 for `wss://`; the test fails if Go ever flips this flag on by default.

Plus a concurrency pin for the new fallback path:

- `TestFallbackHTTPHandler_KeepAlive` + `TestNegotiatingConnUnwrapConcurrent`: the handshake-timer is disarmed once a fallback request arrives, safely under concurrent HTTP/2 streams sharing one `*negotiatingConn`.

### DIFF
--- a/p2p/transport/websocket/fallback_http_handler_test.go
+++ b/p2p/transport/websocket/fallback_http_handler_test.go
@@ -1,12 +1,15 @@
 package websocket
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"fmt"
 	"io"
 	"net"
 	"net/http"
+	"os"
+	"os/exec"
 	"slices"
 	"sync"
 	"sync/atomic"
@@ -16,6 +19,7 @@ import (
 	gws "github.com/gorilla/websocket"
 	"github.com/libp2p/go-libp2p/core/network"
 	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/hpack"
 	"golang.org/x/sync/errgroup"
 
 	ma "github.com/multiformats/go-multiaddr"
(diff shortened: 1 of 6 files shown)


lidel:
@MarcoPolo @sukunrt kind ping to put this on the top of review queue. 

The foundations of `/ws` are VERY shaky, and this PR adds necessary regression tests (see last section in description).  

It also adds opt-in `websocket.WithFallbackHTTPHandler(httpHandler)`  to reduce our dependency on `/ws` and benefit from HTTP/2 multiplexing in IPFS ecosystem, but I'm pinging you because of how shaky `/ws` is and asking if we could prioritize this, so we have regression tests running on CI :pray:  (we really, really dont want `libp2p.direct` peers to break silently when upstream library finally adds h2 support).

p2p/transport/websocket/fallback_http_handler_test.go
```diff
@@ -0,0 +1,502 @@
...
+		fallbackHits.Add(1)
+		// Record the protocol so a regression makes the failure obvious.
+		w.Header().Set("X-Proto", r.Proto)
+		fmt.Fprint(w, "fallback")
+	})
+
+	hostPort := startWSSListener(t, handler)
+
+	// Force ALPN to h2 only, so the connection MUST be HTTP/2 or fail.
+	resp, err := httpsClient([]string{"h2"}).Get("https://" + hostPort + "/")
```
sukunrt: I think you need an extended connect request here to make it fail if gorilla/ws implements h2 websocket support.

lidel: Good catch @sukunrt, you're right. A plain GET never trips `IsWebSocketUpgrade`. Improved and it now sends a real extended CONNECT.

One wrinkle worth noting: the bundled h2 server resets the stream before the handler ever sees an extended CONNECT unless it advertises `ENABLE_CONNECT_PROTOCOL` (golang/go#71128), which is gated by `GODEBUG=http2xconnect=1` and read once at startup. So the test re-execs itself in a child with that set, sends the extended CONNECT over raw frames, and asserts it reaches the fallback handler. Confirmed it goes red if dispatch ever routes extended CONNECT to the WS upgrade path. 

See 9dce1ca. Not a fan, but unsure if this could be done in a better way, given golang limitations.

p2p/transport/websocket/websocket.go
```diff
@@ -82,14 +83,84 @@ func WithHandshakeTimeout(timeout time.Duration) Option {
...
+// share its WebSocket port with an ordinary HTTPS website.
+//
+// Why this matters: a /tls/ws listener is, on the wire, a normal HTTPS
+// endpoint that happens to accept WebSocket upgrades. By default the listener
+// rejects every request that is not a WebSocket upgrade, so a network observer
+// can fingerprint the port as "libp2p WSS" with a single HTTP probe. With a
+// fallback handler installed, the same port also serves real HTTP responses,
+// and the libp2p traffic looks indistinguishable from any other web traffic
+// behind the same TLS certificate. A censor that wants to block libp2p must
+// also block the website served from the fallback handler.
```
sukunrt: We can remove this paragraph. 
We can also possibly rename the option to just WithHTTPHandler.

p2p/transport/websocket/websocket.go
```diff
@@ -82,14 +83,84 @@ func WithHandshakeTimeout(timeout time.Duration) Option {
...
+//   - Every other request reaches the fallback handler, regardless of
+//     HTTP version. On TLS listeners that means HTTP/1.1 and HTTP/2 (Go's
+//     http.Server.ServeTLS negotiates h2 via ALPN automatically). On
+//     plaintext listeners that means HTTP/1.1 and HTTP/2 cleartext (h2c):
+//     the listener is wrapped with [h2c.NewHandler] so reverse proxies
+//     that speak h2c to backends (Caddy, Traefik, nginx with the right
+//     config) get HTTP/2 multiplexing end-to-end, while proxies that only
+//     speak h1 keep working through the same listener.
+//   - If no fallback handler is configured, non-upgrade requests get a 404,
+//     which is the prior behaviour.
```
sukunrt: I'd remove the info on go defaults and just mention we support both http1.1, http2 with both TLS and plaintext.

p2p/transport/websocket/websocket.go
```diff
@@ -82,14 +83,84 @@ func WithHandshakeTimeout(timeout time.Duration) Option {
...
+//     the Connection and Upgrade headers (RFC 7540 §8.1.2.2), and gorilla's
+//     server-side does not implement RFC 8441 (WebSockets over HTTP/2 via
+//     extended CONNECT). So gorilla.IsWebSocketUpgrade always returns false
+//     for h2 requests and they land on the fallback handler.
+//
+// If a future version of gorilla/websocket adds RFC 8441 support, the
+// dispatch automatically extends to it, because this code defers the
+// "is this a WebSocket upgrade?" decision to gorilla. The regression test
+// TestFallbackHTTPHandler_HTTP2NeverReachesWSPath pins the current
+// behaviour so any silent change is caught.
```
sukunrt: This doesn't belong to the option decoumentation. it's implementation detail.

p2p/transport/websocket/listener.go
```diff
@@ -133,6 +143,32 @@ func newListener(a ma.Multiaddr, tlsConf *tls.Config, sharedTcp *tcpreuse.ConnMg
...
+	// must not set on its own.
+	//
+	// The dispatch in ServeHTTP is, however, written to upgrade
+	// transparently: it defers the "is this a WebSocket upgrade?" decision
+	// to ws.IsWebSocketUpgrade. When upstream Go exposes the setting and
+	// gorilla/websocket adds server-side ext-CONNECT support (most likely
+	// by extending IsWebSocketUpgrade and Upgrader.Upgrade to recognise
+	// `:method=CONNECT, :protocol=websocket`), this listener will accept
+	// WS-over-HTTP/2 with no code change — only a gorilla bump and a
+	// future Go upgrade.
```
sukunrt: You can structure this code to be less go defaults dependent by explicitly configuring the http2 server using `http2.ConfigureServer`
And then we can just opt in to Extended Connect always, and don't have to depend on the godebug for tests.

sukunrt (approved):

p2p/transport/websocket/listener.go
```diff
@@ -175,6 +211,35 @@ func (l *listener) extractConnFromContext(ctx context.Context) (*negotiatingConn
...
+		// so the timer must go away once we know a request has arrived.
+		// Unwrap serializes through sync.Once, so concurrent h2 streams
+		// on the same TCP connection all see the same result.
+		if nc, err := l.extractConnFromContext(r.Context()); err == nil {
+			if _, err := nc.Unwrap(); err != nil {
+				// The handshake-timeout fired between Accept and now; the
+				// connection is already closed. Nothing useful to send.
+				log.Debug("connection timed out before fallback request", "remote_addr", r.RemoteAddr)
+				return
+			}
```
sukunrt, line 246: Note: The http handler will share the resource manager here. So the connections will be shared between the http server and libp2p.

sukunrt:
@lidel We can merge it after fixing the verbose comments. But I think this design would be better: https://github.com/libp2p/go-libp2p/pull/3509#discussion_r3334392942

p2p/transport/websocket/websocket.go
sukunrt: you can add all the context and *why* to the commit message

p2p/transport/websocket/websocket.go
lidel: Renamed to WithHTTPHandler in 9762ed14. The "Why this matters" paragraph is gone from the godoc in bb289fcf; its rationale moved to that commit's message.

p2p/transport/websocket/websocket.go
lidel: Done in bb289fcf: the godoc now just says the handler is served over HTTP/1.1 or HTTP/2 on TLS listeners and HTTP/1.1 or h2c on plaintext listeners.

p2p/transport/websocket/websocket.go
lidel: Done in bb289fcf: the HTTP/2 dispatch detail is out of the option godoc. The dispatch rationale lives in the listener comments, and the design rationale in the commit message.

p2p/transport/websocket/listener.go
lidel: Right, the scope is opened at `Accept`, so fallback connections draw from the same libp2p connection budget. 

Pushed 0bf17d2e which bounds their lifetime with default `IdleTimeout`/`ReadHeaderTimeout` so idle web traffic can't pin scope slots, and `WithHTTPServerConfig` lets operators tune them. The shared budget is otherwise inherent to sharing one port; operators size the resource manager accordingly. 

This should be enough knobs to control end-to-end, and people passing `WithHTTPHandler` can tweak limits based on handler use cases.

p2p/transport/websocket/listener.go
lidel: Thanks, I split this into the two things you raised:

**On making it less go-defaults dependent**: done in 0bf17d2e. It configures the server explicitly, dropping the `x/net/http2/h2c` wrapper for `net/http` `Protocols` so one server handles HTTP/1.1, HTTP/2, and h2c, following the `http2.ConfigureServer` pattern. It also adds `WithHTTPServerConfig(func(*http.Server))` so callers set timeouts and HTTP/2 options on a transport-owned server.

**On opting into Extended Connect always**: there is no per-server switch, and we should not advertise it anyway. In `golang.org/x/net/http2` (v0.50.0 and the latest v0.55.0) a single package-global decides whether the server sends `SETTINGS_ENABLE_CONNECT_PROTOCOL`, flipped only by `GODEBUG=http2xconnect=1` at startup; `http2.ConfigureServer`, `http2.Server`, and `net/http.HTTP2Config` expose no field for it (golang/go#71128 turned it off by default). Even with a switch, advertising it would break browser `wss://`: gorilla has no server-side RFC 8441, so dispatch (via `ws.IsWebSocketUpgrade`, which reads only HTTP/1.1 Upgrade headers) would route a browser's `:method=CONNECT, :protocol=websocket` to the fallback handler instead of upgrading. `TestModernBrowserWSSFlow` pins the absence of the flag; the `GODEBUG` re-exec in the ext-CONNECT test only simulates that future to prove dispatch stays correct.

Rationale now recorded in bb289fcf.

lidel:
Thanks @sukunrt, I went through feedback and did my best:

- I trimmed the long `WithHTTPHandler` doc comment down to what it does and how requests are routed. The "why" (censorship resistance and the RFC 8441 details) now lives in the commit message instead (bb289fcf).
- I renamed the option from `WithFallbackHTTPHandler` to `WithHTTPHandler`, and the internal field to match (9762ed14).
- I added `WithHTTPServerConfig(func(*http.Server))` so you can set your own timeouts and HTTP/2 options. The plaintext path now uses `net/http` `Protocols` instead of the `x/net/http2/h2c` wrapper, so a single server handles HTTP/1.1, HTTP/2, and h2c, following the `http2.ConfigureServer` pattern. When only `WithHTTPHandler` is set, it applies safe default timeouts, which also keeps fallback connections from holding on to the resource-manager budget you pointed out (0bf17d2e).

One thing I could not do the way you suggested: turning Extended Connect on by default. There is no per-server switch for it today, and turning it on would break browser `wss://`, because gorilla cannot speak WebSocket-over-HTTP/2 on the server side yet. I wrote up the full reasoning here: https://github.com/libp2p/go-libp2p/pull/3509#discussion_r3343310036

Happy to change anything if you see a better way. Thanks again for the careful look.

sukunrt:
>One thing I could not do the way you suggested: turning Extended Connect on by default. There is no per-server switch for it today, and turning it on would break browser wss://, because gorilla cannot speak WebSocket-over-HTTP/2 on the server side yet. I wrote up the full reasoning here: #3509 (comment)

Thanks @lidel . I just assumed there would be such an option.

sukunrt (approved):

PR #3510, opened 2026-05-19
http/auth: close intermediate peer ID auth handshake responses

Sahil-4555:
This fixes a resource leak in the HTTP peer ID auth client. During a multi-step authentication flow, the client sends intermediate requests before returning the final authenticated response to the caller. Those intermediate responses are only used for their auth headers, so the library owns their bodies and must close them.

Before this change, those handshake-only response bodies were left open. A server that forces first-time authentication or repeatedly rejects cached bearer tokens could make a long-running client retain HTTP transport resources across repeated handshakes.

The fix closes every non-final response body inside the handshake loop while preserving the existing API contract: the final authenticated response is still returned open and remains the caller’s responsibility.

The tests cover both normal first-contact authentication and the cached-token rejection path that triggers reauthentication.


Sahil-4555:
@sukunrt if you got chance can you please review this PR? and can provide feedback

MarcoPolo:
I've made changes to this PR. I'll push them soon.

MarcoPolo:
> @sukunrt if you got chance can you please review this PR? and can provide feedback

Please do not ping maintainers. We can see this in our backlog, and will prioritize it appropriately.

MarcoPolo (approved):

PR #3511, opened 2026-05-25
p2p/net/swarm: prevent shutdown hang on connectedness backpressure

Sahil-4555:
## Issue

While reviewing the swarm connection lifecycle, found that connectedness event backpressure can stall connection notification paths. When a connection is added or removed, the swarm queues a `PeerConnectednessChanged` event through the connection events emitter. That queue is bounded. If the connectedness emitter goroutine is blocked, for example because an eventbus subscriber is slow or stuck, the queue can fill up and later callers can block while trying to enqueue another connectedness event.

That runtime backpressure is intentional. The swarm should not silently drop connectedness events, reject otherwise valid connections, or replace the bounded queue with unbounded pending state during normal operation.

The issue fixed here is the shutdown/lifecycle side of that behavior. During `Swarm.Close`, connection cleanup can also try to enqueue connectedness events. If the queue is full and event delivery is blocked, those cleanup goroutines can remain stuck, which can cause `Swarm.Close` to wait forever.

## Fix

This change keeps the normal runtime behavior conservative and bounded.

During normal operation:

- the connectedness event queue remains bounded,
- `AddConn` / `RemoveConn` still apply backpressure when the queue is full,
- connectedness events are not dropped,
- valid connections are not rejected,
- and no unbounded pending queue is introduced.

To make the runtime stall visible, enqueueing now logs an error if a connection notification remains blocked for 30 seconds, then continues waiting. This preserves the existing semantics while making slow connectedness event consumers easier to diagnose.

During shutdown, the behavior changes intentionally. `Swarm.Close` now marks the connection events emitter as shutting down before closing active connections. In shutdown mode, enqueueing connectedness events becomes best-effort, so cleanup goroutines are not held forever behind a full event queue.

This does not fully remove runtime connection setup backpressure. I stopped short of that because the safer behavior here is to preserve the existing ordering and bounded memory guarantees. The alternatives are not good tradeoffs:

- dropping runtime connectedness events can hide real connection state,
- rejecting valid connections when a small queue is full is too aggressive,
- using an unbounded pending queue can grow badly under churn,
- moving `c.start()` before notification changes existing ordering semantics.

So this patch keeps the safe runtime behavior and fixes the shutdown hang case.

## Tests
Added focused coverage in the existing swarm test files:

- `TestEmitter_AddConnBackpressuresOnFullQueue`
- `TestEmitter_RemoveConnBackpressuresOnFullQueue`
- `TestEmitter_ShutdownUnblocksQueuedAddConn`
- `TestConnectednessBackpressureAllowsSwarmClose`

These tests verify that runtime backpressure is preserved, queued operations resume when capacity is available, shutdown unblocks queued connection notifications, and `Swarm.Close` completes even when connectedness event delivery is backpressured.

### DIFF
--- a/p2p/net/swarm/connection_events_emitter.go
+++ b/p2p/net/swarm/connection_events_emitter.go
@@ -3,11 +3,15 @@ package swarm
 import (
 	"context"
 	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/libp2p/go-libp2p/core/event"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 )
+
+const connectednessEventQueueStallTimeout = 30 * time.Second
 
 type peerConnectednessEventType int
 
@@ -24,9 +28,10 @@ type peerConnectednessEvent struct {
 // connectionEventsEmitter emits PeerConnectednessChanged events and dispatches
 // per-conn Connected / Disconnected callbacks supplied by the owner.
 //
-// We ensure that for any peer we connected to we always sent atleast 1 NotConnected Event after
-// the peer disconnects. This is because peers can observe a connection before they are notified
-// of the connection by a peer connectedness changed event.
+// Under normal operation, we ensure that for any peer we connected to we always
+// send at least 1 NotConnected Event after the peer disconnects. This is
+// because peers can observe a connection before they are notified of the
+// connection by a peer connectedness changed event.
 //
 // For the per-conn callbacks, we guarantee that for a given *Conn,
 // onDisconnected never fires before the corresponding onConnected has finished.
@@ -38,8 +43,9 @@ type peerConnectednessEvent struct {
 // from a goroutine; AddConn always runs onConnected synchronously so callers
 // see Connected fire before they continue.
 type connectionEventsEmitter struct {
-	// peerConnectednessCh carries add and remove events to the run loop, which
-	// uses them to emit PeerConnectednessChanged events.
+	// peerConnectednessCh is intentionally bounded. A slow connectedness event
+	// consumer should apply backpressure instead of allowing unbounded pending
+	// peer state to accumulate in the swarm.
 	peerConnectednessCh chan peerConnectednessEvent
 	// lastConnectednessEvent is the last connectedness event sent for a particular peer.
 	lastConnectednessEvent map[peer.ID]network.Connectedness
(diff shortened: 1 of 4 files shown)


sukunrt:
>During Swarm.Close, connection cleanup can also try to enqueue connectedness events. If the queue is full and event delivery is blocked, those cleanup goroutines can remain stuck, which can cause Swarm.Close to wait forever.

As explained in your previous PR https://github.com/libp2p/go-libp2p/pull/3502 this is not a bug. 

> In shutdown mode, enqueueing connectedness events becomes best-effort, so cleanup goroutines are not held forever behind a full event queue.

There's no reason to do this.

Sahil-4555:
Thanks @sukunrt, I get the point from #3502 that runtime connectedness backpressure is intentional. This PR is not trying to change that part.

The difference I was trying to make here is only around shutdown.

During normal operation, this change still keeps the same behaviour:

- the queue is still bounded,
- `AddConn` / `RemoveConn` still block when the queue is full,
- connectedness events are not dropped,
- valid connections are not rejected,
- and there is no unbounded pending queue.

So the runtime semantics stay the same as today.

The only changed path is after `Swarm.Close` has already started. At that point the swarm is tearing down connections, and these connectedness notifications are coming from cleanup work. My concern is that if an application event consumer is blocked at that point, internal cleanup goroutines can also remain blocked, and then `Swarm.Close` may never return.

If the intended contract is that `Swarm.Close` is allowed to wait indefinitely behind a blocked connectedness event consumer, then I agree this patch is not the right direction.

But if `Swarm.Close` is expected to make progress once shutdown has started, then I think this is a reasonable boundary: keep strict backpressure during normal operation, but don’t let application-level event consumption hold the swarm shutdown forever.

Could you confirm what contract we want here? Should `Swarm.Close` be allowed to block forever on connectedness event consumers, or should shutdown cleanup be allowed to make progress even when the event consumer is stuck?

sukunrt:
There's no reason to change this behaviour in shutdown.
swarm.Close makes progress if eventbus consumers are consuming events. If they aren't swarm.Close shouldn't complete. And the eventbus will log for slow consumers. 

There's also a test for this: `TestCloseWithBlockingConnectedNotifee`

Sahil-4555:
Thanks, got it.

I was looking at this slightly differently because `TestCloseWithBlockingConnectedNotifee` is for the `Notifiee.Connected` callback path, while this PR was touching the `EvtPeerConnectednessChanged` eventbus queue path.

But I understand your point now. The intended behaviour is that `swarm.Close` can remain blocked if the application side is not consuming connection notifications/events. So even during shutdown, we should not bypass that backpressure or make the connectedness events best-effort.

With that contract, this PR is not the right direction. It would allow `Close` to finish by skipping connectedness delivery during shutdown, which changes the existing semantics.

PR #3512, opened 2026-05-28
feat(webrtc): stable /certhash across restarts

lidel:
## Problem

webrtc-direct minted a fresh random ECDSA cert on every process start. The cert's SHA-256 is the `/certhash` in the listen multiaddr, so the advertised address changed on every restart and invalidated anything that had pinned it: peerstore TTLs, DHT records, external address books, dialer caches.

## Fix

Derive the DTLS cert deterministically from the libp2p host key, so the `/certhash` depends only on that key and stays put across restarts:

- HKDF (stdlib `crypto/hkdf`) over the host key yields the serial and the ECDSA P-256 key; signing with a nil rand source is deterministic on Go 1.24+.
- A fixed `2020-01-01` to `2120-01-01` window keeps the DER byte-stable.
- The serial is read unsigned so it's always positive, as RFC 5280 requires. The old signed-then-negate left `math.MinInt64` negative, which `x509.CreateCertificate` rejects; I caught it fuzzing the serial against the RFC. webtransport shared the bug, so it's fixed here too.
  - I think we did not notice this before because new cert on each restart hid this as an ephemeral hiccup.
- A golden-vector test pins the `/certhash` for a fixed key, so any drift in the derivation fails CI instead of silently changing every node's address.

The DTLS cert isn't the identity credential (Noise XX inside DTLS handles that), so a stable cert is safe. Full rationale and spec links live in `p2p/transport/webrtc/cert.go`.

### DIFF
--- a/p2p/transport/webrtc/cert.go
+++ b/p2p/transport/webrtc/cert.go
@@ -70,11 +70,11 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/hkdf"
-	"crypto/rand"
 	"crypto/sha256"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"math/big"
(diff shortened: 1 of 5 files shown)


p2p/transport/webrtc/cert.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	ecdsaSeed := make([]byte, 192/8) // 192 bits of entropy for P-256, per filippo.io/keygen
+	if _, err := io.ReadFull(r, ecdsaSeed); err != nil {
+		return nil, fmt.Errorf("derive ECDSA seed: %w", err)
+	}
+	priv, err := keygen.ECDSA(elliptic.P256(), ecdsaSeed)
+	if err != nil {
+		return nil, fmt.Errorf("derive ECDSA key: %w", err)
+	}
```
sukunrt: Can you use `webrtc.GenerateCertificate(priv)` for the certificate creation?

sukunrt:
@MarcoPolo This looks fine to me, but I'd like your review on this.

p2p/transport/webrtc/cert.go
lidel: Sadly no, it's non-deterministic: random serial, `NotBefore`/`NotAfter` from `time.Now()`, and it signs with `rand.Reader`. The DER bytes change every run, so `/certhash` would shift on each restart, which is the behavior (bug?) this  PR fixes. `NewCertificate` has the same problem since it also signs with `rand.Reader`, so Go 1.24's deterministic-ECDSA path never kicks in. That's why the cert is built by hand with a nil-rand signer and wrapped via `webrtc.CertificateFromX509`.

hpndalem1-droid (approved):

lidel:
Pushed two small follow-ups here, both belong in this PR (nothing that needs a separate change).

The cert serial derivation wasn't RFC 5280 compliant: it read the bytes as a signed int64 and flipped the sign when negative, but math.MinInt64 flips back to itself, so that one value stayed negative and x509.CreateCertificate refuses a negative serial. I ran into it while fuzzing the serial against the RFC. webtransport had the same pattern, so I fixed both rather than leave one half-done.

I also added a golden-vector test that pins the certhash for a fixed key, so the derivation can't drift without CI catching it.

No behavior change otherwise. The webrtc side also moves to the stdlib crypto/hkdf, which produces identical bytes.

sukunrt (approved):

PR #3513, opened 2026-05-28
feat(autonatv2): expose AllowPrivateAddrs via an Option

gmelodie:
## Summary

Expose AutoNATv2's `allowPrivateAddrs` setting through the top-level `libp2p`
package so callers can opt in without patching the library.

### Changes

- `p2p/protocol/autonatv2`
  - Rename the private `allowPrivateAddrs` field/option to the exported
    `AllowPrivateAddrs` (commit `e3a1d72d`).

- `config`, `options.go`
  - Add `Config.AutoNATv2AllowPrivateAddrs`.
  - Add a new `libp2p.AutoNATv2AllowPrivateAddrs()` option.
  - When used alongside `EnableAutoNATv2()`, the fx provider passes
    `autonatv2.AllowPrivateAddrs` to `autonatv2.New`
    (commit `b85ffade`).

- `libp2p_test.go`
  - Add a smoke test verifying that:

    ```go
    New(
        EnableAutoNATv2(),
        AutoNATv2AllowPrivateAddrs(),
    )
    ```

    successfully builds a host.

## Motivation

`nim-libp2p`'s AutoNATv2 interop tests currently maintain a patch against
`go-libp2p` to disable private-address filtering:

- `disable-filtering-of-private-ip-addresses.patch`
  (`interop/autonatv2/go-peer/...`)

The setting already exists internally; this change simply exposes it through the
public API so the patch can be removed (see `nim-libp2p#1699`).

## Notes

- The option is documented as test-only and explicitly discouraged for
  production use.
- No behavioral change for existing callers; the default remains `false`.
- https://github.com/vacp2p/nim-libp2p/pull/2558

### DIFF
--- a/p2p/protocol/autonatv2/options.go
+++ b/p2p/protocol/autonatv2/options.go
@@ -56,7 +56,7 @@ func withDataRequestPolicy(drp dataRequestPolicyFunc) AutoNATOption {
 	}
 }
 
-func allowPrivateAddrs(s *autoNATSettings) error {
+func AllowPrivateAddrs(s *autoNATSettings) error {
 	s.allowPrivateAddrs = true
 	return nil
 }
--- a/p2p/protocol/autonatv2/autonat_test.go
+++ b/p2p/protocol/autonatv2/autonat_test.go
@@ -98,7 +98,7 @@ func TestAutoNATPrivateAddr(t *testing.T) {
 }
 
 func TestClientRequest(t *testing.T) {
-	an := newAutoNAT(t, nil, allowPrivateAddrs)
+	an := newAutoNAT(t, nil, AllowPrivateAddrs)
 	defer an.Close()
 	defer an.host.Close()
 
@@ -133,7 +133,7 @@ func TestClientRequest(t *testing.T) {
 }
 
 func TestClientServerError(t *testing.T) {
-	an := newAutoNAT(t, nil, allowPrivateAddrs)
+	an := newAutoNAT(t, nil, AllowPrivateAddrs)
 	defer an.Close()
 	defer an.host.Close()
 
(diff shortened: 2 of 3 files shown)


sukunrt (approved):

p2p/protocol/autonatv2/options.go
```diff
@@ -4,7 +4,7 @@ import "time"
 
 // autoNATSettings is used to configure AutoNAT
 type autoNATSettings struct {
-	allowPrivateAddrs                    bool
+	AllowPrivateAddrs                    bool
```
MarcoPolo: Why does this field (and other similar fields across this diff) need to be public?

gmelodie: On nim-libp2p we have an integration test between nim-libp2p and go-libp2p for autonat, since that is not yet present in https://github.com/libp2p/unified-testing. For that we need to allow private addresses on go-libp2p (just like you do in your tests), but there's no way to do that as a user currently. What we've been doing is cloning go-libp2p and applying an ugly patch instead. This makes the option public so we no longer have to do that there.

MarcoPolo: But why do you need the field of the autoNATSettings struct to be public? Remember in Go, when you capitalize a field you make it public to users outside the package it is defined in.

gmelodie: We have two scripts that instanciate peers: [one in go](https://github.com/vacp2p/nim-libp2p/tree/master/interop/autonatv2/go-peer) that imports go-libp2p and another one in nim that imports nim-libp2p, they instanciate autonat-capable peers and communicate. For that test, the go script (a "user outside the package", if you will) needs to be able to set autonat settings, specifically the one that allows private ip addresses.

MarcoPolo: I think exposing just [`AllowPrivateAddrs func`](https://github.com/libp2p/go-libp2p/pull/3513/changes#diff-edccfc1e3ab2235a3bf78a6aed5c1fe275a06c839ac55395e1e8b08c4e2b9eb4R59) should be enough, no? Does that make sense? I can share a more concrete patch if that helps.

gmelodie: Done! How does this look?

MarcoPolo (approved):

PR #3516, opened 2026-06-09
fix(peer): Allow `AddrInfo` with empty `ID` field to round-trip through JSON

dimitar-kostadinov:
The `AddrInfo.MarshalJSON` successfully serialise objects with empty `ID` field, but `AddrInfo.UnmarshalJSON` then failed because `peer.ID.UnmarshalJSON` rejects empty string.

The `ID` field in the `addrInfoJson` is tag with json:",omitempty" so an empty ID is omitted on marshal and left as the zero value on unmarshal.


dimitar-kostadinov:
Аny chance to get a feedback on this PR?

PR #3518, opened 2026-06-16
fix: resolve concurrency bugs in eventbus and webtransport

Sahil-4555:
### Issue

 observed two concurrency issues:

1. **Eventbus Deadlock**: When multiple emitters concurrently broadcast events to a slow consumer (full queue), they can deadlock. Both the wildcard subscriber node and standard subscriber node shared a single `slowConsumerTimer` pointer. Concurrent emitters raced on `Reset()`, causing the timer to fire only once. One emitter consumed the tick and blocked on the channel, leaving other emitters permanently blocked on `<-timer.C` even after the queue was drained. This hang occurred under a read-lock (`RLock`), blocking any subscription closure (`sub.Close()`) or node updates from acquiring a write-lock.
2. **WebTransport Cert Manager Data Race**: The `SerializedCertHashes()` getter read the shared `m.serializedCertHashes` slice without acquiring a lock. This caused a read/write data race with the background cert manager thread which updates the slice during certificate rotation. Returning the raw slice by reference also caused slice aliasing, exposing the caller to backing memory being concurrently mutated/resized.

### Fix

1. **Eventbus Timer Isolation**: Removed the shared `slowConsumerTimer` pointer field from `node` and `wildcardNode`. The slow consumer warning logic was refactored to use a local `time.Timer` allocated on the stack within `emitAndLogError`, eliminating shared timer state across concurrent emitters.
2. **WebTransport Thread-Safety**: Added `m.mx.RLock()` to `SerializedCertHashes()` and returned a deep copy of the slice rather than a direct reference to avoid data races and slice aliasing.

### Tests

1. **Eventbus**: Added `TestWildcardSlowConsumerDeadlock` in `p2p/host/eventbus/basic_test.go` which simulates a full queue under concurrent emissions to verify that no deadlocks occur and the emitters safely proceed when the queue is drained.

### DIFF
--- a/p2p/host/eventbus/basic.go
+++ b/p2p/host/eventbus/basic.go
@@ -334,8 +334,6 @@ type wildcardNode struct {
 	nSinks        atomic.Int32
 	sinks         []*namedSink
 	metricsTracer MetricsTracer
-
-	slowConsumerTimer *time.Timer
 }
 
 func (n *wildcardNode) addSink(sink *namedSink) {
@@ -385,12 +383,7 @@ func (n *wildcardNode) emit(evt any) {
 		select {
 		case sink.ch <- evt:
 		default:
-			slowConsumerTimer := emitAndLogError(n.slowConsumerTimer, wildcardType, evt, sink)
-			defer func() {
-				n.Lock()
-				n.slowConsumerTimer = slowConsumerTimer
-				n.Unlock()
-			}()
+			emitAndLogError(wildcardType, evt, sink)
 		}
 	}
 	n.RUnlock()
@@ -410,8 +403,6 @@ type node struct {
 
 	sinks         []*namedSink
 	metricsTracer MetricsTracer
-
-	slowConsumerTimer *time.Timer
 }
 
 func newNode(typ reflect.Type, metricsTracer MetricsTracer) *node {
@@ -440,32 +431,24 @@ func (n *node) emit(evt any) {
 		select {
 		case sink.ch <- evt:
 		default:
-			n.slowConsumerTimer = emitAndLogError(n.slowConsumerTimer, n.typ, evt, sink)
+			emitAndLogError(n.typ, evt, sink)
 		}
 	}
 	n.lk.Unlock()
 }
 
-func emitAndLogError(timer *time.Timer, typ reflect.Type, evt any, sink *namedSink) *time.Timer {
+func emitAndLogError(typ reflect.Type, evt any, sink *namedSink) {
 	// Slow consumer. Log a warning if stalled for the timeout
-	if timer == nil {
-		timer = time.NewTimer(slowConsumerWarningTimeout)
-	} else {
-		timer.Reset(slowConsumerWarningTimeout)
-	}
+	timer := time.NewTimer(slowConsumerWarningTimeout)
+	defer timer.Stop()
 
 	select {
 	case sink.ch <- evt:
-		if !timer.Stop() {
-			<-timer.C
-		}
 	case <-timer.C:
 		log.Warn("subscriber is a slow consumer. This can lead to libp2p stalling and hard to debug issues.", "subscriber_name", sink.name, "event_type", typ)
 		// Continue to stall since there's nothing else we can do.
 		sink.ch <- evt
 	}
-
-	return timer
 }
 
 func sendSubscriberMetrics(metricsTracer MetricsTracer, sink *namedSink) {
--- a/p2p/host/eventbus/basic.go
+++ b/p2p/host/eventbus/basic.go
@@ -3,7 +3,9 @@ package eventbus
 import (
 	"errors"
 	"fmt"
+	"log/slog"
 	"reflect"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -25,6 +27,7 @@ type basicBus struct {
 	nodes         map[reflect.Type]*node
 	wildcard      *wildcardNode
 	metricsTracer MetricsTracer
+	log           *slog.Logger
 }
 
 var _ event.Bus = (*basicBus)(nil)
@@ -65,7 +68,8 @@ func (e *emitter) Close() error {
 func NewBus(opts ...Option) event.Bus {
 	bus := &basicBus{
 		nodes:    map[reflect.Type]*node{},
-		wildcard: &wildcardNode{},
+		wildcard: &wildcardNode{log: log},
+		log:      log,
 	}
 	for _, opt := range opts {
 		opt(bus)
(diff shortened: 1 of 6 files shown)


lidel (approved):
Thanks for this. I'd been debugging the same bug myself, so I'm glad you beat me to it :)

I pushed two small test commits on top:

- Moved the eventbus slow-consumer test over to synctest. It runs on a fake clock now, so it's fast and won't flake on CI the way the sleep-based version could.
- Added two tests for `SerializedCertHashes`: one checks the getter returns a copy, the other runs it against cert rotation under `-race`.

@MarcoPolo fwiw lgtm, and Kubo could benefit from this, but your call

p2p/host/eventbus/basic.go
```diff
@@ -440,32 +431,24 @@ func (n *node) emit(evt any) {
...
-		timer.Reset(slowConsumerWarningTimeout)
-	}
+	timer := time.NewTimer(slowConsumerWarningTimeout)
+	defer timer.Stop()
 
 	select {
 	case sink.ch <- evt:
-		if !timer.Stop() {
-			<-timer.C
-		}
```
sukunrt, line 461: You just need to remove this, no? Do you need to create a new timer always?

p2p/transport/webtransport/cert_manager.go
```diff
@@ -166,7 +166,17 @@ func (m *certManager) AddrComponent() ma.Multiaddr {
...
+	m.mx.RLock()
+	defer m.mx.RUnlock()
+	if len(m.serializedCertHashes) == 0 {
+		return nil
+	}
+	res := make([][]byte, len(m.serializedCertHashes))
+	for i, h := range m.serializedCertHashes {
+		res[i] = make([]byte, len(h))
+		copy(res[i], h)
+	}
```
sukunrt: do you need a deepcopy? Can we just do `slices.Clone`?

p2p/transport/webtransport/cert_manager.go
```diff
@@ -166,7 +166,17 @@ func (m *certManager) AddrComponent() ma.Multiaddr {
...
-	return m.serializedCertHashes
+	m.mx.RLock()
+	defer m.mx.RUnlock()
+	if len(m.serializedCertHashes) == 0 {
+		return nil
+	}
+	res := make([][]byte, len(m.serializedCertHashes))
+	for i, h := range m.serializedCertHashes {
+		res[i] = make([]byte, len(h))
+		copy(res[i], h)
```
sukunrt:
```suggestion
		res[i] = slices.Clone(h)
```

p2p/host/eventbus/basic.go
Sahil-4555: Actually, if we keep using the shared timer and just remove that drain block, the deadlock won't go away. 

The main issue is that multiple concurrent emitters will race on `Reset()` of the same shared timer. If that happens, the timer fires only once and the other emitters get stuck on `<-timer.C` indefinitely. Since they hold the `RLock` during this time, it deadlocks the whole system. 

About allocating a new timer: we only call `emitAndLogError` on the fallback/slow path when the subscriber channel is full. In the normal case, we don't allocate anything at all, so there is no hot-path performance issue. Sticking to a local timer with `defer timer.Stop()` is much safer here.

Kindly let me know if this clarifies!

p2p/transport/webtransport/cert_manager.go
Sahil-4555: Actually, `slices.Clone` performs a shallow copy of the slice. Since the type of `m.serializedCertHashes` is `[][]byte`, using `slices.Clone` will copy the outer slice, but the inner `[]byte` elements will still point to the same underlying byte arrays. 

See, if we do that, there is a risk of slice aliasing. If the caller mutates the returned cert hashes, it will directly modify the internal cache. Also, the background thread reuses and updates these hashes during certificate rotation. If the outer slice is cloned but inner bytes are shared, we can still have data races.

So, to be completely safe against concurrent mutations and data races on the actual byte arrays, we need a proper deep copy here. 

Hope this makes sense!

p2p/host/eventbus/basic.go
sukunrt: how would multiple concurrent emitters run Reset? The entry is guarded by `n.lk.Lock()`?

p2p/transport/webtransport/cert_manager.go
sukunrt: it's an internal struct not exposed to users.

p2p/host/eventbus/basic.go
Sahil-4555: You are correct - for `node.emit` it is protected by `n.lk.Lock()` (full Mutex) so only one emitter can enter at a time. There the shared timer was safe.

But the real problem was in `wildcardNode.emit` - it uses `n.RLock()` (read-lock), so multiple concurrent emitters can enter that path concurrently. They all hit the slow consumer fallback, and race on `Reset()` and `<-timer.C` of the same shared timer pointer. Here was where the deadlock was occurring.

Since both `node` and `wildcardNode` use the same helper `emitAndLogError`, and the wildcard path needs a local timer anyway, I simplified this by removing the shared timer from both structs, and always using a stack-local timer inside `emitAndLogError`. Thus, the helper is safe regardless of the locking semantics the caller has.

p2p/transport/webtransport/cert_manager.go
Sahil-4555: You are right, I checked the callers twice. Only caller is `listener.handshake()` which reads the returned hashes and passes them straight to `newEarlyDataSender` for the Noise handshake - it never mutates the inner byte slices. Also in `cacheSerializedCertHashes`, `multihash.Encode` returns a new `[]byte` each time, so the old inner slices are not mutated in-place, they get replaced wholesale.

So yeah, `slices.Clone` is enough here. The `RLock` already protects the slice header from concurrent read/write and neither the caller nor the writer mutate the inner bytes in place, so a shallow copy is safe. I'll switch this to use `slices.Clone`

sukunrt (approved):

lidel (approved):
SGTM, good to ship from my side.

@sukunrt  fyi pushed two commits on top of your `removeSink` cleanup that make tests more tight:

- dffa0f3e **test(webtransport): assert cert hash ownership**
  The concurrent test only guarded the read lock: `clear()` on a `[][]byte` compiles to `runtime.memclrHasPointers`, which has no race instrumentation, so `-race` never saw the write. An indexed store does, and now catches a missing clone too. Also brought back an ownership test, scoped to the outer slice that `slices.Clone` actually copies.

- 4f6c46c7 **test(eventbus): tighten slow-consumer assertions**
  A failed `require` in a synctest bubble unwinds into the deferred emitter close, which parks on the node lock a stalled emit still holds, so failures hung until `-timeout` rather than failing. Closing the sub first fixes it. Also yielded in the rotation reader loop: 8s at `GOMAXPROCS=1` and 2.2s on a 4 vCPU runner, now 1.1s.

Issue #3499, opened 2026-05-06
WebRTC-Direct transport at risk due to Chrome changes (ICE credential munging): alignment and next steps


johannamoran:
**Description**
Sharing a consolidated update from recent discussions across js-libp2p, py-libp2p and specs to align on implications for go-libp2p.

This needs attention as it impacts core connectivity guarantees for browser and server communication.

**Context**
Recent threads to review:
- Summary by lidel on WebRTC-Direct + Chrome changes (specs#672 comment at https://github.com/libp2p/specs/issues/672#issuecomment-4297060067 )
- feat: webrtc-direct-v2 : https://github.com/libp2p/js-libp2p/pull/3480
- docs: webrtc-direct-v2 : https://github.com/libp2p/specs/pull/715
- Summary by lidel on WebRTC-Direct + Chrome changes (specs#672 comment at https://github.com/libp2p/specs/issues/672#issuecomment-4297060067 )
- Py-libp2p discussion: https://github.com/libp2p/py-libp2p/pull/1309#issuecomment-4337569723 

**Summary of the Problem**
The /webrtc-direct transport is at risk due to upcoming changes in Google Chrome.

- WebRTC-Direct exists primarily to allow:

1. browser to public libp2p server dialing
2. without signalling
3. without CA-signed certificates

- This relies on ICE credential munging (SDP manipulation)
- Chrome is introducing a restriction (WebRTC-NoSdpMangleUfrag) that breaks this mechanism

When this change rolls out:

- Browser to server connectivity via /webrtc-direct may stop working
- The core reason for this transport’s existence is undermined

**Impact**

- Existing assumptions about browser and server direct connectivity may no longer hold
- /webrtc-direct could degrade or become unusable
- Affects:

1. onboarding flows
2. public node access patterns
3. browser-based libp2p applications

**Priority**
High

This is externally driven (Chrome changes) and affects a foundational transport. Delay risks breaking connectivity for browser-based libp2p usage.

lidel:
WebRTC-Direct v2 (no ICE ufrag munging) is ready across all three repos.
If it helps, I tested it end to end by hand using below revisions (so when in doubt, know this was the state that worked):

- Spec: libp2p/specs#715 (libp2p/specs@3587339) closes the v2 gaps: server credentials, version dispatch, ICE/DTLS roles, RFC references.
- Go: #3520 (c398b263) adds the v2 server and dialer.
[...]

PR #3520, opened 2026-06-19
feat(webrtc): support webrtc-direct v2

lidel:
## Problem

`/webrtc-direct` lets a browser dial a public libp2p server with no signalling and no CA-signed cert. That path works by "munging" the local SDP so the server can predict the browser's ICE credentials, and Chrome is taking that ability away (the `WebRTC-NoSdpMangleUfrag` field trial). Once it reaches stable, browsers can no longer open `/webrtc-direct` connections to a go-libp2p server, which is the whole reason the transport exists.

I documented it in https://github.com/libp2p/specs/issues/672#issuecomment-4297060067

## Fix

v2 spec avoids "munging" and is proposed at:
- https://github.com/libp2p/specs/pull/715

This PR implements it while maintaining backward-compatibility with v1 clients too

### Details 

- the listener accepts v1 and v2 dialers on one port, picking the flow from the ICE username fragment prefix and rejecting versions it doesn't know;
- v2 carries the browser's own ICE password in the server ufrag, so the server recovers it from the STUN `USERNAME` with no signalling;
- parsed username fragments are validated (charset and length) before they reach SDP;
- opt-in `WithDialerVersion("v2")` lets a go node dial v2; v1 stays the default for now. can be flipped in future PR, but also does not matter, GO will not do the dials, JS will (libp2p/js-libp2p#3480), so this is just for formal parity and testing

### Testing

This keeps browser-to-server connectivity alive once Chrome flips the trial; verified end to end against a real Chrome (v2, munging off) and js-libp2p both ways. 

Details: https://github.com/libp2p/go-libp2p/issues/3499#issuecomment-4755702789

## References

- This PR is GO (Server) part of #3499
- The v2 spec: libp2p/specs#715
- JS implementation : libp2p/js-libp2p#3480
  - In practice, this will be THE majority of browser clients that do the v2 dials to Kubo

### DIFF
--- a/p2p/transport/webrtc/listener.go
+++ b/p2p/transport/webrtc/listener.go
@@ -32,67 +32,6 @@ var _ network.ConnMultiaddrs = &connMultiaddrs{}
 func (c *connMultiaddrs) LocalMultiaddr() ma.Multiaddr  { return c.local }
 func (c *connMultiaddrs) RemoteMultiaddr() ma.Multiaddr { return c.remote }
 
-// The "libp2p+webrtc+" namespace selects the WebRTC Direct handshake version via
-// the ICE username fragment. v1 ("libp2p+webrtc+v1/") munges the SDP; v2
-// ("libp2p+webrtc+v2/") does not, because Chromium's WebRTC-NoSdpMangleUfrag
-// field trial rejects an offer whose ICE ufrag was munged. A v2 client instead
-// leaves its local credentials untouched and puts its own ICE password in the
-// server ufrag as "libp2p+webrtc+v2/<client_pwd>", which the server reads back
-// from the STUN USERNAME with no SDP exchange. A server MUST reject a version it
-// does not recognize rather than guess one. v2 was introduced in
-// https://github.com/libp2p/specs/pull/715; the spec lives at
-// https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md and the
-// js-libp2p implementation at https://github.com/libp2p/js-libp2p/pull/3480.
-const (
-	ufragPrefixV1 = "libp2p+webrtc+v1/"
-	ufragPrefixV2 = "libp2p+webrtc+v2/"
-)
-
-const (
-	// ICE credential bounds, per RFC 8839 section 5.4
-	// (ufrag = 4*256ice-char, password = 22*256ice-char).
-	iceUfragMinLen   = 4
-	icePwdMinLen     = 22
-	iceCredentialMax = 256
-)
-
-// ice-char is ASCII-only, so any string that passes isICECharString has exactly
-// one byte per character. The len() byte count used by isICEUfrag/isICEPwd then
-// equals both the ice-char count and the UTF-16 length that js-libp2p measures,
-// so the two implementations accept the same credentials; non-ice-char input
-// (including any multi-byte UTF-8) is rejected by both regardless of how each
-// counts length.
-
-// isICEChar reports whether b is in the RFC 8839 ice-char set
-// (ALPHA / DIGIT / "+" / "/").
-func isICEChar(b byte) bool {
-	return (b >= 'a' && b <= 'z') ||
-		(b >= 'A' && b <= 'Z') ||
-		(b >= '0' && b <= '9') ||
-		b == '+' || b == '/'
-}
-
-func isICECharString(s string) bool {
-	for i := 0; i < len(s); i++ {
-		if !isICEChar(s[i]) {
-			return false
-		}
-	}
-	return true
-}
-
-// isICEUfrag reports whether s is a syntactically valid ICE username fragment
-// (RFC 8839 section 5.4: ufrag = 4*256ice-char).
-func isICEUfrag(s string) bool {
-	return len(s) >= iceUfragMinLen && len(s) <= iceCredentialMax && isICECharString(s)
-}
-
-// isICEPwd reports whether s is a syntactically valid ICE password
-// (RFC 8839 section 5.4: password = 22*256ice-char).
-func isICEPwd(s string) bool {
-	return len(s) >= icePwdMinLen && len(s) <= iceCredentialMax && isICECharString(s)
-}
-
 const (
 	candidateSetupTimeout = 10 * time.Second
 	// This is higher than other transports(64) as there's no way to detect a peer that has gone away after
@@ -196,8 +135,8 @@ func (l *listener) listen() {
 
 			conn, err := l.handleCandidate(ctx, candidate)
 			if err != nil {
-				l.mux.RemoveConnByUfrag(candidate.Ufrag)
-				log.Debug("could not accept connection", "ufrag", candidate.Ufrag, "error", err)
+				l.mux.RemoveConnByUfrag(candidate.LocalUfrag)
+				log.Debug("could not accept connection", "ufrag", candidate.LocalUfrag, "error", err)
 				return
 			}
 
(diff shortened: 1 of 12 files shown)


sukunrt:
Is this tested via the interop tester?

lidel:
@sukunrt Not via the interop tester yet. It dials `webrtc-direct` with the default v1 dialer, so the v2 path isn't exercised there. 

I did verify v2 manually end to end against real Chrome (munging off) and js-libp2p both ways: https://github.com/libp2p/go-libp2p/issues/3499#issuecomment-4755702789 (see specific commits listed there)

So iiuc once JS ships dialling v2 (libp2p/js-libp2p#3480) we'll have browser-to-go v2 running as part of CI interop in libp2p/test-plans.

sukunrt:
Can you add an integration test here? 
https://github.com/libp2p/go-libp2p/blob/master/p2p/test/transport/transport_test.go#L342

I think you just need to enable the V2Dialer option.

sukunrt:
> It dials webrtc-direct with the default v1 dialer, so the v2 path isn't exercised there.

Can we enable go v2 dialer to go server meanwhile and add the js dialer when it's merged?

lidel:
Thanks for pointer @sukunrt, added in 53e9c5aa. The transport integration tests now run a `WebRTC-v2` case with `WithDialerVersion(2)`, exercising go's v2 dialer against a go server across the full battery (ping, streams, resets, deadlines, rcmgr). I pinned the existing case to v1 too, so both flows are explicit.

It's go dialer to go server, so self-consistency rather than cross-implementation interop. iiuc the JS dialer side lands in libp2p/test-plans once libp2p/js-libp2p#3480 merges, so this PR can be merged without blocking?

p2p/transport/webrtc/udpmux/mux.go
```diff
@@ -32,8 +32,15 @@ const ReceiveBufSize = 1500
 const maxAddrsPerUfrag = 32
 
 type Candidate struct {
+	// Ufrag is the local (server) ICE ufrag, parsed from the part of the STUN
+	// USERNAME attribute before the ':'. pion retrieves a muxed connection by
+	// its local ufrag, so the mux is keyed on this value.
 	Ufrag string
```
sukunrt: `LocalUfrag`

sukunrt: The comments on the other datastructures keyed on ufrag need to mention which ufrag we're using now, Local or Remote. Can we change the name wherever appropriate to LocalUfrat to make that clear?

p2p/transport/webrtc/udpmux/mux.go
```diff
@@ -243,25 +250,27 @@ type ufragConnKey struct {
...
-		return "", err
+		return "", "", err
 	}
 	index := bytes.Index(attr, []byte{':'})
 	if index == -1 {
-		return "", fmt.Errorf("invalid STUN username attribute")
+		return "", "", fmt.Errorf("invalid STUN username attribute")
 	}
-	return string(attr[index+1:]), nil
+	return string(attr[:index]), string(attr[index+1:]), nil
```
sukunrt: We should either reject empty ufrag here or remove the `ufrag==""` check in `RemoveConnByUfrag`

p2p/transport/webrtc/listener.go
```diff
@@ -196,9 +257,41 @@ func (l *listener) setupConnection(
...
+		// valid ICE password (ice-char, length 22..256) per RFC 8839 section 5.4.
+		if !isICEPwd(pwd) {
+			return nil, fmt.Errorf("invalid v2 ufrag %q: recovered client password is not a valid ICE password", serverUfrag)
+		}
+		clientPwd = pwd
+	case strings.HasPrefix(serverUfrag, ufragPrefixV1):
+		clientPwd = clientUfrag
+	default:
+		return nil, fmt.Errorf("unsupported WebRTC Direct version in ufrag %q", serverUfrag)
+	}
```
sukunrt: Can you move all this validation to `mux.go`. We should do all these checks before we allocate any space for the ufrag etc. This would avoid putting invalid candidates in the candidate accept channel in the listener.

p2p/transport/webrtc/transport.go
```diff
@@ -301,17 +325,38 @@ func (t *WebRTCTransport) dial(ctx context.Context, scope network.ConnManagement
...
+	// browser's would be) and we encode our password into the server ufrag as
+	// "libp2p+webrtc+v2/<pwd>", so the server can recover it from the STUN
+	// USERNAME without us munging our local SDP. See https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md.
+	var localUfrag, localPwd, serverUfrag string
+	switch t.dialerVersion {
+	// TODO: flip the default to v2 roughly 12 months after Chromium ships its
+	// no-munging behavior (the WebRTC-NoSdpMangleUfrag field trial,
+	// https://webrtc-review.googlesource.com/c/src/+/385721) in stable, giving
+	// servers and other implementations a grace period to adopt v2 first.
+	case 0, 1:
```
sukunrt: can we strictly enforce that this value is 1? 0 is confusing.

sukunrt (changes requested):

p2p/transport/webrtc/udpmux/mux.go
lidel: Done in 48464706. `Candidate.Ufrag` is now `LocalUfrag`, and I named the local ufrag the same way across the mux internals (`ufragConnKey.localUfrag`, `muxedConnection.localUfrag`, and the `GetConn`/`getOrCreateConn`/`RemoveConnByUfrag` params). The `ufragMap` and `ufragAddrMap` comments now say they key on the local (server) ufrag.

p2p/transport/webrtc/udpmux/mux.go
lidel: Rejected it upstream instead of keeping the guard. `credentialsFromSTUNMessage` runs `isICEUfrag` on both halves (min length 4), so an empty server ufrag is dropped before any allocation. The `ufrag == ""` check in `RemoveConnByUfrag` was then unreachable, so I removed it. 48464706.

p2p/transport/webrtc/listener.go
lidel: Ack, failing early is safer. Moved to the udpmux package in 48464706. 

`credentialsFromSTUNMessage` now does the ICE-ufrag checks, version dispatch, and client-password recovery before `getOrCreateConn`, so invalid or unknown-version requests are dropped before anything is allocated or queued. The ice-char validators and the v1/v2 prefixes moved into the package too (`udpmux` can't import the transport package), and the listener just reads the validated `LocalUfrag`/`RemoteUfrag`/`RemotePwd` off the `Candidate`.

One thing to flag: I kept the leaf validators and prefixes in a sibling `ufrag.go` and the parse/dispatch in `mux.go`, both in `udpmux`. Happy to fold it all into `mux.go` if you'd rather have it in one file.

p2p/transport/webrtc/transport.go
lidel: Done in 48464706. `New` now defaults `dialerVersion` to 1, so `dial` only handles `case 1` and `case 2` and errors on anything else; 0 never reaches it. `WithDialerVersion` still accepts only 1 or 2.

p2p/transport/webrtc/listener.go
sukunrt: this is fine. I just wanted the check to happen earlier in the processing.

sukunrt (approved):

PR #3524, opened 2026-07-11
perf(swarm): cut allocations in the dial ranker

Sahil-4555:
`DefaultDialRanker` runs on every dial, so I had a look at where its time actually goes. Almost all of it was in `score()`.

The sort in `getAddrDelay` calls `score()` O(n·log n) times per ranking. Each of those calls walked the multiaddr up to five times through `ValueForProtocol`, and allocated a string for the port only to hand it straight to `strconv.Atoi`.

Two changes:

- `score()` now walks the address once and reads the port straight out of the component bytes. TCP and UDP both store it as a big endian uint16, so there is nothing to parse. It doesn't allocate at all now.
- `res` in `DefaultDialRanker` was sized with `len(addrs)`, but `filterAddrs` partitions in place, so by that line `addrs` is only the leftover addresses, not the total that `res` ends up holding. It's sized from the count before filtering now.

Behaviour is unchanged: same scores, same ordering. I added a test that pins the exact score for every address shape we handle and checks the ranking order, plus one that keeps `score()` allocation free so it doesn't quietly regress later.

### Benchmark

master vs this branch, `-benchtime=300ms -count=10 -cpu=1`. The two test binaries were run alternately, ten times each, so CPU frequency drift lands on both sides equally.

```
goos: linux
goarch: amd64
pkg: github.com/libp2p/go-libp2p/p2p/net/swarm
cpu: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz

                           │   master     │            this branch              │
                           │    sec/op    │   sec/op     vs base                │
DefaultDialRanker/addrs=4     2.084µ ± 3%   1.198µ ± 4%  -42.50% (p=0.000 n=10)
DefaultDialRanker/addrs=12    9.652µ ± 3%   4.167µ ± 3%  -56.84% (p=0.000 n=10)
DefaultDialRanker/addrs=30   26.604µ ± 3%   9.336µ ± 5%  -64.91% (p=0.000 n=10)
geomean                       8.118µ        3.599µ       -55.67%

                           │   master     │            this branch               │
                           │     B/op     │     B/op      vs base                │
DefaultDialRanker/addrs=4      480.0 ± 0%     352.0 ± 0%  -26.67% (p=0.000 n=10)
DefaultDialRanker/addrs=12   2.438Ki ± 0%   1.031Ki ± 0%  -57.69% (p=0.000 n=10)
DefaultDialRanker/addrs=30   6.438Ki ± 0%   2.281Ki ± 0%  -64.56% (p=0.000 n=10)
geomean                      1.945Ki          954.0       -52.09%

                           │   master    │            this branch             │
                           │  allocs/op  │ allocs/op   vs base                │
DefaultDialRanker/addrs=4    21.000 ± 0%   5.000 ± 0%  -76.19% (p=0.000 n=10)
DefaultDialRanker/addrs=12   107.00 ± 0%  13.000 ± 0%  -87.85% (p=0.000 n=10)
DefaultDialRanker/addrs=30   319.00 ± 0%  13.000 ± 0%  -95.92% (p=0.000 n=10)
geomean                       89.49        9.454       -89.44%
```

### What this means for a dial

To put those numbers in proportion, one full outbound connection over loopback (handshake, security, muxer) costs:

```
BenchmarkFullDialTCP    3.3ms    442818 B/op    3345 allocs/op
BenchmarkFullDialQUIC   2.1ms    400000 B/op    2717 allocs/op
```

The ranker is called once per dial, so for a peer advertising 30 addresses this takes about **9% of the allocations of a TCP dial** off the table (306 of 3345), and about 3% for a peer with 12 addresses.

On wall clock it's smaller: roughly 0.5% of a loopback dial, and less than that on a real network where the RTTs dominate. So the win here is allocation and GC pressure rather than dial latency, which is what you'd want it to be — the allocation count doesn't shrink when the network gets slower.


Sahil-4555:
heya, @sukunrt @lidel @marten-seemann can you please review this PR when get chance

Sahil-4555:
heya, @sukunrt @lidel @marten-seemann @MarcoPolo can you please review this PR when get chance

PR #3526, opened 2026-07-15
fix(basichost): sort confirmed addrs

lidel:
## Problem

AutoNAT v2 keeps a list of the node's addresses that are confirmed reachable from the internet. The list is built in two parts: primary transports (tcp, quic-v1) first, then secondary ones (webrtc-direct, websocket, webtransport). Each part is sorted, but the combined list is not.

The consumer of that list, `removeNotInSource`, assumes fully sorted input and silently drops entries when it is not. webrtc-direct is the common victim: on a shared UDP port its address sorts before quic-v1, so it lands out of order and gets dropped. `ConfirmedAddrs()` then says the node is not reachable over webrtc-direct even though AutoNAT just confirmed it was, undoing the inheritance from #3435. Kubo announces confirmed addresses to HTTP routers, so browsers never saw the webrtc-direct address (ipfs/kubo#11369) :see_no_evil: 

## Fix

- sort the list before returning it from the probe manager
- experience existential dread
- add regression tests at two levels; both fail without the fix

### DIFF
--- a/p2p/host/basic/addrs_reachability_tracker.go
+++ b/p2p/host/basic/addrs_reachability_tracker.go
@@ -433,7 +433,7 @@ func (m *probeManager) UpdateAddrs(addrs []ma.Multiaddr) {
 	m.mx.Lock()
 	defer m.mx.Unlock()
 
-	slices.SortFunc(addrs, func(a, b ma.Multiaddr) int { return a.Compare(b) })
+	slices.SortFunc(addrs, ma.Multiaddr.Compare)
 	statuses := make(map[string]*addrStatus, len(addrs))
 	for _, addr := range addrs {
 		k := string(addr.Bytes())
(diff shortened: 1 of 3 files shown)


p2p/host/basic/addrs_reachability_tracker.go
```diff
@@ -425,6 +428,16 @@ func (m *probeManager) AppendConfirmedAddrs(reachable, unreachable, unknown []ma
...
+
+	// primaryAddrs and secondaryAddrs are each sorted, but interleave in the
+	// buckets above (a secondary like webrtc-direct sorts before its quic-v1
+	// primary). Unsorted output makes removeNotInSource silently drop
+	// confirmed addrs, and makes removeInSource fail to filter unreachable
+	// addrs out of Addrs().
+	cmp := func(a, b ma.Multiaddr) int { return a.Compare(b) }
+	slices.SortFunc(reachable, cmp)
+	slices.SortFunc(unreachable, cmp)
+	slices.SortFunc(unknown, cmp)
```
lidel: @sukunrt @MarcoPolo this PR is literally just these 3 sorts, ok to merge?

sukunrt (approved):

PR #3528, opened 2026-07-22
fix(basichost): fix some bugs in addrs reachability manager

sukunrt:
These are all minor except https://github.com/libp2p/go-libp2p/commit/7ccb3cdd9b50630a4180e124f306f145f65fb402 which may cause a panic on close. 

All the commits are meaningful. See the commits for individual issues.


lidel (approved):
Pushed two follow-ups from review:

- 7ba8874b **refactor(autonatv2): unlock pickServer via defer**
  so a panic inside the locked section can't leave `an.mx` held and wedge `Close()`

- fad42536 **test(basichost): cancel in-flight refresh probes**
  cancel() used to win the startup race, so the mock client was never entered; the subtest now parks every worker inside the client before cancelling

Otherwise, fixes look solid.  @sukunrt  LGTM, if it looks ok to you too, feel free to merge.

PR #3530, opened 2026-07-27
chore: update quic-go to v0.61.0

lidel:
> [!WARNING]
>
> This feels like a minefield, interfaces changed, abstraction changed, and things may required deeper refactor.
>
> Not adopted here: `ClientConn` session pooling (quic-go/webtransport-go#344) and flow-control config (quic-go/webtransport-go#340). Both are behavior changes and neither is in a tagged release, nor wired in this PR. 
>
> Opening this only for illustration purpose.
> I'll open second PR which uses last "stable" release of webtransport-go instead.

## Problem

quic-go v0.61.0 replaced `http3.ParseCapsule` with the stateful `CapsuleParser` API (quic-go/quic-go#5729), and every tagged webtransport-go release still calls the old one (lol?). No tagged webtransport-go builds against it.

## Fix

- webtransport-go moves to master `23b4cd1`, which carries the port from quic-go/webtransport-go#323. That commit requires quic-go v0.61.0 itself, so the untagged pin rests on a real release rather than another pseudo-version.
- One call site changes: master renamed `Dialer` to `Transport` (quic-go/webtransport-go#343). `Transport.init` and `Close` behave exactly as `Dialer` did, so no lifecycle change.

`TestDialWorkerLoopTCPConnUpgradeWait` and the `holepunch` simnet panic fail on unmodified master too, 3/3 runs each, so they are not from this bump.

Once webtransport-go tags a release with the capsule port, this drops back to an ordinary version bump.


sukunrt:
let's wait for webtransport-go to make a new release? Till then let's use v0.60 quic-go?

lidel:
Ack, opened https://github.com/libp2p/go-libp2p/pull/3532 with that approach. Closing this one.

PR #3533, opened 2026-07-28
test(basichost): add a stress test for reachability logic

sukunrt:
This does catch the bug fixed by https://github.com/libp2p/go-libp2p/pull/3526 and is exhaustive. 

I'd like to add one end to end test with running basichosts with address discovery but making webrtc work with simnet is a lot more work so that'll take more time.


lidel (approved):
LGTM, fine to merge once CI is green.

PR #3540, opened 2026-09-09
fix(http/auth): reject oversized opaque

MarcoPolo:
Client-initiated PeerID auth could embed a large public key in the server opaque blob. When that blob exceeded the 1024-byte scratch buffer, slicing the buffer paniced. Return errTooBig instead.

Co-authored-by: Marco Munizaga <git@marcopolo.io>

Reported-by: <sahilsojitra4555@gmail.com>

### DIFF
--- a/p2p/http/auth/internal/handshake/handshake_test.go
+++ b/p2p/http/auth/internal/handshake/handshake_test.go
@@ -337,6 +337,34 @@ func FuzzServerHandshakeNoPanic(f *testing.F) {
 	})
 }
 
+// TestClientInitiatedOversizedPublicKeyNoPanic ensures a client-initiated
+// handshake with a large public-key param returns an error instead of
+// panicking when the opaque value would exceed the server's fixed scratch buffer.
+func TestClientInitiatedOversizedPublicKeyNoPanic(t *testing.T) {
+	serverKey, _, err := crypto.GenerateEd25519Key(rand.Reader)
+	require.NoError(t, err)
+
+	// Large enough that opaque JSON (HMAC + client-public-key + metadata) exceeds
+	// the 1024-byte scratch buffer, but the Authorization header stays under maxHeaderSize.
+	largeKey := bytes.Repeat([]byte{0xAB}, 900)
+	challenge := bytes.Repeat([]byte{0x11}, challengeLen)
+	hdr := fmt.Sprintf(
+		`libp2p-PeerID challenge-server="%s", public-key="%s"`,
+		base64.URLEncoding.EncodeToString(challenge),
+		base64.URLEncoding.EncodeToString(largeKey),
+	)
+	require.LessOrEqual(t, len(hdr), maxHeaderSize)
+
+	h := PeerIDAuthHandshakeServer{
+		Hostname: "example.com",
+		PrivKey:  serverKey,
+		Hmac:     hmac.New(sha256.New, make([]byte, 32)),
+	}
+	require.NoError(t, h.ParseHeaderVal([]byte(hdr)))
+	err = h.Run()
+	require.ErrorIs(t, err, errTooBig)
+}
+
 func BenchmarkOpaqueStateWrite(b *testing.B) {
 	zeroBytes := [32]byte{}
 	hmac := hmac.New(sha256.New, zeroBytes[:])
--- a/p2p/http/auth/internal/handshake/server.go
+++ b/p2p/http/auth/internal/handshake/server.go
@@ -280,6 +280,11 @@ func (h *PeerIDAuthHandshakeServer) addOpaqueParam() error {
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


p2p/http/auth/internal/handshake/server.go
```diff
@@ -280,6 +280,11 @@ func (h *PeerIDAuthHandshakeServer) addOpaqueParam() error {
 	if err != nil {
 		return err
 	}
+	// Reject when Marshal grew past the fixed scratch buffer; otherwise
+	// h.buf[len(opaqueVal):] panics (e.g. client-initiated large public keys).
+	if len(opaqueVal) > len(h.buf) {
```
Sahil-4555, line 285: Nice catch on the actual root cause here. Guarding on `len(opaqueVal) > len(h.buf)` before the slice is exactly the right fix, and using `>` (not `>=`) is correct too, since `h.buf[1024:]` is still a valid empty slice.

One thing I wanted to raise for discussion: on the client-initiated path this opaque blob carries the client's public key, so any client whose key pushes the opaque past 1024 bytes will now get a `400`. I tested with a real RSA-8192 key (≈1067 bytes marshalled, header still well under `maxHeaderSize`) - earlier it panicked, now it returns `errTooBig` cleanly, which is a big improvement. But it does mean such a peer can effectively never authenticate on this path, and the message `"header value too big"` is slightly misleading, because the header itself is within limits - it is the derived opaque that overflowed.

If we want legitimate large keys to keep working, one option is to just let `writeParamB64` allocate when it does not fit (it already handles a nil scratch):

```go
var scratch []byte
if len(opaqueVal) <= len(h.buf) {
    scratch = h.buf[len(opaqueVal):]
}
h.hb.writeParamB64(scratch, "opaque", opaqueVal)
```

The input is already bounded by `maxHeaderSize`, so the allocation stays small. If the rejection is intentional (to bound the opaque size), then maybe a dedicated error / clearer message would help. Either way is fine with me — just wanted it to be a conscious choice rather than an accidental limit.

p2p/http/auth/internal/handshake/handshake_test.go
```diff
@@ -337,6 +337,34 @@ func FuzzServerHandshakeNoPanic(f *testing.F) {
 	})
 }
 
+// TestClientInitiatedOversizedPublicKeyNoPanic ensures a client-initiated
+// handshake with a large public-key param returns an error instead of
+// panicking when the opaque value would exceed the server's fixed scratch buffer.
+func TestClientInitiatedOversizedPublicKeyNoPanic(t *testing.T) {
```
Sahil-4555, line 343: Small suggestion while we are adding tests here: the existing `FuzzServerHandshakeNoPanic` just above already exercises this exact `ParseHeaderVal -> Run` path, but it never caught this panic — mainly because it has no seed corpus, so the fuzzer almost never constructs a valid `libp2p-PeerID` header with a large enough `public-key` on its own. If we add a single `f.Add(...)` seed with a valid client-initiated header (ideally with a largish key), the fuzzer can then mutate around this boundary and would catch this whole class of overflow in any handshake state, not just the one we are fixing here. The targeted unit tests are good — this would just make sure the next sibling bug does not slip through.

p2p/http/auth/internal/handshake/handshake_test.go
```diff
@@ -337,6 +337,34 @@ func FuzzServerHandshakeNoPanic(f *testing.F) {
...
+	serverKey, _, err := crypto.GenerateEd25519Key(rand.Reader)
+	require.NoError(t, err)
+
+	// Large enough that opaque JSON (HMAC + client-public-key + metadata) exceeds
+	// the 1024-byte scratch buffer, but the Authorization header stays under maxHeaderSize.
+	largeKey := bytes.Repeat([]byte{0xAB}, 900)
```
Sahil-4555, line 349: Minor: using `bytes.Repeat([]byte{0xAB}, 900)` is perfectly fine since the guard is purely size-based. But since the more realistic trigger is an actual large key (e.g. RSA-8192, which libp2p still allows up to 8192 bits), it might be nicer to use a real marshalled RSA-8192 public key here. That way the test also documents the behaviour for a genuinely valid key, and not only for arbitrary bytes.

Sahil-4555 (commented):
Thanks for the quick turnaround on this @MarcoPolo 🙏 I went through the change in a fair bit of detail and also tested it locally — the guard is correct, the panic is gone (I checked with a genuine RSA-8192 key as well, not just random bytes), and the existing `p2p/http/auth/...` suite still passes, so no regressions from what I can see.

Overall this LGTM as a security fix. I have left a few inline comments - the main one worth a look is the large-key behaviour on the client-initiated path (we now reject instead of panicking, which is a clear improvement, but it also means such a peer can no longer authenticate on that path). The rest are just test/robustness suggestions. None of them are blockers.

sukunrt (approved):

